PR #6, opened 2019-05-08
Move the storage details to a separate module

Dexterp37:
This PR moves the rkv implementation details, including Lifetime.Application handling, to
the newly introduced Database module.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -1,0 +1,205 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use rkv::{Rkv, SingleStore, StoreOptions};
+use std::collections::btree_map::Entry;
+use std::collections::BTreeMap;
+use std::fs;
+
+use crate::metrics::Metric;
+use crate::Lifetime;
+
+#[derive(Debug)]
+pub struct Database {
+    rkv: Option<Rkv>,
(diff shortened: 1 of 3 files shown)


glean-core/src/database/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+    // Metrics with 'application' lifetime only live as long
+    // as the application lives: they don't need to be persisted
+    // to disk using rkv. Store them in a map.
+    app_lifetime_data: BTreeMap<String, Metric>,
+}
+
+impl Database {
+    pub fn new() -> Self {
+        Self {
+            rkv: None,
```
badboy: We can probably later move away from option here and instantiate on `new` and then handle the init part on the glean object only

glean-core/src/database/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+            app_lifetime_data: BTreeMap::new(),
+        }
+    }
+
+    // Initialize the data store.
```
badboy: Use `///` (3 slashes) to make it a doc comment.
This gives us:
* rendered docs (when `--document-private-items` is used)
* doc tests

glean-core/src/database/mod.rs
```diff
@@ -0,0 +1,205 @@
...
+
+        let rkv = self.rkv.as_ref().unwrap();
+        let store: SingleStore = rkv
+            .open_single(store_name.as_str(), StoreOptions::create())
+            .unwrap();
+        let writer = rkv.write().unwrap();
+        transaction_fn(writer, store);
+    }
+
+    // Records a metric in the underlying storage system.
```
badboy: As above, 3 slashes.

Same for further down.

PR #15, opened 2019-05-08
Implement API to get the value back from a metric

badboy:

### DIFF
--- a/glean-core/src/metrics/counter.rs
+++ b/glean-core/src/metrics/counter.rs
@@ -1,4 +1,5 @@
 use crate::metrics::Metric;
+use crate::storage::StorageManager;
 use crate::CommonMetricData;
 use crate::Glean;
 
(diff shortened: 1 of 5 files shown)


glean-core/src/metrics/counter.rs
```diff
@@ -24,4 +25,19 @@ impl CounterMetric {
...
+    /// **Test-only API (exported for FFI purposes).**
+    ///
+    /// Get the currently stored value as an integer.
+    ///
+    /// This doesn't clear the stored value.
+    pub fn test_get_value(&self, glean: &Glean, storage_name: &str) -> Option<u64> {
+        let snapshot = StorageManager.snapshot_as_json(glean.storage(), storage_name, false);
+        snapshot
+            .as_object()
+            .and_then(|o| o.get("counter"))
```
Dexterp37: That's a nice way to chain Options!

glean-core/src/metrics/counter.rs
```diff
@@ -24,4 +25,19 @@ impl CounterMetric {
...
+            .and_then(|o| o.as_object())
+            .and_then(|o| o.get(&self.meta.identifier()))
+            .and_then(|o| o.as_i64().map(|i| i as u64))
```
mdboom: In the future we support deserialization in the `Metrics` enum so we could do this generically?  (Not to hold up this PR)

badboy: We might, I need to find a way to make that nice.

PR #16, opened 2019-05-08
Add test coverage for Database record

Dexterp37:
This additionally moves some Database specific
detail (i.e. storage names formatted as "{}#")
back to the Database.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -61,17 +61,20 @@ impl Database {
         rkv
     }
 
-    pub fn iter_store_from<F>(&self, lifetime: Lifetime, iter_start: &str, mut transaction_fn: F)
+    /// Iterates with the provided transaction function on the data from
+    /// the given storage.
+    pub fn iter_store_from<F>(&self, lifetime: Lifetime, storage_name: &str, mut transaction_fn: F)
     where
         F: FnMut(&[u8], &Metric),
     {
+        let iter_start = format!("{}#", storage_name);
         let len = iter_start.len();
 
         // Lifetime::Application data is not persisted to disk
         if lifetime == Lifetime::Application {
             let data = self.app_lifetime_data.read().unwrap();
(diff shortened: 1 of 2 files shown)


glean-core/src/database/mod.rs
```diff
@@ -206,11 +220,36 @@ impl Database {
         store.put(&mut writer, final_key, &value).unwrap();
         let _ = writer.commit();
     }
+
+    /// Clears a storage (only Ping Lifetime).
+    pub fn clear_ping_lifetime_storage(&self, storage_name: &str) {
```
mdboom: Should we make this generic by passing in the lifetime enum?

Dexterp37: I'd rather not do this for now, since all we want is clearing the Ping lifetime. Unless there's some other use-case I've missed?

glean-core/src/database/mod.rs
```diff
@@ -61,17 +61,20 @@ impl Database {
...
     {
+        let iter_start = format!("{}#", storage_name);
         let len = iter_start.len();
 
         // Lifetime::Application data is not persisted to disk
         if lifetime == Lifetime::Application {
             let data = self.app_lifetime_data.read().unwrap();
             for (key, value) in data.iter() {
-                if key.starts_with(iter_start) {
+                if key.starts_with(storage_name) {
```
badboy: Shouldn't this still be `iter_start`?

glean-core/src/database/mod.rs
badboy: We will eventually need a way to delete the `user` storage as well, but I think we can tackle that when we get to it.

glean-core/src/database/mod.rs
Dexterp37: yes, you're right!

glean-core/src/database/mod.rs
Dexterp37: Good point!

PR #38, opened 2019-05-09
Add smoke values for common fields to make valid pings

Dexterp37:

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -42,12 +42,18 @@ impl PingMaker {
         json!({
             "ping_type": storage_name,
             "seq": self.get_ping_seq(storage_name),
+            "start_time": "2019-03-29T09:50-04:00",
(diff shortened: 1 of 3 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -42,12 +42,18 @@ impl PingMaker {
...
         })
     }
 
     fn get_client_info(&self, storage: &Database) -> JsonValue {
         let client_info = StorageManager.snapshot_as_json(storage, "glean_client_info", true);
-        let mut map = json!({});
+        // Add the "telemetry_sdk_build", which is the glean-core version.
+        let version = env!("CARGO_PKG_VERSION");
+        let mut map = json!({
+            "telemetry_sdk_build": version,
```
mdboom: Maybe add a bug that we'll have to coordinate this with the current `a-c` implementation so they are monotonic -- or add another field so we can distinguish the two?

PR #47, opened 2019-05-10
WIP: Use a counter for the sequence number per ping

badboy:
This mis-uses our test-only API right now to get back the value and
encodes the ping name into the key.

NOT READY TO BE MERGED!


badboy:
Superseded by #120

PR #51, opened 2019-05-13
Initialize objects on creation

### LINKED ISSUE #29: Don't use `Option` in main `Glean` object members
We should set those values in `new`, not `initialize` so we don't have to use `Option`.

badboy:
Fixes #29

This is in preparation of the [error handling](https://bugzilla.mozilla.org/show_bug.cgi?id=1551141) so that I don't have to touch these parts again.

### DIFF
--- a/glean-core/examples/sample.rs
+++ b/glean-core/examples/sample.rs
@@ -18,9 +18,7 @@ fn main() {
         root.path().display().to_string()
     };
 
-    let mut glean = Glean::new();
-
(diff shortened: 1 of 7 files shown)


mdboom:
Related, but somewhat orthogonal to this -- we need to do checks for `handle == 0L` from all public API entry points, so they don't end up throwing exceptions.  It's totally possible to have a Kotlin Glean object without a handle to a Rust Glean object in this model (but it was before this PR, too, obviously).

PR #52, opened 2019-05-13
Generically handle error

badboy:
(Based on #51, that needs to be merged first)

This is a first try at removing most `.unwrap`s in the code.
There are some left, mostly in database operations, which should be done in a followup.

The way it works is it's always passing errors up until the FFI layer, where errors are logged and a potential default value for the FFI type is returned (e.g. a null pointer for pointers or 0 for the handle).
The Kotlin side needs to check these values (only relevant for glean object creation and metric constructor)

(This also includes a couple of semi-related commits to update the generated C header and adding the license header where it was missing, hope that's not too distracting)


glean-core/ffi/src/handlemap_ext.rs
```diff
@@ -0,0 +1,114 @@
...
+//! Glean does not pass errors through the FFI component upwards.
+//! Most errors are not propagated and simply only logged for debugging.
+//! Errors should be recoverable and the platform-side should mostly ignore them.
+//! Additionally the platform-side can work asynchronously anyway, leaving us with no good way to
+//! pass back errors.
+//!
+//! The `HandleMapExtension` extension traits adds additional methods that log potential errors.
+//!
+//! **Note: the platform-side still needs to check for null pointers or other default values before
+//! using returned values. This is only relevant for creation of the main object and metrics.
```
Dexterp37: nit: is it worth mentioning that this is true because the handles of these objects are stored in `ConcurrentHandleMap`(s)?

Makefile
```diff
@@ -27,3 +27,7 @@ test:
 	RUST_TEST_THREADS=1 cargo test --all
 	./gradlew test
 .PHONY: test
+
+cbindgen:
```
Dexterp37: Why is this change needed? What does this do?

Dexterp37: I mean, does it serve a specific purpose other than updating header files?

glean-core/ffi/src/handlemap_ext.rs
badboy: It's more because of that's nearly the only part where we actually return something potentially fallible.

Dexterp37: I see, please call that out in the comment there, so that it's clearer (and so that next time I can avoid asking you the same question again! :D )

glean-core/src/error.rs
```diff
@@ -0,0 +1,109 @@
...
+/// exhaustively match against it.
+///
+/// It is used with the [`Error`] struct.
+///
+/// [`Error`]: std.struct.Error.html
+#[derive(Debug, Fail)]
+pub enum ErrorKind {
+    /// Any error not part of this list.
+    #[fail(display = "Generic error.")]
+    Other,
```
mdboom: Does this associate a message with the "Other" errors?  I think we should probably require a message for any misc. error something might return.

badboy: Actually, we currently don't use the `Other` variant at all. We only handle rkv and io errors.
I'll remove that variant for now.

glean-core/ffi/src/handlemap_ext.rs
badboy: Will do!

glean-core/src/error.rs
mdboom: Makes sense.  We can always add something back in if we realize we need our own error type.

PR #57, opened 2019-05-14
Turn stateless PingMaker into free functions

### LINKED ISSUE #31: Make `PingMaker` into a module of free functions
It doesn't have any state, so it can probably doesn't need to be a struct.

badboy:
Fixes #31

### DIFF
--- a/glean-core/examples/sample.rs
+++ b/glean-core/examples/sample.rs
@@ -1,7 +1,7 @@
 use std::env;
 
 use glean_core::metrics::*;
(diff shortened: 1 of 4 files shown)


Dexterp37:
@badboy we might want to put some state here, due to the `start_time` / `end_time`. Maybe we should keep this on hold until we figure out the other part.

badboy:
That's ... a good point. I'll rework the other PR without this then.

PR #58, opened 2019-05-14
Dont send pings if empty

badboy:
Based on #57

### DIFF
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
@@ -8,24 +8,26 @@ import android.content.Context
 import androidx.test.core.app.ApplicationProvider
 import org.junit.Assert.assertEquals
 import org.junit.Test
+import org.junit.Before
 import org.junit.runner.RunWith
 import org.robolectric.RobolectricTestRunner
 import java.io.File
 
 @RunWith(RobolectricTestRunner::class)
 class GleanTest {
-    @Test
-    fun `simple smoke test`() {
-        GleanInternalAPI().initialize(ApplicationProvider.getApplicationContext())
+
+    @Before
+    fun setUp() {
+        resetGlean()
     }
 
     @Test
     fun `send a ping`() {
         val context: Context = ApplicationProvider.getApplicationContext()
-        val glean = GleanInternalAPI()
-        glean.initialize(context)
-        glean.handleBackgroundEvent()
         val pingPath = File(context.applicationInfo.dataDir, "glean_data/pings")
-        assertEquals(2, pingPath.listFiles()?.size)
+
+        Glean.handleBackgroundEvent()
+        // Only the baseline ping should have been written
+        assertEquals(1, pingPath.listFiles()?.size)
     }
 }
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -149,3 +149,20 @@ impl Glean {
         Ok(())
     }
 }
+
+#[cfg(test)]
+mod test {
+    use super::*;
+
+    #[test]
+    fn path_is_constructed_from_data() {
+        let t = tempfile::tempdir().unwrap();
+        let name = t.path().display().to_string();
+        let glean = Glean::new(&name, "org.mozilla.glean").unwrap();
+
+        assert_eq!(
+            "/submit/org-mozilla-glean/baseline/1/this-is-a-docid",
+            glean.make_path("baseline", "this-is-a-docid")
+        );
+    }
+}
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -105,7 +105,9 @@ impl Glean {
     }
 
     pub fn snapshot(&mut self, store_name: &str, clear_store: bool) -> String {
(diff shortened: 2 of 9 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -24,9 +24,13 @@ class GleanTest {
...
         glean.initialize(context)
+
         glean.handleBackgroundEvent()
+
         val pingPath = File(context.applicationInfo.dataDir, "glean_data/pings")
         val path = pingPath.toString()
-        assertEquals(2, pingPath.listFiles()?.size)
+
+        // All pings are empty, thus none should be stored
+        assertNull(pingPath.listFiles())
```
Dexterp37: The `baseline` ping should not be empty: why aren't we getting one here?

badboy: Hah, that is true, we should at least have the `glean.baseline.locale`.

Now this is what happens:

* We directly create an object of GleanInternalAPI
* Internally that initializes core metrics
	* This uses the created metrics (`GleanBaseline.locale.set(...)`)
	* `StringMetricType.set` uses `Glean.handle`, that is: the handle of the singleton (which in this case is ... `0`)

So before we change this and to still keep it testable we might need to pass/store the handle somehow into the metrics (or pass it on setting)

mdboom: Ah, I guess we need to test using the Kotlin-side `Glean` singleton here, rather than making a new one...

badboy: That breaks down now that we have lots of parallel tests, all of them using the singleton object.

mdboom: But we aren't parallel on the Kotlin side, right?  (At least `glean-ac` isn't -- I don't know if we had to set it from the non-default).  I think it's reasonable on the Kotlin side to have tests with global state and use the `resetGlean` command to reset before each test, etc.

badboy: You are right.
We're not parallel, just same-process, therefore re-using the (already initialized) singleton.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -22,10 +22,11 @@ class GleanTest {
     @Test
     fun `send a ping`() {
         val context: Context = ApplicationProvider.getApplicationContext()
-        val glean = GleanInternalAPI()
-        glean.initialize(context)
-        glean.handleBackgroundEvent()
+        Glean.testDestroyGleanHandle()
```
Dexterp37: Instead of doing this and the line below, let's just call the shared `resetGlean`. You can do so by adding this block of code at the beginning of the class, before any test definition:

```
    @Before
    fun setUp() {
        resetGlean()
    }
```

Note: I don't think we need the simple one line "smoke test" anymore

glean-core/src/ping/mod.rs
```diff
@@ -52,40 +52,46 @@ impl PingMaker {
...
         json!(map)
     }
 
-    pub fn collect(&self, storage: &Database, storage_name: &str) -> JsonValue {
+    pub fn collect(&self, storage: &Database, storage_name: &str) -> Option<JsonValue> {
         info!("Collecting {}", storage_name);
 
-        let metrics_data = StorageManager.snapshot_as_json(storage, storage_name, true);
+        let metrics_data = match StorageManager.snapshot_as_json(storage, storage_name, true) {
+            None => return None,
```
Dexterp37: nit: can we also log that we're baling out because the ping is empty?

glean-core/src/lib.rs
```diff
@@ -102,7 +102,9 @@ impl Glean {
     }
 
     pub fn snapshot(&mut self, store_name: &str, clear_store: bool) -> String {
-        StorageManager.snapshot(&self.storage(), store_name, clear_store)
+        StorageManager
```
Dexterp37: Can we add unit test coverage to make sure that `snapshot` behaves as expected?

glean-core/src/lib.rs
```diff
@@ -127,8 +129,17 @@ impl Glean {
         let ping_maker = PingMaker::new();
         let doc_id = Uuid::new_v4().to_string();
         let url_path = self.make_path(ping_name, &doc_id);
-        let ping_content =
-            ::serde_json::to_string_pretty(&ping_maker.collect(self.storage(), ping_name)).unwrap();
+        let ping_content = match ping_maker.collect_string(self.storage(), ping_name) {
```
Dexterp37: Can we add unit test coverage to make sure that `make_path` behaves as expected?

glean-core/src/lib.rs
badboy: We have a couple of tests using the snapshot function in tests/simple.rs.
We need to improve our testing strategy internally anyway. Ok if I take this in a followup?

Dexterp37: Ok!

PR #59, opened 2019-05-14
1551258: Fix Kotlin linting and enable on CI

mdboom:

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -1,0 +1,4 @@
+style:
+  # We want to allow "TODO" in comments
+  ForbiddenComment:
+    active: false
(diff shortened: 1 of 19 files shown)


.detekt.yml
```diff
@@ -0,0 +1,4 @@
+style:
+  # We want to allow "TODO" in comments
```
Dexterp37: This is ok for now, but depending on how we want to approach the integration we might need to settle with whatever [a-c does](https://github.com/mozilla-mobile/android-components/blob/5c56606626d542d282ec2619e75b0d5e2cd64486/config/detekt.yml), otherwise we'll break klint there. Not for this PR, just for future reference.

glean-core/android/src/main/AndroidManifest.xml
```diff
@@ -1,2 +1,2 @@
 <manifest xmlns:android="http://schemas.android.com/apk/res/android"
-    package="org.mozilla.glean_rs" />
+    package="org.mozilla.gleancore" />
```
badboy: Oh noes, that makes it official.

PR #64, opened 2019-05-15
Bug 1551863 - Pass missing disabled flag to Rust side

badboy:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
```diff
@@ -27,7 +27,8 @@ class BooleanMetricType(
                 name = name,
                 send_in_pings = ffiPingsList,
                 send_in_pings_len = sendInPings.size,
-                lifetime = lifetime.ordinal)
+                lifetime = lifetime.ordinal,
+                disabled = if (disabled) { 1 } else { 0 })
```
Dexterp37: What if we write an extension function for this, in `LibGleanFFI.kt`? We could do something like:

```
// Comment about what this does and why we need it
internal fun Boolean.toInt() = if (this) 1 else 0
```

This way the code above can be reduced to:

`disabled = disabled.toInt()`

glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt
```diff
@@ -43,7 +43,8 @@ class CounterMetricType(
                 name = name,
                 send_in_pings = ffiPingsList,
                 send_in_pings_len = sendInPings.size,
-                lifetime = lifetime.ordinal)
+                lifetime = lifetime.ordinal,
+                disabled = if (disabled) { 1 } else { 0 })
```
Dexterp37: nit: same consideration as above, here and in all the other places below.

Dexterp37:
@badboy do you want to take enabling back the tests to a separate PR/bug?

badboy:
> @badboy do you want to take enabling back the tests to a separate PR/bug?

Nope, let me actually re-enable them here

PR #65, opened 2019-05-15
Bug 1551892: Implement the testing API for the StringMetricType and enable the related tests

badboy:


glean-core/android/src/test/java/mozilla/telemetry/glean/private/StringMetricTypeTest.kt
```diff
@@ -58,27 +56,6 @@ class StringMetricTypeTest {
...
-            name = "stringMetric",
-            sendInPings = listOf("store1")
-        )
-
-        // Attempt to store the string.
-        stringMetric.set("value")
-        // Check that nothing was recorded.
-        assertFalse("Strings must not be recorded if they have no lifetime",
-            stringMetric.testHasValue())
-    }
```
mdboom: Why is this test being deleted?  If it's not working yet (because `disabled` isn't implemented) maybe just update the `@Ignore` message?

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -15,6 +15,28 @@ import java.lang.reflect.Proxy
...
+    }
+}
+
+/**
+ * Helper to read a null terminated string out of the pointer.
+ *
+ * Important: doesn't free the pointer, use [getAndConsumeRustString] for that!
+ */
+internal fun Pointer.getRustString(): String {
+    return this.getString(0, "utf8")
```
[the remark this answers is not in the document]
badboy: Correct. The pointer on the Rust side is allocated by Rust and needs to get freed there.
The Kotlin string is constructed by copying the nul-terminated data out.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/StringMetricTypeTest.kt
badboy: See commit message. There's no "no lifetime" thing and the test has the same contents as the following test (except its name)

mdboom: Ah -- missed that.  LGTM.

badboy:
> Somewhat orthogonal: do we want to add some Rust-side test coverage for the string metric while we're in here?

[We have one](https://github.com/mozilla/glean/blob/2d6995a03d9c682e8c6ab1c45b58506cd150c4dc/glean-core/tests/metrics.rs#L22-L34).
We should add more in a follow-up.

PR #66, opened 2019-05-15
Update pull_request_template.md

mdboom:
This makes it more language-agnostic.

I like using the `Makefile` for these conveniences, and I kind of like maintaining the Makefile indefinitely as a source of developer conveniences (as long is it doesn't become the "real" build tool).  But I don't know how well it works on Windows.

### DIFF
--- a/.github/pull_request_template.md
+++ b/.github/pull_request_template.md
@@ -1,8 +1,8 @@
 ### Pull Request checklist ###
 <!-- Before submitting the PR, please address each item -->
 - [ ] **Quality**: This PR builds and tests run cleanly
-  - `cargo clean; cargo test --all` runs without emitting any warnings
-  - `cargo fmt` does not produce any changes to the code
+  - `make test` runs without emitting any warnings
+  - `make lint` runs without emitting any errors
 - [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
 - [ ] **Changelog**: This PR includes a changelog entry or an explanation of why it does not need one
-  - Any breaking changes to Swift or Kotlin binding APIs are noted explicitly
+  - Any breaking changes to language binding APIs are noted explicitly


.github/pull_request_template.md
```diff
@@ -1,8 +1,8 @@
 ### Pull Request checklist ###
 <!-- Before submitting the PR, please address each item -->
 - [ ] **Quality**: This PR builds and tests run cleanly
-  - `cargo clean; cargo test --all` runs without emitting any warnings
-  - `cargo fmt` does not produce any changes to the code
+  - `make test` runs without emitting any warnings
+  - `make lint` runs without emitting any errors
```
badboy: We should make lint depend on fmt, so that we also get that:

```
lint: fmt
    ...
```

.github/pull_request_template.md
```diff
@@ -1,8 +1,8 @@
 ### Pull Request checklist ###
 <!-- Before submitting the PR, please address each item -->
 - [ ] **Quality**: This PR builds and tests run cleanly
-  - `cargo clean; cargo test --all` runs without emitting any warnings
-  - `cargo fmt` does not produce any changes to the code
+  - `make test` runs without emitting any warnings
```
Dexterp37: `make` is not generally available on Windows, or, at least, it's not generally added to the `PATH`, so this won't work. I don't think it's too much of a big deal.

.github/pull_request_template.md
badboy: https://github.com/mozilla/glean/pull/71

.github/pull_request_template.md
mdboom: Yeah -- I've seen other projects include a `.bat` file for this kind of thing, but then we'd have two things to maintain which is also not ideal.

I'm fine with leaving as-is for now until we have more Windows contributors ;)

PR #72, opened 2019-05-16
1552122: Initial port (with some stubbing) of Glean core API

mdboom:
This moves over the public API on the Glean object.  I have also reorganized the other of the methods in `Glean.kt` so it matches glean-ac (which should make it easier to do line-by-line comparisons).

Again, the first commit is copying some files from glean-ac over verbatim.  The second commit then modifies them for our context.


glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
```diff
@@ -0,0 +1,45 @@
...
+import mozilla.telemetry.glean.Glean
+
+/**
+ * This implements the developer facing API for custom pings.
+ *
+ * Instances of this class type are automatically generated by the parsers at build time.
+ *
+ * The Ping API only exposes the [send] method, which schedules a ping for sending.
+ */
+data class PingType(
```
badboy: We should eventually talk about this and decide if this and the Ping registry stay on the kotlin side or move into Rust (-> followup)

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -0,0 +1,44 @@
...
+ */
+internal class GleanLifecycleObserver : LifecycleObserver {
+    /**
+     * Calls the "background" event when entering the background.
+     */
+    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
+    fun onEnterBackground() {
+        // We're going to background, so store how much time we spent
+        // on foreground.
+        GleanBaseline.duration.stopAndSum(this)
```
badboy: Having all this already in here means it will just start working once we implement what's currently stubbed out /o\ I think that's really cool

glean-core/src/ping/mod.rs
```diff
@@ -93,8 +93,10 @@ impl PingMaker {
     }
 
     pub fn collect_string(&self, storage: &Database, storage_name: &str) -> Option<String> {
+        // Use ::to_string, rather than ::to_string_pretty since the ping
+        // uploader relies on the JSON content being on a single line.
```
badboy: Ah, good to know. Well, `pretty` was so far mostly so that we see it nicely in the log output anyway.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -75,6 +107,102 @@ open class GleanInternalAPI internal constructor () {
...
+     * When uploading is disabled, metrics aren't recorded at all and no data
+     * is uploaded.
+     *
+     * When disabling, all pending metrics, events and queued pings are cleared.
+     *
+     * When enabling, the core Glean metrics are recreated.
+     *
+     * @param enabled When true, enable metric collection.
+     */
+    fun setUploadEnabled(enabled: Boolean) {
```
badboy: Do we already have a bug for correctly implementing the logic here?

Dexterp37: Are you referring to a specific logic currently available in glean-ac? Right now, we don't do much more than that.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -75,6 +107,102 @@ open class GleanInternalAPI internal constructor () {
...
+     * @param branch The experiment branch (maximum 30 bytes)
+     * @param extra Optional metadata to output with the ping
+     */
+    fun setExperimentActive(
+        experimentId: String,
+        branch: String,
+        extra: Map<String, String>? = null
+    ) {
+        Log.e(LOG_TAG, "setExperimentActive is a stub")
+        // TODO: stub
```
Dexterp37: Let's please file a bug about this and add the bug number here.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -75,6 +107,102 @@ open class GleanInternalAPI internal constructor () {
...
+    }
+
+    /**
+     * Indicate that an experiment is no longer running.
+     *
+     * @param experimentId The id of the experiment to deactivate.
+     */
+    fun setExperimentInactive(experimentId: String) {
+        Log.e(LOG_TAG, "setExperimentInactive is a stub")
+        // TODO: stub
```
Dexterp37: Let's please file a bug about this and add the bug number here (I think this can be the same bug as above).

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -75,6 +107,102 @@ open class GleanInternalAPI internal constructor () {
...
+    /**
+     * Tests whether an experiment is active, for testing purposes only.
+     *
+     * @param experimentId the id of the experiment to look for.
+     * @return true if the experiment is active and reported in pings, otherwise false
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun testIsExperimentActive(experimentId: String): Boolean {
+        Log.e(LOG_TAG, "testIsExperimentActive is a stub")
+        // TODO: stub
```
Dexterp37: Let's file a bug about this, you know the deal :) Can we please also assert as we do in other unimplemented test API functions?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -75,6 +107,102 @@ open class GleanInternalAPI internal constructor () {
...
+    /**
+    * Returns the stored data for the requested active experiment, for testing purposes only.
+    *
+    * @param experimentId the id of the experiment to look for.
+    * @return the [RecordedExperimentData] for the experiment
+    * @throws [NullPointerException] if the requested experiment is not active
+    */
+    /*
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun testGetExperimentData(experimentId: String): RecordedExperimentData {
```
Dexterp37: I think we can uncomment the function and simply return a bogus `RecordedExperimentData`. Plus add the bug number TODO, log and assert.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -166,19 +260,83 @@ open class GleanInternalAPI internal constructor () {
...
+    private fun sendPing(pingName: String): Boolean {
+        return LibGleanFFI.INSTANCE.glean_send_ping(
             handle,
             pingName,
             (configuration.logPings).toByte()
-        )
-        if (queued != 0.toByte()) {
-            PingUploadWorker.enqueueWorker()
-        }
+        ) != 0.toByte()
```
Dexterp37: I think we should really introduce some extension function for this: let's do something like [this](https://github.com/mozilla/glean/blob/110e44584de68a960e92542903b0c3e093f3eaa4/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt#L16)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Weren't there changes how we handle the flipping of the flag, e.g. deleting stored data?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: In another PR we also said that to take in a follow-up. I think it would be fine to take that in one swoop

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: Hasn't landed yet :)

badboy: I see. Let's keep it in mind then for when it lands

glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
mdboom: Yes, definitely as a follow-up.  I think it will ultimately have to live on the Rust side, since the Rust code will need to be able to send pings on its own (in response to persisted data for example) and will need to know things like whether to send the `client_id`.  I just consider this a stub to help more tests pass for now.

glean-core/src/ping/mod.rs
mdboom: That's a good point.  We may want to use pretty for logging only...

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: Created [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1552474)

glean-core/src/ping/mod.rs
mdboom: @badboy: I've updated this so:
1) Serialization of the JSON happens in `storePing` so there's a single place that is responsible for *exactly* how the ping is serialized.
2) When logging the ping, we still use `pretty`.  (Yes, that means we serialize it twice, but logging pings is a QA-only sort of thing -- that's better than glean-ac where we serialize->deserialize->pretty, I guess)

This meant I made a bigger change and added a new Error type and stuff, so probably worth another quick review of the last commit here.

glean-core/src/error.rs
```diff
@@ -32,12 +34,16 @@ pub enum ErrorKind {
     Handle(HandleError),
 
     /// IO error
-    #[fail(display = "An I/O error occured.")]
+    #[fail(display = "An I/O error occurred.")]
```
badboy: Oops. :D

glean-core/src/error.rs
```diff
@@ -115,3 +121,11 @@ impl From<Error> for ExternError {
         ffi_support::ExternError::new_error(ffi_support::ErrorCode::new(42), format!("{}", error))
     }
 }
+
+impl From<serde_json::error::Error> for Error {
```
badboy: Fun fact: we should never even run into that as we're building our content from JsonValue and kinda guarantee that it maps to valid json. Anyway, the code requires it.

glean-core/src/ping/mod.rs
```diff
@@ -124,7 +124,7 @@ impl PingMaker {
             let mut file = File::create(&temp_ping_path)?;
             file.write_all(url_path.as_bytes())?;
             file.write_all(b"\n")?;
-            file.write_all(ping_content.as_bytes())?;
+            file.write_all(::serde_json::to_string(ping_content)?.as_bytes())?;
```
badboy: Wait, that works with `std::io::Result` here? I'm confused, I would have though we need to switch to our own error type here, but if compile & tests pass it's ok (I'll take a look later to see what's going on)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Whatever way around we land the PRs, this will need to be adopted then.

glean-core/src/error.rs
```diff
@@ -8,6 +8,8 @@ use ffi_support::{handle_map::HandleError, ExternError};
 
 use rkv::error::StoreError;
 
+use serde_json;
```
badboy: Is this import necessary? Paths below should in theory work without it.

glean-core/src/ping/mod.rs
badboy: Is `collect_string` anywhere still used? If not, should we just remove it?

glean-core/src/error.rs
mdboom: Good point.  Seems to be unnecessary.

glean-core/src/error.rs
mdboom: Yeah -- theoretically `to_string` returns one, but it seems unlikely that serialization would fail.  Anyway, this lets us at least not have an `unwrap` hanging there looking suspicious.

mdboom:
`collect_string` is still used from some sample code, but isn't really used internally anymore.  I'll go ahead and update the samples and remove it.

mdboom:
On second thought, `collect_string` is kind of useful for the C example, even though it's no longer used internally.  I think we can just defer this and I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1552479) for once we have better ways of testing from C.

glean-core/src/ping/mod.rs
mdboom: Oh, yeah, I'm confused, too, come to think of it.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -76,6 +109,104 @@ open class GleanInternalAPI internal constructor () {
         return initialized.toInt() != 0
```
badboy: Oh noes, this is a leftover and could be converted to `.toBoolean()` now

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -76,6 +109,104 @@ open class GleanInternalAPI internal constructor () {
...
+     */
+    fun setUploadEnabled(enabled: Boolean) {
+        LibGleanFFI.INSTANCE.glean_set_upload_enabled(handle, enabled.toByte())
+    }
+
+    /**
+     * Get whether or not Glean is allowed to record and upload data.
+     */
+    fun getUploadEnabled(): Boolean {
+        return LibGleanFFI.INSTANCE.glean_is_upload_enabled(handle) != 0.toByte()
```
badboy: `.toBoolean()`

glean-core/src/ping/mod.rs
badboy: This is why: https://docs.rs/serde_json/1.0.39/src/serde_json/error.rs.html#134

mdboom:
> This is why: https://docs.rs/serde_json/1.0.39/src/serde_json/error.rs.html#134

Ah, got it.  It's only `to_string_pretty` that returns a JSON error...

PR #74, opened 2019-05-17
Document how to add a new metric type in the code

badboy:

### DIFF
--- a/docs/dev/core/new-metric-type.md
+++ b/docs/dev/core/new-metric-type.md
@@ -1,6 +1,6 @@
 # Adding a new metric type
 
-Data in Glean is stored in so called metrics.
+Data in Glean is stored in so-called metrics.
 You can find the full list of implemented metric types [in the user overview](/user/metrics/index.md).
 
 Adding a new metric type involves defining the metric type's API, its persisted and in-memory storage as well as its serialization into the ping payload.
@@ -165,7 +165,7 @@ pub extern "C" fn glean_counter_add(glean_handle: u64, metric_id: u64, amount: i
 
 ### Platform-part (Kotlin)
 
-The platform-specific FFI wrapper needs the definitons of these new functions.
+The platform-specific FFI wrapper needs the definitions of these new functions.
 For Kotlin this is in `glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt`:
 
 ```kotlin
--- a/docs/dev/core/new-metric-type.md
+++ b/docs/dev/core/new-metric-type.md
@@ -9,7 +9,7 @@ Adding a new metric type involves defining the metric type's API, its persisted 
 
 A metric type implementation is defined in its own file under `glean-core/src/metrics/`, e.g. `glean-core/src/metrics/counter.rs` for a [Counter](/user/metrics/counter.md).
 
-Start by defining a structure:
+Start by defining a structure to hold the metric's metadata:
 
 ```rust,noplaypen
 #[derive(Debug)]
--- a/docs/dev/core/new-metric-type.md
+++ b/docs/dev/core/new-metric-type.md
@@ -130,14 +130,7 @@ pub extern "C" fn glean_new_counter_metric(
 ) -> u64 {
     COUNTER_METRICS.insert_with_log(|| {
         let send_in_pings = unsafe { from_raw_string_array(send_in_pings, send_in_pings_len) };
-        let lifetime = Lifetime::try_from(lifetime)
-            .map_err(|_| {
-                log::error!(
-                    "[Counter] Failed to convert from lifetime value {}",
-                    lifetime
-                );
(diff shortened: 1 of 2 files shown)


docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,212 @@
...
+    COUNTER_METRICS.insert_with_log(|| {
+        let send_in_pings = unsafe { from_raw_string_array(send_in_pings, send_in_pings_len) };
+        let lifetime = Lifetime::try_from(lifetime)
+            .map_err(|_| {
+                log::error!(
+                    "[Counter] Failed to convert from lifetime value {}",
+                    lifetime
+                );
+            })
+            .unwrap_or(Lifetime::Ping);
```
mdboom: An aside (not for this PR): Is there a way to encapsulate this to not duplicate it in every metric type constructor?

docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,212 @@
...
+        COUNTER_METRICS.call_infallible(metric_id, |metric| {
+            metric.add(glean, amount);
+        })
+    })
+}
+```
+
+### Platform-part (Kotlin)
+
+The platform-specific FFI wrapper needs the definitons of these new functions.
```
mdboom:
```suggestion
The platform-specific FFI wrapper needs the definitions of these new functions.
```

docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,212 @@
...
+        match self {
+            // ...
+            Metric::Counter(c) => json!(c),
+        }
+    }
+}
+```
+
+For more complex serialization consider implementing serialization logic as a function returning a [`serde_json::Value`](https://docs.rs/serde_json/*/serde_json/enum.Value.html)
+or another object that can be serialized.
```
mdboom: I feel like we'll want a concrete example of this, but we can defer that until we actually have one in the code base.  In particular, documenting the convention that the storage struct (as distinct from the metric definition struct) lives in the same module as the metric type etc., etc...

docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,212 @@
+# Adding a new metric type
+
+Data in Glean is stored in so called metrics.
```
mdboom:
```suggestion
Data in Glean is stored in so-called metrics.
```

docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,212 @@
...
+You can find the full list of implemented metric types [in the user overview](/user/metrics/index.md).
+
+Adding a new metric type involves defining the metric type's API, its persisted and in-memory storage as well as its serialization into the ping payload.
+
+## The metric type's API
+
+A metric type implementation is defined in its own file under `glean-core/src/metrics/`, e.g. `glean-core/src/metrics/counter.rs` for a [Counter](/user/metrics/counter.md).
+
+Start by defining a structure:
```
mdboom: ... to hold the metric's metadata.

(Or something, to make it clear that this structure is holding the metric definition, not the data for each recording of the metric)...

docs/dev/core/new-metric-type.md
badboy: Yeah, I was thinking about that but then I was half-way into implementing Datetime and stopped.
I'll file a bug so we replace this once we have a more concrete example.

docs/dev/core/new-metric-type.md
badboy: Yes. Actually we can use `from` and default to `Lifetime::Ping`. I'll change it here and add another PR changing the code.

docs/dev/core/new-metric-type.md
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1552774

docs/dev/core/new-metric-type.md
```diff
@@ -0,0 +1,205 @@
...
+    category: FfiStr,
+    name: FfiStr,
+    send_in_pings: RawStringArray,
+    send_in_pings_len: i32,
+    lifetime: i32,
+    disabled: u8,
+) -> u64 {
+    COUNTER_METRICS.insert_with_log(|| {
+        let send_in_pings = unsafe { from_raw_string_array(send_in_pings, send_in_pings_len) };
+        let lifetime = Lifetime::try_from(lifetime)?;
```
Dexterp37: Looks much nicer :)

badboy: I agree.

docs/dev/core/new-metric-type.md
mdboom: Yeah, sounds good.  I think we don't fully know all the details on this until we implement the first one anyway...

PR #75, opened 2019-05-17
1552136: Convert CounterStorageEngineTests.kt to Rust

mdboom:

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -12,7 +12,7 @@ emulator:
 .PHONY: install
 
 lint: fmt
-	cargo clippy --all
+	cargo clippy --all --all-targets --all-features -- -D warnings
 	./gradlew ktlint detekt
 .PHONY: lint
 
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -1,0 +1,11 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+pub fn tempdir() -> (tempfile::TempDir, String) {
+    let t = tempfile::tempdir().unwrap();
+    let name = t.path().display().to_string();
+    (t, name)
+}
+
+pub const GLOBAL_APPLICATION_ID: &str = "org.mozilla.glean.test.app";
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -4,6 +4,10 @@
 
 use glean_core::Glean;
 
+// #[allow(dead_code)] is required on functions in this module as a workaround
+// for https://github.com/rust-lang/rust/issues/46379
+
+#[allow(dead_code)]
 pub fn tempdir() -> (tempfile::TempDir, String) {
     let t = tempfile::tempdir().unwrap();
     let name = t.path().display().to_string();
@@ -14,6 +18,7 @@ pub const GLOBAL_APPLICATION_ID: &str = "org.mozilla.glean.test.app";
 
 // Create a new instance of Glean with a temporary directory.
 // We need to keep the `TempDir` alive, so that it's not deleted before we stop using it.
+#[allow(dead_code)]
 pub fn new_glean() -> (Glean, tempfile::TempDir) {
     let dir = tempfile::tempdir().unwrap();
     let tmpname = dir.path().display().to_string();
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use glean_core::Glean;
+
 pub fn tempdir() -> (tempfile::TempDir, String) {
     let t = tempfile::tempdir().unwrap();
     let name = t.path().display().to_string();
@@ -9,3 +11,14 @@ pub fn tempdir() -> (tempfile::TempDir, String) {
 }
 
 pub const GLOBAL_APPLICATION_ID: &str = "org.mozilla.glean.test.app";
+
+// Create a new instance of Glean with a temporary directory.
+// We need to keep the `TempDir` alive, so that it's not deleted before we stop using it.
+pub fn new_glean() -> (Glean, tempfile::TempDir) {
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().display().to_string();
+
+    let glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID).unwrap();
+
+    (glean, dir)
+}
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -2,12 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// #[allow(dead_code)] is required on this module as a workaround for
+// https://github.com/rust-lang/rust/issues/46379
+#![allow(dead_code)]
 use glean_core::Glean;
 
-// #[allow(dead_code)] is required on functions in this module as a workaround
-// for https://github.com/rust-lang/rust/issues/46379
-
-#[allow(dead_code)]
 pub fn tempdir() -> (tempfile::TempDir, String) {
     let t = tempfile::tempdir().unwrap();
     let name = t.path().display().to_string();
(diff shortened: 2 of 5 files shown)


glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+            ..Default::default()
+        });
+
+        metric.add(&glean, 1);
+
+        let snapshot = StorageManager
+            .snapshot_as_json(glean.storage(), "store1".into(), true)
+            .unwrap();
+        assert_eq!(
+            "{\"counter\":{\"telemetry.counter_metric\":1}}",
```
badboy: Instead of constructing a string we should be able to build the JSON value like:

```
json!({"counter": {"telemetry.counter_metric": 1}})
```

and compare it to  `snapshot`.

glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+    {
+        let glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID).unwrap();
+
+        let metric = CounterMetric::new(CommonMetricData {
+            name: "counter_metric".into(),
+            category: "telemetry".into(),
+            send_in_pings: vec!["store1".into()],
+            disabled: false,
+            lifetime: Lifetime::User,
+            ..Default::default()
```
badboy: The default here is unneeded, we're filling all fields now.

glean-core/tests/common/mod.rs
```diff
@@ -0,0 +1,11 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+pub fn tempdir() -> (tempfile::TempDir, String) {
+    let t = tempfile::tempdir().unwrap();
+    let name = t.path().display().to_string();
+    (t, name)
+}
+
+pub const GLOBAL_APPLICATION_ID: &str = "org.mozilla.glean.test.app";
```
badboy: Creating the `glean` object is the same across all tests.
In `tests/metrics.rs` we have;

```rust
fn new_glean() -> (Glean, TempDir) {
    let dir = tempfile::tempdir().unwrap();
    let tmpname = dir.path().display().to_string();

    let glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID).unwrap();

    (glean, dir)
}
```

Can we use that everywhere instead of just the `tempdir` function?

glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+        assert_eq!(
+            "{\"counter\":{\"telemetry.counter_metric\":1}}",
+            ::serde_json::to_string(&snapshot).unwrap(),
+        );
+    }
+
+    // Make a new glean instance here, which should force reloading of the data from disk
+    // so we can ensure it persisted, because it has User lifetime
+    {
+        let glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID).unwrap();
```
badboy: Ah, we need a new glean object (re `new_glean` from above).
Maybe we expose both things and in this test create the Glean objects ourselves, but otherwise use `new_glean` taking care of that?

glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+
+    metric.add(&glean, 1);
+
+    for store_name in store_names {
+        let snapshot = StorageManager
+            .snapshot_as_json(glean.storage(), &store_name, true)
+            .unwrap();
+
+        assert_eq!(1, snapshot.as_object().unwrap().len());
+        println!("{:?}", snapshot);
```
badboy: Stray `println` in test code.

glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+        assert_eq!(1, snapshot["counter"]["telemetry.counter_metric"]);
```
badboy: oooooh, that ... works.

We should use the same style everywhere though. I think building the JSON object as above (`json!(...)`) would be sensible.

glean-core/tests/counter.rs
```diff
@@ -0,0 +1,152 @@
...
+    let snapshot = StorageManager.snapshot(glean.storage(), "store1", true);
+    assert!(snapshot.is_some());
+    // Check that getting a new snapshot for "store1" returns an empty store.
+    assert!(StorageManager
+        .snapshot(glean.storage(), "store1", false)
+        .is_none());
+    // Check that we get the right data from both the stores. Clearing "store1" must
+    // not clear "store2" as well.
+    let snapshot2 = StorageManager.snapshot(glean.storage(), "store2", true);
+    assert!(snapshot.unwrap().len() != 0);
```
badboy: Instead of checking the length, we can just check that it `is_some`, just as above.
Also checking `snapshot` again seems weird, that check was done in line 111.

glean-core/tests/counter.rs
mdboom: In this case I actually think it's valuable to test against a string -- we want to make sure we're serializing with no unnecessary whitespace and in a single line etc.  Could move that out into its own test instead, of course...

glean-core/tests/counter.rs
mdboom: Make sense.

glean-core/tests/common/mod.rs
mdboom: Sure, that's a good idea.  Do you think that belongs in this file or in `common/mod.rs`?

badboy: If we re-use it in mod.rs.

glean-core/tests/counter.rs
badboy: But the string-serialization happens in this test, not in library code. We therefore test only that `serde_json::to_string` kinda does what we expect, but don't test that anything internally actually uses that.

glean-core/tests/counter.rs
mdboom: It's not quite checking the same thing, though.  The function could easily return `Some(Object::new())` which would pass `is_some()`, but fail `unwrap().len() != 0`.  I think it's important to assert that *if* it returns `Some` that `Object` also has at least one entry.

Reordering things makes sense, though.

glean-core/tests/counter.rs
mdboom: Ah, good point. :slightly_frowning_face: Indeed, makes more sense to compare JSON objects then.

glean-core/tests/counter.rs
badboy: Hm.... `snapshot` already has this check though and it serializes a HashMap. Even an empty hashmap is at least serialized as `{}`, therefore defeating your test.

IMO it would be much more sensible to test that we get `None` when we expect it and `Some` to indicate it contains data and to look into the data if needed.

(And at least `len() != 0` should be written as `!...is_empty()`)

mdboom: Got it -- I forgot that `snapshot` is snapshotting as a string.

Makefile
```diff
@@ -12,7 +12,7 @@ emulator:
 .PHONY: install
 
 lint: fmt
-	cargo clippy --all
+	cargo clippy --all --all-targets --all-features -- -D warnings
```
badboy: 🎉

glean-core/tests/common/mod.rs
```diff
@@ -4,6 +4,10 @@
 
 use glean_core::Glean;
 
+// #[allow(dead_code)] is required on functions in this module as a workaround
+// for https://github.com/rust-lang/rust/issues/46379
+
+#[allow(dead_code)]
```
badboy: We should be able to put that globally for this file:

`#![allow(dead_code)]` as the first non-comment line in the file.

glean-core/tests/common/mod.rs
```diff
@@ -0,0 +1,29 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use glean_core::Glean;
+
+// #[allow(dead_code)] is required on functions in this module as a workaround
+// for https://github.com/rust-lang/rust/issues/46379
```
mdboom: @badboy: The issue sort of makes sense given that test modules are compiled as separate crates. I still think the upside of putting this kind of code in a common place outweighs the downsides of having declare everything with #[allow(dead_code)], but if you have any alternative suggestions, let me know.

glean-core/tests/common/mod.rs
```diff
@@ -2,12 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+// #[allow(dead_code)] is required on this module as a workaround for
+// https://github.com/rust-lang/rust/issues/46379
+#[allow(dead_code)]
```
badboy: Missing the bang:

```
help: if you just forgot a `!`, use: `#![allow(dead_code)]`
```

glean-core/tests/common/mod.rs
```diff
@@ -0,0 +1,27 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// #[allow(dead_code)] is required on this module as a workaround for
+// https://github.com/rust-lang/rust/issues/46379
+#[allow(dead_code)]
```
badboy:
```suggestion
#![allow(dead_code)]
```

PR #76, opened 2019-05-20
Default to a ping lifetime during conversion

badboy:

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -4,6 +4,8 @@
 
 use std::convert::TryFrom;
 
+use crate::error::{Error, ErrorKind};
+
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Lifetime {
     /// The metric is reset with each sent ping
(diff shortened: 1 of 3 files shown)


glean-core/src/common_metric_data.rs
```diff
@@ -30,15 +28,13 @@ impl Lifetime {
...
+    fn from(value: i32) -> Lifetime {
         match value {
-            0 => Ok(Lifetime::Ping),
-            1 => Ok(Lifetime::Application),
-            2 => Ok(Lifetime::User),
-            _ => Err(()),
+            0 => Lifetime::Ping,
+            1 => Lifetime::Application,
+            2 => Lifetime::User,
+            _ => Lifetime::Ping,
```
Dexterp37: Mh, what's the rationale behind this? In general, we try not to have "fallback" behaviours in Glean, but rather have clear erroring paths. What are we adjusting here?

badboy: Hm, I see. 
I was trying to simplify the FFI component.
So far we did error report and still fallback to `Lifetime::Ping`.

I can see the use of "log error and bail out" though. Let me try something.

Dexterp37: Cheers! Thanks for 🐻 ing with me :P

badboy:
@Dexterp37 What do you think about this?

PR #77, opened 2019-05-20
Bug 1552511 - Use finalizers to destruct metrics

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1552511

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
@@ -9,6 +9,7 @@ import com.sun.jna.StringArray
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.rust.LibGleanFFI
 import mozilla.telemetry.glean.rust.toByte
+import mozilla.telemetry.glean.rust.RustError
 
 class BooleanMetricType(
     disabled: Boolean,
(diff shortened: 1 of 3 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
```diff
@@ -32,6 +33,13 @@ class BooleanMetricType(
                 disabled = disabled.toByte())
     }
 
+    protected fun finalize() {
```
Dexterp37: @badboy, is this a Rust specific behaviour/trait? When does this get called?

Dexterp37: Oh wait, this is Kotlin code. So it must be what Mike mentioned [here](https://kotlinlang.org/docs/reference/java-interop.html#finalize). But that should be for imported Java object: did you check this works?

badboy: Kinda. I added a print statement and run the tests and checked the stdout log to contain the finalize thing, so yeah, it gets called (and didn't crash).

Of course finalizers are not guaranteed to run (or its at least nondeterministic _when_ they run).

badboy:
> When I filed the bug, I considered it pretty low-priority in practice because these objects are all essentially static lifetime singletons (except in unit tests), but I like the idea of being "correct" here.

Yeah, I spent ~5 minutes after clicking on the link in the bug and considered it worth doing earlier today to start into the day :D

PR #79, opened 2019-05-20
Redirect logging to stdout in tests

Dexterp37:
This enables a logger implementation on platforms that are not Android: we couldn't see any log beyond FFI for this reason, since tests are x86/x64.

This additionally redirects Kotlin logcat to stdout in tests.

### DIFF
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -60,6 +60,9 @@ fn initialize_logging() {
     {
         match env_logger::try_init() {
             Ok(_) => log::debug!("stdout logging should be hooked up!"),
+            // Please note that this is only expected to fail during unit tests,
+            // where the logger might have already been initialized by a previous
+            // test. So it's fine to print with the "logger".
             Err(_) => log::debug!("stdout was already initialized"),
         };
     }
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -40,8 +40,9 @@ unsafe fn from_raw_string_array(arr: RawStringArray, len: i32) -> Vec<String> {
         .collect()
 }
 
-#[no_mangle]
-pub extern "C" fn glean_initialize(data_dir: FfiStr, application_id: FfiStr) -> u64 {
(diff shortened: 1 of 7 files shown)


glean-core/ffi/src/lib.rs
```diff
@@ -52,6 +53,21 @@ pub extern "C" fn glean_initialize(data_dir: FfiStr, application_id: FfiStr) ->
...
         });
     }
+    // Make sure logging does something on non Android platforms as well. Use
+    // the RUST_LOG environment variable to set the desired log level, e.g.
+    // setting RUST_LOG=debug sets the log level to debug.
+    #[cfg(not(target_os = "android"))]
+    {
+        match env_logger::try_init() {
+            Ok(_) => log::debug!("stdout logging should be hooked up!"),
+            Err(_) => log::debug!("stdout was already initialized"),
```
badboy: It's ... not very helpful to log the error that the log setup failed using the same log mechanism.
IMO it's fine to call `env_logger::init()` here, it will panic if that fails, therefore ultimately killing the process.

Dexterp37: This is only supposed to fail in tests, where it will try to init the logger multiple times. If that panics, the test itself will fail. 

When this fails, the logger should still be up and running, given that was previously initialized already.

badboy: Hm, fair enough. Guess that makes sense then. We should just keep it in mind once we add iOs to the mix, we will then need new `#[cfg...]`s anyway

Dexterp37:
> 
> 
> Is all of the following true?
> 
>     * If running the Rust unit tests natively, without android in the mix (`cargo test`), logging is controlled by the `RUST_LOG` env variable.
> 
>     * Logging from Rust inside the Android environment goes through Android logging, and therefore (1) shows up in stderr when running the Android tests (`./gradlew test`), otherwise shows up in logcat.
> 
> 
> Maybe it would be worthwhile documenting this, if so. (I realize this is all making things work how Rust and Android developers would expect, but still might be useful to have _short_ docs on this).

Yes, that is true. And good point about documenting this.

PR #84, opened 2019-05-21
Bug 1552850 - Implement shouldRecord on the Kotlin side

badboy:
Posting this because the code part was easy, but I have some hesitations:

* The `should_record` check _also_ happens on the Rust side
	* We do avoid putting any work in the Dispatcher queue with this.
* The (Rust-side) check is on the metric and requires access to the glean object, therefore we need to re-implement it for every type
	* There might be a way of using traits here to implement it just once (we will need to "re-implement" exposing the common metric data though)
	* I opted for not doing that now (it's a refactoring, not blocking anything and the code duplication so far is simple)
*  We also need to duplicate code on the Kotlin side, because we need a `*_should_record` function per type. Unfortunately I don't see a way to avoid this (again, duplication here contains no complex logic so for now I'm fine with it)

The biggest thing remains the first statement: Do we _need_ the check on the Kotlin side or is Rust-side only actually enough? (Dispatch enqueue cost?) 

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
@@ -63,11 +63,15 @@ internal interface LibGleanFFI : Library {
 
     fun glean_boolean_set(glean_handle: Long, metric_id: Long, value: Byte)
 
+    fun glean_boolean_should_record(glean_handle: Long, metric_id: Long): Byte
+
     fun glean_boolean_test_get_value(glean_handle: Long, metric_id: Long, storage_name: String): Byte
 
     fun glean_boolean_test_has_value(glean_handle: Long, metric_id: Long, storage_name: String): Byte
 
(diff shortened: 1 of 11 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -63,12 +63,16 @@ internal interface LibGleanFFI : Library {
 
     fun glean_boolean_set(glean_handle: Long, metric_id: Long, value: Byte)
 
+    fun glean_boolean_should_record(glean_handle: Long, metric_id: Long): Byte
```
Dexterp37: Instead of increasing the FFI API surface, since this function always returns a `Byte` and has the same arguments, I just wish we had a way to pass in Enums through the FFI :|

badboy: weeeeell, .... not impossible, just right now not worth it.

PR #85, opened 2019-05-21
Move PingType metadata to the Rust side

mdboom:
I'm just posting this here for discussion -- either here or in our next meeting if that's more efficient.  FFI and Kotlin side not implemented yet.

I first considered putting the ping registry on the Glean object, since it sort of logically makes sense there.  However, libraries need to register pings at any time during startup, including before `Glean::initialize` is called.  Since we've tied `new` and `initialize` together, we couldn't add pings to the ping registry before the Glean singleton is initialized, since before that there is no Rust-side Glean object.  An alternative to make that work would be to split `new` and `initialize` apart again, which isn't great.

So I came up with this solution which uses an actual global object and state on the Rust side.  It's not bad, but it does mean for testing that there is state carried over between tests.

Tl;dr: I thought this could use more feedback / brainstorming before filling out the rest of the pieces.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -68,6 +68,7 @@ open class GleanInternalAPI internal constructor () {
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
      */
+    @Synchronized
     fun initialize(
         applicationContext: Context,
         configuration: Configuration = Configuration()
@@ -323,10 +324,11 @@ open class GleanInternalAPI internal constructor () {
      * guaranteed to happen immediately, as that depends on the upload
      * policies.
      *
-     * If the ping currently contains no content, it will not be sent.
+     * If the ping currently contains no content, it will not be assembled and
+     * queued for sending.
      *
      * @param pings List of pings to send.
-     * @return The async Job performing the work of assembling the ping
+     * @return The async [Job] performing the work of assembling the ping
      */
     internal fun sendPings(pings: List<PingType>): Job? {
         val sendPing: (PingType) -> Boolean = {
@@ -350,10 +352,11 @@ open class GleanInternalAPI internal constructor () {
      * guaranteed to happen immediately, as that depends on the upload
      * policies.
      *
-     * If the ping currently contains no content, it will not be sent.
+     * If the ping currently contains no content, it will not be assembled and
+     * queued for sending.
      *
      * @param pingNames List of ping names to send.
-     * @return The async Job performing the work of assembling the ping
+     * @return The async [Job] performing the work of assembling the ping
      */
     internal fun sendPingsByName(pingNames: List<String>): Job? {
         val sendPing: (String) -> Boolean = {
@@ -397,7 +400,7 @@ open class GleanInternalAPI internal constructor () {
      * Register a [PingType] in the registry associated with this [Glean] object.
      */
     internal fun registerPingType(pingType: PingType) {
-        if (handle == 0L) {
+        if (!this.isInitialized()) {
             pingTypeQueue.add(pingType)
         } else {
             LibGleanFFI.INSTANCE.glean_register_ping_type(
@@ -410,9 +413,9 @@ open class GleanInternalAPI internal constructor () {
     /**
      * Returns true if a ping by this name is in the ping registry.
      *
-     * For testing only.
+     * For internal testing only.
      */
-    // @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     internal fun testHasPingType(pingName: String): Boolean {
         return LibGleanFFI.INSTANCE.glean_test_has_ping_type(handle, pingName).toBoolean()
     }
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -399,6 +399,7 @@ open class GleanInternalAPI internal constructor () {
     /**
      * Register a [PingType] in the registry associated with this [Glean] object.
      */
+    @Synchronized
     internal fun registerPingType(pingType: PingType) {
         if (!this.isInitialized()) {
             pingTypeQueue.add(pingType)
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -51,6 +51,9 @@ open class GleanInternalAPI internal constructor () {
     // TODO: 1551159 Integrate MetricsPingScheduler
     // internal lateinit var metricsPingScheduler: MetricsPingScheduler
 
+    // Keep track of ping types that have been registered before Glean is initialized.
+    private val pingTypeQueue: MutableList<PingType> = mutableListOf()
+
     /**
      * Initialize Glean.
      *
@@ -80,6 +83,12 @@ open class GleanInternalAPI internal constructor () {
 
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
         handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName, uploadEnabled.toByte())
+
+        // If any pings were registered before initializing, do so now
+        this.pingTypeQueue.forEach { this.registerPingType(it) }
+        if (!Dispatchers.API.testingMode) {
+            this.pingTypeQueue.clear()
+        }
 
         // TODO: on glean-legacy we perform other actions before initialize the metrics (e.g.
         // init the engines), then init the core metrics, and finally kick off the metrics
@@ -121,7 +130,7 @@ open class GleanInternalAPI internal constructor () {
      */
     fun registerPings(pings: Any) {
         // Instantiating the Pings object to send this function is enough to
-        // call the constructor and have it registered in [PingType.pingRegistry].
+        // call the constructor and have it registered through [Glean.registerPingType].
         Log.i(LOG_TAG, "Registering pings for ${pings.javaClass.canonicalName}")
     }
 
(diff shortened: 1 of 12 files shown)


mdboom:
@Dexterp37 , @badboy : Totally new approach as discussed the other day.  Ping types exist on the Glean object, but we queue them on the Kotlin side until `Glean.initialized` is called.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -368,6 +380,30 @@ open class GleanInternalAPI internal constructor () {
         LibGleanFFI.INSTANCE.glean_destroy_glean(handle, e)
         handle = 0L
     }
+
+    /**
+     * Register a [PingType] in the registry associated with this [Glean] object.
+     */
+    internal fun registerPingType(pingType: PingType) {
```
Dexterp37: Does this need to be `@Synchronized` together with `initialize`?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -368,6 +380,30 @@ open class GleanInternalAPI internal constructor () {
...
+        if (handle == 0L) {
```
Dexterp37: I'd use `isInitialized()` here.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -368,6 +380,30 @@ open class GleanInternalAPI internal constructor () {
...
+            )
+        }
+    }
+
+    /**
+     * Returns true if a ping by this name is in the ping registry.
+     *
+     * For testing only.
+     */
+    // @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
Dexterp37: nit: let's remove the commented line if not needed.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -368,6 +380,30 @@ open class GleanInternalAPI internal constructor () {
...
+                handle,
+                pingType.handle
+            )
+        }
+    }
+
+    /**
+     * Returns true if a ping by this name is in the ping registry.
+     *
+     * For testing only.
```
Dexterp37: nit: "For internal testing only"

glean-core/src/lib.rs
```diff
@@ -154,6 +157,41 @@ impl Glean {
...
         }
     }
+
+    /// Send a ping by name.
+    ///
+    /// The ping content is assembled as soon as possible, but upload is not
+    /// guaranteed to happen immediately, as that depends on the upload
+    /// policies.
+    ///
+    /// If the ping currently contains no content, it will not be sent.
```
Dexterp37: nit: not be sent or not be assembled? :)

glean-core/src/lib.rs
```diff
@@ -154,6 +157,41 @@ impl Glean {
...
+    ///
+    /// Returns true if a ping was sent, false otherwise.
```
Dexterp37: nit: sent or assembled?

glean-core/src/lib.rs
```diff
@@ -120,26 +124,25 @@ impl Glean {
...
 
-    /// Send a ping by name.
+    /// Send a ping.
     ///
     /// The ping content is assembled as soon as possible, but upload is not
     /// guaranteed to happen immediately, as that depends on the upload
     /// policies.
     ///
-    /// TODO: (Verify this is correct):
     /// If the ping currently contains no content, it will not be sent.
```
Dexterp37: Since we're here, can we fix the language here? I think we need to use 'assembled' in most places in which we used 'sent'

glean-core/src/metrics/ping.rs
```diff
@@ -0,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::error::Result;
+use crate::Glean;
+
+#[derive(Debug, Clone)]
+pub struct PingType {
```
Dexterp37: Can we add documentation to this explaining what this is for?

glean-core/tests/ping.rs
```diff
@@ -10,7 +10,10 @@ use glean_core::CommonMetricData;
 
 #[test]
 fn write_ping_to_disk() {
```
Dexterp37: Can we file a bug about porting over the tests from `PingMakerTest.kt` (from [glean-ac](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/src/test/java/mozilla/components/service/glean/ping/PingMakerTest.kt)) to Rust? They cover most of the implementation here. Doesn't need to block this PR, it can be done as a follow-up.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -77,6 +80,12 @@ open class GleanInternalAPI internal constructor () {
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, "glean_data")
         handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName)
 
+        // If any pings were registered before initializing, do so now
+        this.pingTypeQueue.forEach { this.registerPingType(it) }
+        if (!Dispatchers.API.testingMode) {
```
badboy: Why do we keep this queue around in testing mode?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -305,6 +304,30 @@ open class GleanInternalAPI internal constructor () {
...
+     * Send a list of pings.
+     *
+     * The ping content is assembled as soon as possible, but upload is not
+     * guaranteed to happen immediately, as that depends on the upload
+     * policies.
+     *
+     * If the ping currently contains no content, it will not be sent.
+     *
+     * @param pings List of pings to send.
+     * @return The async Job performing the work of assembling the ping
```
badboy: Is `Job` here uppercased because that's the actual type of what `sendPingsGeneric` returns?

glean-core/src/lib.rs
```diff
@@ -154,6 +157,41 @@ impl Glean {
...
+    }
+
+    pub fn register_ping_type(&mut self, ping: &PingType) {
+        if self.ping_registry.contains_key(&ping.name) {
+            log::error!("Duplicate ping named {}", ping.name)
+        }
+
+        // TODO: This just clones the PingType.  This means we have a copy of the PingType
+        // for the ping registry and another managed by handles on the language-specific
+        // side, but these objects are small and this seems like a simple solution...
```
badboy: For now they are small. We can optimize this when it becomes a problem.

badboy:
One thing I forgot: Can you run `make cbindgen` to also update the C header?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: Good point.  I think it probably does.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: Oh, shoot.  It should be there, just committed without removing my playing around here.

glean-core/src/metrics/ping.rs
```diff
@@ -0,0 +1,25 @@
...
+    pub name: String,
+    pub include_client_id: bool,
+}
+
+impl PingType {
+    pub fn new<A: Into<String> + Copy>(name: A, include_client_id: bool) -> Self {
```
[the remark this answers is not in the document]
mdboom: My bad -- I think this is a leftover from earlier struggles to get this to work.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: It's, uh, complicated...

PingTypes are registered by virtue of their being created.  They are singletons.  Therefore, they only get registered once (regardless of when `Glean.registerPings()` is called -- it only works the first time).  This is because we don't want to use introspection and don't want to have an explicit data structure listing all the available pings.

So... in testing, we want to "rerun" the queue every time.  Outside of testing, this is just wasted memory.

But I'd be happy to just delete this clause altogether.  It's not a lot of wasted memory.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: Yes.  Should probably make this `[Job]` (which will link in the generated docs).

glean-core/src/lib.rs
mdboom: If you agree this is ok, it's ok for me.  I felt this was a little "we couldn't figure out how to disable the borrow checker, so we gave up", and there is probably a way to use handles here.  But the duplication isn't a real problem, other than very small amounts of memory, since they are immutable "data classes" essentially.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: This needs to be `@Synchronized` as well.

glean-core/src/metrics/ping.rs
```diff
@@ -0,0 +1,28 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::error::Result;
+use crate::Glean;
+
+#[derive(Debug, Clone)]
+/// Stores information about a ping.
```
Dexterp37: Can we explain why this is needed, e.g. to make it possible to send pings with partial information, for example, no client id.

mdboom:
> Let's merge if CI is green. How comes that we're not getting which test is failing when running Android tests?

I'm not sure, but it seemed to be an intermittent failure and I can no longer reproduce.  :cat: :fire:

Dexterp37:
> 
> 
> > Let's merge if CI is green. How comes that we're not getting which test is failing when running Android tests?
> 
> I'm not sure, but it seemed to be an intermittent failure and I can no longer reproduce. 🐱 🔥

OH NOES, THEY ARE BACK (and gone again)!

PR #88, opened 2019-05-22
1553195 labeled metrics stub

badboy:
First try at the labeled metrics.
Makes use of the trait implementation in #86 (needs rebase when that gets merged).

Notably lacking: Any Kotlin-side handling.


glean-core/src/metrics/labeled.rs
```diff
@@ -0,0 +1,46 @@
...
+            name,
+            ..self.meta.clone()
+        };
+
+        T::with_meta(meta)
+    }
+
+    pub fn get(&self, label: &str) -> T {
+        self.new_metric_with_name(format!("{}/{}", self.meta.name, label))
+    }
```
mdboom: This will be an interesting one to FFI, I suspect.

glean-core/src/storage/mod.rs
```diff
@@ -31,14 +31,30 @@ impl StorageManager {
...
-            let map = snapshot
-                .entry(metric.category())
-                .or_insert_with(HashMap::new);
             let metric_name = String::from_utf8_lossy(metric_name).into_owned();
-            map.insert(metric_name, metric.as_json());
+            if metric_name.contains("/") {
+                let category = format!("labeled_{}", metric.category());
+                let map = snapshot.entry(category).or_insert_with(HashMap::new);
+
+                let mut s = metric_name.splitn(2, '/');
```
mdboom: MICROOPTIMIZATION TO FREELY IGNORE: Maybe there's a way to do this above so we don't have to scan through the string twice?

glean-core/src/storage/mod.rs
```diff
@@ -31,14 +31,30 @@ impl StorageManager {
...
+                let metric_name = s.next().expect("No metric name");
+                let label = s.next().expect("No metric label");
+
+                let obj = map.entry(metric_name.into()).or_insert_with(|| json!({}));
+                if let Some(obj) = obj.as_object_mut() {
```
mdboom: Won't this always be the case, given the `or_insert_with` above?

badboy: In theory yes. Handling it here shields us against broken data (where we _don't_ get an object for some reason .... wait, now that I'm writing this down we're constructing this object completely on our own so we provide the guarantees on what's in there)

glean-core/src/storage/mod.rs
```diff
@@ -31,14 +31,30 @@ impl StorageManager {
         store_name: &str,
         clear_store: bool,
     ) -> Option<JsonValue> {
-        let mut snapshot: HashMap<&str, HashMap<String, JsonValue>> = HashMap::new();
+        let mut snapshot: HashMap<String, HashMap<String, JsonValue>> = HashMap::new();
```
Dexterp37: To learn a bit more: why is this change needed?

glean-core/src/metrics/labeled.rs
```diff
@@ -0,0 +1,46 @@
...
+}
+
+impl<T> LabeledMetric<T>
+where
+    T: MetricType,
+{
+    pub fn new(meta: CommonMetricData, labels: Option<Vec<String>>) -> LabeledMetric<T> {
+        LabeledMetric {
+            meta,
+            labels,
```
Dexterp37: With the current design, how are we going to cope with dynamic labels?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -0,0 +1,171 @@
...
+ * Instances of this class type are automatically generated by the parsers at build time,
+ * allowing developers to record values that were previously registered in the metrics.yaml file.
+ *
+ * Unlike most metric types, LabeledMetricType does not have its own corresponding storage engine,
+ * but records metrics for the underlying metric type T in the storage engine for that type.  The
+ * only difference is that labeled metrics are stored with the special key `$category.$vname/$label`.
+ * The |StorageEngineManager.collect| method knows how to pull these special values back out of the
+ * individual storage engines and rearrange them correctly in the ping.
+ */
+class LabeledMetricType<T>(
```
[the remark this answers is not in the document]
badboy: On it.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -0,0 +1,171 @@
...
+    val name: String,
+    val sendInPings: List<String>,
+    val subMetric: T,
+    val labels: Set<String>? = null
+) {
+
+    companion object {
+        private const val MAX_LABELS = 16
+        private const val OTHER_LABEL = "__other__"
+        private val labelRegex = Regex("^[a-z_][a-z0-9_]{0,29}$")
```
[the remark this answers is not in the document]
badboy: yeah, once we have this file it should be easy to diff and change it here as well

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -0,0 +1,171 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
```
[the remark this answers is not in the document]
badboy: wait, didn't I do it? or did I end up squashing it before pushing? sorry about that

glean-core/src/metrics/labeled.rs
```diff
@@ -0,0 +1,46 @@
...
+use crate::CommonMetricData;
+use crate::Glean;
+
+#[derive(Debug)]
+pub struct LabeledMetric<T> {
+    meta: CommonMetricData,
+    labels: Option<Vec<String>>,
+    // Type of the underlying metric
+    // We don't hold on to any of these, but we need to be able to create them.
+    typ: PhantomData<T>,
```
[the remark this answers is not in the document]
badboy: `underlying_type` ... I take care of proper naming in the followup

glean-core/src/metrics/labeled.rs
badboy: not fleshed out at all.

glean-core/src/storage/mod.rs
badboy: Because before we were using a `&'static str`, now we have either a `&'static str` or a `&'a str` (derived from something inside our loop). We need to have a single type (though we might reduce overhead with [`Cow`](https://doc.rust-lang.org/std/borrow/enum.Cow.html))

PR #90, opened 2019-05-23
Bug 1553424 - Fix C example (was: log on invalid handles, but don't crash)

badboy:
We never return an error and we certainly don't expect too many.
Yet we should log them _when we_ encounter them.

In debug mode we crash, in release we don't (that's what `debug_assert` is).
Now for the handle thing, do we want to avoid crashing even in debug mode when we get an invalid one? 
It certainly points to an error which we might want to see in debug (and the thing that made us aware, calling `set_upload_enabled` before init will get fixed in another way anyway).

Plus, so far we initialize logging during init, saving us a check at other points, but this means our `invalid handle used` log might actually get lost when it happens before the init.

wdyt?

### DIFF
--- a/glean-core/ffi/examples/glean_app.c
+++ b/glean-core/ffi/examples/glean_app.c
@@ -1,28 +1,40 @@
 #include <stdio.h>
 #include <stdlib.h>
 #include <string.h>
+
+// Not exported in glean.h right now.
+typedef struct ExternError {
+    uint32_t code;
+    char *message;
+} ExternError;
+
 #include "glean.h"
 
 int main(void)
 {
-  glean_initialize("/tmp/glean_data");
+  uint64_t glean = glean_initialize("/tmp/glean_data", "c-app");
 
-  printf("Glean upload enabled? %d\n", glean_is_upload_enabled());
-  glean_set_upload_enabled(0);
-  printf("Glean upload enabled? %d\n", glean_is_upload_enabled());
-  glean_set_upload_enabled(1);
+  printf("Glean upload enabled? %d\n", glean_is_upload_enabled(glean));
+  glean_set_upload_enabled(glean, 0);
+  printf("Glean upload enabled? %d\n", glean_is_upload_enabled(glean));
+  glean_set_upload_enabled(glean, 1);
 
-  ExternError err;
-  uint64_t metric = glean_new_counter_metric("local", "counter", &err);
+  const char *pings[2];
+  pings[0] = "store1";
+  pings[1] =  NULL;
+  uint64_t metric = glean_new_counter_metric("local", "counter", pings, 1, 0, 0);
   printf("Created counter: %llu\n", metric);
 
-  glean_counter_add(metric, 2, &err);
+  glean_counter_add(glean, metric, 2);
 
-  char *payload = glean_ping_collect("core", &err);
+  char *payload = glean_ping_collect(glean, "store1");
   printf("Payload:\n%s\n", payload);
   glean_str_free(payload);
 
-  glean_destroy_boolean_metric(metric, &err);
+  ExternError err;
+  glean_destroy_counter_metric(metric, &err);
+
+  glean_destroy_glean(glean, &err);
 
   return 0;
 }


glean-core/ffi/src/handlemap_ext.rs
```diff
@@ -111,7 +112,20 @@ impl<T> HandleMapExtension for ConcurrentHandleMap<T> {
...
         let res = self.call_with_output(&mut error, h, callback);
-        debug_assert!(error.get_code().is_success());
+
+        if !error.get_code().is_success() {
+            if let ErrorCode::INVALID_HANDLE = error.get_code() {
+                // Intentionally left blank
+                // ffi-support already logged a warning.
+            } else {
+                log::warn!(
+                    "Encounterend an unexpected error. Code: {:?}",
```
mdboom:
```suggestion
                    "Encountered an unexpected error. Code: {:?}",
```

glean-core/ffi/src/handlemap_ext.rs
```diff
@@ -122,7 +136,20 @@ impl<T> HandleMapExtension for ConcurrentHandleMap<T> {
...
         let res = self.call_with_output_mut(&mut error, h, callback);
-        debug_assert!(error.get_code().is_success());
+
+        if !error.get_code().is_success() {
+            if let ErrorCode::INVALID_HANDLE = error.get_code() {
+                // Intentionally left blank
+                // ffi-support already logged a warning.
+            } else {
+                log::warn!(
+                    "Encounterend an unexpected error. Code: {:?}",
```
mdboom:
```suggestion
                    "Encountered an unexpected error. Code: {:?}",
```

badboy:
> I like the idea of crashing when errors hit the FFI boundary in Debug mode, and just logging in release, so 👍 from me, but definitely let @Dexterp37 chime in before merging.

That's the thing: if we want to keep that we don't need this PR (except fixing the C app)

Dexterp37:
> 
> 
> We never return an error and we certainly don't expect too many.
> Yet we should log them _when we_ encounter them.
> 
> In debug mode we crash, in release we don't (that's what `debug_assert` is).
> Now for the handle thing, do we want to avoid crashing even in debug mode when we get an invalid one?
> It certainly points to an error which we might want to see in debug (and the thing that made us aware, calling `set_upload_enabled` before init will get fixed in another way anyway).

I agree: we should keep crashing, in debug mode, on these kind of problems as they point to specific bigger errors.

> Plus, so far we initialize logging during init, saving us a check at other points, but this means our `invalid handle used` log might actually get lost when it happens before the init.

Good point. Any chance we could have that *always* initialized when the library is loaded, on the Rust side?

@badboy so this keeps the crash in debug and additionally prints a message (for release builds), but doesn't fix the log init, right?

badboy:
> Good point. Any chance we could have that _always_ initialized when the library is loaded, on the Rust side?

Yes we can. [After loading the lib, just as a-s does](Good point. Any chance we could have that always initialized when the library is loaded, on the Rust side?).

> @badboy so this keeps the crash in debug and additionally prints a message (for release builds), but doesn't fix the log init, right?

This PR does not crash anymore on invalid handles (e.g. handle=0 when things get called before we created the glean object), but crashes for other bugs.
Without the PR we crash even for invalid handles (IMO the way to go)

Dexterp37:
> 
> 
> > Good point. Any chance we could have that _always_ initialized when the library is loaded, on the Rust side?
> 
> Yes we can. [After loading the lib, just as a-s does](Good point. Any chance we could have that always initialized when the library is loaded, on the Rust side?).

Cool, let's do this.

> > @badboy so this keeps the crash in debug and additionally prints a message (for release builds), but doesn't fix the log init, right?
> 
> This PR does not crash anymore on invalid handles (e.g. handle=0 when things get called before we created the glean object), but crashes for other bugs.
> Without the PR we crash (and log) even for invalid handles (IMO the way to go)

Yes, I agree. Let's crash and log for invalid handles, no special handling.

badboy:
D'oh! Managed to copy text instead of the link in the comment above. Fixed that.
Will update this PR to only fix the C example and do logging in a follow-up

PR #92, opened 2019-05-24
Enable logging at load time

badboy:
Based on #90, will rebase when that is merged.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
@@ -46,7 +46,9 @@ internal interface LibGleanFFI : Library {
         private val JNA_LIBRARY_NAME = "glean_ffi"
 
         internal var INSTANCE: LibGleanFFI = try {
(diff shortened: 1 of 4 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -46,7 +46,9 @@ internal interface LibGleanFFI : Library {
         private val JNA_LIBRARY_NAME = "glean_ffi"
 
         internal var INSTANCE: LibGleanFFI = try {
-            Native.loadLibrary(JNA_LIBRARY_NAME, LibGleanFFI::class.java) as LibGleanFFI
+            val lib = Native.loadLibrary(JNA_LIBRARY_NAME, LibGleanFFI::class.java) as LibGleanFFI
```
Dexterp37: Ah, I thought there was a Rust specific way to do that in glean-core, rather than in the Kotlin side. No way I guess, @badboy ?

badboy: No better way than calling `enable_logging` as the first thing in every function. Has the overhead of the constant internal check if it is enabled.

PR #93, opened 2019-05-24
Bug 1553424 - Setting uploadEnabled before initialization should not crash

badboy:
Follow up to #90 in which we now keep track of the flag before we initialize.

We can eventually deprecate the "call before init" way and move the flag into the initializer.
For now this is the easy way out of crashing.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -44,7 +44,7 @@ open class GleanInternalAPI internal constructor () {
     private lateinit var gleanDataDir: File
 
     // Keep track of this flag before Glean is initialized
-    private var uploadEnabled: Boolean? = null
+    private var uploadEnabled: Boolean = true
 
     // This object holds data related to any persistent information about the metrics ping,
     // such as the last time it was sent and the store name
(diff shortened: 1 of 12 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -137,7 +144,11 @@ open class GleanInternalAPI internal constructor () {
      * @param enabled When true, enable metric collection.
      */
     fun setUploadEnabled(enabled: Boolean) {
-        LibGleanFFI.INSTANCE.glean_set_upload_enabled(handle, enabled.toByte())
+        if (isInitialized()) {
```
Dexterp37: What about `getUploadEnabled`? We should have the same non-crash guarantees there.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -78,6 +81,10 @@ open class GleanInternalAPI internal constructor () {
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
         handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName)
```
Dexterp37: Sorry for not catching this before, but can we change `glean_initialize`  to have `uploadEnabled` passed in here? The outer interface will change later on, and it will make our intent more explicit internally. Less errors, hopefully.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -43,6 +43,9 @@ open class GleanInternalAPI internal constructor () {
 
     private lateinit var gleanDataDir: File
 
+    // Keep track of this flag before Glean is initialized
+    private var uploadEnabled: Boolean? = null
```
Dexterp37: Instead of making this a nullable, can we make it `private var uploadEnabled: Boolean = true`? The default value is merely for consistency with glean-ac. We don't need this to be nullable, given the suggested changes below.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -78,6 +81,10 @@ open class GleanInternalAPI internal constructor () {
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
         handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName)
 
+        uploadEnabled?.let {
```
Dexterp37: This is not needed anymore, let's always pass the `uploadEnabled` status through the initialize method.

PR #96, opened 2019-05-28
`should_record` fixup

Dexterp37:

### DIFF
--- a/glean-core/src/metrics/mod.rs
+++ b/glean-core/src/metrics/mod.rs
@@ -22,11 +22,11 @@ pub use self::uuid::UuidMetric;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum Metric {
-    String(String),
     Boolean(bool),
     Counter(i32),
-    Uuid(String),
+    String(String),
     StringList(Vec<String>),
+    Uuid(String),
 }
 
 pub trait MetricType {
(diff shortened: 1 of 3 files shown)


glean-core/src/metrics/mod.rs
```diff
@@ -22,11 +22,11 @@ pub use self::uuid::UuidMetric;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum Metric {
-    String(String),
     Boolean(bool),
     Counter(i32),
-    Uuid(String),
+    String(String),
     StringList(Vec<String>),
+    Uuid(String),
```
badboy: Just for ... alphabetic reasons?

glean-core/src/metrics/string_list.rs
```diff
@@ -17,13 +18,19 @@ pub struct StringListMetric {
     meta: CommonMetricData,
 }
 
+impl MetricType for StringListMetric {
```
badboy: He, I explicitly left these untouched for whoever fully implements these types.
I don't mind getting them fixed now though.

glean-core/src/metrics/mod.rs
Dexterp37: Mh, yes, feels weird to have a random order there

glean-core/src/metrics/string_list.rs
Dexterp37: Yes, I thought about that too, but then I figured that I was copy/pasting outdated code. I think we should either keep things up-to-speed or remove partial implementations. We can have these bits there, the rest is mostly tests and logic that the implementor would need to do anyway.

PR #95, opened 2019-05-28
Bug 1552873 - Implement labeled metrics

badboy:
A first try of implementing a (good enough) solution for the labeled metrics.

I want to write a few more comments before I make this a full PR.

### Pull Request checklist ###

- [x] **Quality**: This PR builds and tests run cleanly
  - `make test` runs without emitting any warnings
  - `make lint` runs without emitting any errors
- [x] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry or an explanation of why it does not need one
  - Any breaking changes to language binding APIs are noted explicitly

PR #98, opened 2019-05-28
Snapshot a single metric

badboy:
This avoids going awkwardly through a JSON-serialized storage and trying to find the right place.

Extracted from #95 

### DIFF
--- a/glean-core/src/storage/mod.rs
+++ b/glean-core/src/storage/mod.rs
@@ -55,4 +55,26 @@ impl StorageManager {
             Some(json!(snapshot))
         }
     }
+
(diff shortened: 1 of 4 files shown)


glean-core/src/storage/mod.rs
```diff
@@ -55,4 +55,26 @@ impl StorageManager {
             Some(json!(snapshot))
         }
     }
+
+    pub fn snapshot_metric(
```
Dexterp37: Can you please add documentation about what does the function do, its parameters and the return value? :)

badboy: Ha, damn, you got me.

PR #99, opened 2019-05-28
Add error reporting for labeled metrics (and counters)

badboy:
Based on #95, needs a rebase when that lands.

The last 3 commits are relevant. We now have all things in place for error reporting using labeled metrics.
Even though the error reporting itself doesn't use the labeled metrics, we simply depend on the same "encoding" of labels into the key.

### DIFF
--- a/glean-core/src/error_recording.rs
+++ b/glean-core/src/error_recording.rs
@@ -2,18 +2,34 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! # Error Recording
+//!
+//! Glean keeps track of errors that occured due to invalid labels or invalid values when recording
+//! other metrics.
+//!
+//! Error counts are stored in labeled counters in the `glean.error` category.
+//! The labeled counter metrics that store the errors are defined in the `metrics.yaml` for documentation purposes,
+//! but are not actually used directly, since the `send_in_pings` value needs to match the pings of the metric that is erroring (plus the "metrics" ping),
+//! not some constant value that we could define in `metrics.yaml`.
+
+use std::fmt::Display;
+
 use crate::metrics::CounterMetric;
 use crate::CommonMetricData;
 use crate::Glean;
 use crate::Lifetime;
 
+/// The possible error types for metric recording.
 #[derive(Debug)]
 pub enum ErrorType {
+    /// For when the value to be recorded does not match the metric-specific restrictions
     InvalidValue,
+    /// For when the label of a labeled metric does not match the restrictions
     InvalidLabel,
 }
 
 impl ErrorType {
+    /// The error type's metric name
     pub fn to_string(&self) -> &'static str {
         match self {
             ErrorType::InvalidValue => "invalid_value",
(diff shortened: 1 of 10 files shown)


mdboom:
What about passing in the message to `record_error`?  That would enforce consistency that every error reported also gets a log message, and that it goes to the same log level etc.

badboy:
That ... actually makes sense.

badboy:
@mdboom / @Dexterp37:
Please take a look at the last 2 commits:

* https://github.com/mozilla/glean/pull/99/commits/b9a93d5c98e80077f9bf3cf58c2181f573fb0eea
* https://github.com/mozilla/glean/pull/99/commits/4cbb2f08205a8d1b116207fba6b7be7e19aeb1cf

These expose a testing API the same way Glean AC does it.

Advantage: 

* for the Kotlin side we don't even need to go through the FFI again, we can construct the metric type and ask it directly
* on the Kotlin-side it's nicely "testing only" through the annotation

Downsides:

* Kotlin now has some logic and knows how labeled metrics are encoded
* the Rust side can't be "test-only" (aka `#[cfg(test)]`), because we need it in the integration tests (those in the `test/` directory). This makes me sad.

Please give me feedback if we want to have these in this way.
If so, we can add calls to the testing API in other parts back again in followups.

glean-core/android/src/main/java/mozilla/telemetry/glean/error/ErrorRecording.kt
```diff
@@ -0,0 +1,61 @@
...
+import mozilla.telemetry.glean.Dispatchers
+import mozilla.telemetry.glean.private.CounterMetricType
+import mozilla.telemetry.glean.private.Lifetime
+
+// Test-only internal object.
+//
+// This allows to check the number of errors recorded per metric.
+
+@VisibleForTesting(otherwise = VisibleForTesting.NONE)
+object ErrorRecording {
```
Dexterp37: Does this really need to be an `object`? Doesn't look like we're storing any persisted state, maybe we could just make this a `class` with a `companion object` or just have no class at all and just have `ErrorType`, `GLEAN_ERROR_NAMES` and `testGetNumRecordedErrors` be loose in the file. Can you file a bug about this?

glean-core/android/src/main/java/mozilla/telemetry/glean/error/ErrorRecording.kt
```diff
@@ -0,0 +1,61 @@
...
+    /**
+    * Get the number of recorded errors for the given metric and error type.
+    *
+    * @param identifier The full metric identifier
+    * @param errorType The type of error
+    * @param pingName The name of the ping.
+    * @return The number of errors reported
+    */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    internal fun testGetNumRecordedErrors(
```
Dexterp37: I think I'd lean for going through the FFI with this. There's no reason why we should leak or duplicate the recording logic. If this becomes a problem (e.g. expensive to do), we can investigate ways to fix it.

glean-core/android/src/main/java/mozilla/telemetry/glean/error/ErrorRecording.kt
mdboom: I don't mind this sticking around as a "namespace" class just for organization (it's the same way in glean-ac), but don't feel strongly either way.

glean-core/android/src/main/java/mozilla/telemetry/glean/error/ErrorRecording.kt
mdboom: Particularly since this is testing only (where performance is less critical).  Or, maybe I'm reading that wrong?

glean-core/android/src/main/java/mozilla/telemetry/glean/error/ErrorRecording.kt
Dexterp37: Right, that's true. I'd still file a bug, just for discussing this again "in the future".

Dexterp37:
@badboy , any chance you could add error logging to `datetime` as well? here or in a follow-up :)

badboy:
> @badboy , any chance you could add error logging to `datetime` as well? here or in a follow-up :)

will do

badboy:
I removed the Kotlin API for now.
I'll also file a bug to followup on that.

PR #102, opened 2019-05-29
Implement the DatetimeMetricType

Dexterp37:
This implements the full stack for the `datetime` metric: from the Rust implementation to the Kotlin platform specific code.


glean-core/src/metrics/mod.rs
```diff
@@ -16,17 +19,20 @@ use crate::Glean;
...
+pub use self::time_unit::TimeUnit;
 pub use self::uuid::UuidMetric;
 
 #[derive(Serialize, Deserialize, Debug, Clone)]
 pub enum Metric {
-    String(String),
     Boolean(bool),
     Counter(i32),
-    Uuid(String),
+    Datetime(DateTime<Utc>, TimeUnit),
```
mdboom: I think we want to record the offset from Utc here, not drop it, so use `DateTime<FixedOffset>`?

glean-core/tests/datetime.rs
```diff
@@ -0,0 +1,57 @@
...
+
+        // `2014-07-08T09:10:11Z`
+        let dt = Utc.ymd(2014, 7, 8).and_hms(9, 10, 11);
+        metric.set(&glean, dt);
+
+        let snapshot = StorageManager
+            .snapshot_as_json(glean.storage(), "store1", true)
+            .unwrap();
+        assert_eq!(
+            json!({"datetime": {"telemetry.datetime_metric": "2014-07-08T09:10:11Z"}}),
```
mdboom: We want the datetime to have an offset here (not `Z`, but a `+` or `-` followed by `MM:SS`)

glean-core/src/metrics/time_unit.rs
```diff
@@ -0,0 +1,56 @@
...
+    Hour,
+    Day
+}
+
+impl TimeUnit {
+    /// TODO
+    pub fn format_pattern(&self) -> &'static str {
+        use TimeUnit::*;
+        match self {
+            Nanosecond => "%Y-%m-%dT%H:%M:%S.%.2f%z",
```
badboy:
```suggestion
            Nanosecond => "%Y-%m-%dT%H:%M:%S%.3f%z",
```

3 digits, includes the dot.
See https://docs.rs/chrono/0.4.6/chrono/format/strftime/index.html

Dexterp37: Thank you! That was it :) I had to slightly change the enums though, to support different resolution.

glean-core/tests/datetime.rs
Dexterp37: Good catch!

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -81,6 +81,25 @@ internal interface LibGleanFFI : Library {
...
+    fun glean_datetime_set(
+        glean_handle: Long,
+        metric_id: Long,
+        year: Int,
+        month: Int,
+        day: Int,
+        hour: Int,
+        minute: Int,
+        second: Int,
+        nano: Int, // FIXME?
```
Dexterp37: This should really be a `u32` :|

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,119 @@
...
+    meta: CommonMetricData,
+    time_unit: TimeUnit,
+}
+
+impl MetricType for DatetimeMetric {
+    fn with_meta(meta: CommonMetricData) -> Self {
+        Self {
+            meta,
+            time_unit: TimeUnit::Day,
+        } // FIXME: How do we handle this?
```
Dexterp37: I'm not sure how to handle this case. This is relevant for the labelled metrics. @mdboom @badboy , thoughts?

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,119 @@
...
+        nano: u32,
+        offset_seconds: i32,
+    ) {
+        let datetime_obj = FixedOffset::east(offset_seconds)
+            .ymd_opt(year, month, day)
+            .and_hms_nano_opt(hour, minute, second, nano);
+
+        match datetime_obj.single() {
+            Some(d) => self.set(glean, d),
+            _ => log::warn!("DatetimeMetric::set: invalid input data. Not recording."), // FIXME: record error?
```
Dexterp37: This should be updated once the error recording merges. I'll rebase

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
mdboom: I think are limited to the [types that JNI supports](https://docs.oracle.com/javase/7/docs/technotes/guides/jni/spec/types.html#wp428).  So maybe use `Long`/`i64` at the FFI layer (which is big enough to hold any `u32` value), and truncate to `u32` on the Rust side inside `glean_datetime_set`?

mdboom: Oh, and hopefully iOS has the same subset of numeric types available :)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
```diff
@@ -114,11 +140,16 @@ class DatetimeMetricType(
...
+        if (!testHasValue(pingName)) {
+            throw NullPointerException()
+        }
+        val ptr = LibGleanFFI
+            .INSTANCE
+            .glean_datetime_test_get_value_as_string(Glean.handle, this.handle, pingName)!!
+        return ptr.getAndConsumeRustString()
     }
 
     /**
```
Dexterp37: Note to self: the `testGetValue(pingName: String = sendInPings.first()): Date` must be implemented and will probably require importing `DateUtils.jt` from glean-ac. That's fine, as this testing API function is platform specific anyway.

glean-core/src/metrics/datetime.rs
badboy: Yeah, so I knew about this shortcoming when I implemented this for labeled metrics.

We essentially have 2 options I think (didn't test them yet):

1. We move `time_unit` into common metric data as an optional field (and then have to do that with multiple of the other optional fields)
2. We keep around the data and in labeled metrics we pass in a constructed instance instead of just keeping the type around (has additional memory overhead)

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,121 @@
...
+    /// The precision of this value is truncated to the `time_unit`
+    /// precision.
+    ///
+    /// This doesn't clear the stored value.
+    pub fn test_get_value(
+        &self,
+        _glean: &Glean,
+        _storage_name: &str,
+    ) -> Option<DateTime<FixedOffset>> {
+        panic!("This is not yet implemented. Please consider using `test_get_value_as_string");
```
badboy: What's the reason to have this in the first place?

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,121 @@
...
+use crate::metrics::Metric;
+use crate::metrics::MetricType;
+use crate::storage::StorageManager;
+use crate::util::get_iso_time_string;
+use crate::CommonMetricData;
+use crate::Glean;
+
+use chrono::{DateTime, FixedOffset, TimeZone};
+
+#[derive(Debug)]
```
badboy: While you're at it: Can you already add docs here?

```rust
/// # A datetime metric
///
/// Used to record an absolute date and time, such as the time the user first ran the application.
```

glean-core/src/metrics/time_unit.rs
```diff
@@ -0,0 +1,58 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use serde::{Deserialize, Serialize};
+use std::convert::TryFrom;
+
+use crate::error::{Error, ErrorKind};
+
+/// Enumeration of different resolutions supported by the
```
badboy: Wording suggestion: Just say "Different time resolutions supported by ..." (that is: leave out `enumeration`)

glean-core/src/metrics/time_unit.rs
```diff
@@ -0,0 +1,58 @@
...
+            Second => "%Y-%m-%dT%H:%M:%S%:z",
+            Minute => "%Y-%m-%dT%H:%M%:z",
+            Hour => "%Y-%m-%dT%H%:z",
+            Day => "%Y-%m-%d%:z",
+        }
+    }
+}
+
+/// Trait implementation for converting an integer value
+/// to a TimeUnit. This is used in the FFI code.
```
badboy: It's rare that trait implementations get a comment in Rust, but I guess it doesn't hurt.

Should there be a note that the values should match the ordering on the kotlin side?

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,121 @@
...
+
+impl DatetimeMetric {
+    pub fn new(meta: CommonMetricData, time_unit: TimeUnit) -> Self {
+        Self { meta, time_unit }
+    }
+
+    /// Public facing API for setting the metric to a date/time which
+    /// includes the timezone offset.
+    ///
+    /// Arguments:
```
badboy: Below and in other places we tend to use a second-level header for this:
```suggestion
    /// ## Arguments:
```

glean-core/src/util.rs
```diff
@@ -22,18 +26,72 @@ pub fn sanitize_application_id(application_id: &str) -> String {
...
+            sanitize_application_id("org.mozilla..test---app")
+        );
+        assert_eq!(
+            "org-mozilla-test-app",
+            sanitize_application_id("org-mozilla-test-app")
+        );
+    }
+
+    #[test]
+    fn test_get_iso_time_string() {
```
badboy: Yey for inline tests :)

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -81,6 +81,25 @@ internal interface LibGleanFFI : Library {
...
+        offset_seconds: Int
+    )
+
+    fun glean_datetime_should_record(glean_handle: Long, metric_id: Long): Byte
+
+    fun glean_datetime_test_has_value(glean_handle: Long, metric_id: Long, storage_name: String): Byte
+
+    fun glean_datetime_test_get_value_as_string(glean_handle: Long, metric_id: Long, storage_name: String): Pointer?
```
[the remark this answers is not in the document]
Dexterp37: We provide the very same testing API in platform specific code, with the same naming scheme. I'd keep this for now and consider changing this in the future. What do you think?

glean-core/src/metrics/datetime.rs
Dexterp37: Feature parity with the platform specific testing API. We can remove this, since there's no consumer for it and the Rust API it's not very close on the roadmap. What do you think?

glean-core/src/metrics/datetime.rs
Dexterp37: > 
>     2. We keep around the data and in labeled metrics we pass in a constructed instance instead of just keeping the type around (has additional memory overhead)

I think I would favour this, but I'm not quite sure what do you mean exactly by it. Can you expand a bit?

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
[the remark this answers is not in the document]
badboy: 👍 - fine with me

glean-core/src/metrics/datetime.rs
badboy: Yeah, seeing panic-y code is not much more helpful than not seeing this code.
 ack for removal.

glean-core/ffi/src/lib.rs
```diff
@@ -299,6 +300,117 @@ pub extern "C" fn glean_boolean_test_get_value(
...
+    minute: u32,
+    second: u32,
+    nano: i64,
+    offset_seconds: i32,
+) {
+    use std::convert::TryInto;
+    // Convert and truncate the nanos to u32, as that's what the underlying
+    // library uses. Unfortunately, not all platform have unsigned integers
+    // so we need to work with what we have.
+    if nano < std::u32::MIN.try_into().unwrap() || nano > std::u32::MAX.try_into().unwrap() {
```
badboy: The first part is just a very fancy way of saying `0`.
The second part is just `std::u32::MAX as i64` (as we [statically] upcast we don't have to worry about any wrappings))

glean-core/src/metrics/datetime.rs
Dexterp37: This will require #99

glean-core/src/metrics/datetime.rs
badboy: Yeah, so this is the change: https://github.com/mozilla/glean/commit/46e490835784fba1459ed751a14f454828b6767e

(It's independent from your part, so we could easily land that before/after)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
```diff
@@ -19,19 +23,47 @@ import java.util.Date
...
+
+    protected fun finalize() {
+        if (this.handle != 0L) {
+            val error = RustError.ByReference()
+            LibGleanFFI.INSTANCE.glean_destroy_datetime_metric(this.handle, error)
+        }
+    }
 
-    // private val logger = Logger("glean/DatetimeMetricType")
+    fun shouldRecord(): Boolean {
```
mdboom: I guess this should probably be `private`, also.  (Though this pattern is throughout the code -- file a separate bug if that makes more sense).

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/DateUtils.kt
```diff
@@ -0,0 +1,102 @@
...
+        val dateFormat = SimpleDateFormat(format, Locale.US)
+        try {
+            return dateFormat.parse(correctedDate)
+        } catch (e: java.text.ParseException) {
+            continue
+        }
+    }
+
+    return null
+}
```
mdboom: This feels like a lot of fiddly code to go outside of the Rust core.  On the other hand, it's fairly well tested in glean-ac, so it's fine to go in now.  Maybe file a bug?  It won't really matter until we have to duplicate this in Swift and C++ and ...

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,110 @@
...
+    /// ## Arguments:
+    ///
+    /// * `glean`: the Glean instance this metric belongs to.
+    /// * `year`: the year to set the metric to.
+    /// * `month`: the month to set the metric to (1-12).
+    /// * `day`: the day to set the metric to (1-based).
+    /// * `hour`: the hour to set the metric to.
+    /// * `minute`: the minute to set the metric to.
+    /// * `second`: the second to set the metric to.
+    /// * `nano`: the nanosecond fraction to the last whole second.
```
mdboom: Missing `offset_seconds` parameter in docstring.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
Dexterp37: Yes, we should refactor this. I filed [bug 1556650](https://bugzilla.mozilla.org/show_bug.cgi?id=1556650) for it.

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/DateUtils.kt
Dexterp37: You have a point. However, right now, this is only needed for the `parseISOTimeString` which is used by the `testGetValue(): Date`, which is platform/language specific, that's why I didn't port it over (I did port over `getISOTimeString` though). I thought about removing it from the file, but I wanted to keep changes to the bare minimum for now. Filed [bug 1556651](https://bugzilla.mozilla.org/show_bug.cgi?id=1556651) for tracking this.

glean-core/src/metrics/datetime.rs
Dexterp37: Thanks, I've rebased on #114 !

glean-core/src/metrics/datetime.rs
Dexterp37: I think this might be as well tackled in #99.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
```diff
@@ -40,45 +72,40 @@ class DatetimeMetricType(
...
-        @Suppress("EXPERIMENTAL_API_USAGE")
-        Dispatchers.API.launch {
-            // Delegate storing the datetime to the storage engine.
-            DatetimesStorageEngine.set(
-                this@DatetimeMetricType,
-                value
-            )
-        }*/
-        Log.e(LOG_TAG, "DatetimeMetricType.set is a stub")
+        val cal = Calendar.getInstance()
```
badboy: I didn't pay close attention to this previously, but do now:

Given that most of the time this gets called without a value set, we just use the current date.
I wonder if we should move this "some date or use now" into Rust? This doesn't have to happen right now, but we should keep it in mind.
(we would still need to support passing in the `Date` anyway)

glean-core/ffi/src/lib.rs
```diff
@@ -299,6 +300,116 @@ pub extern "C" fn glean_boolean_test_get_value(
     })
 }
 
+// *** Start of the Datetime FFI part ***
```
badboy: Yeah, we really need to reorganize this file and split it up.

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,118 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![allow(clippy::too_many_arguments)]
```
badboy: :D

glean-core/src/metrics/datetime.rs
```diff
@@ -0,0 +1,118 @@
...
+    ///
+    /// * `glean`: the Glean instance this metric belongs to.
+    /// * `value`: the date/time value, with offset, to set the metric to.
+    pub fn set(&self, glean: &Glean, value: DateTime<FixedOffset>) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let value = Metric::Datetime(value, self.time_unit);
+        glean.storage().record(&self.meta, &value)
```
badboy: I do like how simple the actualy data storage is.

glean-core/ffi/src/lib.rs
Dexterp37: Yes! We have a bug for  that already, see [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1553800).

glean-core/src/metrics/datetime.rs
Dexterp37: That was my way to say "STOP IT CLIPPY"

glean-core/src/metrics/datetime.rs
Dexterp37: Yup :) It's making me feel good about our choice. I fear Desktop, though :P

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
Dexterp37: This is a very good point. I filed [bug 1556677](https://bugzilla.mozilla.org/show_bug.cgi?id=1556677) to track this.

glean-core/src/metrics/datetime.rs
mdboom: Curious: Dare I ask why you fear desktop?

Dexterp37: Depending on how we approach there, we might need to abstract away the "lifetime" from the storage, since these concepts there are a bit more ... confused :)

PR #104, opened 2019-05-29
Fetch seen labels from the database

badboy:
This goes through the database and looks for all stored data for the
named labeled metric and extracts the labels.

I thought about extracting the extraction part into a function but couldn't come up with a proper name for it.


glean-core/src/metrics/labeled.rs
```diff
@@ -67,8 +69,30 @@ where
...
+
+            let lifetime = self.meta.lifetime;
+            for store in &self.meta.send_in_pings {
+                glean
+                    .storage()
+                    .iter_store_from(lifetime, store, &mut snapshotter);
+            }
+
+            // Swap the newly seen labels with the old (empty) storage.
+            mem::swap(&mut self.seen_labels, &mut seen_labels);
```
mdboom: Learning question: Is this just so this is transactional, or is there a ownership reason to do this?

badboy: Actually ... it was for ownership reasons, but I now see how to refactor it.

badboy:
Failed because:

```
            > Could not GET 'https://jcenter.bintray.com/com/squareup/javawriter/2.5.0/javawriter-2.5.0.pom'. Received status code 504 from server: Gateway Time-out
```

I rekicked the test.

glean-core/src/metrics/labeled.rs
```diff
@@ -67,8 +68,27 @@ where
         }
     }
 
-    fn dynamic_label<'a>(&mut self, label: &'a str) -> &'a str {
-        // TODO(bug 1554970): Fetch seen_labels from the database if empty
+    fn dynamic_label<'a>(&mut self, glean: &Glean, label: &'a str) -> &'a str {
```
Dexterp37: I know this isn't a public function, but can you please add documentation for what this is supposed to do?

glean-core/tests/labeled.rs
```diff
@@ -251,3 +252,70 @@ fn dynamic_labels_regex_mimsatch() {
         snapshot
     );
 }
+
+#[test]
+fn seen_labels_get_reloaded_from_disk() {
```
Dexterp37: Will this allow us to enable [this test](https://github.com/mozilla/glean/blob/93736f8598b607f77c96bf9cda266c7716dde987/glean-core/android/src/test/java/mozilla/telemetry/glean/private/LabeledMetricTypeTest.kt#L436) again?

badboy: once I understand the mocking part

Dexterp37: I can help you with that, no worries :) ping me about i!

badboy: I took a look and can't really see how to re-enable that tests (or even that we _should_ re-enable it):

* It's mocking sharedpreferences, which we are not using anymore
* We can't easily inject data into the store, without going through the rudst part anyway.
* This case is covered by the Rust test

How about removing that kotlin test case?
wdyt, @Dexterp37?

Dexterp37: I think you're right. Let's remove the test and add something like this in its place:

```
// SKIPPED `test seen labels get reloaded from disk`
// REASON XYZ HERE
```

(here or in a follow-up)

PR #107, opened 2019-05-29
1554275: Port PingMakerTests.kt to the rust side

mdboom:

### DIFF
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -32,21 +32,16 @@ pub fn new_glean() -> (Glean, tempfile::TempDir) {
 }
 
 /// Convert an iso8601::DateTime to a chrono::DateTime<FixedOffset>
-pub fn iso8601_to_chrono(
-    datetime: &iso8601::DateTime,
-) -> Option<chrono::DateTime<chrono::FixedOffset>> {
+pub fn iso8601_to_chrono(datetime: &iso8601::DateTime) -> chrono::DateTime<chrono::FixedOffset> {
     if let YMD { year, month, day } = datetime.date {
-        Some(
-            chrono::FixedOffset::east(datetime.time.tz_offset_hours * 3600)
-                .ymd(year, month, day)
-                .and_hms_milli(
-                    datetime.time.hour,
-                    datetime.time.minute,
-                    datetime.time.second,
-                    datetime.time.millisecond,
-                ),
-        )
-    } else {
-        None
-    }
+        return chrono::FixedOffset::east(datetime.time.tz_offset_hours * 3600)
+            .ymd(year, month, day)
+            .and_hms_milli(
+                datetime.time.hour,
+                datetime.time.minute,
+                datetime.time.second,
+                datetime.time.millisecond,
+            );
+    };
+    panic!("Unsupported datetime format");
 }
--- a/glean-core/tests/common/mod.rs
+++ b/glean-core/tests/common/mod.rs
@@ -7,6 +7,11 @@
 #![allow(dead_code)]
 use glean_core::Glean;
 
+use chrono;
+use chrono::offset::TimeZone;
+use iso8601;
+use iso8601::Date::YMD;
+
 pub fn tempdir() -> (tempfile::TempDir, String) {
     let t = tempfile::tempdir().unwrap();
     let name = t.path().display().to_string();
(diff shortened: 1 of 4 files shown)


glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+
+#[test]
+fn ping_info_must_contain_a_nonempty_start_and_end_time() {
+    let (glean, ping_maker, ping_type, _t) = set_up_basic_ping();
+
+    let content = ping_maker.collect(glean.storage(), &ping_type).unwrap();
+    let ping_info = content["ping_info"].as_object().unwrap();
+
+    let start_time_str = ping_info["start_time"].as_str().unwrap();
+    let start_time_date = iso8601_to_chrono(&iso8601::datetime(start_time_str).unwrap()).unwrap();
```
Dexterp37: Ah, this is a neat util. It might be useful for the DatetimeMetric too.

glean-core/tests/common/mod.rs
```diff
@@ -25,3 +30,23 @@ pub fn new_glean() -> (Glean, tempfile::TempDir) {
 
     (glean, dir)
 }
+
+/// Convert an iso8601::DateTime to a chrono::DateTime<FixedOffset>
+pub fn iso8601_to_chrono(
+    datetime: &iso8601::DateTime,
+) -> Option<chrono::DateTime<chrono::FixedOffset>> {
```
badboy: It's for tests only. Can we just not wrap it in option and instead `panic!` when we get something we can't handle? Would save us some unwraps further down.

glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+use glean_core::{CommonMetricData, Glean, Lifetime};
+
+fn set_up_basic_ping() -> (Glean, PingMaker, PingType, tempfile::TempDir) {
+    let (t, tmpname) = tempdir();
+    let mut glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID, true).unwrap();
+    let ping_maker = PingMaker::new();
+    let ping_type = PingType::new("store1", true);
+    glean.register_ping_type(&ping_type);
+
+    let metric = BooleanMetric::new(CommonMetricData {
```
badboy: Should this have a note that we record something to ensure the ping will contain data?

glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+        name: "boolean_metric".into(),
+        category: "telemetry".into(),
+        send_in_pings: vec!["store2".into()],
+        disabled: false,
+        lifetime: Lifetime::User,
+    });
+    metric.set(&glean, true);
+
+    let mut results: Vec<i64> = Vec::new();
+    for _ in 1..3 {
```
badboy: `1..3` is 1-to-exclusive-3.
`1..=2` is 1-to-inclusive-2, which sometimes makes it easier to understand. Whatever you prefer though.

glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+        for ping_name in ["store1", "store2"].iter() {
+            let ping_type = PingType::new(*ping_name, true);
+            let content = ping_maker.collect(glean.storage(), &ping_type).unwrap();
+            let seq_num = content["ping_info"]["seq"].as_i64().unwrap();
+            results.push(seq_num);
```
badboy: Can't we test the sequence numbers right here?
E.g. from above change to
```rust
for i in 0..=1 {
...
```

then `assert_eq!(i, seq_num);`.

glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+        }
+    }
+
+    assert_eq!(0, results[0]);
```
badboy: At the very least this should have a comment block explaining what results contains and why we expect `0, 0, 1, 1`

glean-core/tests/ping_maker.rs
```diff
@@ -0,0 +1,120 @@
...
+    let unknown_ping_type = PingType::new("unknown", true);
+    glean.register_ping_type(&ping_type);
+
+    assert!(ping_maker
+        .collect(glean.storage(), &unknown_ping_type)
+        .is_none());
+}
+
+#[test]
+#[ignore] // sequence numbers aren't implemented
```
badboy: I like having the test already written, but just ignored. 🎉

glean-core/tests/common/mod.rs
```diff
@@ -7,6 +7,11 @@
 #![allow(dead_code)]
 use glean_core::Glean;
 
+use chrono;
```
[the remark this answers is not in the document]
mdboom: Yeah -- I liked having the fully-qualified names in this file below since there is a function that uses both `chrono::DateTime` and `iso8601::DateTime`.

glean-core/tests/common/mod.rs
mdboom: That's a good idea.

glean-core/tests/common/mod.rs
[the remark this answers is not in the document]
badboy: But crates are always in scope, so `chrono::DateTime` will work without that `use chrono` line.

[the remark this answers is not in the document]
mdboom: Oh, cool.  TIL.

PR #112, opened 2019-05-31
Upgrade the WorkManager dependency to 2.0.1

Dexterp37:
This fixes a crash occurring in glean when the connectivity is disabled and then enabled. See
https://github.com/mozilla-mobile/android-components/issues/3110 for context.


mdboom:
I suppose when we officially combine this with the wrapper in a-c, we'll be forced to use the same versions of everything that a-c does anyway...  So, besides from fixing the bug, it's good to track this so there's fewer "surprises" later.

PR #114, opened 2019-06-03
Store an instance of the underlying metric and clone it on request

badboy:
This has the overhead of storing a metric instance per labeled metric.
However, these should be rather small.

And on the upside this allows us to store metric-type-specific data
(e.g. the time unit for DateTime) for that metric only instead of inside
the common metric data.


Dexterp37:
> How will this work with metrics with extra parameters (such as the DateTime @Dexterp37 is working on?)

This will actually make it possible, see [this comment](https://github.com/mozilla/glean/pull/102#discussion_r289879657)

PR #117, opened 2019-06-04
1552842: Document how to run tests

mdboom:

### DIFF
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -1,0 +1,34 @@
+# Running the tests
+
+## Running all tests
+
+The tests for all languages my be run from the command line (on Unix platforms):
+
+```
+make test
+```
+
+## Running the Rust tests
+
+The Rust tests may be run with the following command:
+
+```
+cargo test --all
+```
+
+## Running the Kotlin/Android tests
+
+### From the command line
+
+The full Android test suite may be run from the command line with:
+
+```
+./gradlew test
+```
+
+### From Android Studio
+
+To run the full Android test suite, in the "Gradle" pane, navigate to `glean-core` -> `Tasks` -> `verification` and double-click either `testDebugUnitTest` or `testReleaseUnitTest` (depending on whether you want to run in Debug or Release mode).
+You can save this task permanently by opening the task dropdown in the toolbar and selecting "Save glean.rs:glean-core:android [testDebugUnitTest] Configuration".
+
+To run a single Android test, navigate to the file containing the test, and right click on the green arrow in the left margin next to the test.  There you have a choice of running or debugging the test.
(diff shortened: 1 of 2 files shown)


docs/dev/testing.md
```diff
@@ -0,0 +1,34 @@
+# Running the tests
+
+## Running all tests
+
+The tests for all languages my be run from the command line (on Unix platforms):
```
chutten:
```suggestion
The tests for all languages may be run from the command line (on Unix platforms):
```

chutten: Also, this works on Windows as well.

mdboom: Oh, interesting.  I vaguely recall @Dexterp37 reporting that it didn't...

chutten: Works in my mozillabuild shell (just set it up yesterday). Maybe it depends what env you're running it in...

I've been keeping notes on what it's taken to get up and running. Maybe Alessio and I need to compare Windows notes.

Dexterp37: Yup. I'm not considering `mozillabuild` shell a requirement for this project to be honest. My *personal* definition of "working on Windows" is you can open a command prompt, call commands in it, and it will just work.

By default, as far as I know, Windows doesn't know what a `Makefile` is, so can't run that out of the box. However, the commands executed by `make test` work just fine  when executing in the usual Windows command prompt.

An alternative solution is to use the Linux subsystem for Windows (aka bash for Windows). `make test` will work there, without requiring us an implicit dependency on `mozillabuild`

docs/dev/testing.md
```diff
@@ -0,0 +1,34 @@
+# Running the tests
```
Dexterp37: Can we add a note for Windows users (see [my comment](https://github.com/mozilla/glean/pull/117#discussion_r290659092)) or file a follow-up for adding it?

mdboom:
@Dexterp37, @chutten: Added a Windows-specific note -- trying to just add pointers rather than detail on that -- our Windows-using audience probably generally has experience and preferences around these things anyway.

docs/dev/testing.md
```diff
@@ -0,0 +1,40 @@
...
+
+The tests for all languages may be run from the command line:
+
+```
+make test
+```
+
+---
+
+_Windows Note:_ If `make` is not installed, it may be obtained from the [MozillaBuild shell](https://wiki.mozilla.org/MozillaBuild) or by using the [Windows Subsystem for Linux](https://docs.microsoft.com/en-us/windows/wsl/faq).
```
Dexterp37: Let's leave the MozillaBuild shell out for now, I'd rather not advise people to do so, if possible.

PR #119, opened 2019-06-05
Initialize the logger for all (integration) tests

badboy:
This initializes the logger through a "constructor", that is called before `main`.
This is safe to use, re-initializing the logger has no effect.

Log ouptut can be controlled via the environment variable `RUST_LOG` for the `glean_core` crate,

```
export RUST_LOG=glean_core=debug
```

When running tests you need to tell cargo to not suppress output:

```
cargo test -- --nocapture
```

Tests run in parallel by default, leading to interleaving log lines.
This makes it harder to understand what's going on.
For debugging you can force single-threaded tests:

```
cargo test -- --nocapture --test-threads=1
```


glean-core/tests/common/mod.rs
```diff
@@ -12,6 +12,20 @@ use chrono::offset::TimeZone;
...
+/// Initialize the logger for all tests without individual tests requiring to call the init code.
+/// Log ouptut can be controlled via the environment variable `RUST_LOG` for the `glean_core` crate,
+/// e.g.:
+///
+/// ```
+/// export RUST_LOG=glean_core=debug
+/// ```
+#[ctor]
+fn init() {
+    let _ = env_logger::builder().is_test(true).try_init();
```
Dexterp37: @badboy , should this call in [`glean_enable_logging`](https://github.com/mozilla/glean/blob/c96af194a24759b741148b4b179e6d9b30e2b63a/glean-core/ffi/src/lib.rs#L50)?

If not, why?

glean-core/tests/common/mod.rs
```diff
@@ -12,6 +12,20 @@ use chrono::offset::TimeZone;
...
 use iso8601::Date::YMD;
 
+use ctor::ctor;
+
+/// Initialize the logger for all tests without individual tests requiring to call the init code.
+/// Log ouptut can be controlled via the environment variable `RUST_LOG` for the `glean_core` crate,
+/// e.g.:
+///
+/// ```
+/// export RUST_LOG=glean_core=debug
```
Dexterp37: Should this bit of information go in #117 ? If that's merged before, can you please add it to the document in this PR?

mdboom:
This is to address https://bugzilla.mozilla.org/show_bug.cgi?id=1556820, I presume.

glean-core/tests/common/mod.rs
Dexterp37: Oh, I now understand. That's required due to what @mdboom mentioned in his comment, right? Still, let's add a comment here to disambiguate and mention what's the relation with the other logging function

badboy: `glean_enable_logging` is from the FFI component. We don't have that available in the `glean_core` tests.

badboy: @Dexterp37 	there's a big comment above the function (but actually I should rename the function). Do you think that comment is enough?

Makefile
```diff
@@ -31,6 +31,9 @@ test:
 	./gradlew test
 .PHONY: test
 
+test-rust-with-logs:
+	RUST_LOG=glean_core=debug cargo test --all -- --nocapture --test-threads=1
+
```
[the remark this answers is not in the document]
badboy: Oversight indeed.
[Phony targets](https://www.gnu.org/software/make/manual/html_node/Phony-Targets.html) are used to tell make that this task always needs to run, ignoring any file of a similar name (for things like `test-rust-with-logs` where the chance a file of that name exists is ~0 it doesn't matter much though)

glean-core/tests/common/mod.rs
Dexterp37: @badboy I'm no sure I understand how the comment relates to `glean_enable_logging`, sorry :( What I'm trying to understand is: do we still need the logger to be instantiated in a platform specific way (thus calling the same code `glean_enable_logging` has) also when running unit tests? If the answer is "no", is it because we want unit tests to always print to stdout/stderrr regardless of the platform they run on? If that's correct, can you mention it in a comment?

badboy: Ah, I didn't get that from the first comment. I'll address this.

badboy:
I am not really sure what failed.
There are some `IllegalStateException`s logged, no idea if that's relevant.

Log in https://circleci.com/gh/mozilla/glean/3302.

Restarting test to see if that's an interkitten.

badboy:
It seems it was an interkitten.

PR #113, opened 2019-05-31
1555975: Move markdown docs from glean-ac to glean-core

mdboom:
This moves all of the doc content in glean-ac in markdown (i.e. not docstrings) over to glean-core -- specifically, this is everything in `components/service/glean/docs` as well as `components/service/glean/README.md`, but reorganized into smaller chunks.  I tried to update things that have changed, but not much has since we're keeping the same public API.

For the documentation of the specific metric types, I tried to split the difference between the "narrative" style in glean-ac and the systematic style with standard headers that @badboy started over here.  I basically added a `Configuration` section (talking about metrics.yaml content), an `API` section talking about using the metric and a `Reference` section linking to the javadoc.

It's clear from this that eventually we'll need a way to handle documenting Kotlin, Swift, Rust and beyond -- maybe [something like this](https://json-schema.org/understanding-json-schema/conventions.html#language-specific-notes) -- but for now, this is all completely Kotlin/Android-specific.

This adds a linkchecking step to CI -- I found it really helpful as I was moving pieces around from other repos to automate the process of making sure internal links go somewhere.  I considered using [mdbook-linkcheck](https://github.com/Michael-F-Bryan/mdbook-linkcheck) for this, but that doesn't deal with the fact that we have docs generated from 3 sources: mdbook, rustdoc and dokka.  Instead, I'm using the pretty old and slow but popular "linkchecker" tool, mainly just because it has a Debian package.  **NOTE**: This is only checking internal links.  Checking external links is much slower, and there's a lot of false positives to things in the generated docs that we can't control.

This also had me looking at the Kotlin docs and noting that some things should be private but aren't.  Sorry if that's kind of out-of-scope -- happy to separate that out to another PR.

### Pull Request checklist ###

- [ ] **Quality**: This PR builds and tests run cleanly
  - `make test` runs without emitting any warnings
  - `make lint` runs without emitting any errors
- [ ] **Tests**: This PR includes thorough tests or an explanation of why it does not
- [ ] **Changelog**: This PR includes a changelog entry or an explanation of why it does not need one
  - Any breaking changes to language binding APIs are noted explicitly

PR #120, opened 2019-06-05
Implement sequence numbers as a counter

badboy:
Sequence numbers are stored as a counter under a name that includes the storage name:

    <storage name>#sequence

Everything is stored in the special `glean.internal` storage.
This is not used for anything else.

### DIFF
--- a/glean-core/examples/sample.rs
+++ b/glean-core/examples/sample.rs
@@ -63,11 +63,10 @@ fn main() {
 
     let ping_maker = PingMaker::new();
     let ping = ping_maker
-        .collect_string(glean.storage(), glean.get_ping_by_name("baseline").unwrap())
+        .collect_string(&glean, glean.get_ping_by_name("baseline").unwrap())
         .unwrap();
     println!("Baseline Ping:\n{}", ping);
 
-    let ping =
-        ping_maker.collect_string(glean.storage(), glean.get_ping_by_name("metrics").unwrap());
+    let ping = ping_maker.collect_string(&glean, glean.get_ping_by_name("metrics").unwrap());
     println!("Metrics Ping: {:?}", ping);
 }
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -9,9 +9,9 @@ use std::path::{Path, PathBuf};
 use log::info;
 use serde_json::{json, Value as JsonValue};
 
-use crate::database::Database;
 use crate::metrics::PingType;
 use crate::storage::StorageManager;
+use crate::Glean;
 
 pub struct PingMaker;
 
@@ -39,20 +39,20 @@ impl PingMaker {
         Self
     }
 
-    fn get_ping_seq(&self, _storage: &str) -> usize {
+    fn get_ping_seq(&self, glean: &Glean, storage_name: &str) -> usize {
         1
     }
 
-    fn get_ping_info(&self, storage_name: &str) -> JsonValue {
+    fn get_ping_info(&self, glean: &Glean, storage_name: &str) -> JsonValue {
         json!({
             "ping_type": storage_name,
-            "seq": self.get_ping_seq(storage_name),
+            "seq": self.get_ping_seq(glean, storage_name),
             "start_time": "2019-03-29T09:50-04:00",
             "end_time": "2019-03-29T09:53-04:00"
         })
     }
 
-    fn get_client_info(&self, storage: &Database, include_client_id: bool) -> JsonValue {
+    fn get_client_info(&self, glean: &Glean, include_client_id: bool) -> JsonValue {
         // Add the "telemetry_sdk_build", which is the glean-core version.
         let version = env!("CARGO_PKG_VERSION");
         let mut map = json!({
@@ -61,7 +61,7 @@ impl PingMaker {
 
         // Flatten the whole thing.
         if let Some(client_info) =
-            StorageManager.snapshot_as_json(storage, "glean_client_info", true)
+            StorageManager.snapshot_as_json(glean.storage(), "glean_client_info", true)
         {
             let client_info_obj = client_info.as_object().unwrap(); // safe, snapshot always returns an object.
             for (_key, value) in client_info_obj {
@@ -76,10 +76,11 @@ impl PingMaker {
         json!(map)
     }
 
-    pub fn collect(&self, storage: &Database, ping: &PingType) -> Option<JsonValue> {
+    pub fn collect(&self, glean: &Glean, ping: &PingType) -> Option<JsonValue> {
         info!("Collecting {}", ping.name);
 
-        let metrics_data = match StorageManager.snapshot_as_json(storage, &ping.name, true) {
+        let metrics_data = match StorageManager.snapshot_as_json(glean.storage(), &ping.name, true)
+        {
             None => {
                 info!("Storage for {} empty. Bailing out.", ping.name);
                 return None;
@@ -87,8 +88,8 @@ impl PingMaker {
             Some(data) => data,
         };
 
-        let ping_info = self.get_ping_info(&ping.name);
-        let client_info = self.get_client_info(storage, ping.include_client_id);
+        let ping_info = self.get_ping_info(glean, &ping.name);
+        let client_info = self.get_client_info(glean, ping.include_client_id);
 
         Some(json!({
             "ping_info": ping_info,
@@ -97,8 +98,8 @@ impl PingMaker {
         }))
     }
 
-    pub fn collect_string(&self, storage: &Database, ping: &PingType) -> Option<String> {
-        self.collect(storage, ping)
+    pub fn collect_string(&self, glean: &Glean, ping: &PingType) -> Option<String> {
+        self.collect(glean, ping)
             .map(|ping| ::serde_json::to_string_pretty(&ping).unwrap())
     }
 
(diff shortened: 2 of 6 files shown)


glean-core/examples/sample.rs
```diff
@@ -62,7 +62,7 @@ fn main() {
 
     let ping_maker = PingMaker::new();
     let ping = ping_maker
-        .collect_string(glean.storage(), glean.get_ping_by_name("core").unwrap())
+        .collect_string(&glean, glean.get_ping_by_name("core").unwrap())
```
Dexterp37: nit: it's totally unrelated, but can we rename `"core"` to `"baseline"` or something else?

glean-core/src/ping/mod.rs
```diff
@@ -40,7 +41,30 @@ impl PingMaker {
     }
 
     fn get_ping_seq(&self, glean: &Glean, storage_name: &str) -> usize {
-        1
+        // A internal category, not to be touched by anything else
+        let internal_storage = "glean.internal";
```
Dexterp37: We should be careful about this: we currently export some metrics in [`glean.internal.metrics`](https://github.com/mozilla/glean/blob/efdcd37604b4a6244478f406f5d5d9ad80604eff/glean-core/android/metrics.yaml#L42). I would do a few things:

- consider adding the Glean owned storage names to an internal document [like this](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/docs/internal.md). If @mdboom has ported that over, we should maybe add this to that doc (was it #113 ?).
- check with @mdboom if we're forbidding `glean.internal` in glean_parser.

glean-core/src/ping/mod.rs
```diff
@@ -40,7 +41,30 @@ impl PingMaker {
...
+        // Sequence numbers are stored as a counter under a name that includes the storage name
+        let seq = CounterMetric::new(CommonMetricData {
+            name: format!("{}#sequence", storage_name),
+            category: "".into(),
+            send_in_pings: vec![internal_storage.into()],
+            lifetime: Lifetime::User,
+            ..Default::default()
+        });
+
+        seq.add(glean, 1);
```
Dexterp37: What if we move this after the snapshotting below...

glean-core/src/ping/mod.rs
```diff
@@ -40,7 +41,30 @@ impl PingMaker {
...
+
+        let current_seq = match StorageManager.snapshot_metric(
+            glean.storage(),
+            internal_storage,
+            &seq.meta().identifier(),
+        ) {
+            Some(Metric::Counter(i)) => i - 1, // We start at 0 and need the last value
```
Dexterp37: ... so that this could just become `Some(Metric::Counter(i)) => i,` ...

glean-core/src/ping/mod.rs
```diff
@@ -40,7 +41,30 @@ impl PingMaker {
...
+            _ => unreachable!("Sequence number just set, but now missing"),
```
Dexterp37: ...and this `_ => 0`?

glean-core/tests/ping_maker.rs
```diff
@@ -88,7 +88,6 @@ fn collect_must_report_none_when_no_data_is_stored() {
 }
 
 #[test]
-#[ignore] // sequence numbers aren't implemented
 fn seq_number_must_be_sequential() {
```
Dexterp37: nit: can we slightly augment this test so that after the store increase together one store keeps increasing? This is to test that we don't return the same sequences for all the stores :)

glean-core/src/ping/mod.rs
badboy: Yup, consider it done.

glean-core/examples/sample.rs
badboy: Taking that into #122

glean-core/src/ping/mod.rs
mdboom: `glean.internal.metrics` is a special category that the `glean_parser` only lets Glean itself use.  (Glean's internal build passes a special commandline flag to `glean_parser`).

Maybe we could just document this like the other internal metrics by adding it to `metrics.yaml` and set `send_in_pings` to `[]` so it will never be sent by accident on any pings?

Dexterp37: I'd like this, but then we'd be advertising a pattern (i.e. `send_in_pings: []`) which should be only used by us. I'm not sure of the top off my head if that has implications on the storage, too. @badboy , thoughts?

badboy:
From https://github.com/mozilla/glean/pull/120#discussion_r290744427:

> glean.internal.metrics is a special category that the glean_parser only lets Glean itself use. (Glean's internal build passes a special commandline flag to glean_parser).

Ah, I managed to confuse myself and all of you at the same time by using `category`.
The metrics itself are actually placed into the "empty string" category. That's the same as e.g. the "client_id" in the code (see [internal_metrics.rs](https://github.com/mozilla/glean/blob/5f2fbbb3caff08f26b1b4f7eff52c67180143236/glean-core/src/internal_metrics.rs#L18)).
Also `client_id` and others are documented to be in the `glean.internal.metrics` category.
This is fine, as they get special handling (`send_in_pings = ["glean_client_info"]` and they end up in the `client_info` part)

The `glean.internal` as used in this code is just the _ping name_ the data is stored in (equivalent to `glean_client_info` for other things).
I'm fine with renaming that to e.g. `glean_internal_info` and documenting that this is a reserved ping name (reserve all `glean_*` ping names?)

> Maybe we could just document this like the other internal metrics by adding it to metrics.yaml and set send_in_pings to [] so it will never be sent by accident on any pings?

Two problems:

1. Right now I'm using dynamic names of the format `<ping name>#sequence` (and similar for the start/end time. We therefore can't properly document that (we sould document just `sequence` with a note in the description that it is actually differently used?).
	* We can't store the sequence for the ping name itself, otherwise it would end up in the data payload (or needs yet another special handling there)
	* We need the ping name in the key to have one sequence per actual ping
2. Due to above we need `send_in_pings = ["glean_internal_info"]` in the code to actually store the data and get it out.
	* If we go the above way and document it, we could still have the `send_in_pings` be empty as a documentation purpose, though we don't do this for `client_id` either.



Sorry for the now long comment, but I hope this clears up the confusion and helps us move forward.
wdyt, @Dexterp37, @mdboom?

Dexterp37:
> 
> 
> From [#120 (comment)](https://github.com/mozilla/glean/pull/120#discussion_r290744427):
> 
> > glean.internal.metrics is a special category that the glean_parser only lets Glean itself use. (Glean's internal build passes a special commandline flag to glean_parser).
> 
> Ah, I managed to confuse myself and all of you at the same time by using `category`.

Good job :-D 

> The metrics itself are actually placed into the "empty string" category. That's the same as e.g. the "client_id" in the code (see [internal_metrics.rs](https://github.com/mozilla/glean/blob/5f2fbbb3caff08f26b1b4f7eff52c67180143236/glean-core/src/internal_metrics.rs#L18)).
> Also `client_id` and others are documented to be in the `glean.internal.metrics` category.
> This is fine, as they get special handling (`send_in_pings = ["glean_client_info"]` and they end up in the `client_info` part)
> 
> The `glean.internal` as used in this code is just the _ping name_ the data is stored in (equivalent to `glean_client_info` for other things).

Ok, great catch! Looks like this is following the pattern as other internal metrics. I'd just add a comment there, with what you just wrote here, and change the name to something like `glean_internal` (replacing the dot).

> I'm fine with renaming that to e.g. `glean_internal_info` and documenting that this is a reserved ping name (reserve all `glean_*` ping names?)

That's even better, `glean_internal_info` it is! Yes, let's add this to an internal document. Please note that this also requires a change to the parser itself (and the related docs).

> > Maybe we could just document this like the other internal metrics by adding it to metrics.yaml and set send_in_pings to [] so it will never be sent by accident on any pings?
> 
> Two problems:
> 
>     1. Right now I'm using dynamic names of the format `<ping name>#sequence` (and similar for the start/end time. We therefore can't properly document that (we sould document just `sequence` with a note in the description that it is actually differently used?).
>        
>        * We can't store the sequence for the ping name itself, otherwise it would end up in the data payload (or needs yet another special handling there)
>        * We need the ping name in the key to have one sequence per actual ping

Yes, I see no easy solution to that without additional special handling. Eugh.

>     2. Due to above we need `send_in_pings = ["glean_internal_info"]` in the code to actually store the data and get it out.
>        
>        * If we go the above way and document it, we could still have the `send_in_pings` be empty as a documentation purpose, though we don't do this for `client_id` either.

Yeah, this would be a bit odd and possibly cause confusion.
 
> Sorry for the now long comment, but I hope this clears up the confusion and helps us move forward.
> wdyt, @Dexterp37, @mdboom?

What if we keep the current status, with changing the `send_in_pings`, changing the parsers and documenting the changes on both sides? @badboy @mdboom ?

badboy:
> That's even better, `glean_internal_info` it is! Yes, let's add this to an internal document. Please note that this also requires a change to the parser itself (and the related docs).

Why would it? The parser already doesn't handle `glean_client_info` in any special way.

Dexterp37:
> 
> 
> > That's even better, `glean_internal_info` it is! Yes, let's add this to an internal document. Please note that this also requires a change to the parser itself (and the related docs).
> 
> Why would it? The parser already doesn't handle `glean_client_info` in any special way.

I was wrong, this is not required.

PR #121, opened 2019-06-05
Start documenting public parts of all code

badboy:
This is nowhere near complete, but it gets us started.

### DIFF
--- a/glean-core/src/metrics/boolean.rs
+++ b/glean-core/src/metrics/boolean.rs
@@ -8,7 +8,7 @@ use crate::storage::StorageManager;
 use crate::CommonMetricData;
 use crate::Glean;
 
-/// # A boolean metric
+/// A boolean metric.
 ///
 /// Records a simple flag.
 #[derive(Clone, Debug)]
@@ -27,12 +27,17 @@ impl MetricType for BooleanMetric {
 }
 
 impl BooleanMetric {
-    /// Create a new metric
+    /// Create a new boolean metric.
     pub fn new(meta: CommonMetricData) -> Self {
         Self { meta }
     }
 
-    /// Set to the specified boolean value
+    /// Set to the specified boolean value.
+    ///
+    /// ## Arguments
+    ///
+    /// * `glean` - the Glean instance this metric belongs to.
+    /// * `value` - the value to set.
     pub fn set(&self, glean: &Glean, value: bool) {
         if !self.should_record(glean) {
             return;
--- a/glean-core/src/metrics/boolean.rs
+++ b/glean-core/src/metrics/boolean.rs
@@ -8,6 +8,9 @@ use crate::storage::StorageManager;
 use crate::CommonMetricData;
 use crate::Glean;
 
+/// # A boolean metric
+///
+/// Records a simple flag.
 #[derive(Clone, Debug)]
 pub struct BooleanMetric {
     meta: CommonMetricData,
@@ -24,10 +27,12 @@ impl MetricType for BooleanMetric {
 }
 
 impl BooleanMetric {
+    /// Create a new metric
     pub fn new(meta: CommonMetricData) -> Self {
         Self { meta }
     }
 
+    /// Set to the specified boolean value
     pub fn set(&self, glean: &Glean, value: bool) {
         if !self.should_record(glean) {
             return;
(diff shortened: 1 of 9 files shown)


glean-core/src/metrics/counter.rs
```diff
@@ -25,10 +28,14 @@ impl MetricType for CounterMetric {
 }
 
 impl CounterMetric {
+    /// Create a new metric
     pub fn new(meta: CommonMetricData) -> Self {
         Self { meta }
     }
 
+    /// Increase the counter by `amount`
+    ///
```
Dexterp37: Can we document arguments here, using the format we're using in other parts of the code?

glean-core/src/metrics/mod.rs
```diff
@@ -2,6 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! # Metrics
+//!
+//! Glean supports different metric types to store data.
```
Dexterp37: nit: should this be more like "The different metric types supported by the Glean SDK to handle data"?

glean-core/src/metrics/mod.rs
```diff
@@ -30,27 +34,44 @@ pub use self::string_list::StringListMetric;
 pub use self::time_unit::TimeUnit;
 pub use self::uuid::UuidMetric;
 
+/// The available metrics
+///
+/// This is the in-memory and persisted layout of a metric
 #[derive(Serialize, Deserialize, Clone, Debug)]
 pub enum Metric {
+    /// A boolean metric. See [`BooleanMetric`](struct.BooleanMetric.html) for more information.
```
Dexterp37: Ah, doesn't autolink/discover things?

glean-core/src/metrics/mod.rs
```diff
@@ -30,27 +34,44 @@ pub use self::string_list::StringListMetric;
...
+    ///
+    /// This depends on the metrics own state, as determined by its metadata,
+    /// and whether upload is enabled on the Glean object.
     fn should_record(&self, glean: &Glean) -> bool {
         glean.is_upload_enabled() && self.meta().should_record()
     }
 }
 
 impl Metric {
+    /// The category the metric fits into
```
Dexterp37: Can you expand on what is this used for? (i.e. to decide where to put stuff in the ping?)

glean-core/src/metrics/string.rs
```diff
@@ -11,6 +11,9 @@ use crate::Glean;
 
 const MAX_LENGTH_VALUE: usize = 100;
 
+/// # A string metric
+///
+/// Record an Unicode string value with arbitrary content.
```
Dexterp37: Should we mention that it's length limited and point to the `MAX_LENGTH_VALUE` above?

glean-core/src/metrics/string.rs
```diff
@@ -27,10 +30,12 @@ impl MetricType for StringMetric {
 }
 
 impl StringMetric {
+    /// Create a new metric
     pub fn new(meta: CommonMetricData) -> Self {
         Self { meta }
     }
 
+    /// Set to the specified value
```
Dexterp37: nit: document the parameters

glean-core/src/metrics/boolean.rs
```diff
@@ -24,10 +27,12 @@ impl MetricType for BooleanMetric {
 }
 
 impl BooleanMetric {
+    /// Create a new metric
```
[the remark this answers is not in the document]
badboy: That is an excellent question. Let me see what the standard library does.

glean-core/src/metrics/mod.rs
badboy: So there is an [intra-doc links RFC](https://github.com/rust-lang/rfcs/blob/master/text/1946-intra-rustdoc-links.md) and it seems part of this is implemented, but I don't think it's stable yet (at least I didn't get it to work, I'll ask around)

Dexterp37: Ah, interesting. I'm fine with leaving this as it is for now, as long as we file a bug to fix it in the future :)

glean-core/src/metrics/boolean.rs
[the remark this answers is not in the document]
badboy: Let's require it (e.g. [HashMap does it](https://doc.rust-lang.org/src/std/collections/hash/map.rs.html#684))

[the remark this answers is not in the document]
Dexterp37: Thank you for checking this out!

PR #122, opened 2019-06-05
Sample should be limited

badboy:
The limit was recently increased. In order to record something into the
metrics ping, we now need to add more.

### DIFF
--- a/glean-core/examples/sample.rs
+++ b/glean-core/examples/sample.rs
@@ -67,12 +67,7 @@ fn main() {
         .unwrap();
     println!("Core Ping:\n{}", ping);
 
-    let mut long_string = std::iter::repeat('x').take(49).collect::<String>();
-    long_string.push('a');
-    long_string.push('b');
-    local_metric.set(&glean, long_string);
-    let ping = ping_maker
-        .collect_string(glean.storage(), glean.get_ping_by_name("metrics").unwrap())
-        .unwrap();
-    println!("Metrics Ping:\n{}", ping);
+    let ping =
+        ping_maker.collect_string(glean.storage(), glean.get_ping_by_name("metrics").unwrap());
+    println!("Metrics Ping: {:?}", ping);
 }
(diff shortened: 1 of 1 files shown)


glean-core/examples/sample.rs
```diff
@@ -64,14 +65,16 @@ fn main() {
     let ping = ping_maker
         .collect_string(glean.storage(), glean.get_ping_by_name("core").unwrap())
         .unwrap();
-    println!("Ping:\n{}", ping);
+    println!("Core Ping:\n{}", ping);
 
-    let mut long_string = std::iter::repeat('x').take(49).collect::<String>();
+    // Cause some error reporting, which will land in the metrics ping.
+    // The `local_metric` itself is not in the metrics ping.
+    let mut long_string = std::iter::repeat('x').take(99).collect::<String>();
```
Dexterp37: I think there was a test case for this already, why should we clutter this sample a bit more?

badboy: Hm, it was in there before and was useful for a quick one-off test.
But I'm also fine with keeping the sample much more minimal and thus removing this part.

Dexterp37: I personally feel like we should keep examples to the bare minimum :)

glean-core/examples/sample.rs
```diff
@@ -19,59 +19,55 @@ fn main() {
...
     list.add(&glean, "once");
     list.add(&glean, "upon");
 
     let ping_maker = PingMaker::new();
     let ping = ping_maker
-        .collect_string(glean.storage(), glean.get_ping_by_name("core").unwrap())
+        .collect_string(glean.storage(), glean.get_ping_by_name("baseline").unwrap())
         .unwrap();
-    println!("Ping:\n{}", ping);
+    println!("Core Ping:\n{}", ping);
```
Dexterp37: nit: Baseline Ping?

PR #123, opened 2019-06-05
Implement start_time/end_time for pings

badboy:
Implement start and end times for pings

The start time can be stored and persisted as a datetime metric.
The end time is only ever needed in-memory for the moment the ping is
assembled and stored.

We also store the object creation time in the Glean object itself (which
is used as the first start time), because:

* that was the easy way
* glean-ac does something similar (storing it in the PingMaker object)
* we don't have global metrics in the core library itself on which we
  could just set the date

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -10,8 +10,9 @@ use log::info;
 use serde_json::{json, Value as JsonValue};
 
 use crate::common_metric_data::{CommonMetricData, Lifetime};
-use crate::metrics::{CounterMetric, Metric, MetricType, PingType};
+use crate::metrics::{CounterMetric, DatetimeMetric, Metric, MetricType, PingType, TimeUnit};
 use crate::storage::StorageManager;
+use crate::util::{get_iso_time_string, local_now_with_offset};
 use crate::Glean;
 
 // An internal ping name, not to be touched by anything else
@@ -69,12 +70,44 @@ impl PingMaker {
         current_seq as usize
     }
 
+    /// Get the formatted start and end times for this ping and update for the next ping.
+    fn get_start_end_times(&self, glean: &Glean, storage_name: &str) -> (String, String) {
+        let time_unit = TimeUnit::Minute;
+
+        let start_time = DatetimeMetric::new(
+            CommonMetricData {
+                name: format!("{}#start", storage_name),
+                category: "".into(),
+                send_in_pings: vec![INTERNAL_STORAGE.into()],
+                lifetime: Lifetime::User,
+                ..Default::default()
+            },
+            time_unit,
+        );
+
+        // "start_time" is the time the ping was generated the last time.
+        // If not available, we use the date the Glean object was initialized.
+        let start_time_data = start_time
+            .get_value(glean, INTERNAL_STORAGE)
+            .unwrap_or_else(|| glean.start_time());
+        let end_time_data = local_now_with_offset();
+
+        // Update the start time with the current time.
+        start_time.set(glean, end_time_data);
+
+        // Format the times.
+        let start_time_data = get_iso_time_string(start_time_data, time_unit);
+        let end_time_data = get_iso_time_string(end_time_data, time_unit);
+        (start_time_data, end_time_data)
+    }
+
     fn get_ping_info(&self, glean: &Glean, storage_name: &str) -> JsonValue {
+        let (start_time, end_time) = self.get_start_end_times(glean, storage_name);
         json!({
             "ping_type": storage_name,
             "seq": self.get_ping_seq(glean, storage_name),
-            "start_time": "2019-03-29T09:50-04:00",
-            "end_time": "2019-03-29T09:53-04:00"
+            "start_time": start_time,
+            "end_time": end_time,
         })
     }
 
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -8,6 +8,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 
+use chrono::{DateTime, FixedOffset};
 use uuid::Uuid;
 
 mod common_metric_data;
@@ -23,13 +24,13 @@ mod util;
 
 pub use crate::common_metric_data::{CommonMetricData, Lifetime};
 use crate::database::Database;
+pub use crate::error::{Error, Result};
 pub use crate::error_recording::{test_get_num_recorded_errors, ErrorType};
 use crate::internal_metrics::CoreMetrics;
+use crate::metrics::PingType;
 use crate::ping::PingMaker;
 use crate::storage::StorageManager;
-use crate::util::sanitize_application_id;
-pub use error::{Error, Result};
-use metrics::PingType;
+use crate::util::{local_now_with_offset, sanitize_application_id};
 
 const GLEAN_SCHEMA_VERSION: u32 = 1;
 
@@ -42,6 +43,7 @@ pub struct Glean {
     data_path: PathBuf,
     application_id: String,
     ping_registry: HashMap<String, PingType>,
+    start_time: DateTime<FixedOffset>,
 }
 
 impl Glean {
@@ -61,6 +63,7 @@ impl Glean {
             data_path: PathBuf::from(data_path),
             application_id,
             ping_registry: HashMap::new(),
+            start_time: local_now_with_offset(),
         };
         glean.initialize_core_metrics()?;
         glean.initialized = true;
@@ -188,6 +191,11 @@ impl Glean {
         }
 
         self.ping_registry.insert(ping.name.clone(), ping.clone());
+    }
+
+    /// Get create time of the Glean object.
+    pub(crate) fn start_time(&self) -> DateTime<FixedOffset> {
+        self.start_time
     }
 }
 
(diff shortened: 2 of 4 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+            glean.storage(),
+            INTERNAL_STORAGE,
+            &start_time.meta().identifier(),
+        ) {
+            Some(Metric::Datetime(t, tu)) => {
+                log::info!("datetime start found: {:?}", t);
+                Metric::Datetime(t, tu)
+            }
+            _ => {
+                let now: DateTime<Local> = Local::now();
```
Dexterp37: Can we refactor this into some utils file as `get_local_time_with_offset` (or some better naming)?

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+                let naive = now.naive_utc();
+                let fixed_tz = Local.offset_from_utc_datetime(&naive);
+                let now_fixed: DateTime<FixedOffset> = fixed_tz.from_utc_datetime(&naive);
+                log::info!("datetime start missing, creating now: {:?}", now_fixed);
+                Metric::Datetime(now_fixed, time_unit)
+            }
+        };
+
+        let now: DateTime<Local> = Local::now();
```
Dexterp37: Then we could use the same function here.

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+        let naive = now.naive_utc();
+        let fixed_tz = Local.offset_from_utc_datetime(&naive);
+        let now_fixed: DateTime<FixedOffset> = fixed_tz.from_utc_datetime(&naive);
+
+        log::info!("datetime end setting to now: {:?}", now_fixed);
+        end_time.set(glean, now_fixed);
+
+        let end_time_data = match StorageManager.snapshot_metric(
```
Dexterp37: Maybe we have a case for refactoring this in `DatetimeMetric`, having it as package private (I hope it's the right name :D ). We'd use an identical code for the testing function, so we might as well just write it once.

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+                log::info!("datetime start found: {:?}", t);
+                Metric::Datetime(t, tu)
+            }
+            _ => {
+                let now: DateTime<Local> = Local::now();
+                let naive = now.naive_utc();
+                let fixed_tz = Local.offset_from_utc_datetime(&naive);
+                let now_fixed: DateTime<FixedOffset> = fixed_tz.from_utc_datetime(&naive);
+                log::info!("datetime start missing, creating now: {:?}", now_fixed);
+                Metric::Datetime(now_fixed, time_unit)
```
Dexterp37: This should not really be `now`, but the time glean-core was initialized. See the [glean-ac implementation](https://github.com/mozilla-mobile/android-components/blob/e2a019f6f8ffd3d2fbf514af0f79fa7d2c21cc25/components/service/glean/src/main/java/mozilla/components/service/glean/ping/PingMaker.kt#L83-L92). I'm afraid we were lacking test coverage for that, so it would be great if you could add it here ;)

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+        let end_time_data = match StorageManager.snapshot_metric(
+            glean.storage(),
+            INTERNAL_STORAGE,
+            &end_time.meta().identifier(),
+        ) {
+            Some(Metric::Datetime(t, tu)) => {
+                start_time.set(glean, t);
+                Metric::Datetime(t, tu)
+            }
+            _ => unreachable!("End time just set, but now missing"),
```
Dexterp37: Is this possible in practice? Does the `unreachable!` string get printed to the logs in case of problems? I'd like it to show on CI logs, if this ever happens

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+                name: format!("{}#start", storage_name),
+                category: "".into(),
+                send_in_pings: vec![INTERNAL_STORAGE.into()],
+                lifetime: Lifetime::User,
+                ..Default::default()
+            },
+            time_unit,
+        );
+
+        let end_time = DatetimeMetric::new(
```
Dexterp37: Do we need this one? I think we only require the `start_time` one to be stored in memory, `end_time` will always be `Metric::Datetime(fixed_now, time_unit)`

glean-core/src/ping/mod.rs
```diff
@@ -67,12 +70,80 @@ impl PingMaker {
...
+            }
+        };
+
+        let now: DateTime<Local> = Local::now();
+        let naive = now.naive_utc();
+        let fixed_tz = Local.offset_from_utc_datetime(&naive);
+        let now_fixed: DateTime<FixedOffset> = fixed_tz.from_utc_datetime(&naive);
+
+        log::info!("datetime end setting to now: {:?}", now_fixed);
+        end_time.set(glean, now_fixed);
```
Dexterp37: I don't think we need this, we can just set `let end_time_data = Metric::Datetime(fixed_now, time_unit);` and then do `start_time.set(glean, t);`

badboy:
This is still lacking the tests, but I'm working on these now. Please give the implementation a look though

badboy:
Unfortunately I'm not sure what to test here.
[We do test that start and end time are present and are roughly in order](https://github.com/mozilla/glean/blob/87270470e210ddbfabe84468e033bce1cd55d72a/glean-core/tests/ping_maker.rs#L41-L47).
But because it's minute-resolution we can't really test that it increases for subsequent pings.

* We can't really test in the same file that the first date matches the creation date, unless we make that public (which I was trying to avoid).
* We could test this as a "unit test" inside `ping/mod.rs`, but that would require setting up a bunch of stuff such as the `Glean` object, and therefore making it not much of a "unit test" any more.
* We could defer the test to happen on the Kotlin side (where we currently don't test this, but we could copy over glean AC tests?)

glean-core/src/ping/mod.rs
```diff
@@ -69,12 +70,44 @@ impl PingMaker {
         current_seq as usize
     }
 
+    /// Get the formatted start and end times for this ping and update for the next ping.
+    fn get_start_end_times(&self, glean: &Glean, storage_name: &str) -> (String, String) {
```
Dexterp37: What I would try to test here:

- do we get the Glean object start time if no data is stored?
- does the `start_time` get set to the `end_time`?

Dexterp37:
> 
> 
> Unfortunately I'm not sure what to test here.
> [We do test that start and end time are present and are roughly in order](https://github.com/mozilla/glean/blob/87270470e210ddbfabe84468e033bce1cd55d72a/glean-core/tests/ping_maker.rs#L41-L47).
> But because it's minute-resolution we can't really test that it increases for subsequent pings.
> 
>     * We can't really test in the same file that the first date matches the creation date, unless we make that public (which I was trying to avoid).
> 
>     * We could test this as a "unit test" inside `ping/mod.rs`, but that would require setting up a bunch of stuff such as the `Glean` object, and therefore making it not much of a "unit test" any more.

Oh, I see your concerns about testing. What about the things I [highlighted here](https://github.com/mozilla/glean/pull/123#discussion_r291190353)? 

>     * We could defer the test to happen on the Kotlin side (where we currently don't test this, but we could copy over glean AC tests?)

The test coverage on the Kotlin side doesn't exist in glean-ac :(

> 
> wdyt, @Dexterp37, @mdboom?

glean-core/src/ping/mod.rs
badboy: 1. That would require setting up the Glean object. in `glean-core/tests/*` that is easy to do, but there we can't (right now) access `start_time` of that object as it's not exposed (I can make it public, though I would not want to). Plus, it gets truncated to minutes and therefore `Glean::new().start_time()` formatted to minutes should be the same as `Local::now()`, therefore not testing much. (We don't [currently] have any mocking in place to change this)
	* We _could_ make `get_start_end_times` return the `DateTime<FixedOffset>` instead, and defer the string formatting.
2. Again, in all our tests `start_time` and `end_time` tend to be within the same minute already (we could get the full timestamp from the database and compare that, but to me that seems a bit far-reaching)

mdboom: I don't know if this is possible in Rust, but could we mock `Local::now()` to return whatever time we want?  Then we could simulate many minutes passing.

We do something similar for the `TimingManager` in glean-ac.

badboy: Not easily. 
There are some mocking frameworks (I never used one), but I assume they all depend on mocking by using traits in places.

Dexterp37: I'd vote for moving on and filing a bug for adding this specific testing (and one for testing mocking frameworks.. :D )

badboy: On it!

badboy:
And filed https://bugzilla.mozilla.org/show_bug.cgi?id=1557354 for expanding the testing PAI

PR #128, opened 2019-06-06
Correctly set first_run_date as a datetime metric

badboy:

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -72,9 +72,7 @@ impl Glean {
 
     fn initialize_core_metrics(&mut self) -> Result<()> {
         if first_run::is_first_run(&self.data_path)? {
-            self.core_metrics
-                .first_run_date
-                .set(self, "2019-05-09-04:00");
+            self.core_metrics.first_run_date.set(self, None);
         }
         self.core_metrics.client_id.generate_if_missing(self);
         Ok(())
--- a/glean-core/src/metrics/datetime.rs
+++ b/glean-core/src/metrics/datetime.rs
@@ -9,7 +9,7 @@ use crate::metrics::time_unit::TimeUnit;
 use crate::metrics::Metric;
 use crate::metrics::MetricType;
 use crate::storage::StorageManager;
-use crate::util::get_iso_time_string;
+use crate::util::{get_iso_time_string, local_now_with_offset};
 use crate::CommonMetricData;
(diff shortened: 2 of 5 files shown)


glean-core/src/lib.rs
```diff
@@ -72,9 +72,7 @@ impl Glean {
 
     fn initialize_core_metrics(&mut self) -> Result<()> {
         if first_run::is_first_run(&self.data_path)? {
```
mdboom: Just an FYI: In https://github.com/mozilla-mobile/android-components/pull/2838, we ultimately got rid of this side-channel information for this and just set first_run whenever that metric doesn't exist in the database.  I think we defer that until we port that PR to glean-core, but just wanted to note it here.

glean-core/src/metrics/datetime.rs
```diff
@@ -108,6 +108,21 @@ impl DatetimeMetric {
         glean.storage().record(&self.meta, &value)
     }
 
+    /// Set the metric to the current date/time, including the local timezone offset.
+    ///
+    /// ## Arguments:
+    ///
+    /// * `glean` - the Glean instance this metric belongs to.
+    pub fn set_now(&self, glean: &Glean) {
```
mdboom: I don't feel strongly, but would this be better as `set(&self, glean: &Glean, value: Option<DateTime<FixedOffset>>)` to more closely hew to the Kotlin API?  I'm mainly thinking ahead to when we have a proper public Rust API and we want to document it easily in parallel with the Kotlin API.

glean-core/src/metrics/datetime.rs
```diff
@@ -108,6 +108,21 @@ impl DatetimeMetric {
...
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let value = local_now_with_offset();
+        let value = Metric::Datetime(value, self.time_unit);
```
mdboom: Maybe call `self.set(glean, value)` here instead to reduce code duplication?

glean-core/src/lib.rs
badboy: Do we have a bug for that in Glean-core?

mdboom: I couldn't find one, so just made one: https://bugzilla.mozilla.org/show_bug.cgi?id=1557366

badboy:
Folded the `set_now` into `set` by requiring an option.

PR #129, opened 2019-06-06
Bug 1552862 - Implement StringList metric type in glean-core

chutten:

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
@@ -4,18 +4,58 @@
 
 package mozilla.telemetry.glean.private
 
-import android.util.Log
 import androidx.annotation.VisibleForTesting
+import com.sun.jna.StringArray
+import mozilla.components.support.ktx.android.org.json.toList
+import mozilla.telemetry.glean.Dispatchers
+import mozilla.telemetry.glean.Glean
+import mozilla.telemetry.glean.rust.getAndConsumeRustString
+import mozilla.telemetry.glean.rust.LibGleanFFI
+import mozilla.telemetry.glean.rust.RustError
+import mozilla.telemetry.glean.rust.toBoolean
+import mozilla.telemetry.glean.rust.toByte
+import org.json.JSONArray
 
 class StringListMetricType(
-    disabled: Boolean,
-    category: String,
-    lifetime: Lifetime,
-    name: String,
+    private var handle: Long,
     private val sendInPings: List<String>
 ) {
-    companion object {
-        private val LOG_TAG: String = "glean/StringListMetricType"
+    /**
+     * The public constructor used by automatically-generated metrics.
+     */
+    constructor(
+        disabled: Boolean,
+        category: String,
+        lifetime: Lifetime,
+        name: String,
+        sendInPings: List<String>
+    ) : this(handle = 0, sendInPings = sendInPings) {
+        val ffiPingsList = StringArray(sendInPings.toTypedArray(), "utf-8")
+        this.handle = LibGleanFFI.INSTANCE.glean_new_string_list_metric(
+            category = category,
+            name = name,
+            send_in_pings = ffiPingsList,
+            send_in_pings_len = sendInPings.size,
+            lifetime = lifetime.ordinal,
+            disabled = disabled.toByte())
+    }
+
+    protected fun finalize() {
+        if (this.handle != 0L) {
+            val error = RustError.ByReference()
+            LibGleanFFI.INSTANCE.glean_destroy_string_list_metric(this.handle, error)
+        }
+        // Do nothing with the error, for now.
+        // It is expected to only ever error if this.handle's invalid.
+    }
+
+    private fun shouldRecord(): Boolean {
+        // Don't record metrics if Glean isn't initialized.
+        if (!Glean.isInitialized()) {
+            return false
+        }
+
+        return LibGleanFFI.INSTANCE.glean_string_list_should_record(Glean.handle, this.handle).toBoolean()
     }
 
     /**
@@ -26,20 +66,17 @@ class StringListMetricType(
      *              this string is [MAX_STRING_LENGTH].
      */
     fun add(value: String) {
-        // if (!shouldRecord(logger)) {
-        //     return
-        // }
+        if (!shouldRecord()) {
+            return
+        }
 
-        // @Suppress("EXPERIMENTAL_API_USAGE")
-        // Dispatchers.API.launch {
-        //     // Delegate storing the string to the storage engine.
-        //     StringListsStorageEngine.add(
-        //         metricData = this@StringListMetricType,
-        //         value = value
-        //     )
-        // }
-        // TODO: stub
-        Log.e(LOG_TAG, "StringListMetricType is a stub")
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.launch {
+            LibGleanFFI.INSTANCE.glean_string_list_add(
+                Glean.handle,
+                this@StringListMetricType.handle,
+                value)
+        }
     }
 
     /**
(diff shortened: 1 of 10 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
```diff
@@ -98,11 +135,24 @@ class StringListMetricType(
...
+        val ptr = LibGleanFFI.INSTANCE.glean_string_list_test_get_value_as_string(Glean.handle, this.handle, pingName)!!
+        try {
+            jsonRes = JSONArray(ptr.getAndConsumeRustString())
+        } catch (e: org.json.JSONException) {
+            throw NullPointerException()
+        }
+        val stringList: MutableList<String> = mutableListOf()
+        for (i in 0..(jsonRes.length() - 1)) {
+            stringList.add(jsonRes.getString(i))
+        }
```
mdboom: Kotlin has this neat/horrifying feature that lets you add new methods to external classes.  In android-components, they use that to add [utility methods](https://github.com/mozilla-mobile/android-components/blob/742786c25f88ce8ab9dd3203a04104082e67eca0/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/org/json/JSONArray.kt#L34) on top of the `org.json` library.  We could do something like that here.

Even if we don't go that far, maybe we could use the functional approach [in this toList function](https://github.com/mozilla-mobile/android-components/blob/742786c25f88ce8ab9dd3203a04104082e67eca0/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/org/json/JSONArray.kt#L34).

chutten: Oh geez, that's... well that's a thing isn't it. I'm not a Kotlin dev by trade, so I'll bow to whatever is most Kotlin-idiomatic.

That being said, where does `asSequence` come from? I skim-read the [`JSONArray` docs](https://developer.android.com/reference/kotlin/org/json/JSONArray) and couldn't find any way to iterate over it except taking the length and doing a ranged `for`.

mdboom: Oh, geeze, it looks like `asSequence` is a [a-c custom thing](https://github.com/mozilla-mobile/android-components/blob/742786c25f88ce8ab9dd3203a04104082e67eca0/components/support/ktx/src/main/java/mozilla/components/support/ktx/android/org/json/JSONArray.kt#L13), too.  Doing some sort of iterator magic.  Given all that, I'd say what you've got here is more than good enough...

glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
```diff
@@ -4,20 +4,63 @@
...
 class StringListMetricType(
-    disabled: Boolean,
-    category: String,
-    lifetime: Lifetime,
-    name: String,
+    private var handle: Long,
     private val sendInPings: List<String>
 ) {
     companion object {
         private val LOG_TAG: String = "glean/StringListMetricType"
```
Dexterp37: nit: let's remove this if it's not used anymore.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
Dexterp37: We're [importing a-c support libraries](https://github.com/mozilla/glean/blob/ff85c3cfc969668123de95a83ebd468f3821c68b/glean-core/android/build.gradle#L135), so we could use it ;)

glean-core/src/metrics/string_list.rs
```diff
@@ -133,4 +134,17 @@ impl StringListMetric {
         let value = Metric::StringList(value);
         glean.storage().record(&self.meta, &value);
     }
+
+    /// **Test-only API (exported for FFI purposes).**
+    ///
+    /// Get the currently-stored values as a JSON String.
```
Dexterp37: nit: what's the expected format? `["string1", "string2", ..., "string20"]`? If so, can you please add it here, in the docs, as an example?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
badboy: If it really is easy to use, we should totally do that (do we currently depend on any other imported functionality from that lib?)

@Dexterp37: Could you advise chutten on what to import here?

If that doesn't work out though staying with the loop here doesn't seem too bad (afterall it's the test function)

glean-core/ffi/src/lib.rs
```diff
@@ -456,6 +481,74 @@ pub extern "C" fn glean_string_test_get_value(
...
+        STRING_LIST_METRICS.call_infallible(metric_id, |metric| {
+            metric
+                .test_get_value_as_string(glean, storage_name.as_str())
+                .is_some()
+        })
+    })
+}
+
+#[no_mangle]
+pub extern "C" fn glean_string_list_test_get_value_as_string(
```
badboy: That's quite a name.
I wonder if `_as_json` would be a better naming here?

glean-core/src/error_recording.rs
```diff
@@ -100,7 +100,7 @@ pub fn test_get_num_recorded_errors(
     meta: &CommonMetricData,
     error: ErrorType,
     ping_name: Option<&str>,
-) -> i32 {
+) -> Result<i32, String> {
```
badboy: Normally we do have a dedicated error type, but given that this should be test only and we therefore don't _want_ to expose the error in general, opting for a String error here seems ok.

glean-core/tests/string_list.rs
```diff
@@ -23,17 +26,142 @@ fn list_can_store_multiple_items() {
     let snapshot = glean.snapshot("core", false);
     assert!(snapshot.contains(r#""local.list": ["#));
     assert!(snapshot.contains(r#""first""#));
+    assert!(!snapshot.contains(r#""second""#));
```
badboy: Yikes, we really should convert this to the proper testing functions instead of doing ugly string matching against a potential json format.

[I filed a bug so we can fix this in a follow-up](https://bugzilla.mozilla.org/show_bug.cgi?id=1557648)

glean-core/tests/string_list.rs
```diff
@@ -23,17 +26,142 @@ fn list_can_store_multiple_items() {
...
+        disabled: false,
+        lifetime: Lifetime::Ping,
+    });
+
+    let test_string = "0123456789".repeat(20);
+    metric.add(&glean, test_string.clone());
+
+    // Ensure the string was truncated to the proper length.
+    assert_eq!(
+        format!("[\"{}\"]", test_string[..50].to_string()),
```
badboy: Instead of constructing the JSON string here ourselves how about we can expose a proper `test_get_value` function that returns `Vec<String>`

glean-core/src/metrics/string_list.rs
```diff
@@ -133,4 +134,17 @@ impl StringListMetric {
...
+    ///
+    /// This doesn't clear the stored value.
+    pub fn test_get_value_as_string(&self, glean: &Glean, storage_name: &str) -> Option<String> {
```
badboy: Follow up from below: Let's implement a `test_get_value` function returning `Vec<String>` to use in Rust tests.

this `as_string` function can then call that as well.

glean-core/tests/string_list.rs
```diff
@@ -23,17 +26,142 @@ fn list_can_store_multiple_items() {
...
+
+    metric.set(&glean, vec!["test_string_2".repeat(20)]);
+
+    // Ensure the stringlist was not set.
+    assert_eq!(None, metric.test_get_value_as_string(&glean, "store1"));
+
+    // Ensure no error was recorded.
+    assert!(
+        test_get_num_recorded_errors(&glean, metric.meta(), ErrorType::InvalidValue, None).is_err()
+    );
```
badboy: This lacks a test that we limit the number of elements in a list.
We should check that

* We can add N items
* On adding the N+1 item the list stays unchanged (and I think we log an error?)

glean-core/android/src/test/java/mozilla/telemetry/glean/private/StringListMetricTypeTest.kt
```diff
@@ -33,7 +32,6 @@ class StringListMetricTypeTest {
         resetGlean()
     }
 
-    @Ignore("StringListMetricType is a stub")
     @Test
```
badboy: Yey for the fact that simply enabling them works!

glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
Dexterp37: I think this could be as easy as `jsonRes.toList<String>()` (see [the test code](https://github.com/mozilla-mobile/android-components/blob/742786c25f88ce8ab9dd3203a04104082e67eca0/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/org/json/JSONArrayTest.kt#L59))

badboy: Ah, but this might require an `import mozilla.components.support.ktx.android.org.json.toList` in this package right?

Dexterp37: Yup, shouldn't be too much of a big deal. We're already importing the dependency

mdboom: Cool.  I forgot we had that dependency already.  In that case, let's use it.

chutten:
Oh, I should clarify: This is ready for another look.

PR #131, opened 2019-06-11
Bug 1557828 - Record error on too many list items

badboy:
Followup for #129.

Also see commit messages for details.

### DIFF
--- a/glean-core/src/metrics/string_list.rs
+++ b/glean-core/src/metrics/string_list.rs
@@ -66,25 +66,29 @@ impl StringListMetric {
             value
         };
 
+        let mut error = None;
         glean
             .storage()
             .record_with(&self.meta, |old_value| match old_value {
                 Some(Metric::StringList(mut old_value)) => {
                     if old_value.len() == MAX_LIST_LENGTH {
-                        //let msg = format!(
-                        //    "String list length of {} exceeds maximum of {}",
-                        //    old_value.len() + 1,
-                        //    MAX_LIST_LENGTH
-                        //);
-                        // TODO (bug 1557828) - record the error when it doesn't deadlock.
-                        // record_error(glean, &self.meta, ErrorType::InvalidValue, msg);
-                        return Metric::StringList(old_value);
+                        let msg = format!(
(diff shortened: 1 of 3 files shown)


glean-core/src/metrics/string_list.rs
```diff
@@ -66,25 +66,29 @@ impl StringListMetric {
...
-                        //);
-                        // TODO (bug 1557828) - record the error when it doesn't deadlock.
-                        // record_error(glean, &self.meta, ErrorType::InvalidValue, msg);
+                        let msg = format!(
+                            "String list length of {} exceeds maximum of {}",
+                            old_value.len() + 1,
+                            MAX_LIST_LENGTH
+                        );
+                        error = Some(msg);
                         return Metric::StringList(old_value);
```
mdboom: I find the early return inside an anonymous function a bit confusing.

Maybe move the `old_value.push(value.clone())` into an `else` clause, but always return `Metric::StringList(old_value)` in the same place at the end of the expression?

glean-core/tests/string_list.rs
```diff
@@ -195,7 +195,10 @@ fn string_lists_dont_exceed_max_items() {
 
     // TODO (bug 1557828) - Uncomment when the error starts being recorded again.
```
Dexterp37: nit: let's remove the comment here

glean-core/tests/string_list.rs
```diff
@@ -211,7 +214,7 @@ fn string_lists_dont_exceed_max_items() {
 
     // TODO (bug 1557828) - Increment by 1 when the previous error starts being recorded again.
```
Dexterp37: nit: let's remove the comment here

badboy:
> If I understand this correctly we're forced to use a mut transform because that's the only way we can get the error out?

Yes, before we took an `Fn`, which is a closure without an environment, which therefore it cannot read or write to anything outside of it.
The storage doesn't really care about what the closure does, so it's fine relaxing the bound to `FnMut` which allows the closure to reference stuff on its outside (by actually taking references to stuff outside).

PR #133, opened 2019-06-11
Bug 1554973 - Document all public items

badboy:

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -6,7 +6,7 @@ use std::convert::TryFrom;
 
 use crate::error::{Error, ErrorKind};
 
-/// Different lifetimes for metrics.
+/// The supported metrics' lifetimes.
 ///
 /// A metric's lifetime determines when its stored data gets reset.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -6,6 +6,9 @@ use std::convert::TryFrom;
 
 use crate::error::{Error, ErrorKind};
 
+/// Different lifetimes for metrics.
+///
+/// A metric's lifetime determines when its stored data gets reset.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
 pub enum Lifetime {
     /// The metric is reset with each sent ping
@@ -23,6 +26,7 @@ impl Default for Lifetime {
 }
 
 impl Lifetime {
+    /// String representation of the lifetime.
     pub fn as_str(self) -> &'static str {
         match self {
             Lifetime::Ping => "ping",
@@ -45,17 +49,25 @@ impl TryFrom<i32> for Lifetime {
     }
 }
 
+/// The common set of data shared across all different metric types.
 #[derive(Default, Debug, Clone)]
 pub struct CommonMetricData {
+    /// The metric's name.
     pub name: String,
+    /// The metric's category.
     pub category: String,
+    /// List of pings to include this metric in.
     pub send_in_pings: Vec<String>,
+    /// The metric's lifetime.
     pub lifetime: Lifetime,
+    /// Whether or not the metric is disabled.
+    ///
+    /// Disabled metrics are never recorded.
     pub disabled: bool,
 }
 
 impl CommonMetricData {
-    /// Create a new metadata object
+    /// Create a new metadata object.
     pub fn new<A: Into<String>, B: Into<String>, C: Into<String>>(
         category: A,
         name: B,
@@ -69,6 +81,10 @@ impl CommonMetricData {
         }
     }
 
+    /// The metric's unique identifier.
+    ///
+    /// If `category` is empty, it's just the name.
+    /// Otherwise, it's the combination of the metric's `category` and `name`.
     pub fn identifier(&self) -> String {
         if self.category.is_empty() {
             self.name.clone()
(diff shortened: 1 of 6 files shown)


glean-core/src/common_metric_data.rs
```diff
@@ -45,17 +49,25 @@ impl TryFrom<i32> for Lifetime {
     }
 }
 
+/// The common set of data shared across all different metric types.
 #[derive(Default, Debug, Clone)]
 pub struct CommonMetricData {
+    /// The metric's name.
     pub name: String,
+    /// The metric's category.
```
mdboom: Is it worth adding `may contain dots to specify subcategories`?

glean-core/src/common_metric_data.rs
```diff
@@ -45,17 +49,25 @@ impl TryFrom<i32> for Lifetime {
...
     pub category: String,
+    /// List of pings to include this metric in.
```
mdboom:
```suggestion
    /// List of ping names to include this metric in.
```

To disambiguate from the `PingType` objects.  (It's clear from the data type, of course, so feel free to ignore...

glean-core/src/lib.rs
```diff
@@ -34,6 +43,30 @@ use crate::util::{local_now_with_offset, sanitize_application_id};
 
 const GLEAN_SCHEMA_VERSION: u32 = 1;
 
+/// The object holding meta information about a Glean instance.
```
mdboom: Is it worth saying something like: "In the language bindings, this is usually a singleton and all metric recording goes to a single instance of this object.  In the Rust core, it is possible to make instances of this to make testing easier."  (Maybe that's a developer note, though, not a docstring...)  I just feel it's important to call out because it somewhat informs the design.

glean-core/src/lib.rs
```diff
@@ -89,18 +122,32 @@ impl Glean {
...
     pub fn get_data_path(&self) -> &Path {
         &self.data_path
     }
 
+    /// Get a handle to the database.
     pub fn storage(&self) -> &Database {
         &self.data_store
     }
 
+    /// Take a snapshot for the givven store and optionally clear it.
```
mdboom:
```suggestion
    /// Take a snapshot for the given store and optionally clear it.
```

glean-core/src/lib.rs
```diff
@@ -89,18 +122,32 @@ impl Glean {
...
+    ///
+    /// ## Arguments
+    ///
+    /// * `store_name` - The store to snapshot.
+    /// * `clear_store` - Whether to clear the store after snapshotting.
+    ///
+    /// ## Return value
+    ///
+    /// Returns the snapshot as a JSON string.
```
mdboom:
```suggestion
    /// Returns the snapshot in a string encoded as JSON.
```

"JSON string" is a bit ambiguous -- it could mean [one of these](https://docs.serde.rs/serde_json/enum.Value.html#variant.String)

glean-core/src/metrics/time_unit.rs
```diff
@@ -11,12 +11,19 @@ use crate::error::{Error, ErrorKind};
 /// metric types (e.g. DatetimeMetric).
 #[derive(Copy, Clone, Debug, Deserialize, Serialize)]
 pub enum TimeUnit {
+    /// Format to nanosecond precision.
```
mdboom: I would say "truncate to".  It both (a) expresses that we use truncation and not rounding, (b) is generic between DateTime (where format makes sense) and Timespan (where it doesn't).

glean-core/src/common_metric_data.rs
```diff
@@ -6,6 +6,9 @@ use std::convert::TryFrom;
 
 use crate::error::{Error, ErrorKind};
 
+/// Different lifetimes for metrics.
```
Dexterp37: What about "The supported metrics' lifetimes"?

glean-core/src/ping/mod.rs
```diff
@@ -157,6 +172,17 @@ impl PingMaker {
         }))
     }
 
+    /// Collect a snapshot for the given ping from storage and attach required meta information as a JSON string.
```
mdboom:
```suggestion
    /// Collect a snapshot for the given ping from storage and attach required meta information, returning as a string containing JSON.
```

glean-core/src/storage/mod.rs
```diff
@@ -37,6 +40,18 @@ fn snapshot_labeled_metrics(
...
+    ///
+    /// ## Arguments
+    ///
+    /// * `storage` - the database to read from.
+    /// * `store_name` - the store to snapshot.
+    /// * `clear_store` - whether to clear the data after snapshotting.
+    ///
+    /// ## Return value
+    ///
+    /// Returns a JSON string of the stored data.
```
mdboom: Ditto on JSON string (if you agree).

glean-core/src/lib.rs
Dexterp37: "givven" -> "given"

glean-core/src/ping/mod.rs
```diff
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+//! Ping collection, assembly & sending.
```
Dexterp37: nit: sending.. maybe "submission"?

glean-core/src/ping/mod.rs
```diff
@@ -40,6 +43,7 @@ impl Default for PingMaker {
 }
 
 impl PingMaker {
+    /// Create a new PingMaker
```
Dexterp37: nit: missing trailing dot `.`

glean-core/src/common_metric_data.rs
badboy: Hm, to me that implies that we somehow handle "sub"categories, which we don't. Even in the YAML this it not another nested object, it's just a flat name with dots.

mdboom: Yeah -- that's fine.  I don't feel too strongly about this -- particularly at this point (it's more important for folks writing metrics.yaml files).  Let's leave it as-is.

PR #148, opened 2019-06-21
Bug 1552863 - Implement timespan metric

badboy:
https://github.com/mozilla-mobile/android-components/pull/3488 should land first.

TODO:

* [x] `setRawNanos` method
* [x] correct encoding in ping (it should include the string representation of the time unit)


glean-core/src/metrics/timespan.rs
```diff
@@ -0,0 +1,106 @@
...
+        }
+    }
+
+    /// Start tracking time for the provided metric.
+    pub fn start(&self, glean: &Glean) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let mut start_time = self.start_time.lock().unwrap();
```
mdboom: Since this code is run off of the main thread in a dispatch queue, this could happen significantly after the "start" of the thing being measured.  Even though it's less clean, it's probably better to capture the current time on the main thread on the Kotlin side and then pass it in over FFI.

badboy: Hm, what about _not_ dispatching this? It doesn't hit the database here, so I/O should not be a problem.

(Unless ... we record an error, then it _does_ hit the database :/)

badboy: Ah, but then we have the same problem on stop. Hmpf.

Dexterp37: I agree with @mdboom , good catch. Let's keep the dispatcher (at least for now) and propagate the time over FFI.

badboy: Ok, fixed that now. Now missing double-checking that I have all the tests and same behavior everywhere.

glean-core/tests/timespan.rs
```diff
@@ -0,0 +1,230 @@
...
+use serde_json::json;
+
+use glean_core::metrics::*;
+use glean_core::storage::StorageManager;
+use glean_core::{CommonMetricData, Glean, Lifetime};
+
+// Tests ported from glean-ac
+
+#[test]
+fn serializer_should_correctly_serialize_counters() {
```
mdboom:
```suggestion
fn serializer_should_correctly_serialize_timespans() {
```

glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimespanMetricTypeTest.kt
```diff
@@ -149,14 +143,59 @@ class TimespanMetricTypeTest {
...
-        metric.stopAndSum(this)
+        metric.start()
+        metric.start()
+        metric.stop()
 
         // Check that data was properly recorded in the second ping.
         assertTrue(metric.testHasValue("store2"))
         assertTrue(metric.testGetValue("store2") >= 0)
         // TODO
         // assertEquals(1, testGetNumRecordedErrors(metric, ErrorType.InvalidValue))
```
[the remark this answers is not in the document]
badboy: Nope, [error test api is not exposed yet](https://bugzilla.mozilla.org/show_bug.cgi?id=1556963)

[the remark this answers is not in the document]
Dexterp37: Ok, thanks. Would you kindly add the bug number near the TODO here?

PR #150, opened 2019-06-25
Fix showing "not implemented" status in docs

georgf:
The current docs show *"status: not implemented"* for a few features (metrics, pings, debug activity).
However, we now point from android-components to these docs as the main docs, so this is misleading.

We should prioritize providing correct and helpful docs for the a-c glean sdk, as this is actively getting used.


georgf:
It turns out i should be saving all the files i changed before pushing a PR...

PR #156, opened 2019-07-05
Use created temporary directory

badboy:
glean AC uses `context.cacheDir`, but I didn't want to pass that through the FFI, so now we build our own temp directory to which we can write files.

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -195,6 +195,12 @@ impl PingMaker {
         Ok(pings_dir)
     }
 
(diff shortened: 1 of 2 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -195,6 +195,12 @@ impl PingMaker {
         Ok(pings_dir)
     }
 
+    fn get_tmp_dir(&self, data_path: &Path) -> std::io::Result<PathBuf> {
```
Dexterp37: Would you kindly add documentation for this function? IMHO it's also worth explaining why we're not simply saving files to /tmp

PR #157, opened 2019-07-05
Use a closure to get the ping's name

badboy:
When we switched to either a string name or a ping type object we
introduced a generic function, but didn't adjust the log message.
This results in the object's string representation to be used (which is
most likely just `PingType0x12345` or something along those lines).

The non-generic wrapper functions do know how to get to the name, so we
can use that knowledge and pass it in.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
@@ -17,7 +17,7 @@ import mozilla.telemetry.glean.rust.toByte
  * The Ping API only exposes the [send] method, which schedules a ping for sending.
  */
 class PingType(
-    name: String,
+    internal val name: String,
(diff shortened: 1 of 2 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
```diff
@@ -17,7 +17,7 @@ import mozilla.telemetry.glean.rust.toByte
  * The Ping API only exposes the [send] method, which schedules a ping for sending.
  */
 class PingType(
-    name: String,
+    public val name: String,
```
mdboom: Maybe we should make this `internal` so it isn't part of the public API?  (It's low risk even if it is, though, so I don't feel too strongly about it).

Dexterp37: Yes, I think this is a good idea.

PR #158, opened 2019-07-05
Make the linter happy

badboy:
ktlint was accumulating some non-allowed stuff which we can easily get rid off.
(we allow up to 10 lint failures apparently)

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -85,7 +85,11 @@ open class GleanInternalAPI internal constructor () {
         this.configuration = configuration
 
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
-        handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName, uploadEnabled.toByte())
+        handle = LibGleanFFI.INSTANCE.glean_initialize(
+            this.gleanDataDir.path,
+            applicationContext.packageName,
+            uploadEnabled.toByte()
+        )
 
         // If any pings were registered before initializing, do so now
         this.pingTypeQueue.forEach { this.registerPingType(it) }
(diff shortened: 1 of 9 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
```diff
@@ -102,8 +102,10 @@ class DatetimeMetricType internal constructor(
                 hour = value.get(Calendar.HOUR_OF_DAY),
                 minute = value.get(Calendar.MINUTE),
                 second = value.get(Calendar.SECOND),
-                nano = value.get(Calendar.MILLISECOND) * 1000000L,
-                offset_seconds = value.get(Calendar.ZONE_OFFSET).div(1000)
+                nano = java.util.concurrent.TimeUnit.MILLISECONDS.toNanos(value.get(Calendar.MILLISECOND).toLong()),
```
Dexterp37: nit: you could do `import java.util.concurrent.TimeUnit as AndroidTimeUnit` above in the file and then use `AndroidTimeUnit.MILLISECONDS.toNanos` to make this a bit more readable

badboy: Ha, that was the missing piece.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -85,7 +85,10 @@ open class GleanInternalAPI internal constructor () {
         this.configuration = configuration
 
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
-        handle = LibGleanFFI.INSTANCE.glean_initialize(this.gleanDataDir.path, applicationContext.packageName, uploadEnabled.toByte())
+        handle = LibGleanFFI.INSTANCE.glean_initialize(
+            this.gleanDataDir.path,
+            applicationContext.packageName,
+            uploadEnabled.toByte())
```
mdboom: Maybe final `)` on the next line?  Maybe ktlint doesn't care, but I think that's more readable (and consistent with rustfmt FWIW).

badboy:
Ah, detekt it is then. There are still 2 failures left (too many return statements).
I will probably need to add an exception marker there, but I will take that in a follow-up

badboy:
I ended up adding it here as well

PR #160, opened 2019-07-08
Document the behavior of Glean pings on crash

Dexterp37:

### DIFF
--- a/docs/user/pings/baseline.md
+++ b/docs/user/pings/baseline.md
@@ -3,6 +3,7 @@
 ## Description
 
 This ping is intended to provide metrics that are managed by the library itself, and not explicitly set by the application or included in the application's `metrics.yaml` file.
+No `baseline` ping is sent if the application crashes, with the data from before the crash.
 
 ## Scheduling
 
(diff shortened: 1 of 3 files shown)


docs/user/pings/baseline.md
```diff
@@ -3,6 +3,7 @@
 ## Description
 
 This ping is intended to provide metrics that are managed by the library itself, and not explicitly set by the application or included in the application's `metrics.yaml` file.
+No `baseline` ping is sent if the application crashes, with the data from before the crash.
```
badboy: This sentence construction reads weird.
Below we always start with "if the application crashes", which would make sense here as well.
To me it's not clear what happens with baseline data. baseline can potentially include metrics, those behave just the same as the metrics ping would?

Dexterp37: The fact that metrics can be added to the `baseline` ping is more of an edge case, rather than a feature we'd like to push forward heavily.

With that said, the above sentence is referring to, for example, the duration field: you won't receive a 'baseline' ping with the duration of the session before the crash.

PR #161, opened 2019-07-08
1552872: Implement events

mdboom:


glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// that the memory and disk representations are in sync.
+    ///
+    /// Secondly, if this is the first time the application has been run since
+    /// rebooting, any pings containing events are assembled into pings and cleared
+    /// immediately, since their timestamps won't be compatible with the timestamps
+    /// we would create during this boot of the device.
+    ///
+    /// # Arguments
+    ///
+    /// * `glean`:  The Glean instance.
```
badboy:
```suggestion
    /// * `glean` - The Glean instance.
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+        }
+    }
+
+    fn load_events_from_disk(&self) -> Result<()> {
+        let mut db = self.event_stores.write().unwrap();
+        for entry in fs::read_dir(&self.path)? {
+            let entry = entry?;
+            if entry.file_type()?.is_file() {
+                let store_name = entry.file_name().into_string()?;
+                let file = BufReader::new(OpenOptions::new().read(true).open(entry.path())?);
```
badboy: `File::open` does open in read mode only: https://doc.rust-lang.org/std/fs/struct.File.html#method.open

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                db.insert(
+                    store_name,
+                    file.lines()
+                        .filter_map(|line| line.ok())
+                        .filter_map(|line| {
+                            println!("{}", line);
```
badboy: Stray println

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                            serde_json::from_str::<JsonValue>(&line).ok()
```
badboy: Why build a JsonValue first and then deserializing it into `RecordedEventData`? We can go directly to `RecordedEventData` thanks to serde.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                );
+            }
+        }
+    }
+
+    /// Record an event in the desired stores.
+    ///
+    /// # Arguments
+    ///
+    /// - `glean`: The Glean instance.
```
badboy:
```suggestion
    /// * `glean` - The Glean instance.
```

That's the formatting we use in other places. applies to the following lines as well.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                    store_name,
+                    err
+                );
+            }
+        }
+    }
+
+    /// Writes an event to a single store on disk.
+    ///
+    /// This assumes that the write lock in `self.event_stores` is already held.
```
badboy: Why is this relevant? If it is, we should build this function in a way that this is required.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// - `store_name`: The name of the store.
```
badboy:
```suggestion
    /// * `store_name` - The name of the store.
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+        {
+            log::error!("Error writing event to store {}: {}", store_name, err);
+        }
+    }
+
+    /// Get a snapshot of the stored event data as a JsonValue.
+    ///
+    /// # Arguments
+    ///
+    /// - `store_name` The name of the desired store.
```
badboy:
```suggestion
    /// * `store_name` - The name of the desired store.
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+    /// - `clear_store` Whether to clear the store afterward.
+    ///
+    /// # Returns
+    ///
+    /// The JsonValue as an array of events, if any.
```
badboy:
```suggestion
    /// The array of events, JSON encoded, if any.
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+    pub fn snapshot_as_json(&self, store_name: &str, clear_store: bool) -> Option<JsonValue> {
+        let result = {
+            let db = self.event_stores.read().unwrap();
+            db.get(&store_name.to_string()).and_then(|store| {
+                if !store.is_empty() {
+                    let first_timestamp = store[0].timestamp;
+                    Some(json!(store
```
badboy: Weird formatting here. Did rustfmt run?

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+                        .iter()
+                        .map(|e| e.serialize_relative(first_timestamp))
+                        .collect::<Vec<JsonValue>>()))
```
badboy: I have the feeling we can be much better about serialization here.
We should only ever do the json serialization at the I/O points, and then always on the full values.

glean-core/src/lib.rs
```diff
@@ -95,11 +100,13 @@ impl Glean {
         let mut glean = Self {
             upload_enabled,
             data_store: Database::new(data_path)?,
+            event_data_store: EventDatabase::new(data_path)?,
             core_metrics: CoreMetrics::new(),
             data_path: PathBuf::from(data_path),
             application_id,
             ping_registry: HashMap::new(),
             start_time: local_now_with_offset(),
+            max_events: DEFAULT_MAX_EVENTS,
```
badboy: Are there any plans to make this configurable? If not, should we just keep it as a constant and not a config option?

glean-core/src/metrics/event.rs
```diff
@@ -0,0 +1,132 @@
...
+    ///   is looked up.
+    pub fn record(&self, glean: &Glean, timestamp: u64, extra: Option<HashMap<usize, String>>) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let extra_strings = extra.and_then(|extra| {
+            Some(
+                extra
+                    .iter()
```
badboy: We should be able to consume the map with `into_iter`, then we don't need the `.to_string()` below.

glean-core/tests/common/mod.rs
```diff
@@ -63,3 +68,29 @@ pub fn iso8601_to_chrono(datetime: &iso8601::DateTime) -> chrono::DateTime<chron
     };
     panic!("Unsupported datetime format");
 }
+
+/// Get a vector of the currently queued pings
+pub fn test_get_queued_pings(data_path: &Path) -> Result<Vec<(String, JsonValue)>> {
```
badboy: This is already only available to tests, we don't need the `test_` prefix?

glean-core/tests/common/mod.rs
```diff
@@ -63,3 +68,29 @@ pub fn iso8601_to_chrono(datetime: &iso8601::DateTime) -> chrono::DateTime<chron
...
+    let pings_dir = data_path.join("pings");
+    let entries = read_dir(&pings_dir)?;
+    Ok(entries
+        .filter_map(|entry| entry.ok())
+        .filter(|entry| match entry.file_type() {
+            Ok(file_type) => file_type.is_file(),
+            Err(_) => false,
+        })
+        .filter_map(|entry| OpenOptions::new().read(true).open(entry.path()).ok())
```
badboy: `File::open`

glean-core/ffi/src/event.rs
```diff
@@ -0,0 +1,127 @@
...
+    name: FfiStr,
+    send_in_pings: RawStringArray,
+    send_in_pings_len: i32,
+    lifetime: i32,
+    disabled: u8,
+    extra_keys: RawStringArray,
+    extra_keys_len: i32,
+) -> u64 {
+    EVENT_METRICS.insert_with_log(|| {
+        let send_in_pings = unsafe { from_raw_string_array(send_in_pings, send_in_pings_len) };
```
badboy: Uff, I really need to work on these unsafe parts to make them safer.

glean-core/ffi/src/event.rs
```diff
@@ -0,0 +1,127 @@
...
+
+#[no_mangle]
+pub extern "C" fn glean_event_should_record(glean_handle: u64, metric_id: u64) -> u8 {
+    GLEAN.call_infallible(glean_handle, |glean| {
+        EVENT_METRICS.call_infallible(metric_id, |metric| metric.should_record(&glean))
+    })
+}
+
+/// Create a HashMap<usize, String> from a pair of C arrays.
+unsafe fn from_raw_int_string_arrays(
```
badboy: Can we move this into `ffi/src/lib.rs`? We already have `from_raw_string_array` there and it would be a good idea to keep these utility functions close to each other.

badboy: The name of the function is already not conveying what it does.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -64,45 +98,36 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+            return
+        }
+
+        // We capture the event time now, since we don't know when the async code below
+        // might get executed.
+        val timestamp = SystemClock.elapsedRealtime()
+
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.launch {
+            // The Map is sent over FFI as a pair of parallel Arrays
```
badboy: What are parallel arrays?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -117,15 +142,39 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+            Glean.handle,
+            this.handle,
+            pingName
+        ).toBoolean()
+    }
+
+    /**
+    * Deserializes an event in JSON into a RecordedEventData object.
+    *
+    * @param jsonContent The JSONObject containing the data for the event.
```
Dexterp37: nit: can you document the expected structure of the JSONobject?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -140,13 +189,29 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+        Dispatchers.API.assertInTestingMode()
+
+        val ptr = LibGleanFFI.INSTANCE.glean_event_test_get_value_as_json_string(
+            Glean.handle,
+            this.handle,
+            pingName
+        )!!
+
+        var jsonRes: JSONArray
+        try {
```
Dexterp37: Can't this be: `val jsonRes = try {} catch (...) {}`?

glean-core/ffi/src/event.rs
Dexterp37: @mdboom could you file a bug about this? Or @badboy is there any low-hanging fruit we could grab in this PR?

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+/// may have come from a previous boot of the device, and therefore will not be
+/// compatible with any newly-collected events.
+#[derive(Debug)]
+pub struct EventDatabase {
+    pub path: PathBuf,
+    // The in-memory list of events
+    event_stores: RwLock<HashMap<String, Vec<RecordedEventData>>>,
+}
+
+impl EventDatabase {
```
Dexterp37: Do we really need this, here? Maybe we could leverage the database we already have for this? We used this approach in glean-ac mostly to cope with crashes and to make sure we were not loosing events. But do we have the same risk here, given the storage we're using now, which should be a bit more resilient?

@badboy thoughts?

glean-core/src/lib.rs
Dexterp37: Yes, we need this for feature parity with glean-ac for now, [see here](https://github.com/mozilla-mobile/android-components/blob/8ba1a8d0ca4521fc95a371f308ffcf1075ca2ce1/components/service/glean/src/main/java/mozilla/components/service/glean/config/Configuration.kt#L51). We might revisit this decision in the future, though, as no product is using it right now.

glean-core/src/metrics/event.rs
```diff
@@ -0,0 +1,132 @@
...
+                    .collect(),
+            )
+        });
+
+        glean
+            .event_storage()
+            .record(glean, &self.meta, timestamp, extra_strings);
+    }
+
+    fn truncate_value<'a>(&self, glean: &Glean, key: &str, value: &'a str) -> &'a str {
```
Dexterp37: nit: can you add a tiny comment describing what does this do? It currently does a bit more than just truncating (i.e. error reporting). Plus, it returns the truncated value.

mdboom:
Thanks for the extensive review.

> One thing I would like to see is an explanation (in the commit message?) why events are handled differently from every other metric type.
> Especially the fact that we are adding additional files that we (try to) keep in sync and re-read on startup should be called out.

This is documented with the `EventDatabase` struct.  Would you have expected to find it somewhere else (or linked to from somewhere else)?

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,375 @@
...
+        )
+    }
+
+    /// Internal function to perform the serialization to JSON
+    fn serialize_parts(
+        timestamp: u64,
+        category: &str,
+        name: &str,
+        extra: &Option<HashMap<String, String>>,
+    ) -> JsonValue {
```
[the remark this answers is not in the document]
mdboom: Thanks -- indeed we can.  Simplifies the code considerably.

glean-core/src/event_database/mod.rs
mdboom: We don't want multiple threads writing to files at the same time.  On glean-ac, this was handled using an explicit sync variable *just* for the filesystem operations.  In this implementation, we always hold a lock on the in-memory data structure whenever we do file system operations anyway.  We could add a RwLock just for the on-disk stuff for clarity, but it would be redundant.

mdboom: I've updated this to have an explicit lock for file operations, separate from the in-memory data structure's lock.

glean-core/src/event_database/mod.rs
mdboom: Yep -- `rustfmt` is insisting on this :man_shrugging:

glean-core/src/event_database/mod.rs
mdboom: In this instance (and actually only this place), we also need to update the timestamps to be relative to the first timestamp.  It seems more efficient to do this on the fly, rather than create a whole second Vec of `RecordedEventData` objects prior to JSON serialization.  This mirrors `StorageManager.snapshot_as_json` which also does a bunch of JSON-specific fixups before passing along to the `PingMaker` to put together.

But maybe I'm also just not understanding your comment...

glean-core/src/lib.rs
mdboom: Ah -- I had meant to come back around to this.  I had initially made this a parameter to the Glean constructor, but that isn't going to scale well to the ~dozen config parameters we'll probably end of having, and would require updating calls to the Glean constructor in *all* of the tests.

A better pattern might be to have an explicit `Configuration` struct that's passed into the `Glean` constructor (as glean-ac does).  Ideas and brainstorming [in this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1564505), but I'd advocate doing that as a follow-on to this PR.

glean-core/src/metrics/event.rs
mdboom: I think the `to_string` is still required to handle the fact that it could be the original string or the truncated string -- or at least I can't find the way through with that...

glean-core/ffi/src/event.rs
mdboom: I'm not clear on what could be done, but I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1564523).

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
mdboom: Sure -- I can clarify in the comment.  [Parallel arrays](https://en.wikipedia.org/wiki/Parallel_array) are basically records where the fields have been separated out into separate arrays.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
mdboom: Good catch.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
mdboom: It's the same as the event in a data ping, but it doesn't hurt to document it here.

glean-core/src/event_database/mod.rs
mdboom: (Some of this was discussed in the meeting, but wanted to summarize here for the record)...

rkv doesn't seem to have a good way to handle data that is an appended list of entries, rather than key/values.  We *could* put the index in the key if we wanted to, but just figuring out what the next index should be when adding an event would be a non-constant operation.

Storing events on disk is required not just for crashes -- the OS routinely shuts down the app without warning even if it is behaving correctly, so we need to persist (through some mechanism) regardless.

A possible simplification (and the approach taken in `service-telemetry`) is to *only* store events on disk.  That was [reviewed out](https://github.com/mozilla-mobile/android-components/pull/2328#pullrequestreview-213441306) in the glean-ac implementation since it would require I/O on the same thread as other Glean operations (but never on the main thread).  In glean-core, we don't fire off I/O threads anyway -- maybe we should, but if not, that simplification might make a lot of sense here and significantly reduce the complexity of the implementation.

glean-core/ffi/src/event.rs
mdboom: I've moved and renamed this -- hopefully a little better now.

glean-core/src/event_database/mod.rs
mdboom: It occurs to me that if we wanted to use a robust data store that meets the requirements, we could use `sqlite`, but that would be an extra dependency.  It ships on all Android devices, so not *really* a new dependency, but the Rust wrapper would need to be configured to use it, etc., which could be thorny.

Dexterp37: Yeah, plus we don't know how the story looks on iOS yet. It's probably not worth the effort :)

glean-core/ffi/src/event.rs
```diff
@@ -0,0 +1,105 @@
...
+#[no_mangle]
+pub extern "C" fn glean_new_event_metric(
+    category: FfiStr,
+    name: FfiStr,
+    send_in_pings: RawStringArray,
+    send_in_pings_len: i32,
+    lifetime: i32,
+    disabled: u8,
+    extra_keys: RawStringArray,
+    extra_keys_len: i32,
```
mdboom: @badboy: So here's a use case for the fancy new FFI-generating macro that I couldn't find a solution for.  This is a pair of arguments that need to convert to a single argument in Event::new().  Tips online for this seem to suggest passing the argument as a [struct over FFI](http://jakegoulding.com/rust-ffi-omnibus/tuples/), and then of course we can implement `Into` on that struct, but I couldn't figure out how to create and pass that struct from Kotlin...

Barring that, it looks like `new` is required in the macro, so couldn't just skip that one.

So, I've left this file as not using the macro at all for now (which still works, but lacks the cool factor...)

badboy: Ah yeah. A single macro will never cover all the cases we have (or become such a big macro that _I_ don't want to maintain it).

The macro should cover the common cases. If we need to fall back to manual implementation here and there that seems fine to me IMO.
(We might still have another/reduced macro that allows skipping the `new` part)

badboy: `new` can be optional: #174

glean-core/src/event_database/mod.rs
badboy: FxA has wrappers around sqlite, but it's a bit more complicated and I don't see the need in bringing in another database when the filesystem ought to be good enough for now.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+/// ping. There is one file per store.
+///
+/// When restarting the application, these on-disk files are checked, and if any are
+/// found, they are loaded, queued for sending and flushed immediately before any
+/// further events are collected. This is because the timestamps for these events
+/// may have come from a previous boot of the device, and therefore will not be
+/// compatible with any newly-collected events.
+#[derive(Debug)]
+pub struct EventDatabase {
+    pub path: PathBuf,
```
badboy: How about adding a comment here like "Path to on-disk database file"

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    // The in-memory list of events
```
badboy:
```suggestion
    /// The in-memory list of events
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        create_dir_all(&path)?;
+
+        Ok(Self {
+            path,
+            event_stores: RwLock::new(HashMap::new()),
+            file_lock: RwLock::new(()),
+        })
+    }
+
+    /// Initialize events storage. This must be called once on application startup,
```
badboy: The first sentence here doesn't reflect much of the actual function name. Maybe we can clarify this further?

"Initialize events storage after Glean is fully initialized and ready to send pings" or something along the lines.
(Just making it clear in the first sentence what it does and when it gets called)

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+    /// * `glean` - The Glean instance.
+    pub fn on_ready_to_send_pings(&self, glean: &Glean) {
+        match self.load_events_from_disk() {
+            Ok(_) => self.send_all_events(glean),
+            Err(err) => log::error!("Error loading pings from disk: {}", err),
+        }
+    }
+
+    fn load_events_from_disk(&self) -> Result<()> {
+        let _lock = self.file_lock.read().unwrap();
```
badboy: What are we locking here?
The nice thing about locks in Rust that they usually wrap the data that ought to be protected.
Now that we have nothing in here, it's not clear what's protected.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        let event = RecordedEventData {
+            timestamp,
+            category: meta.category.to_string(),
+            name: meta.name.to_string(),
+            extra,
+        };
+        let event_json = serde_json::to_string(&event).unwrap();
+        let mut stores_to_send: Vec<&str> = Vec::new();
+
+        {
```
badboy: The actual storing of the event is hidden a bit in that multi-line block below.
How about we mark the different phases with a comment?
E.g. above here:

```rust
// Store the event in-memory and persist to disk
```

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+                    .or_insert_with(|| Vec::new());
+                store.push(event.clone());
+                self.write_event_to_disk(store_name, &event_json);
+                if store.len() == glean.get_max_events() {
+                    stores_to_send.push(&store_name);
+                }
+            }
+        }
+
+        for store_name in stores_to_send {
```
badboy:
```rust
// Send pings with a full event queue
```

glean-core/src/event_database/mod.rs
badboy: Yikes.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+            log::error!("Error writing event to store {}: {}", store_name, err);
+        }
+    }
+
+    /// Get a snapshot of the stored event data as a JsonValue.
+    ///
+    /// # Arguments
+    ///
+    /// * `store_name` - The name of the desired store.
+    /// * `clear_store` - Whether to clear the store afterward.
```
badboy:
```suggestion
    /// * `clear_store` - Whether to clear the store after snapshotting.
```

that's the phrasing we use in other parts.

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+                        .map(|e| e.serialize_relative(first_timestamp))
+                        .collect::<Vec<JsonValue>>()))
+                } else {
+                    None
+                }
+            })
+        };
+
+        if clear_store {
+            {
```
badboy: These double-nested blocks look weird.
It's just to shorten the scope of the locks, right?

badboy: (which means we can at least leave out the 2nd block?

glean-core/src/event_database/mod.rs
```diff
@@ -0,0 +1,325 @@
...
+        let value: Vec<RecordedEventData> = self
+            .event_stores
+            .read()
+            .unwrap()
+            .get(&store_name.to_string())
+            .into_iter()
+            .flatten()
+            .filter(|event| event.name == meta.name && event.category == meta.category)
+            .cloned()
+            .collect();
```
badboy: Chain it ALL!
(Not saying this is wrong or anything, it's just weird seeing such a long call chain :D)

glean-core/src/lib.rs
badboy: yup, let's do a followup.

glean-core/src/lib.rs
```diff
@@ -114,6 +121,14 @@ impl Glean {
         Ok(())
     }
 
+    /// Called when Glean is initialized to the point where it can correctly
+    /// assemble pings. Usually called from the language specific layer after all
+    /// of the core metrics have been set and the ping types have been
+    /// registered.
+    pub fn on_ready_to_send_pings(&self) {
+        self.event_data_store.on_ready_to_send_pings(&self);
```
badboy: Bikeshed:

I wonder whether we eventually have other things to call in here.
IMO then the `event_data_store.on_ready_to_send_pings` name can change to something more clearly identifying what the data store is doing there and the `on_ready_to_send_pings` is only the public name bundling it all

glean-core/src/metrics/event.rs
```diff
@@ -0,0 +1,127 @@
...
+    ///
+    /// ## Arguments
+    ///
+    /// * `glean` - The Glean instance this metric belongs to.
+    /// * `timestamp` - A monotonically increasing timestamp, in nanoseconds.
+    ///   This must be provided since the actual recording of the event may
+    ///   happen some time later than the moment the event occurred.
+    /// * `extra` - A HashMap of (key, value) pairs. The key is an index
+    ///   into the metric's `allowed_extra_keys` vector where the key's string
+    ///   is looked up.
```
badboy: Should this documentation state that events are not recorded when their extra keys are not allowed?

glean-core/src/metrics/event.rs
```diff
@@ -0,0 +1,127 @@
...
+    pub fn record(&self, glean: &Glean, timestamp: u64, extra: Option<HashMap<i32, String>>) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let extra_strings = extra.and_then(|extra| {
+            Some(
+                extra
+                    .into_iter()
+                    .map(|(k, v)| (self.allowed_extra_keys.get(k as usize).unwrap(), v))
```
badboy: This unwrap can panic when `k` is invalid.
We should do better error handling here (e.g. bailing out early on invalid extra arguments).

Another note when the above is tackled:
Iterator combinators in Rust are nice, but sometimes a good ol' for loop, pushing into the result map is better (and more easily allows bailing out on errors)

glean-core/tests/event.rs
```diff
@@ -0,0 +1,245 @@
...
+        },
+        vec!["key1".into(), "key2".into()],
+    );
+
+    let extra: HashMap<i32, String> = [(0, "value1".into()), (1, "value2".into())]
+        .iter()
+        .cloned()
+        .collect();
+
+    metric.record(&glean, 1000, Some(extra));
```
badboy: Seeing this now, there's a trick in Rust to make both `metric.record(&g, 1000, None)` and `metric.record(&g, 1000, extra)` work.

```rust
fn record<M: Into<Option<HashMap<...>>>(glean: &Glean, timestamp: u64, extra: M) {
  let extra = extra.into()
  ...
}
```

Just a suggestion, maybe we don't need this here, it's not made to be the nice user-facing API (yet)

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -194,7 +200,7 @@ class EventMetricTypeTest {
             firstEvent.timestamp < secondEvent.timestamp)
     }
 
-    @Ignore("EventMetricType is a stub")
+    @Ignore("Metric clearing when uploading disabled is not implemented")
```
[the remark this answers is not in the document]
mdboom: Yep.  In all likelihood, we'll merge this first and then fix this as part of #169.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -223,4 +229,57 @@ class EventMetricTypeTest {
...
+            lifetime = Lifetime.Ping,
+            sendInPings = listOf("events"),
+            allowedExtraKeys = listOf("someExtra")
+        )
+
+        event.record(extra = mapOf(SomeExtraKeys.SomeExtra to "bar"))
+        assertEquals(1, event.testGetValue().size)
+
+        // Clear the in-memory storage only to mock being loaded in a fresh process
+        // EventsStorageEngine.eventStores.clear()
```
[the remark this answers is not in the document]
mdboom: Yeah -- comment is out-of-date.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -223,4 +229,57 @@ class EventMetricTypeTest {
...
+            name = "test_event",
+            lifetime = Lifetime.Ping,
+            sendInPings = listOf("events"),
+            allowedExtraKeys = listOf("someExtra")
+        )
+
+        event.record(extra = mapOf(SomeExtraKeys.SomeExtra to "bar"))
+        assertEquals(1, event.testGetValue().size)
+
+        // Clear the in-memory storage only to mock being loaded in a fresh process
```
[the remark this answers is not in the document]
mdboom: Exactly.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -223,4 +229,57 @@ class EventMetricTypeTest {
...
+                logPings = true
+            ),
+            clearStores = false
+        )
+
+        // Trigger worker task to upload the pings in the background
+        PingUploadWorker.enqueueWorker()
+        triggerWorkManager()
+
+        val request = server.takeRequest(20L, TimeUnit.SECONDS)
```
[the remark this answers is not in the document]
mdboom: Oof -- copied that from glean-ac without noticing.  Maybe let's make it 1 second and change if it becomes a problem...

glean-core/src/event_database/mod.rs
mdboom: We're locking for operations on the filesystem.  I just added a comment to this field in the struct to clarify this.

In this case, we don't want to read in events from files when we might also be deleting them in snapshot_as_json.  Since Android is Linux, the files themselves have locking semantics.  What doesn't work is iterating over a directory while deleting those files in some other thread.

In practice right now, this shouldn't happen given how we queue work on the Kotlin side, but if that were to change we could be in trouble, and since these functions don't have a mutable reference to the EventDatabase instance, I don't think Rust would enforce the synchronization.

All of this comes from actual problems we had with intermittent tests in glean-ac...

glean-core/src/event_database/mod.rs
mdboom: Exactly.  The first block actually can be converted to a single statement.  And then, yes, the second is redundant because it's the last block in the outer block.  So, can be much simplified.

glean-core/src/event_database/mod.rs
mdboom: I feel that's like most of the Rust code I see in the wild, but again, I'm not sure if it's a good or a bad thing.  If it really seems unmanageable, let me know.  I guess the alternative is something really deeply nested so :man_shrugging: ?

glean-core/src/lib.rs
mdboom: That's not a bad suggestion.  I suppose the thing would be called `flush_events_from_previous_run`?

glean-core/src/metrics/event.rs
mdboom: Yes, that should be the behavior.  In practice, this is enforced at build time on the Kotlin side (since the keys are enums), but it's entirely possible that something got messed up through the FFI layer.

glean-core/src/metrics/event.rs
mdboom: I've rejiggered this with a for loop.

glean-core/tests/event.rs
mdboom: Cool trick.  Might as well do this, I guess...

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -117,15 +143,44 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+
+        return LibGleanFFI.INSTANCE.glean_event_test_has_value(
+            Glean.handle,
+            this.handle,
+            pingName
+        ).toBoolean()
+    }
+
+    /**
+    * Deserializes an event in JSON into a RecordedEventData object.
```
badboy: Small nit. We usually align the first star below the one above.

glean-core/ffi/src/event.rs
badboy: Let's merge this as is and fix that in a followup (after https://github.com/mozilla/glean/pull/179 lands)

glean-core/ffi/src/lib.rs
```diff
@@ -4,6 +4,8 @@
 
 #![allow(clippy::redundant_closure)]
 
+use std::collections::HashMap;
+use std::iter::Iterator;
```
badboy: Do we need that import?

glean-core/ffi/src/lib.rs
```diff
@@ -50,6 +54,28 @@ unsafe fn from_raw_string_array(arr: RawStringArray, len: i32) -> Vec<String> {
         .collect()
 }
 
+/// Create a HashMap<T, String> from a pair of C int and string arrays.
```
badboy: It's `HashMap<i32, String>` (a concrete `T`)

glean-core/ffi/src/lib.rs
```diff
@@ -50,6 +54,28 @@ unsafe fn from_raw_string_array(arr: RawStringArray, len: i32) -> Vec<String> {
...
+    }
+
+    let keys_ptrs = std::slice::from_raw_parts(keys, len as usize);
+    let values_ptrs = std::slice::from_raw_parts(values, len as usize);
+
+    Some(
+        keys_ptrs
+            .iter()
+            .zip(values_ptrs.iter())
+            .map(|(&k, &v)| (k, FfiStr::from_raw(v).into_string()))
```
badboy: This can potentially fail, see #179, but as mentioned above let's fix that in a follow up

glean-core/src/event_database/mod.rs
badboy: Maybe during backlog/tech debt cleanup we can take another look if we can solve this differently. For now I don't worry too much.

glean-core/src/event_database/mod.rs
badboy: All good!

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -64,45 +98,37 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+        Dispatchers.API.launch {
+            // The Map is sent over FFI as a pair of arrays, one containing the
+            // keys, and the other containing the values.
+            var keys: IntArray? = null
+            var values: StringArray? = null
+            var len: Int = 0
+            if (extra != null) {
+                val extraList = extra.toList()
+                keys = IntArray(extra.size, { extraList[it].first.ordinal })
+                values = StringArray(Array<String>(extra.size, { extraList[it].second }), "utf-8")
```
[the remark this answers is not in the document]
mdboom: I'm not sure -- `keys` and `values` are not guaranteed to come back in any particular order, and they are lazily evaluated.  So you could get the keys and values arrays in a different order (and they would no longer be parallel).  It seems unlikely to happen if the Map isn't mutated in the middle, but the API makes no guarantees about it.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -64,45 +98,37 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>>(
...
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.launch {
+            // The Map is sent over FFI as a pair of arrays, one containing the
+            // keys, and the other containing the values.
+            var keys: IntArray? = null
+            var values: StringArray? = null
+            var len: Int = 0
+            if (extra != null) {
+                val extraList = extra.toList()
+                keys = IntArray(extra.size, { extraList[it].first.ordinal })
```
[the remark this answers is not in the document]
mdboom: See above.  That *could* be incorrect.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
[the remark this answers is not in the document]
Dexterp37: Oh, I see. This makes sense. I think I need to change this in #173 too.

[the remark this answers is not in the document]
badboy: That is actually a very good explanation. Can we leave this as a comment somewhere (in the code or in the commit message)?

glean-core/ffi/src/lib.rs
mdboom: Looks like no.  I wonder why clippy didn't complain...

mdboom:
I've fixed this last round of nits.  For the parallel arrays of keys and values, I left the code as-is but added a comment.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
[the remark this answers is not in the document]
mdboom: I added this in the latest commit:

```
 // In Kotlin, Map.keys and Map.values are not guaranteed to return the entries
 // in any particular order. Therefore, we iterate over the pairs together and
 // create the keys and values arrays step-by-step.
```

PR #168, opened 2019-07-10
Use a macro to reduce boilerplate in FFI layer

badboy:
This should be considered a discussion ground.
Notable things:

* It doesn't handle the `test_(get|has)_value` functions yet
	* In part because some of those are a bit more complex
* It can't wrap all actual record functions
	* It can do some, see e.g. the counter
	* It might be possible to include some more
* The chosen syntax in the macro is arbitrary and can be changed


mdboom:
This definitely seems like an improvement.  I'd be in favor of merging something at this level, and then iterating later as we see more opportunities for reduction of boilerplate.

Too bad Kotlin doesn't have macros...  That side of things is still kind of annoying...

PR #169, opened 2019-07-10
1557366: Clear metrics when uploading is disabled

mdboom:


glean-core/src/lib.rs
```diff
@@ -42,6 +42,10 @@ use crate::storage::StorageManager;
 use crate::util::{local_now_with_offset, sanitize_application_id};
 
 const GLEAN_SCHEMA_VERSION: u32 = 1;
+lazy_static! {
+    static ref KNOWN_CLIENT_ID: Uuid =
+        Uuid::parse_str("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0").unwrap();
```
mdboom: :coffee:

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -152,7 +157,26 @@ open class GleanInternalAPI internal constructor () {
...
+            Dispatchers.API.launch {
+                // glean_set_upload_enabled might delete all of the queued pings. We
+                // therefore need to obtain the lock from the PingUploader so that
+                // iterating over and deleting the pings doesn't happen at the same time.
+                synchronized(PingUploader.pingQueueLock) {
+                    LibGleanFFI.INSTANCE.glean_set_upload_enabled(handle, enabled.toByte())
+                }
+            }
+
+            Dispatchers.API.launch {
```
badboy: Couldn't we merge the two dispatcher blocks into one?

glean-core/src/lib.rs
```diff
@@ -251,16 +351,137 @@ impl Glean {
...
+                .get_value(&glean, "glean_client_info")
+                .unwrap()
+        );
+    }
+
+    #[test]
+    fn client_id_is_set_to_random_value_when_uploading_disabled_at_start() {
+        let dir = tempfile::tempdir().unwrap();
+        let tmpname = dir.path().display().to_string();
+        let glean = Glean::new(&tmpname, GLOBAL_APPLICATION_ID, true).unwrap();
```
badboy: Upload here is enabled (= `true`). Test name says differently.

glean-core/src/lib.rs
```diff
@@ -251,16 +351,137 @@ impl Glean {
...
-        let glean = Glean::new(&name, "org.mozilla.glean", true).unwrap();
+        let (glean, _) = new_glean();
 
         assert_eq!(
-            "/submit/org-mozilla-glean/baseline/1/this-is-a-docid",
+            "/submit/org-mozilla-glean-test-app/baseline/1/this-is-a-docid",
             glean.make_path("baseline", "this-is-a-docid")
         );
     }
+
```
badboy: Should we have a test ensuring that nothing changes if `set_upload_enabled` is called twice with the same value (`true, true` or `false, false`)?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: You're totally right.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -152,7 +157,26 @@ open class GleanInternalAPI internal constructor () {
...
+                if (originalEnabled != getUploadEnabled()) {
+                    if (getUploadEnabled()) {
```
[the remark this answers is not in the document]
mdboom: Yeah, this could all be simplified.

glean-core/src/lib.rs
```diff
@@ -127,6 +161,72 @@ impl Glean {
...
+        // TODO: 1552872: Clear the event store, once the event metric lands.
+
+        // This does not clear the experiments store (which isn't managed by the
+        // StorageEngineManager), since doing so would mean we would have to have the
+        // application tell us again which experiments are active if telemetry is
+        // re-enabled.
+
+        {
+            // We need to briefly set upload_enabled to true here so that `set`
+            // is not a no-op.
```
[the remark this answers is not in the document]
mdboom: Good idea.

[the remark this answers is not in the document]
mdboom: Additionally, this is safe because we've already cleared the pending pings and the language specific side is holding a lock on the ping uploader.

glean-core/src/lib.rs
mdboom: Yes -- glean-ac had this test but it involved mocking to make sure specific methods weren't called.  I'll have to figure out how to do that in Rust.

mdboom: Mocking in Rust is (for obvious reasons) awfully difficult and invasive and seemed like overkill for this.  So I just ended up adding a return value to `set_upload_enabled`...  Not ideal, but it works...

glean-core/src/lib.rs
```diff
@@ -101,23 +105,61 @@ impl Glean {
...
+    ///
+    /// If the value of this flag is not actually changed, this is a no-op.
+    ///
+    /// # Arguments
+    ///
+    /// * `flag` - When true, enable metric collection.
+    ///
+    /// # Returns
+    ///
+    /// * `did_work` - When true, the flag was different from the current value,
```
Dexterp37: nit: what's `did_work`?

glean-core/src/lib.rs
Dexterp37: Should we have a bug (if we don't have one already) about investigating mocking options for Rust?

glean-core/tests/common/mod.rs
```diff
@@ -63,3 +68,29 @@ pub fn iso8601_to_chrono(datetime: &iso8601::DateTime) -> chrono::DateTime<chron
     };
     panic!("Unsupported datetime format");
 }
+
+/// Get a vector of the currently queued pings.
```
Dexterp37: nit: let's document the parameters and return value

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -62,8 +63,11 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
          * @return true if process was successful
          */
         internal fun uploadPings(): Boolean {
-            val httpPingUploader = HttpPingUploader()
-            return httpPingUploader.process()
+            if (Glean.getUploadEnabled()) {
```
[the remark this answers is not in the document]
mdboom: There was a corresponding change when this "clearing metrics" thing was added to [glean-ac](https://github.com/mozilla-mobile/android-components/pull/2838).  There has always been a check to not collect and assemble a ping when telemetry is disabled.  This just prevents any already queued ones from being sent.  But even then, any queued pings should be deleted from disk when disabling telemetry before we even get here, so it's kind of like an extra insurance policy.

There is a test (in this PR) to check that when telemetry is disabled that the queued pings are cleared.

However, I don't think there's anything that explicitly tests this block -- it will be a bit hard to do, because you'd have to disable telemetry without clearing the queued pings, which we can't really do from the Kotlin side...  I'll have to think on how to test that...

glean-core/src/lib.rs
```diff
@@ -101,23 +105,61 @@ impl Glean {
...
+    fn initialize_core_metrics(&mut self) {
+        let need_new_client_id = match self
+            .core_metrics
+            .client_id
+            .get_value(self, "glean_client_info")
+        {
+            None => true,
+            Some(uuid) => uuid == *KNOWN_CLIENT_ID,
+        };
+        if need_new_client_id {
```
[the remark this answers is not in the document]
mdboom: Thanks for pointing that out -- I actually think it would be good to do that as part of this PR, just so we're testing even more...

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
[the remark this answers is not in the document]
Dexterp37: Nah, thanks for clarifying. No need to test this specifically, not in this PR at least. On the nit side, what if this becomes:

```
return if (Glean.getUploadEnabled()) {
  HttpPingUploader().process()
} else {
  false
}
```

It should prevent a Kotlin warning.

glean-core/src/lib.rs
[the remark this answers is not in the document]
Dexterp37: Good point. Kindly close that bug as the duplicate of 1557366 once you're done ;)

glean-core/src/lib.rs
```diff
@@ -128,6 +170,76 @@ impl Glean {
...
+        let ping_maker = PingMaker::new();
+        if let Err(err) = ping_maker.clear_pending_pings(self.get_data_path()) {
+            log::error!("Error clearing pending pings: {}", err);
+        }
+
+        // Delete all stored metrics.
+        // Note that this also includes the ping sequence numbers, so it has
+        // the effect of resetting those to their initial values.
+        self.data_store.clear_all();
+        // TODO: 1552872: Clear the event store, once the event metric lands.
```
Dexterp37: Since you're working on both the PRs, let's make sure this either happens here or in #161 (so that it doesn't get lost). You could file a follow-up bug, if you think is appropriate.

glean-core/src/metrics/uuid.rs
```diff
@@ -53,32 +55,31 @@ impl UuidMetric {
...
-    pub fn generate_and_set(&self, storage: &Glean) -> uuid::Uuid {
-        let uuid = uuid::Uuid::new_v4();
+    pub fn generate_and_set(&self, storage: &Glean) -> Uuid {
+        let uuid = Uuid::new_v4();
         self.set(storage, uuid);
         uuid
     }
 
-    /// Generate a new random UUID if none is stored yet.
+    /// Get the stored Uuid value.
```
Dexterp37: Let's call out that this is not expected to be user-facing API, but rather an internal thing for a specific purpose (state the purpose too :) ).

glean-core/src/lib.rs
mdboom: I plan to merge #161 first, then this can be added here.

glean-core/src/metrics/uuid.rs
badboy: IMO that's communicated by `pub(crate)`.

PR #173, opened 2019-07-11
Add support for the experiments API

Dexterp37:
This is a PR for adding experiments support to glean-core:

- Introduces an `ExperimentMetric` (not meant to be used directly).
- Provides serialization support through Serde.
- Provides the Glean.set_experiment* API.
- Enables Glean-ac test coverage.
- Ports tests from ExperimentsStorageEngineTest.kt to rust


glean-core/src/database/mod.rs
```diff
@@ -233,6 +233,14 @@ impl Database {
             writer.commit().unwrap();
         });
     }
+
+    /// Removes a metric from the storage (only Application Lifetime).
+    pub fn remove_application_lifetime_metric(&self, storage_name: &str, metric_name: &str) {
+       let final_key = format!("{}#{}", storage_name, metric_name);
```
mdboom: Maybe factor this out into its own function so it won't become out of sync with the code that *adds* to app_lifetime_data?

glean-core/src/database/mod.rs
```diff
@@ -233,6 +233,14 @@ impl Database {
             writer.commit().unwrap();
         });
     }
+
+    /// Removes a metric from the storage (only Application Lifetime).
+    pub fn remove_application_lifetime_metric(&self, storage_name: &str, metric_name: &str) {
```
mdboom: Maybe it's better to have a generic `remove` method, to be symmetrical with `record`?  It seems weird that the caller would have to care about the lifetime of the metric.

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,41 @@ impl StorageManager {
...
+                let name = metric_name.splitn(2, '#').next().unwrap(); // safe unwrap, first field of a split always valid
+                let map = snapshot
+                    .entry(metric.category().into())
+                    .or_insert_with(HashMap::new);
+                map.insert(name.to_string(), metric.as_json());
+            }
+        };
+
+        // FIXME: glean_internal_info should probably not be known here or, at least, shared.
+        storage.iter_store_from(Lifetime::Application, "glean_internal_info".into(), &mut snapshotter);
```
mdboom: I'm probably missing a detail, but... if `glean_internal_info` *only* contains experiments, then would we need a special case snapshot function?  Could we just call `snapshot_as_json(glean, "glean_internal_info")`?

Dexterp37: `glean_internal_info` is the same storage used for sequence numbers and start_time, so it contains other things as well. However, they live in PingMaker

mdboom: Ah.  Would using a different store name *just* for experiments help?  i.e. `glean_experiments`?

glean-core/src/ping/mod.rs
```diff
@@ -107,12 +107,20 @@ impl PingMaker {
...
-        json!({
+        let mut map = json!({
             "ping_type": storage_name,
             "seq": self.get_ping_seq(glean, storage_name),
             "start_time": start_time,
             "end_time": end_time,
-        })
+        });
+
+        // Get the experiment data, if available.
```
badboy: What's the format of the experiment data in this section?

glean-core/src/metrics/experiment.rs
```diff
@@ -0,0 +1,113 @@
...
+
+use crate::error_recording::{record_error, ErrorType};
+use crate::metrics::Metric;
+use crate::metrics::MetricType;
+use crate::storage::StorageManager;
+use crate::CommonMetricData;
+use crate::Glean;
+use crate::Lifetime;
+
+// FIXME: this should be shared?
```
badboy: Shared with what?

glean-core/src/metrics/experiment.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// ## Arguments
+    ///
+    /// * `glean` - The Glean instance this metric belongs to.
+    /// * `branch` -  the active branch of the experiment.
+    /// * `extra` - an optional, user defined String to String map used to provide richer
+    ///             experiment context if needed.
+    ///
+    /// ## Notes
+    ///
+    /// Logs an error if the `amount` is 0 or negative.
```
badboy: Stray comment that doesn't apply here.

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,41 @@ impl StorageManager {
...
+    ///  Snapshot the experiments.
+    ///
+    /// ## Arguments:
+    ///
+    /// * `storage` - The database to get data from.
+    ///
+    /// ## Return value
+    ///
+    /// Returns a JSON representation of the stored data.
+    /// Returns `None` if no data for the store exists.
```
badboy: This should be adjusted. There's no "store" passed in

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,41 @@ impl StorageManager {
...
+
+    ///  Snapshot the experiments.
+    ///
+    /// ## Arguments:
+    ///
+    /// * `storage` - The database to get data from.
+    ///
+    /// ## Return value
+    ///
+    /// Returns a JSON representation of the stored data.
```
badboy: This can be more explicit and say "experiment data" or something

glean-core/src/metrics/mod.rs
```diff
@@ -26,6 +27,7 @@ use crate::Glean;
 pub use self::boolean::BooleanMetric;
 pub use self::counter::CounterMetric;
 pub use self::datetime::DatetimeMetric;
+pub use self::experiment::ExperimentMetric; // FIXME: should this be crate-only? But then how to run tests? Using Glean.set*?
```
badboy: Yeah, in general it sounds sensible to make this private. Maybe it is enough to test the public API on this?

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,41 @@ impl StorageManager {
...
+        storage: &Database
+    ) -> Option<JsonValue> {
+        let mut snapshot: HashMap<String, HashMap<String, JsonValue>> = HashMap::new();
+
+        let mut snapshotter = |metric_name: &[u8], metric: &Metric| {
+            let metric_name = String::from_utf8_lossy(metric_name).into_owned();
+            if metric_name.ends_with("#experiment") {
+                let name = metric_name.splitn(2, '#').next().unwrap(); // safe unwrap, first field of a split always valid
+                let map = snapshot
+                    .entry(metric.category().into())
```
badboy: We do know the category beforehand. Maybe we can just make use of that knowledge.

glean-core/src/metrics/mod.rs
```diff
@@ -81,6 +85,7 @@ impl Metric {
             Metric::Boolean(_) => "boolean",
             Metric::Counter(_) => "counter",
             Metric::Datetime(_, _) => "datetime",
+            Metric::Experiment(_) => "experiments",
```
badboy: If we don't need to ask the metric for its category further below, maybe we should make this panic, so we catch when this gets called erroneously.

glean-core/src/database/mod.rs
Dexterp37: Good point!

glean-core/src/storage/mod.rs
Dexterp37: I guess so, but I think we should value consistency: other things do not have their own storage, we might want to think about this: do we want to transition other things (sequence numbers, start/end_time) to their own storage or keep them that way? What's the cleanest way going forward, in case we might want to add more things? Do we want an abstraction layer over this, e.g. "InternalMetricsDatabase" or something?

glean-core/src/storage/mod.rs
Dexterp37: You're right. I've also documented the returned data structure here.

glean-core/src/ping/mod.rs
Dexterp37: I added the docs about the format in the relevant StorageManager function.

glean-core/src/metrics/experiment.rs
Dexterp37: With the functions in `StorageManager` using the same constant :(

glean-core/src/metrics/mod.rs
Dexterp37: I followed your suggestion, however I have another problem with `RecordedExperimentData`. See the new comment.

glean-core/src/metrics/mod.rs
Dexterp37: Mh, what do you mean here? That we should remove the entry from `as_json` and then panic if we trip on Experiments in this function?

glean-core/src/storage/mod.rs
Dexterp37: Do you mean instead of calling `metric.category()`?

glean-core/src/metrics/mod.rs
```diff
@@ -28,6 +29,10 @@ use crate::Glean;
 pub use self::boolean::BooleanMetric;
 pub use self::counter::CounterMetric;
 pub use self::datetime::DatetimeMetric;
+pub(crate) use self::experiment::ExperimentMetric;
+// Note: clippy complains about this being useless but, if I remove
+// the next line, then tests in glean-core\src\lib.rs will fail to build.
+pub(crate) use self::experiment::RecordedExperimentData;
```
badboy: Two possibilities:

1. Only export it in test mode:

```suggestion
#[cfg(test)]
pub(crate) use self::experiment::RecordedExperimentData;
```

2. Below in the `pub enum Metric` just use `RecordedExperimentData`, not `experiment::RecordedExperimentData`.

Both should make the clippy complain go away

glean-core/src/metrics/mod.rs
badboy:
```suggestion
            Metric::Experiment(_) => panic!("Experiments should not be serialized through this"),
```

* bikeshed on that wording.

than wherever we serialize experiments, avoid calling `category()` on them

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+    /// ## Arguments:
+    ///
+    /// * `storage` - The database to get data from.
+    /// * `store_name` - The store name to look into.
+    ///
+    /// ## Return value
+    ///
+    /// Returns a JSON representation of the experiment data, in the following format:
+    ///
+    /// {
```
badboy:
```suggestion
    /// ```
    /// {
```

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+    ///  'experiment-id': {
+    ///    'branch': 'branch-id',
+    ///    'extra': {
+    ///      'additional': 'property',
+    ///      // ...
+    ///    }
+    ///  }
+    /// }
```
badboy:
```suggestion
    /// }
    /// ```
```

so that it is nicely rendered in the docs

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+    ///
+    /// * `storage` - The database to get data from.
+    /// * `store_name` - The store name to look into.
+    ///
+    /// ## Return value
+    ///
+    /// Returns a JSON representation of the experiment data, in the following format:
+    ///
+    /// {
+    ///  'experiment-id': {
```
badboy: probably should use double-quotes, because that's what JSON uses.

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+    ///  }
+    /// }
+    ///
+    /// Returns `None` if no data for experiments exists.
+    pub fn snapshot_experiments_as_json(
+        &self,
+        storage: &Database,
+        store_name: &str,
+    ) -> Option<JsonValue> {
+        let mut snapshot: HashMap<String, HashMap<String, JsonValue>> = HashMap::new();
```
badboy:
```suggestion
        let mut snapshot: HashMap<String, JsonValue>> = HashMap::new();
```

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+
+        let mut snapshotter = |metric_name: &[u8], metric: &Metric| {
+            let metric_name = String::from_utf8_lossy(metric_name).into_owned();
+            if metric_name.ends_with("#experiment") {
+                let name = metric_name.splitn(2, '#').next().unwrap(); // safe unwrap, first field of a split always valid
+                let map = snapshot
```
badboy: Remove this `let map = ...` expression.

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+                    .entry(metric.category().into())
+                    .or_insert_with(HashMap::new);
+                map.insert(name.to_string(), metric.as_json());
```
badboy:
```suggestion
                snapshot.insert(name.to_string(), metric.as_json());
```

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+            }
+        };
+
+        // FIXME: glean_internal_info should probably not be known here or, at least, shared.
+        storage.iter_store_from(Lifetime::Application, store_name, &mut snapshotter);
+
+        if snapshot.is_empty() {
+            None
+        } else {
+            Some(json!(snapshot))
```
badboy:
```suggestion
            Some(json!({"experiments": snapshot}))
```

badboy: Wait, do we need the outer object? The new doc comment says no, but we previously had it, so it depends on how it is handled when inserted into the ping payload.

glean-core/src/storage/mod.rs
```diff
@@ -143,4 +143,92 @@ impl StorageManager {
...
+            .collect();
+
+        let metric = ExperimentMetric::new("some-experiment".to_string());
+
+        metric.set_active(&glean, "test-branch".to_string(), Some(extra));
+        let snapshot = StorageManager
+            .snapshot_experiments_as_json(glean.storage(), "glean_internal_info")
+            .unwrap();
+        assert_eq!(
+            json!({"experiments": {"some-experiment": {"branch": "test-branch", "extra": {"test-key": "test-value"}}}}),
```
badboy: Ah, there the `{"experiments": ...}` exists, so we should be clear about which way we need.

glean-core/src/ping/mod.rs
```diff
@@ -107,12 +107,21 @@ impl PingMaker {
...
+        if let Some(experiment_data) =
+            StorageManager.snapshot_experiments_as_json(glean.storage(), INTERNAL_STORAGE)
+        {
+            merge(&mut map, &experiment_data);
```
badboy:
```suggestion
            map.as_object_mut().unwrap().insert("experiments", experiment_data);
```

If we don't have the outer object around the snapshotted data we should be able to do that (see other comments on the `snapshot_experiments_as_json` function)

glean-core/src/metrics/mod.rs
Dexterp37: Thanks for this. I had to go for the (1), the (2) would make the build fail.

glean-core/src/storage/mod.rs
Dexterp37: Let's remove the outer `"experiments"`, I like your suggestion more :)

glean-core/src/storage/mod.rs
Dexterp37: Removed the outer block here :)

glean-core/ffi/src/lib.rs
```diff
@@ -50,6 +51,33 @@ unsafe fn from_raw_string_array(arr: RawStringArray, len: i32) -> Vec<String> {
...
+unsafe fn from_raw_string_array_and_string_array(
+    keys: RawStringArray,
+    values: RawStringArray,
+    len: i32,
+) -> Option<HashMap<String, String>> {
+    if keys.is_null() || values.is_null() || len == 0 {
+        return None;
+    }
+
+    let keys_ptrs = std::slice::from_raw_parts(keys, len as usize);
```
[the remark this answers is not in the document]
Dexterp37: I'm happy to hold back landing this and merge as soon as #179 lands.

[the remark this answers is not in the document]
badboy: I'm making changes to how #179 deals with improper values. I'm fine of doing the extra work of changing it for your function too after this PR lands.

glean-core/ffi/src/lib.rs
```diff
@@ -139,5 +167,49 @@ pub extern "C" fn glean_ping_collect(glean_handle: u64, ping_type_handle: u64) -
...
+    branch: FfiStr,
+    extra_keys: RawStringArray,
+    extra_values: RawStringArray,
+    extra_len: i32,
+) {
+    GLEAN.call_infallible(glean_handle, |glean| {
+        let extra =
+            unsafe { from_raw_string_array_and_string_array(extra_keys, extra_values, extra_len) };
+        glean.set_experiment_active(
+            experiment_id.as_str().to_string(),
```
badboy: I'm filing an issue for us to double-check which parts of the FFI can potentially panic (this can).

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1567110

glean-core/src/metrics/experiment.rs
```diff
@@ -0,0 +1,157 @@
...
+pub struct RecordedExperimentData {
+    pub branch: String,
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub extra: Option<HashMap<String, String>>,
+}
+
+/// An experiment metric.
+///
+/// Used to store active experiments.
+/// This is used through the set_experiment_active/set_experiment_inactive Glean SDK API.
```
badboy:
```suggestion
/// This is used through the `set_experiment_active`/`set_experiment_inactive` Glean SDK API.
```

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -195,9 +197,31 @@ open class GleanInternalAPI internal constructor () {
...
+        // The 'extra' map is sent over FFI as a pair of arrays, one containing
+        // the keys and the other containing the values.
+        var keys: StringArray? = null
+        var values: StringArray? = null
+        var numKeys = 0
+
+        extra?.let {
+            numKeys = extra.size
+            keys = StringArray(extra.keys.toTypedArray(), "utf-8")
+            values = StringArray(extra.values.toTypedArray(), "utf-8")
```
mdboom: As discussed in #161, technically according to the API, this could result in the keys and values being in a different order.  I think to be certain, we need to iterate over the key/value pairs.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -219,26 +246,59 @@ open class GleanInternalAPI internal constructor () {
...
+    private fun getMapFromJSONObject(jsonRes: JSONObject): Map<String, String>? {
+        return jsonRes.optJSONObject("extra")?.let {
+            val map = mutableMapOf<String, String>()
+            it.names()?.let { names ->
+                for (i in 0 until names.length()) {
+                    map[names.getString(i)] = it.getString(names.getString(i))
+                }
+            }
+            map
+        }
```
mdboom: I'm *shocked* this isn't somehow built-in to the Android JSON library, but I can't find it...

glean-core/ffi/src/lib.rs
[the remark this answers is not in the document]
Dexterp37: Thanks!

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: Likewise! I spent some time looking for it and was tempted to add an extension for it :P

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: Thank you, fixed this.

PR #174, opened 2019-07-12
Make constructor function optional in macro.

badboy:
This allows skipping the automatic implementation of the constructor
function.
It then has to be implemented manually (which might be required if the
type has some extra work to do during construction).

cc @mdboom 	this might solve your issues in #161 

### DIFF
--- a/glean-core/ffi/src/macros.rs
+++ b/glean-core/ffi/src/macros.rs
@@ -20,7 +20,7 @@
 #[macro_export]
 macro_rules! define_metric {
     ($metric_type:ident => $metric_map:ident {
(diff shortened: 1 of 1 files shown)


glean-core/ffi/src/macros.rs
```diff
@@ -20,7 +20,7 @@
 #[macro_export]
 macro_rules! define_metric {
     ($metric_type:ident => $metric_map:ident {
-        new -> $new_fn:ident($($new_argname:ident: $new_argtyp:ty),* $(,)*),
+        $(new -> $new_fn:ident($($new_argname:ident: $new_argtyp:ty),* $(,)*),)?
```
Dexterp37: Let's update the comments above to state that `$new_fn(...)` is now optional

PR #176, opened 2019-07-15
Don't overwrite existing value/running timer with raw API

badboy:
This adds tests to verify the behavior [we discussed in bug 1562859](https://bugzilla.mozilla.org/show_bug.cgi?id=1562859).

### DIFF
--- a/glean-core/src/metrics/timespan.rs
+++ b/glean-core/src/metrics/timespan.rs
@@ -108,7 +108,12 @@ impl TimespanMetric {
         }
 
         if self.start_time.is_some() {
-            log::warn!("Timespan already running. Raw value not recorded.");
+            record_error(
+                glean,
+                &self.meta,
+                ErrorType::InvalidValue,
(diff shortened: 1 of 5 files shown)


glean-core/src/metrics/timespan.rs
```diff
@@ -108,7 +108,12 @@ impl TimespanMetric {
         }
 
         if self.start_time.is_some() {
-            log::warn!("Timespan already running. Raw value not recorded.");
+            record_error(
+                glean,
+                &self.meta,
+                ErrorType::InvalidValue,
```
mdboom: One could argue this isn't "Invalid Value" but is more like "Invalid State".  But that's a question for another bug...

Dexterp37: Yup, let's file a new bug for this ;)

badboy: Done: https://bugzilla.mozilla.org/show_bug.cgi?id=1566380

PR #179, opened 2019-07-16
Safer string array

badboy:

### DIFF
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -107,30 +107,47 @@ fn from_raw_int_array_and_string_array(
 }
 
 /// Create a HashMap<String, String> from a pair of C string arrays.
-unsafe fn from_raw_string_array_and_string_array(
+///
+/// Returns an error if any of the strings contain invalid UTF-8 characters.
+///
+/// ## Safety
+///
+/// * We check the array pointer for validity (non-null).
+/// * FfiStr checks each individual char pointer for validity (non-null).
+/// * We discard invalid char pointers (null pointer).
+/// * Invalid UTF-8 in any string will return an error from this function.
+fn from_raw_string_array_and_string_array(
     keys: RawStringArray,
     values: RawStringArray,
     len: i32,
-) -> Option<HashMap<String, String>> {
-    if keys.is_null() || values.is_null() || len == 0 {
-        return None;
-    }
+) -> glean_core::Result<Option<HashMap<String, String>>> {
+    unsafe {
+        if keys.is_null() || values.is_null() || len == 0 {
+            return Ok(None);
+        }
 
-    let keys_ptrs = std::slice::from_raw_parts(keys, len as usize);
-    let values_ptrs = std::slice::from_raw_parts(values, len as usize);
+        let keys_ptrs = std::slice::from_raw_parts(keys, len as usize);
+        let values_ptrs = std::slice::from_raw_parts(values, len as usize);
 
-    Some(
-        keys_ptrs
+        let res: glean_core::Result<_> = keys_ptrs
             .iter()
             .zip(values_ptrs.iter())
             .map(|(&k, &v)| {
-                (
-                    FfiStr::from_raw(k).into_string(),
-                    FfiStr::from_raw(v).into_string(),
-                )
+                let k = FfiStr::from_raw(k)
+                    .as_opt_str()
+                    .map(|s| s.to_string())
+                    .ok_or_else(glean_core::Error::utf8_error)?;
+
+                let v = FfiStr::from_raw(v)
+                    .as_opt_str()
+                    .map(|s| s.to_string())
+                    .ok_or_else(glean_core::Error::utf8_error)?;
+
+                Ok((k, v))
             })
-            .collect(),
-    )
+            .collect();
+        res.map(Some)
+    }
 }
 
 /// Initialize the logging system based on the target platform. This ensures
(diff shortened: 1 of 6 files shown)


glean-core/ffi/src/lib.rs
```diff
@@ -37,17 +37,28 @@ lazy_static! {
...
-        return vec![];
+///
+/// ## Safety
+///
+/// * We check the array pointer for validity (non-null)
+/// * FfiStr checks each individual char pointer for validity (non-null)
+/// * We discard invalid char pointers (null pointer)
+/// * invalid UTF-8 is replaced with the replacement character (guaranteed by ffi-support)
+fn from_raw_string_array(arr: RawStringArray, len: i32) -> Vec<String> {
+    unsafe {
```
Dexterp37: Curiosity: is it common practice to wrap the whole function in `unsafe` rather then restricting this to the "scary parts"? I could imagine `arr.is_null()` being kind of safe.

badboy: Yes and no. Preference differs.
I do this because:

* `unsafe` marks a block where _some_ parts can't be proven by the compiler (it still does tons of checks though ;))
* `unsafe` does not introduce unsafe code, but merely saying "what follows the compiler can't proof safe, but I, the dev, guarantee that all invariants are held and that no _safe_ use of this block can cause unsafety"

The code _around_ the actual unsafe function calls (e.g. `from_raw_parts`) are still relevant for the call itself (that is: `from_raw_parts` would explode if we pass it a null pointer, so we need to check that first).

Dexterp37: Thank you very much for the explanation, much clearer now!

glean-core/ffi/src/lib.rs
```diff
@@ -37,17 +37,28 @@ lazy_static! {
...
+        if arr.is_null() || len == 0 {
+            return vec![];
+        }
+
+        let arr_ptrs = std::slice::from_raw_parts(arr, len as usize);
+        arr_ptrs
+            .iter()
+            .flat_map(|&p| {
+                // Drop invalid strings
+                FfiStr::from_raw(p).into_opt_string()
```
mdboom: I assume the typical failure here would be invalid UTF-8 codepoints?  I'm not sure whether the best behavior is to 

1) drop invalid strings (as here)
2) replace them with empty strings
3) fail the entire conversion (by returning an error result)

I could see this introducing a whole other class of error downstream if the array of strings suddenly had a different length than expected.  I think I would lean on option 3 here, with 2 being a close second.  Though a lot of this depends on expected usage -- but it's awfully hard for the caller to make sense of this and act accordingly from the outside here.

glean-core/ffi/src/lib.rs
mdboom: So. philosophically, the bulk of this change is really just hiding the `unsafe` within the function, rather than doing anything that's inherently safer?  (Fine either way, just seems like an interesting "Rustism")...

badboy: We can't skip unsafety completely, we're dealing with raw pointers here.
[As explained here](https://github.com/mozilla/glean/pull/179#discussion_r303951239) `unsafe` doesn't mean it's all actually unsafe, it's just necessary to explicitly deal with all the cases the compiler can't prove.

glean-core/ffi/src/lib.rs
badboy: Right.
Up to now glean-core just ... explodes on invalid UTF-8 codepoints, so we're not much better off.

And acutally, we're not dropping strings with invalid UTF-8 codepoints, we replace those codepoints:

>  Note that unlike FfiStr::as_opt_str, invalid UTF-8 is replaced with the replacement character instead of causing us to return None.

So ... I guess we should decide first what we want to do. I'm very much ok with returning a result (and thus bailing out early)

badboy:
So, now up for vote:

1. 🎉 Drop strings that contain invalid UTF-8 codepoints
2. 🚀 Replace invalid UTF-8 codepoints in strings with the replacement character
3. 👀 Fail the array decoding and thus fail any further operation. Log the error (no crash in production obviously, crashes in debug mode though)

Vote by emoji response to this comment (or leave a comment)

cc @mozilla/glean

Dexterp37:
> 
>     3. 👀 Fail the array decoding and thus fail any further operation. Log the error (no crash in production obviously, crashes in debug mode though)

Would that allow us to do proper reporting for the metrics using the Glean SDK error reporting system?

mdboom:
> Would that allow us to do proper reporting for the metrics using the Glean SDK error reporting system?

Yes -- though it might not be possible to associate it with a particular metric, which is how that error reporting generally works.  (i.e. if the failure happened while instantiating a metric).  But reporting this error somehow would be good.

Maybe I'm missing something, but it should be exceedingly rare -- I don't think it could happen from a misuse of the Kotlin API.  It should already be type checked at that level, and if it's a valid Kotlin string, it should encode to valid UTF-8.  I think it would require a bug in the FFI layer itself, or memory corruption...

badboy:
> Maybe I'm missing something, but it should be exceedingly rare -- I don't think it could happen from a misuse of the Kotlin API. It should already be type checked at that level, and if it's a valid Kotlin string, it should encode to valid UTF-8. I think it would require a bug in the FFI layer itself, or memory corruption...

True! It's essentially covering us against certain memory corruptions.

badboy:
> Would that allow us to do proper reporting for the metrics using the Glean SDK error reporting system?

Not easily (for now) as this is in the FFI.

mdboom:
Now that #161 is merged, there is a second FFI helper function of this type.  I don't know if you want to apply the same transformation to it as part of this PR or defer it, but we should not forget to do it...

badboy:
@mdboom I'll take that here

badboy:
> Looks good! Would you like to rebase this on master, since #173 has been merged, or do you want to have a follow-up?

I'll rebase and include that.

PR #180, opened 2019-07-17
Bug 1566382 - List error recording values for timespan

badboy:
This also includes documenting the raw API

### DIFF
--- a/docs/user/metrics/timespan.md
+++ b/docs/user/metrics/timespan.md
@@ -67,6 +67,23 @@ assertTrue(Auth.loginTime.testHasValue())
 assertTrue(Auth.loginTime.testGetValue() > 0)
 ```
 
+### Raw API
+
+> **Note**: The raw API was designed to support a specific set of use-cases.
+> Please consider using the higher level APIs listed above.
+
+It's possible to explicitly set the timespan value, in nanoseconds.
+This API should only be used if your library or application requires recording times in a way that can not make use of `start`/`stop`/`cancel`.
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.HistorySync
+
+val duration = SyncResult.status.syncs.took.toLong()
+HistorySync.setRawNanos(duration)
+```
+
+The raw API will not overwrite a running timer or existing timespan value.
+
 ## Limits
 
 * None.
--- a/docs/user/metrics/timespan.md
+++ b/docs/user/metrics/timespan.md
@@ -77,9 +77,12 @@ assertTrue(Auth.loginTime.testGetValue() > 0)
 
 ## Recorded errors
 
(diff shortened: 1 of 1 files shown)


docs/user/metrics/timespan.md
```diff
@@ -78,8 +93,10 @@ assertTrue(Auth.loginTime.testGetValue() > 0)
 ## Recorded errors
 
 * `invalid_value`: If recording a negative timespan.
```
Dexterp37: nit: could this be a two layers list?

- `invalid_value`:
  - If starting a timer..
  - If stopping a timer...

docs/user/metrics/timespan.md
```diff
@@ -67,6 +67,21 @@ assertTrue(Auth.loginTime.testHasValue())
 assertTrue(Auth.loginTime.testGetValue() > 0)
 ```
 
+### Raw API
+
+> **Note**: It's recommended to use the timer API listed above.
```
Dexterp37: nit: maybe we should slightly rephrase this to something like "The raw API was designed to support a specific set of use-cases. Please consider using the higher level APIs listed above."

docs/user/metrics/timespan.md
```diff
@@ -67,6 +67,21 @@ assertTrue(Auth.loginTime.testHasValue())
...
+
+It's possible to explicitly set the timespan value, in nanoseconds.
+This API should only be used if your library or application requires recording times in a way that can not make use of `start`/`stop`/`cancel`.
```
Dexterp37: nit: should we add a specific example here? e.g. storing the values of low-level os timers?

badboy: Hm, it's currently nowhere used so I have a hard time coming up with an example

Dexterp37: Sync was planning on use this to report the time it took for some sync operation to complete on the Rust side. Another example could be the use of some low level API that just exposes elapsed times

docs/user/metrics/timespan.md
```diff
@@ -67,6 +67,23 @@ assertTrue(Auth.loginTime.testHasValue())
...
+> Please consider using the higher level APIs listed above.
+
+It's possible to explicitly set the timespan value, in nanoseconds.
+This API should only be used if your library or application requires recording times in a way that can not make use of `start`/`stop`/`cancel`.
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.HistorySync
+
+val duration = SyncResult.status.syncs.took.toLong()
+HistorySync.setRawNanos(duration)
```
badboy: @Dexterp37 	What do you think about this (kinda constructed) example?

Dexterp37: Good enough for me! Let's see what @mdboom thinks about this too ;)

PR #181, opened 2019-07-17
Skip Android tests if commit contains marker

badboy:
Here's an idea to reduce CI time for doc-only changes:

For doc-only changes we might not require the full CI run.
The Android tests are what's taking the longest, so let's make this
easily skippable.

Just add `[skip android]` somewhere in the commit message.

* Should we do this?
* If so, where to document?
* If so, skipping only Android or also rust tests? What about lints?
    * Everything else is reasonably fast, which is why I didn't add a way to skip it yet
* Naming: `skip android`, `doc only`, ...
    * not possible: `ci skip` - circleci uses that to skip the full build (but we require the build to deploy docs)

**Update**

* Docs added.
* Marker changed to `[doc only]`

### DIFF
--- a/docs/dev/testing.md
+++ b/docs/dev/testing.md
@@ -90,4 +90,4 @@ It is possible to completely skip running CI on a given push by including the fo
 [ci skip]
 ```
 
-This should only be used for meatdata files, such as those in `.github`, `LICENSE` or `CODE_OF_CONDUCT.md`.
+This should only be used for metadata files, such as those in `.github`, `LICENSE` or `CODE_OF_CONDUCT.md`.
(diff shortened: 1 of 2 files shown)


mdboom:
Another reason to never skip the doc build is that it performs an internal link check -- as in, doc-only changes can still introduce errors that we can (and should) catch.

Dexterp37:
> 
> 
> I agree this needs docs. As to where, I don't see one obvious place. I would be equally ok with adding a "Testing in CI" section to `dev/testing.md`, or adding a `Special commit tokens` section to `dev/contributing.md`. (And we could also document `[skip ci]` there, even though that's a built-in feature of Circle).

I agree that both locations for the docs are equally appealing, I (g)lean a bit more toward "Testing in CI", but I have no strong opinion on this.

docs/dev/testing.md
```diff
@@ -56,3 +56,38 @@ To run the full Android test suite, in the "Gradle" pane, navigate to `glean-cor
...
+
+### Skipping CI completely
+
+It is possible to completely skip running CI on a given push by including the following literal string in the commit message:
+
+```
+[ci skip]
+```
+
+This should only be used for meatdata files, such as those in `.github`, `LICENSE` or `CODE_OF_CONDUCT.md`.
```
mdboom:
```suggestion
This should only be used for metadata files, such as those in `.github`, `LICENSE` or `CODE_OF_CONDUCT.md`.
```

PR #182, opened 2019-07-17
Add documentation for custom ping testing

Dexterp37:
This additionally adds a batch file to build docs on Windows and an explicit mention to the mdbook requirement.

### DIFF
--- a/docs/user/pings/testing-custom-pings.md
+++ b/docs/user/pings/testing-custom-pings.md
@@ -13,10 +13,6 @@ One possible strategy could be to wrap the Glean SDK API call to send the ping i
 Let us start by defining a custom ping with a sample metric in it. Here is the `pings.yaml` file:
 
 ```yaml
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
 $schema: moz://mozilla.org/schemas/glean/pings/1-0-0
 
 my_custom_ping:
@@ -36,19 +32,14 @@ my_custom_ping:
 And here is the `metrics.yaml`
 
 ```yaml
-# This Source Code Form is subject to the terms of the Mozilla Public
-# License, v. 2.0. If a copy of the MPL was not distributed with this
-# file, You can obtain one at http://mozilla.org/MPL/2.0/.
-
 $schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
 
-my_custom_ping:
+custom_ping_data:
   sample_string:
     type: string
     lifetime: ping
     description: >
-      An hashed and salted version of the Google Advertising ID from the device.
-      This will never be sent in a ping that also contains the client_id.
+      A sample string metric for demonstrating unit tests for custom pings.
     send_in_pings:
       - my_custom_ping
     bugs:
@@ -64,7 +55,7 @@ A potential usage of the Glean SDK generated API could be the following:
 
 ```kotlin
 import my.component.GleanMetrics.Pings
-import my.component.GleanMetrics.MyCustomPing
+import my.component.GleanMetrics.CustomPingData
 
 class MyCustomPingScheduler {
   /**
@@ -77,7 +68,7 @@ class MyCustomPingScheduler {
    */
   fun addSomeData() {
     // Record some sample data.
-    MyCustomPing.sampleString.set("test-data")
+    CustomPingData.sampleString.set("test-data")
   }
 
   /**
@@ -94,7 +85,7 @@ class MyCustomPingScheduler {
    */
   @VisibleForTesting(otherwise = VisibleForTesting.NONE)
   internal fun sendPing() {
-    MyCustomPing.send()
+    Pings.MyCustomPing.send()
   }
 }
 ```
@@ -102,6 +93,14 @@ class MyCustomPingScheduler {
 Finally, here is a simple unit test that intercepts the `MyCustomPingScheduler.schedulePing()` call in order to perform the validation on the data. This specific example uses Mockito, but any other framework would work.
 
 ```kotlin
+// Metrics and pings definitions.
+import my.component.GleanMetrics.Pings
+import my.component.GleanMetrics.CustomPingData
+
+// Mockito imports for using spies.
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.`when`
+
 @Test
 fun `verify custom ping metrics`() {
   // Enable testing mode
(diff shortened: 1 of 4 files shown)


Dexterp37:
@linacambridge would you kindly give feedback on this too? :)

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+
+my_custom_ping:
+  sample_string:
+    type: string
+    lifetime: ping
+    description: >
+      An hashed and salted version of the Google Advertising ID from the device.
+      This will never be sent in a ping that also contains the client_id.
+    send_in_pings:
+      - my_custom_ping
```
mdboom: Sphinx has a way to [highlight specific lines in a source code listing](http://www.sphinx-doc.org/en/master/usage/restructuredtext/directives.html#directive-code-block).  Wish we could do that here, because it's really just this line that's important, but I don't think mdbook supports that.

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+  internal fun sendPing() {
+    MyCustomPing.send()
+  }
+}
+```
+
+Finally, here is a simple unit test that intercepts the `MyCustomPingScheduler.schedulePing()` call in order to perform the validation on the data. This specific example uses Mockito, but any other framework would work.
+
+```kotlin
+@Test
```
mdboom: Let's put the imports necessary for `spy` here...  I always forget which mocking library to use that's compatible with a-c.

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+One possible strategy could be to wrap the Glean SDK API call to send the ping in a function that can be mocked in the unit test. This would allow for checking the status and the values of the metrics contained in the ping at the time in which the application would have sent it.
+
+## Example testing of a custom ping
+
+Let us start by defining a custom ping with a sample metric in it. Here is the `pings.yaml` file:
+
+```yaml
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
```
mdboom: I feel like the license header just adds clutter to the docs, but don't feel strongly about it...

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+And here is the `metrics.yaml`
+
+```yaml
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+my_custom_ping:
```
badboy: Let's use another category here so that it's not to be confused with the ping declaration from the other file.

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+  sample_string:
+    type: string
+    lifetime: ping
+    description: >
+      An hashed and salted version of the Google Advertising ID from the device.
+      This will never be sent in a ping that also contains the client_id.
```
badboy: Let's use a sample description and not something that mentiongs the GA ID

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+    notification_emails:
+      - custom-ping-owner@example.com
+    expires: "2019-10-01"
+```
+
+A potential usage of the Glean SDK generated API could be the following:
+
+```kotlin
+import my.component.GleanMetrics.Pings
+import my.component.GleanMetrics.MyCustomPing
```
badboy: See, it's already confusing me here. `MyCustomPing` is the namespace containing the metrics.

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+    sendPing()
+  }
+
+  /**
+   * Internal function to only be overriden in tests. This
+   * calls the Glean SDK API to send custom pings.
+   */
+  @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+  internal fun sendPing() {
+    MyCustomPing.send()
```
badboy: Shouldn't this call something like `Pings.MyCustomPing.send()`?

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,121 @@
...
+fun `verify custom ping metrics`() {
+  // Enable testing mode
+  // (Perhaps called from a @Before method so it precedes every test in the suite.)
+  Glean.enableTestingMode()
+
+  val scheduler = spy(MyCustomPingScheduler())
+  `when`(scheduler.sendPing()).then {
+    // Here we validate the content of the ping.
+    assertTrue(MyCustomPing.sampleString.testHasValue())
+    assertEquals("test-data", MyCustomPing.sampleString.testGetValue())
```
badboy: If `MyCustomPing` is not the ping, but the metric namespace, than I think it's more appropriate to say "Here we validate the content that goes into the ping".

Does `then` replace the defined function or does it just add additional code?
If the former, that also means the storage is not cleared, so it can't properly test multiple successive pings.

docs/user/pings/testing-custom-pings.md
Dexterp37: Good point :) Let me remove this

docs/user/pings/testing-custom-pings.md
Dexterp37: Ah, that would be neat. The only similar thing I found was [hiding code lines](https://github.com/rust-lang-nursery/mdBook/blob/master/book-example/src/format/theme/syntax-highlighting.md), but it's only available for rust code blocks.

docs/user/pings/testing-custom-pings.md
Dexterp37: Oh, great idea. I can see how this could get confusing.

docs/user/pings/testing-custom-pings.md
Dexterp37: Yup, sorry that slipped through my copy-pasta.

docs/user/pings/testing-custom-pings.md
Dexterp37: Ouch, I just sorted that out.

docs/user/pings/testing-custom-pings.md
Dexterp37: Yes. You're correct!

docs/user/pings/testing-custom-pings.md
Dexterp37: Right, we need to make sure the original function gets called. I slightly modified this.

Dexterp37:
> 
> 
> All blocking questions cleared.
> 
> One more: sending in test mode doesn't actually cause an HTTP request, right?

That depends on the application. Even if it did, the `applicationId` would be different: CI doesn't pollute release tables.

badboy:
> That depends on the application. Even if it did, the `applicationId` would be different: CI doesn't pollute release tables.

Right, but an external HTTP request could make the test fail.

badboy:
(anyway, that's not blocking this PR)

docs/user/pings/testing-custom-pings.md
```diff
@@ -0,0 +1,125 @@
...
+
+// Mockito imports for using spies.
+import org.mockito.Mockito.spy
+import org.mockito.Mockito.`when`
+
+@Test
+fun `verify custom ping metrics`() {
+  // Enable testing mode
+  // (Perhaps called from a @Before method so it precedes every test in the suite.)
+  Glean.enableTestingMode()
```
linacambridge: Does the first test also need to call `Glean.initialize`, or initialize the work manager? Or does that all happen automatically now?

Dexterp37: Oh, darn, good catch. That's definitely required and missing. Updating this! Thank you so much for the feedback!

PR #184, opened 2019-07-17
Provide timespan value both in the specified unit and in nanoseconds

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1562848

Todo:

- [ ] double-check current tests and change to `AsNanos` where appropriate.

cc @mozilla/glean 
What do you think about having both methods?


badboy:
See linked bug. Behavior differs currently.
Exposing that as a parameter is possible I guess (`Option<TimeUnit>` and use the metric's time unit if `None`).

PR #188, opened 2019-07-19
Regenerate header with updated and expanded config

badboy:

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -25,4 +25,4 @@ Carthage
 # local stuff
 data/
 glean-core/data
-glean_app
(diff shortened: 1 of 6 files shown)


badboy:
> Should we file a bug to revisit the use of nightly here in the future?

I think once we get to the Swift part, we're gonna move around this file anyway and need to properly integrate that into the build. A separate bug is therefore just noise at this point IMO.

mdboom:
> A separate bug is therefore just noise at this point IMO.

Fair enough.  Consider this approved, then.

PR #190, opened 2019-07-19
Upgrade ffi-support to get pointer support

badboy:

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -207,7 +207,7 @@ dependencies = [
 
 [[package]]
 name = "ffi-support"
-version = "0.3.4"
+version = "0.3.5"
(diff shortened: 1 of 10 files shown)


Cargo.lock
```diff
@@ -207,7 +207,7 @@ dependencies = [
 
 [[package]]
 name = "ffi-support"
-version = "0.3.4"
+version = "0.3.5"
```
mdboom: Shouldn't we also specify 0.3.5 explicitly in Cargo.toml if we really need that patch revision (also seems like a semver abuse in ffi-support, but that's a separate question...)

badboy: It's < 1.0, so ... all bets are off anyway :P

PR #191, opened 2019-07-19
[doc only] 1557700: Document some best practices for metrics

mdboom:

### DIFF
--- a/docs/user/adding-new-metrics.md
+++ b/docs/user/adding-new-metrics.md
@@ -36,12 +36,12 @@ toolbar:
       - http://example.com/path/to/data-review
     expires:
       - 2019-06-01  # <-- Update to a date in the future
-    
-  metric2:
+
+  double_click:
     ...
-    
+
 category2.subcategory:  # Categories can contain subcategories
-  metric3:
+  metric:
     ...
 
 ```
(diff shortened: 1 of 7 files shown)


docs/user/metrics/boolean.md
```diff
@@ -12,6 +12,7 @@ flags:
     type: boolean
     description: >
       Records whether a11y is enabled on the device.
+    lifetime: application
```
mdboom: Added these sort of changes to the examples to "dogfood" the recommendations :)

docs/user/adding-new-metrics.md
```diff
@@ -50,7 +50,32 @@ The details of the metric parameters are described in [metric parameters](metric
 
 The `metrics.yaml` file is used to generate `Kotlin` code that becomes the public API to access your application's metrics.
 
-## Metric naming
+## Recommendations for defining new metrics
+
+“There are only two hard things in Computer Science: cache invalidation and naming things” -- attributed to Phil Karlton.
```
mdboom: Maybe this is too cute, but literally these are the two hard things in the recommendations.  Probably won't scale when we have more recommendations.

badboy: ... in which case we simply add the following addendum to the quote:

> and off-by-one errors.

docs/user/adding-new-metrics.md
```diff
@@ -50,7 +50,32 @@ The details of the metric parameters are described in [metric parameters](metric
...
+### Lifetimes
+
+The `lifetime` parameter of a metric defines when it will be reset. There are three options available:
+
+- `ping` (default): The metric is reset each time it is sent in the ping.
+  This is the most common case, and should be used for metrics that are highly dynamic, such as things computed in response to the user's interaction with the application.
+- `user`: The metric is part of the user's profile.
+  This should be used for things that change only when the user's profile is created.
+  It is rare to use this lifetime outside of some metrics that are built in to Glean, such as `client_id`.
+- `application`: The metric is related to an application run, and is reset only when the application restarts.
```
badboy: IMO I would order them by their ... lifetime.
ping < application < user

docs/user/adding-new-metrics.md
```diff
@@ -50,7 +50,32 @@ The details of the metric parameters are described in [metric parameters](metric
...
+  This should be used for things that are constant during the run of an application, such as the operating system version.
+  In practice, these metrics are generally set during application startup.
+  A common mistake---using the `ping` lifetime for these type of metrics---means that they will only be included in the first ping sent during a particular run of the application.
+
+### Naming things
+
+Choose the category and names of metrics to be as specific as possible.
```
badboy: I don't think we currently make any specific use of the fact that we do have categories.
Is there any plan to be able to group things by category in the analysis? Is that something we should call out?

docs/user/adding-new-metrics.md
Dexterp37: +1 on this

mdboom: Good idea.

docs/user/adding-new-metrics.md
mdboom: I'm not aware of any use of them other than as an organizational tool so the metrics aren't just in a "flat" namespace.  But that could change I suppose?  What exactly did you have in mind to call out?  "Categories are a way to organize metrics."?

badboy: I ... don't know really. I guess we should leave it as is and extend this when/if we use categories in some other way eventually

PR #192, opened 2019-07-22
1567914: String truncation shouldn't truncate in the middle of a codepoint

mdboom:
This works by starting one character beyond the truncation point and moves back until it finds a character boundary.  The string is then truncated to the byte immediately preceding that boundary.

An alternative would have been to iterate over characters from the beginning, but that would be O(n), whereas this is O(1) (since barring corrupted string data, it should only have to look at at most 3 bytes).

### DIFF
--- a/glean-core/src/metrics/string.rs
+++ b/glean-core/src/metrics/string.rs
@@ -2,10 +2,10 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use crate::error_recording::{record_error, ErrorType};
 use crate::metrics::Metric;
 use crate::metrics::MetricType;
 use crate::storage::StorageManager;
+use crate::util::truncate_string_at_boundary_with_error;
 use crate::CommonMetricData;
 use crate::Glean;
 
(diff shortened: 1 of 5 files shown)


glean-core/src/metrics/string.rs
```diff
@@ -51,18 +51,7 @@ impl StringMetric {
...
-                "Value length {} exceeds maximum of {}",
-                s.len(),
-                MAX_LENGTH_VALUE
-            );
-            record_error(glean, &self.meta, ErrorType::InvalidValue, msg);
-            s[0..MAX_LENGTH_VALUE].to_string()
-        } else {
-            s
-        };
+        let s = truncate_string_at_boundary_with_error(glean, &self.meta, value, MAX_LENGTH_VALUE);
```
Dexterp37: Would you kindly also update [experiments](https://github.com/mozilla/glean/blob/104d49e8d9fd215e02469c09f6c0a8ab4e63f5e6/glean-core/src/metrics/experiment.rs#L60)?

mdboom: Thanks.  Missed that one.

mdboom:
Thanks.  This is Rusty enough that I'll wait for an approval from @badboy as well.

glean-core/src/util.rs
```diff
@@ -121,4 +183,19 @@ mod test {
             fixed_now
         );
     }
+
+    #[test]
```
badboy: Can we get one test on an ASCII value, just to make sure? e.g. `"1234567890abcdef"` truncated to 10.

badboy:
(actually approving, r+wc)

PR #195, opened 2019-07-23
Add a common histogram base class to `TimingDistributionMetricType`

Dexterp37:
This is porting over the changes from mozilla-mobile/android-components#3829 to glean-core.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
@@ -7,6 +7,7 @@ package mozilla.telemetry.glean.private
 import androidx.annotation.VisibleForTesting
 import android.os.SystemClock
 import com.sun.jna.StringArray
+import mozilla.components.service.glean.private.HistogramBase
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.GleanTimerId
 import mozilla.telemetry.glean.rust.getAndConsumeRustString
@@ -25,7 +26,7 @@ import mozilla.telemetry.glean.rust.toBoolean
 class TimingDistributionMetricType internal constructor(
     private var handle: Long,
     private val sendInPings: List<String>
-) {
+) : HistogramBase {
     /**
      * The public constructor used by automatically generated metrics.
      */
@@ -141,6 +142,26 @@ class TimingDistributionMetricType internal constructor(
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.launch {
             LibGleanFFI.INSTANCE.glean_timing_distribution_cancel(this@TimingDistributionMetricType.handle, timerId)
+        }
+    }
+
+    override fun accumulateSamples(samples: LongArray) {
+        if (!shouldRecord()) {
+            return
+        }
+
+        samples.forEach {
+            println("Kotlin sample: ${it}")
+        }
+
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.launch {
+            LibGleanFFI.INSTANCE.glean_timing_distribution_accumulate_samples(
+                Glean.handle,
+                this@TimingDistributionMetricType.handle,
+                samples,
+                samples.size
+            )
         }
     }
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
@@ -7,7 +7,6 @@ package mozilla.telemetry.glean.private
 import androidx.annotation.VisibleForTesting
 import android.os.SystemClock
 import com.sun.jna.StringArray
-import mozilla.components.service.glean.private.HistogramBase
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.GleanTimerId
 import mozilla.telemetry.glean.rust.getAndConsumeRustString
@@ -72,14 +71,12 @@ class TimingDistributionMetricType internal constructor(
     }
 
     /**
-     * Start tracking time for the provided metric and associated object. This
+     * Start tracking time for the provided metric and [GleanTimerId]. This
      * records an error if it’s already tracking time (i.e. start was already
      * called with no corresponding [stopAndAccumulate]): in that case the original
      * start time will be preserved.
      *
-     * @param timerId The object to associate with this timing.  This allows
-     * for concurrent timing of events associated with different objects to the
-     * same timespan metric.
+     * @return The [GleanTimerId] object to associate with this timing.
      */
     fun start(): GleanTimerId? {
         if (!shouldRecord()) {
@@ -98,12 +95,12 @@ class TimingDistributionMetricType internal constructor(
     }
 
     /**
-     * Stop tracking time for the provided metric and associated object. Add a
+     * Stop tracking time for the provided metric and associated timer id. Add a
      * count to the corresponding bucket in the timing distribution.
      * This will record an error if no [start] was called.
      *
-     * @param timerId The object to associate with this timing.  This allows
-     * for concurrent timing of events associated with different objects to the
+     * @param timerId The [GleanTimerId] to associate with this timing.  This allows
+     * for concurrent timing of events associated with different ids to the
      * same timespan metric.
      */
     fun stopAndAccumulate(timerId: GleanTimerId?) {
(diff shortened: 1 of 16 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
```diff
@@ -144,6 +142,30 @@ class TimingDistributionMetricType internal constructor(
         }
     }
 
+    override fun accumulateSamples(samples: LongArray) {
+        if (!shouldRecord()) {
```
Dexterp37: There's one tricky part here: in glean-ac, if any sample is negative, we report an error. However, doesn't look like we allow for error reporting in Kotlin-land.

mdboom: Could we handle it in the FFI layer by checking the "sign bit" on the u64 coming in?  Something like `if value & 0x8000000000000000` ... report error...?

badboy: We can just take a `i64` properly instead, matching `Long` from the Kotlin side, then do any underflow checks

Dexterp37: All right, I'll go for the `i64` route!

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -171,41 +171,90 @@ impl TimingDistributionMetric {
...
+    ///
+    /// Discards any negative value in `samples` and report an `ErrorType::InvalidValue`
+    /// for each of them.
+    pub fn accumulate_samples_signed(&mut self, glean: &Glean, samples: Vec<i64>) {
+        let sample_count = samples.len();
+        let positive_samples: Vec<u32> = samples
+            .into_iter()
+            .filter(|s| *s >= 0)
+            .map(|s| s as u32)
+            .collect();
```
mdboom: How big do we expect these vectors to be?  If large, maybe it's an issue that it does a copy...?  (Alternatively, `accumulate_samples` could take an iterator and avoid the copy, but then the negative detection would have to happen in a slightly different way...)

Dexterp37: The size of the vectors is to be defined: this is a potential hot path though, so anything we could do here to optimize might be a good win.

badboy: IMO we should simplify this and even avoid rapidly calling record_with all the time.
Just use a for loop, do an if check in there and only accumulate if it is positive.

badboy:
```rust
let mut num_errors = 0;
glean
.storage()
    .record_with(&self.meta, |old_value| {
        let (mut hist, time_unit) = match old_value {
            Some(Metric::TimingDistribution(mut hist, time_unit)) => {
                (hist, time_unit)
            }
            _ => {
                (Histogram::default(), self.time_unit);
            }
        };

        for sample in samples { 
            if sample >= 0 {
                hist.accumulate(sample);
            } else {
                num_errors += 1;
            }
        }


        Metric::TimingDistribution(hist, time_unit)
    });    

let msg = format!("Accumulated {} negative samples", num_errors);
record_error(glean, &self.meta, ErrorType::InvalidValue, msg, num_errors);
```

Dexterp37: Thanks. I had to slightly tweak this because `samples` was not getting captured. Using `move` created another set of problems, which I was not able to deal with :)

glean-core/ffi/src/lib.rs
```diff
@@ -150,6 +151,26 @@ fn from_raw_string_array_and_string_array(
...
+///
+/// * We check the array pointer for validity (non-null).
+fn from_raw_int64_array(values: RawInt64Array, len: i32) -> glean_core::Result<Vec<i64>> {
+    unsafe {
+        if values.is_null() || len == 0 {
+            return Ok(vec![]);
+        }
+
+        let values_ptrs = std::slice::from_raw_parts(values, len as usize);
+        Ok(values_ptrs.iter().map(|&v| v as i64).collect())
```
badboy: This can be just: 

```suggestion
        Ok(values_ptrs.to_vec())
```

(I'd consider a different variable name though, maybe just re-use `values`? Or `value_slice`?)

glean-core/ffi/src/lib.rs
```diff
@@ -150,6 +151,26 @@ fn from_raw_string_array_and_string_array(
     }
 }
 
+/// Create a Vec<u32> from a raw C uint64 array.
+///
+/// This will never return an `Error`. However, this is returning
+/// a `Result` for consistency with `from_raw_string_array` and
```
badboy: We shouldn't return a Result just for consistency with other functions. This only makes us `unwrap` unnecessary below and brings the need to comment why the unwrap is safe.

glean-core/ffi/src/timing_distribution.rs
```diff
@@ -48,6 +50,28 @@ pub extern "C" fn glean_timing_distribution_cancel(metric_id: u64, timer_id: Tim
...
+    metric_id: u64,
+    raw_samples: RawInt64Array,
+    num_samples: i32,
+) {
+    GLEAN.call_infallible(glean_handle, |glean| {
+        TIMING_DISTRIBUTION_METRICS.call_infallible_mut(metric_id, |metric| {
+            // The Kotlin code is sending Long(s), which are 64 bits, as there's
+            // currently no stable UInt type. The positive part of [Int] would not
+            // be enough to represent the values coming in:.
+            // Here Long(s) are handled as i64 and then casted to u32.
```
badboy: The casting happens inside `accumulate_samples_signed` (after appropriate negativity checks). That should be made more clear in the comment.

glean-core/src/error_recording.rs
```diff
@@ -79,7 +81,7 @@ pub fn record_error(
     });
 
     log::warn!("{}: {}", identifier, message);
-    metric.add(glean, 1);
+    metric.add(glean, num_errors.into().unwrap_or(1) as i32);
```
badboy: Why the back-and-forth-casting? We could just take `O: Into<Option<i32>>`.
Call sites with just a number should be fine.

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -198,16 +198,62 @@ impl TimingDistributionMetric {
...
+
+        glean.storage().record_with(&self.meta, |old_value| {
+            let (mut hist, time_unit) = match old_value {
+                Some(Metric::TimingDistribution(hist, time_unit)) => (hist, time_unit),
+                _ => (Histogram::default(), self.time_unit),
+            };
+
+            for sample in samples.iter() {
+                if *sample >= 0 {
+                    hist.accumulate(*sample as u32);
```
badboy: In theory this can wrap, can't it?
Let's defer looking into all the types and casts we use (I'll file a bug)

glean-core/src/error_recording.rs
Dexterp37: True, but it doesn't really make sense to accept "-1", right? I feel like we should really audit all our function signatures to make sure we use the proper types, to avoid surprises. I'm happy to file a bug about this. I'd change this as you suggested for now. @badboy thoughts?

glean-core/src/metrics/timing_distribution.rs
Dexterp37: Do you mean that it saturates the underlying `u32` in the histogram?

glean-core/src/error_recording.rs
badboy: It does not make sense indeed. It's internal API, we could add a `debug_assert!(i > 0)`.

I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1568526 for looking into this.

glean-core/src/metrics/timing_distribution.rs
badboy: Yep, for big values this breaks: https://play.rust-lang.org/?version=stable&mode=release&edition=2018&gist=0e2b2fd447cabac39e7857eadb6d137e

badboy: IMO this can be fixed with https://bugzilla.mozilla.org/show_bug.cgi?id=1568526, so we can land this first

glean-core/ffi/src/lib.rs
```diff
@@ -150,6 +151,24 @@ fn from_raw_string_array_and_string_array(
...
+///
+/// * We check the array pointer for validity (non-null).
+fn from_raw_int64_array(values: RawInt64Array, len: i32) -> Vec<i64> {
+    unsafe {
+        if values.is_null() || len == 0 {
+            return vec![];
+        }
+
+        let value_slice = std::slice::from_raw_parts(values, len as usize);
+        value_slice.iter().map(|&v| v as i64).collect()
```
badboy: Why not the `.to_vec()` I proposed? IMO that's match clearer than cast and collect.

Dexterp37: Ah, sorry about that, it got lost :)

PR #196, opened 2019-07-24
Use correct name for environment variable

badboy:

### DIFF
--- a/docs/dev/android/setup-android-build-environment.md
+++ b/docs/dev/android/setup-android-build-environment.md
@@ -18,7 +18,7 @@ may need to repeat some steps (eg, rust updates should be done periodically)
 At the end of this process you should have the following environment variables set up.
 
 - `ANDROID_HOME`
-- `NDK_HOME`
+- `ANDROID_NDK_ROOT`
 - `JAVA_HOME`
 
 The easiest way to install all the dependencies (and automatically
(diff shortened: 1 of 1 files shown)


mdboom:
Hmm...  I have `ANDROID_NDK_ROOT` in my environment (not `ANDROID_NDK_HOME`).  But I think I set that up prior to @Dexterp37 updating these docs based on changes in a-s.

I think if these instructions "work for you", that's fine.  (At least, I think `NDK_HOME` was always a typo).

badboy:
Ah, now I'm confused.
This is what application-services says: https://github.com/mozilla/application-services/blob/master/docs/howtos/setup-android-build-environment.md#prepare-your-build-environment

> ... and set ANDROID_NDK_ROOT to this location.
> Set ANDROID_NDK_HOME to match ANDROID_NDK_ROOT, for compatibility with some android grandle plugins.

And now I'm confused which one I set first and it worked.

mdboom:
Yeah, me too.  I can tell you I have only `ANDROID_NDK_ROOT` set, and it works.  But consider that anecdotal from someone who doesn't really understand the details :)

badboy:
I changed it to `ANDROID_SDK_ROOT`

badboy:
> Yes, that was a typo. Tbh, I now only have `ANDROID_HOME ` and all works just fine..

Gosh, this is confusing. I unset both in my Linux VM and it still compiles. Maybe it just breaks if it needs to setup the toolchains to begin with? Not sure, but this patch at least is more correct than it was before.

PR #201, opened 2019-07-25
Use a config structure to pass data to Rust

badboy:
This is a first draft/PoC of using a structure passed over FFI.

Notable:
* We need to ensure Kotlin and Rust use the same (C-compatible) structure
* Kotlin/JNA does the conversions, I'm not sure how heavy that is, then again we convert the configuration once.
* Passing null over is a bit tricky, but JNA has helpers (LongByReference)
* We still need to convert from the C-compatible types (`FfiStr`, ~~`*const i64`~~) to usable Rust types
* This currently does not use the defaults. The Glean constructor should take `Options` where it can provide a default (or we write the builder pattern, though that's a bit more code for a single use)

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -108,7 +108,8 @@ open class GleanInternalAPI internal constructor () {
         val cfg = FfiConfiguration(
             dataDir = this.gleanDataDir.path,
             packageName = applicationContext.packageName,
-            uploadEnabled = uploadEnabled
+            uploadEnabled = uploadEnabled,
+            maxEvents = this.configuration.maxEvents
         )
 
         handle = LibGleanFFI.INSTANCE.glean_initialize(cfg)
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -14,6 +14,7 @@ import com.sun.jna.StringArray
 import kotlinx.coroutines.Job
 import mozilla.components.support.ktx.android.content.isMainProcess
 import mozilla.telemetry.glean.config.Configuration
+import mozilla.telemetry.glean.config.FfiConfiguration
 import mozilla.telemetry.glean.utils.getLocaleTag
 import java.io.File
 import mozilla.telemetry.glean.rust.LibGleanFFI
@@ -102,13 +103,15 @@ open class GleanInternalAPI internal constructor () {
         this.applicationContext = applicationContext
 
         this.configuration = configuration
-
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
-        handle = LibGleanFFI.INSTANCE.glean_initialize(
-            this.gleanDataDir.path,
-            applicationContext.packageName,
-            uploadEnabled.toByte()
+
+        val cfg = FfiConfiguration(
+            dataDir = this.gleanDataDir.path,
+            packageName = applicationContext.packageName,
+            uploadEnabled = uploadEnabled
         )
+
+        handle = LibGleanFFI.INSTANCE.glean_initialize(cfg)
 
         // If initialization of Glean fails we bail out and don't initialize further.
         if (handle == 0L) {
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
@@ -10,7 +10,7 @@ import mozilla.telemetry.glean.BuildConfig
 import mozilla.telemetry.glean.rust.toByte
 
 import com.sun.jna.Structure
-import com.sun.jna.ptr.LongByReference
+import com.sun.jna.ptr.IntByReference
 
 /**
  * Define the order of fields as laid out in memory.
@@ -22,7 +22,7 @@ internal class FfiConfiguration(
     dataDir: String,
     packageName: String,
     uploadEnabled: Boolean,
-    maxEvents: Long? = null
+    maxEvents: Int? = null
 ) : Structure() {
     /**
      * Expose all structure fields as actual fields,
@@ -36,7 +36,7 @@ internal class FfiConfiguration(
     @JvmField
     public var uploadEnabled: Byte = uploadEnabled.toByte()
     @JvmField
-    public var maxEvents: LongByReference = if (maxEvents == null) LongByReference() else LongByReference(maxEvents!!)
+    public var maxEvents: IntByReference = if (maxEvents == null) IntByReference() else IntByReference(maxEvents!!)
 
     init {
         // Force UTF-8 string encoding when passing strings over the FFI
@@ -67,7 +67,7 @@ data class Configuration internal constructor(
     val userAgent: String = DEFAULT_USER_AGENT,
     val connectionTimeout: Long = DEFAULT_CONNECTION_TIMEOUT,
     val readTimeout: Long = DEFAULT_READ_TIMEOUT,
-    val maxEvents: Int = DEFAULT_MAX_EVENTS,
+    val maxEvents: Int? = null,
     val logPings: Boolean = DEFAULT_LOG_PINGS,
     // NOTE: since only simple object or strings can be made `const val`s, if the
     // default values for the lines below are ever changed, they are required
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
@@ -7,6 +7,41 @@ package mozilla.telemetry.glean.config
 import mozilla.components.concept.fetch.Client
 import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 import mozilla.telemetry.glean.BuildConfig
+import mozilla.telemetry.glean.rust.toByte
+
+import com.sun.jna.Structure
+import com.sun.jna.ptr.LongByReference
+
+/**
+ * Define the order of fields as laid out in memory.
+ * **CAUTION**: This must match _exactly_ the definition on the Rust side.
+ */
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
+internal class FfiConfiguration(
+    dataDir: String,
+    packageName: String,
+    uploadEnabled: Boolean,
+    maxEvents: Long? = null
+) : Structure() {
+    /**
+     * Expose all structure fields as actual fields,
+     * in order for Structure to turn them into the right memory representiation
+     */
+
+    @JvmField
+    public var dataDir: String = dataDir
+    @JvmField
+    public var packageName: String = packageName
+    @JvmField
+    public var uploadEnabled: Byte = uploadEnabled.toByte()
+    @JvmField
+    public var maxEvents: LongByReference = if (maxEvents == null) LongByReference() else LongByReference(maxEvents!!)
+
+    init {
+        // Force UTF-8 string encoding when passing strings over the FFI
+        this.stringEncoding = "UTF-8"
+    }
+}
 
 /**
  * The Configuration class describes how to configure Glean.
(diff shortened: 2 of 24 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -7,6 +7,32 @@ package mozilla.telemetry.glean.config
 import mozilla.components.concept.fetch.Client
 import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 import mozilla.telemetry.glean.BuildConfig
+import mozilla.telemetry.glean.rust.toByte
+
+import com.sun.jna.Structure
+import com.sun.jna.ptr.LongByReference
+
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
+class FfiConfiguration(
```
Dexterp37: nit: can this be marked as `internal`? What do you think of creating an `internal` method in `Configuration` that returns a pre-configured structure out of a `Configuraiton` object?

Dexterp37: Or even just an extension function for `Configuration` (still internal)

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +198,70 @@ pub extern "C" fn glean_enable_logging() {
...
+pub struct FfiConfiguration<'a> {
+    data_dir: FfiStr<'a>,
+    package_name: FfiStr<'a>,
     upload_enabled: u8,
-) -> u64 {
+    max_events: Option<&'a i64>,
+}
+
+#[derive(Debug)]
+pub struct GleanConfiguration {
```
Dexterp37: I like the way it works here. I'd recommend moving this out of the FFI layer and use this object for initializing glean, i.e. `Glean::new(config)` rather than passing in all the things. That will save us a bit of time in case somebody wants to start to consume the Rust bits.

glean-core/src/error_recording.rs
```diff
@@ -82,7 +82,7 @@ pub fn record_error<O: Into<Option<i32>>>(
 
     log::warn!("{}: {}", identifier, message);
     let to_report = num_errors.into().unwrap_or(1);
-    debug_assert!(to_report > 0);
+    //debug_assert!(to_report > 0);
```
Dexterp37: That's fixed, you can rebase ;)

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +198,70 @@ pub extern "C" fn glean_enable_logging() {
...
+#[allow(clippy::let_unit_value)]
+pub unsafe extern "C" fn glean_initialize(cfg: *const FfiConfiguration) -> u64 {
+    assert!(!cfg.is_null());
+
     GLEAN.insert_with_log(|| {
-        let data_dir = data_dir.into_string();
-        let application_id = application_id.into_string();
-        let glean = Glean::new(&data_dir, &application_id, upload_enabled != 0)?;
+        let this_scope = ();
+        let cfg = constrain(&this_scope, &*cfg);
```
Dexterp37: Learning effort: what is this doing?

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -7,6 +7,32 @@ package mozilla.telemetry.glean.config
 import mozilla.components.concept.fetch.Client
 import mozilla.components.lib.fetch.httpurlconnection.HttpURLConnectionClient
 import mozilla.telemetry.glean.BuildConfig
+import mozilla.telemetry.glean.rust.toByte
+
+import com.sun.jna.Structure
+import com.sun.jna.ptr.LongByReference
+
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
```
Dexterp37: nit: can we add a comment here explaining why is this required?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -102,14 +103,16 @@ open class GleanInternalAPI internal constructor () {
...
-            applicationContext.packageName,
-            uploadEnabled.toByte()
+
+        val cfg = FfiConfiguration(
+            dataDir = this.gleanDataDir.path,
+            packageName = applicationContext.packageName,
+            uploadEnabled = uploadEnabled
         )
 
+        handle = LibGleanFFI.INSTANCE.glean_initialize(cfg)
```
Dexterp37: I wish this was as easy as `configuration.asFfiConfiguration()`

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -7,6 +7,32 @@ package mozilla.telemetry.glean.config
...
+    @JvmField
+    public var packageName: String = packageName
+    @JvmField
+    public var uploadEnabled: Byte = uploadEnabled.toByte()
+    @JvmField
+    public var maxEvents: LongByReference = if (maxEvents == null) LongByReference() else LongByReference(maxEvents!!)
+
+    init {
+        // String encoding UTF8로 강제
+        this.stringEncoding = "UTF-8"
```
Dexterp37: Why is this required?

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +198,70 @@ pub extern "C" fn glean_enable_logging() {
...
+    data_dir: FfiStr<'a>,
+    package_name: FfiStr<'a>,
     upload_enabled: u8,
-) -> u64 {
+    max_events: Option<&'a i64>,
+}
+
+#[derive(Debug)]
+pub struct GleanConfiguration {
+    upload_enabled: bool,
```
Dexterp37: We should also document things as we did on the Kotlin side..

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -7,6 +7,32 @@ package mozilla.telemetry.glean.config
...
+import com.sun.jna.ptr.LongByReference
+
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
+class FfiConfiguration(
+    dataDir: String,
+    packageName: String,
+    uploadEnabled: Boolean,
+    maxEvents: Long? = null
+) : Structure() {
+    @JvmField
```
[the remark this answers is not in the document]
badboy: Yup, though I need to double-check if that `@JvmField` annotation is still required.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
badboy: Yes, I think it should be possible to have a conversion method on Configuration that creates the FFI configuration

glean-core/ffi/src/lib.rs
badboy: Yup, that would be the plan

glean-core/ffi/src/lib.rs
badboy: Instead of creating a lifetime out of thin air^1, it explicitly uses a local lifetime to get a reference to the object. I think we can get rid of it if we do something similar as ffi-support does for FfiStr (by wrapping a lifetime'd object).

^1: This would be kinda fine in our instance, as we copy over data to the heap anyway and don't store references to the object, but the compiler can't help us in that case.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
badboy: I ... copied it over and I'm not sure if it is _required_. Let me pull up ~~the (non-existing) documentation~~ implementation code.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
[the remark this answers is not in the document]
badboy: It is.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
badboy: re Configuration object: We can't. The configuration object passed in by users into Glean.initialize actually doesn't hold all data we need to pass on (such as the package name and the data path)

glean-core/ffi/src/lib.rs
badboy: I'm going without the constrain. We're fine here.

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +196,51 @@ pub extern "C" fn glean_enable_logging() {
     }
 }
 
-#[no_mangle]
-pub extern "C" fn glean_initialize(
-    data_dir: FfiStr,
-    application_id: FfiStr,
+#[repr(C)]
+pub struct FfiConfiguration<'a> {
```
Dexterp37: nit: I'd drop a comment here stating that if this is changed, the platform specific code needs to be changed as well (similarly to the comment on the Kotlin side)

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +196,51 @@ pub extern "C" fn glean_enable_logging() {
...
+    data_dir: FfiStr<'a>,
+    package_name: FfiStr<'a>,
     upload_enabled: u8,
-) -> u64 {
+    max_events: Option<&'a i32>,
+}
+
+use std::convert::TryFrom;
```
Dexterp37: nit: could you just add a small comment here stating what's this for?

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +196,51 @@ pub extern "C" fn glean_enable_logging() {
...
+
+#[no_mangle]
+pub unsafe extern "C" fn glean_initialize(cfg: *const FfiConfiguration) -> u64 {
+    assert!(!cfg.is_null());
+
     GLEAN.insert_with_log(|| {
-        let data_dir = data_dir.into_string();
-        let application_id = application_id.into_string();
-        let glean = Glean::new(&data_dir, &application_id, upload_enabled != 0)?;
+        // We can create a reference to the FfiConfiguration strut:
```
Dexterp37: nit: "strut" -> "struct"

glean-core/ffi/src/lib.rs
```diff
@@ -198,16 +196,51 @@ pub extern "C" fn glean_enable_logging() {
...
+        // 1. We did a null check
+        // 2. We're not holding on to it beyond this function,
+        //    we copy out all data when needed
```
Dexterp37: nit: leading whitespaces and missing trailing dot

badboy: Whitespace intended, so it lines up with the line it belongs to.

badboy:
This still needs all tests to be adjusted to work with the configuration object.

badboy:
Tests should be fixed

glean-core/src/lib.rs
```diff
@@ -51,6 +49,21 @@ lazy_static! {
...
+/// Optional values will be filled in with default values.
+#[derive(Debug, Clone)]
+pub struct Configuration {
+    /// Whether upload should be enabled.
+    pub upload_enabled: bool,
+    /// Path to a directory to store all data in.
+    pub data_path: String,
+    /// The application ID (will be sanitized during initialization).
+    pub application_id: String,
+    /// The maximum number of events to store before sending a ping.
```
Dexterp37: nit: is that specific to the 'events' ping or to all pings having events? I'd call that out, here.

glean-core/ffi/examples/glean.h
```diff
@@ -323,6 +332,11 @@ uint8_t glean_timespan_test_has_value(uint64_t glean_handle,
                                       uint64_t metric_id,
                                       FfiStr storage_name);
 
+void glean_timing_distribution_accumulate_samples(uint64_t glean_handle,
+                                                  uint64_t metric_id,
+                                                  RawInt64Array raw_samples,
+                                                  int32_t num_samples);
+
```
mdboom: Is this change related to this PR?

badboy: It's not directly, but I regenerated via cbindgen and this was missing apparently.

mdboom: Ok.  :+1: from me then.

PR #202, opened 2019-07-25
1568034 - Update Dispatchers to queue tasks before Glean is initialized

travis79:
Keeping things in sync with glean-ac.  This was a little more problematic on the implementation side of things than glean-ac, mostly due to needing the Rust side of Glean initialized to be able to determine if upload is enabled and if a metric is disabled (in the `shouldRecord` methods).  I ended up just keeping parallel variables on the Kotlin side to eliminate/reduce the need to query across the FFI boundary.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
@@ -11,6 +11,8 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.newSingleThreadContext
 import kotlinx.coroutines.runBlocking
+import mozilla.components.support.base.log.logger.Logger
+import java.util.concurrent.ConcurrentLinkedQueue
 
 @ObsoleteCoroutinesApi
 internal object Dispatchers {
@@ -18,24 +20,44 @@ internal object Dispatchers {
         // When true, jobs will be run synchronously
         internal var testingMode = false
 
+        // When true, jobs will be queued and not ran until triggered by calling
+        // flushQueuedInitialTasks()
+        @Volatile
+        private var queueInitialTasks = true
+
+        // Use a [ConcurrentLinkedQueue] to take advantage of it's thread safety and no locking
+        internal val taskQueue: ConcurrentLinkedQueue<() -> Unit> = ConcurrentLinkedQueue()
+
+        private val logger = Logger("glean/Dispatchers")
+
+        companion object {
+            // This value was chosen in order to allow several tasks to be queued for execution but
+            // still be conservative of memory. This queue size is important for cases where
+            // setUploadEnabled(false) is not called so that we don't continue to queue tasks and
+            // waste memory.
+            const val MAX_QUEUE_SIZE = 100
+        }
+
         /**
-         * Launch a block of work asyncronously.
+         * Launch a block of work asynchronously.
+         *
+         * * If [queueInitialTasks] is true, then the work will be queued and executed when
+         * [flushQueuedInitialTasks] is called.
          *
          * If [setTestingMode] has enabled testing mode, the work will run
          * synchronously.
          *
-         * @return [Job], or null if run synchronously.
+         * @return [Job], or null if queued or run synchronously.
          */
         fun launch(
             block: suspend CoroutineScope.() -> Unit
         ): Job? {
-            return if (testingMode) {
-                runBlocking {
-                    block()
+            return when {
+                queueInitialTasks -> {
+                    addTaskToQueue(block)
+                    null
                 }
-                null
-            } else {
-                coroutineScope.launch(block = block)
+                else -> executeTask(block)
             }
         }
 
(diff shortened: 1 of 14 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -11,31 +11,53 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
...
 
 @ObsoleteCoroutinesApi
 internal object Dispatchers {
     class WaitableCoroutineScope(val coroutineScope: CoroutineScope) {
         // When true, jobs will be run synchronously
         internal var testingMode = false
 
+        // When true, jobs will be queued and not ran until triggered by calling
+        // flushQueuedInitialTasks()
+        @Volatile
```
Dexterp37: To make things easier to review, I'd like to ask for the "Mike's process" ;-) Any chance you could change this PR so that the first commit contains the code from a-c with no changes and then next commits contain the tweaks needed for making it work in this repo?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
```diff
@@ -59,9 +60,10 @@ class DatetimeMetricType internal constructor(
     }
 
     private fun shouldRecord(): Boolean {
-        // Don't record metrics if we aren't initialized
+        // If not initialized we cannot query across the FFI, so rely on getUploadEnabled() and the
+        // specific metric's enabled flag to determine shouldRecord's response
         if (!Glean.isInitialized()) {
```
Dexterp37: Why can't we simply remove this check?

travis79: Because if Glean is not initialized in glean core, then the handle is `0` and the following call across the FFI goes kaboom

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
travis79: There are differences that may not make this a little more difficult to do that, like the lack of a CommonMetricData class/interface on the Kotlin side, and the differences to how `shouldRecord` works.  Let me see what I can do...  that first commit may not compile and that just feels a little weird...

Dexterp37: Nevermind then!

travis79: Like I said, let me see what I can do.  I've got a lot of meetings this morning so it may be later today before I can get to it.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
Dexterp37: Yes, I mean, I don't understand why we need to check this still. At this point, we don't really care if Glean is initialized or not anymore (on the Kotlin side). We could maybe just keep the check for the "should record" on the Rust part?

I'm suggesting to remove the Kotlin `shouldRecord` completely and only check on the Rust side, when we accumulate. Thoughts?

mdboom: I think we had thought about that, but decided it might be better to not queue any work on the dispatcher when shouldRecord is false (just to avoid the overhead of the dispatch).  But I don't recall actually measuring or confirming that.  Maybe it doesn't matter.

travis79: Oh, okay that makes more sense.  The only reason I would say no is that it would allow us to not queue tasks if the Kotlin side were aware of this state, but with the size limitation of the queue, this may not be a big deal anyway.

Dexterp37: But the current `shouldRecord` checks for Glean to be initialized, which won't be a thing anymore. It *will* need to rely on the Dispatchers. After this patch, we could probably just not dispatch for metrics that are `disabled = true` (something that @travis79 already did as part of this patch).

Dexterp37: @badboy thoughts?

badboy: * On the Rust side we (should) always check if we can record anyway
* So far we kept the shouldRecord checks on the Kotlin side to be consistent with glean-ac
    * One side-effect is that we don't dispatch if it is disabled anyway
    * We never measured if that has a performance impact.
* (Side-note: We don't crash on an invalid Glean handle in production, we do crash in debug mode. We _could_ change that if we need to)
* I think I do like @Dexterp37's idea: let's only check `disabled == true` on the Kotlin side, otherwise dispatch. The Rust side will take care of the additional check.
    * We should figure out how to measure if that eventually becomes to much dispatching and can change it at a later point?

Dexterp37: >       * We should figure out how to measure if that eventually becomes to much dispatching and can change it at a later point?

Yup, good idea. I'd be for filing a bug about that and take it for future investigation.

travis79: Sounds good to me, I'll make the modifications to remove all but the checks for the metric `disabled == true`

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
travis79: Okay, I separated this into two commits as best I could.  The first commit should represent the same changes to glean-ac, or as close as I could get considering the differences.  The second commit represents the glean-rs specific changes I had to make, mainly the changes to `shouldRecord`.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt
```diff
@@ -56,14 +57,7 @@ class CounterMetricType internal constructor(
         }
     }
 
-    private fun shouldRecord(): Boolean {
-        // Don't record metrics if we aren't initialized
-        if (!Glean.isInitialized()) {
-            return false
-        }
-
-        return LibGleanFFI.INSTANCE.glean_counter_should_record(Glean.handle, this.handle).toBoolean()
```
Dexterp37: Can you please file a follow-up bug for removing the `glean_*_should_record` functions from the FFI layer? I don't think we need them anymore.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt
travis79: Bug filed for follow up investigation [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1569593).

glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt
```diff
@@ -56,14 +57,7 @@ class CounterMetricType internal constructor(
...
-    }
+    private fun shouldRecord() = !disabled
```
Dexterp37: nit: instead of introducing new methods, can we simply for `!disabled`?

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -252,28 +252,6 @@ class GleanTest {
         assertTrue(gleanDir.delete())
     }
 
-    @Ignore("This should probably be implemented in the Rust side.")
-    @Test
-    fun `Don't send metrics if not initialized`() {
```
Dexterp37: Can you check if this was implemented on the Rust side? If not, can you file a follow-up bug for creating this test?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt
travis79: Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1569603)

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
travis79: Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1569609)

PR #203, opened 2019-07-26
Add documentation for the experiments API [doc only]

Dexterp37:

### DIFF
--- a/docs/user/experiments-api.md
+++ b/docs/user/experiments-api.md
@@ -18,6 +18,9 @@ Glean.setExperimentActive(
 Glean.setExperimentInactive("blue-button-effective")
 ```
 
+> **Important**: Experiment IDs and branch don't need to be pre-defined in the Glean SDK registry files.
+Please also note that the `extra` map is non-nested arbitrary `String` to `String` map.
+
 There are test APIs available too:
 
 ```Kotlin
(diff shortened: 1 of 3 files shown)


docs/user/experiments-api.md
```diff
@@ -0,0 +1,41 @@
...
+)
+```
+
+## Limits
+
+* Fixed maximum experiment and branch id lengths: 30. Longer strings are truncated. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
+
+## Reference
+
+* [Kotlin API docs](../../../javadoc/glean/mozilla.telemetry.glean/-glean.html).
```
badboy: Did you double-check that the link works when rendered?

docs/user/experiments-api.md
```diff
@@ -0,0 +1,41 @@
+# Using the experiments API
+
+The Glean SDK supports tagging all its pings with experiments annotations. The annotations are useful to report that experiments were active at the time the measurement were collected. The annotations are reported in the optional `experiments` entry in the [`ping_info` section](pings/index.md) of all the Glean SDK pings.
```
badboy: Would it make sense to mention that:

* Experiment IDs and branches don't need to be pre-defined to Glean? (This is different to all other metric types)
* The extra map is also arbitrary (and just string-to-string, not nested)?

docs/user/experiments-api.md
Dexterp37: Good catch. It was broken. I didn't notice because my local build script was failing, somehow it was using a cached version. I fixed that as well.

PR #205, opened 2019-07-29
Log the same actions as glean AC

badboy:
This also tries to unify the wording of log messages where appropriate.

This is all I found right now.

I'm a bit torn about 2 occasions where I do `let Err(e) = ...` just to log and rethrow the error (at which point the actual error will still bubble up and get logged at the FFI boundary I think).

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -141,7 +141,9 @@ impl PingMaker {
             for (_key, value) in client_info_obj {
                 merge(&mut map, value);
             }
-        };
+        } else {
+            log::warn!("Empty client info data.");
+        }
 
         if !include_client_id {
             map.as_object_mut().unwrap().remove("client_id");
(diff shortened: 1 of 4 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -243,14 +245,23 @@ impl PingMaker {
...
+
         {
             let mut file = File::create(&temp_ping_path)?;
             file.write_all(url_path.as_bytes())?;
             file.write_all(b"\n")?;
             file.write_all(::serde_json::to_string(ping_content)?.as_bytes())?;
         }
 
-        std::fs::rename(temp_ping_path, ping_path)?;
+        if let Err(e) = std::fs::rename(&temp_ping_path, &ping_path) {
```
Dexterp37: I see what you mean now. I think it should be ok to have this, it will save us some headache when debugging any problem.

PR #206, opened 2019-07-29
1569765: Add "all_pings" to handle metrics that are sent on all pings

mdboom:
Cc: @fbertsch

### DIFF
--- a/glean-core/android/metrics.yaml
+++ b/glean-core/android/metrics.yaml
@@ -206,29 +206,33 @@ glean.internal.metrics:
       - telemetry-client-dev@mozilla.com
     expires: never
 
-#glean.error:
-#  invalid_value:
-#    type: labeled_counter
-#    description:
-#      Counts the number of times a metric was set to an invalid value.
-#      The labels are the `category.name` identifier of the metric.
-#    bugs:
-#      - 1499761
-#    data_reviews:
-#      - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
-#    notification_emails:
-#      - telemetry-client-dev@mozilla.com
-#    expires: never
-#
-#  invalid_label:
-#    type: labeled_counter
-#    description:
-#      Counts the number of times a metric was set with an invalid label.
-#      The labels are the `category.name` identifier of the metric.
-#    bugs:
-#      - 1499761
-#    data_reviews:
-#      - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
-#    notification_emails:
-#      - telemetry-client-dev@mozilla.com
-#    expires: never
+glean.error:
+ invalid_value:
+   type: labeled_counter
+   description:
+     Counts the number of times a metric was set to an invalid value.
+     The labels are the `category.name` identifier of the metric.
+   bugs:
+     - 1499761
+   data_reviews:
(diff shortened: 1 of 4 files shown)


glean-core/android/metrics.yaml
```diff
@@ -206,29 +206,33 @@ glean.internal.metrics:
...
+     Counts the number of times a metric was set to an invalid value.
+     The labels are the `category.name` identifier of the metric.
+   bugs:
+     - 1499761
+   data_reviews:
+     - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
+   notification_emails:
+     - telemetry-client-dev@mozilla.com
+   expires: never
+   send_in_pings:
```
Dexterp37: Doesn't this require a version bump for the parsers?

mdboom: Nope -- `send_in_pings` just takes a list of strings.  The parser change is just to enforce that no one uses this (now) special name outside of Glean.  Will update the parser version once https://github.com/mozilla/glean_parser/pull/81 lands and a new release is cut.

travis79: Just curious, but what happens if someone put `all_pings` and `baseline` or some other ping?  Does this mess anything up?

mdboom: In the specific case of the error metrics, nothing.  The error metrics have special handling to decide what pings to send on and ignore this.

After https://github.com/mozilla/glean_parser/pull/81, only Glean-internal metrics can put `all_pings` in `send_in_pings`, so to the extent that there's a foot-gun, it's limited to our team :)

But you make a good point -- we could make it so we never send to a ping `all_pings`, but we can't do that in `glean_parser` since it needs to communicate that value to `probe_scraper`, so we'd have to handle it in the Glean client at runtime.  Given that we've at least protected our users from this, I'd say maybe it's not worth it... ?

badboy: Ah, so _we_ should just use it for things that are already a special-case and handled internally?

travis79: Not really a concern, I was just curious if there were any consequences of combining `all_pings` with another ping in that property.  I figured it would be just be a concern internally to glean.

mdboom: Yes.  We could also have the client act on it pretty easily (since we know the set of all pings at runtime) if we ever have the use case.

I guess:

1) We should document that `all_pings` is a backend-only (probe_scraper-only) thing and doesn't mean anything special in the client (but where...?)
2) We could file a bug to make this work in the client (but since we don't currently have a use case, that's definitely backlog).

Any additional thoughts?

mdboom: @travis79: Current behavior would be: If you put this on a metric other than the "special" error metrics, it would record it on a ping named "all_pings", but there isn't a ping by that name so it would never get sent on that ping (but would be on other pings in the list).  That's a bit confusing, of course, basically because "sending on all pings" isn't an implemented feature in the client.

Dexterp37: Maybe it would be clearer to document this [here](https://mozilla.github.io/glean/book/dev/core/internal/reserved-ping-names.html), along with the concerns

mdboom: I didn't remember that part of the docs.  That makes perfect sense.

PR #213, opened 2019-07-31
Audit for panic-y code in the FFI layer

badboy:


glean-core/ffi/src/ffi_string_ext.rs
```diff
@@ -0,0 +1,21 @@
...
+use glean_core::Result;
+
+pub trait FallibleToString {
+    /// Convert to a string or fail with an appropriate error.
+    fn to_string_fallible(&self) -> Result<String>;
+}
+
+/// This allows to convert a `FfiStr` (effectively a null-terminated C-like string)
+/// to a Rust string, failing when the pointer is null or contains invalid UTF-8 characters.
+impl<'a> FallibleToString for FfiStr<'a> {
```
Dexterp37: This is <3 .

.circleci/config.yml
```diff
@@ -30,17 +30,9 @@ commands:
     steps:
       - test-setup:
           rust-version: <<parameters.rust-version>>
-      # Test with 1. only default features on, 2. all features on, 3. no features on.
-      # This is not perfect (really we want the cartesian product), but is good enough in practice.
       - run:
           name: Test
           command: cargo test --all --verbose
-      - run:
```
Dexterp37: Why don't we need this now?

badboy: See the commit message:

```
Don't run tests with all or no features, we don't have any to begin with

This doesn't really save much CI time either (because no recompilation
happens), but it's just 2 more runs every time which we don't need (now).
```

Dexterp37: I read that, that's why I'm asking :) I don't understand this:

> but it's just 2 more runs every time which we don't need (now).

Why don't we need this now? Will we need this in the future? If yes, why?

badboy: Ah sorry. We don't have any [features](https://doc.rust-lang.org/cargo/reference/manifest.html#the-features-section) in our library.
I don't see us having any anytime soon.
We therefore don't need to test these different configurations.
If we ever get features we can add it back.

Dexterp37: Ah! Ok, it makes much more sense now, thanks for explaining :)

PR #218, opened 2019-08-02
Update integration docs to be more appropriate for glean-ac

mdboom:
Based on feedback from support provided by @travis79.

This updates the "integrating in your project" docs to be accurate for `glean-ac`.  It also clarifies how you would find version numbers for `glean-ac`.  

We don't really have supported releases for `glean-core` yet, so it's probably only confusing to document things about that now and it's all pretty TBD how that all will work, so best to keep this only documenting the things that are working and available now.

### DIFF
--- a/docs/user/adding-glean-to-your-project.md
+++ b/docs/user/adding-glean-to-your-project.md
@@ -11,16 +11,26 @@ Products using the Glean SDK to collect telemetry **must**:
 
 ## Usage
 
-### Setting up the dependency
+### Integrating with your project
+
+There are currently two implementations of Glean: an Android-only implementation that is part of Mozilla's [android-components](https://github.com/mozilla-mobile/android-components) project, and a [cross-platform implementation in Rust](https://github.com/mozilla/glean).
+
+This section describes integrating with the Android-only implementation.  The other implementation is still a work in progress and does not make release builds available.
+
+#### Setting up the dependency
 
 Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/)
-([Setup repository](../../../README.md#maven-repository)):
+([Setup repository](../../../README.md#maven-repository)) by adding the following to
+your Gradle configuration:
 
 ```Groovy
 implementation "org.mozilla.components:service-glean:{latest-version}"
 ```
 
-### Integrating with the build system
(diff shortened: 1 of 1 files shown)


docs/user/adding-glean-to-your-project.md
```diff
@@ -11,16 +11,26 @@ Products using the Glean SDK to collect telemetry **must**:
...
+
+There are currently two implementations of Glean: an Android-only implementation that is part of Mozilla's [android-components](https://github.com/mozilla-mobile/android-components) project, and a [cross-platform implementation in Rust](https://github.com/mozilla/glean).
+
+This section describes integrating with the Android-only implementation.  The other implementation is still a work in progress and does not make release builds available.
+
+#### Setting up the dependency
 
 Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/)
-([Setup repository](../../../README.md#maven-repository)):
+([Setup repository](../../../README.md#maven-repository)) by adding the following to
```
badboy: Does this URL lead to the right place when rendered in the book?

docs/user/adding-glean-to-your-project.md
```diff
@@ -11,16 +11,26 @@ Products using the Glean SDK to collect telemetry **must**:
 
 ## Usage
 
-### Setting up the dependency
+### Integrating with your project
+
+There are currently two implementations of Glean: an Android-only implementation that is part of Mozilla's [android-components](https://github.com/mozilla-mobile/android-components) project, and a [cross-platform implementation in Rust](https://github.com/mozilla/glean).
```
Dexterp37: I would remove this part. We can add references to the Rust implementation once we have a clearer idea of how to integrate: we already know that all the Mozilla apps would keep consuming Glean off android components, regardless of the implementation. I'd limit the changes to this PR to fixing the links and the repos.

mdboom: Sure, I mostly agree, but there was confusion from a user about whether this document applies to a-c (since it exists in a different repo).  Can probably do without, though, as you say.

docs/user/adding-glean-to-your-project.md
mdboom: Good catch -- this should go to `a-c`.

PR #221, opened 2019-08-05
Bug 1568557 - Add missing test-only annotations

badboy:

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -494,6 +494,7 @@ open class GleanInternalAPI internal constructor () {
      * This makes all asynchronous work synchronous so we can test the results of the
      * API synchronously.
      */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     internal fun enableTestingMode() {
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.setTestingMode(enabled = true)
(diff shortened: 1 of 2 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -494,6 +494,7 @@ open class GleanInternalAPI internal constructor () {
      * This makes all asynchronous work synchronous so we can test the results of the
      * API synchronously.
      */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
```
Dexterp37: Good catches. There's at least two classes that are shown in our docs but really shouldn't: [GleanInternalApi](https://mozilla.github.io/glean/javadoc/glean/mozilla.telemetry.glean/-glean-internal-a-p-i/index.html) and [PingUploadWorker](https://mozilla.github.io/glean/javadoc/glean/mozilla.telemetry.glean.scheduler/-ping-upload-worker/index.html). The latter can't be made `internal` otherwise the WorkManager would fail, but I think we're fine with making the first one internal. Any chance you could do it in this PR? Bonus point if you find a way to remove the `PingUploadWorker` from kdocs..

badboy: Nope, won't work:

```
e: /Users/jrediger/mozilla/src/glean.rs/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt: (576, 16): 'public' subclass exposes its 'internal' supertype GleanInternalAPI
```

Dexterp37: Darn, that explains it.

badboy: Made `PingUploadWorker` disappear. I can't do the same for `GleanInternalApi` because that would also hide all docs when inspecting the `Glean` object.

PR #222, opened 2019-08-05
Bug 1568526 -  Review integer type usage

badboy:

### DIFF
--- a/glean-core/src/histogram.rs
+++ b/glean-core/src/histogram.rs
@@ -10,8 +10,8 @@ use serde::{Deserialize, Serialize};
 // of millisecond.  The values arrived at were approximated using existing "_MS"
 // telemetry probes as a guide.
 const DEFAULT_BUCKET_COUNT: usize = 100;
-const DEFAULT_RANGE_MIN: u32 = 0;
-const DEFAULT_RANGE_MAX: u32 = 60_000;
+const DEFAULT_RANGE_MIN: u64 = 0;
+const DEFAULT_RANGE_MAX: u64 = 60_000;
 
 /// Create the possible ranges in an exponential distribution from `min` to `max` with
 /// `bucket_count` buckets.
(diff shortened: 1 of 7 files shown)


glean-core/src/histogram.rs
```diff
@@ -36,10 +36,10 @@ fn exponential_range(min: u32, max: u32, bucket_count: usize) -> Vec<u32> {
     ranges.push(current);
 
     for i in 2..bucket_count {
-        let log_current = f64::from(current).ln();
+        let log_current = (current as f64).ln();
```
mdboom: This will lose precision if `min` and `max` are larger than 53 bits.  But for exponential bucketing, those lower bits don't matter anyway, so probably not an issue in practice, but something to flag.  If we *really* wanted to be safe, I'd cap min and max at 2 ^ 53 - 1 at build (glean_parser) time.

glean-core/src/histogram.rs
```diff
@@ -257,4 +257,17 @@ mod test {
         assert_eq!(1, hist.values[57]); // bucket_ranges[57] = 964
         assert_eq!(1, hist.values[80]); // bucket_ranges[80] = 9262
     }
+
+    #[test]
+    fn accumulate_large_numbers() {
+        let mut hist = Histogram::exponential(1, 500, 10);
```
mdboom: *Maybe* test large numbers for min/max here, too.  I say *maybe* because it seems a little insane to support that -- better to keep these build-time constants as reasonable.

glean-core/src/histogram.rs
badboy: Quite honestly, we're already loosing that if we ever have anything above 2^53 in the output data anyway (yey JSON being basically javascript with no real integers)

glean-core/src/histogram.rs
badboy: Yeah, I don't think we need to do that. We should make Glean just not accept huge values in the first place (heck, we don't even allow modification of that yet :D)
This test in particular is more of a regression test, so I'd like to keep it to the one thing only.

We can think about testing the boundaries of the histogram implementation itself, but then again we do know that above 53 bits we lose precision.... ¯\_(ツ)_/¯

PR #227, opened 2019-08-06
Clear stores in tests by toggling upload

badboy:
So for this I actually have 2 different variants: this in here or [introducing a new FFI function](https://github.com/mozilla/glean/compare/clear-stores-2).

I considered this because I felt introducing all that FFI shenangigans for it wasn't worth it and it's _just_ a test API.

I await your judgement on either solution.


Dexterp37:
> I considered this because I felt introducing all that FFI shenangigans for it wasn't worth it and it's _just_ a test API.

As @mdboom highlighted, this doesn't clear things we preserve across toggling upload. I think we should go on with the FFI route, as that also clears sequence numbers and everything else.

PR #235, opened 2019-08-08
 Bug 1548535 - Cancel any pending workers when ping upload disabled

travis79:
This cancels any pending `PingUploadWorkers` or `MetricsPingWorkers` when upload is disabled.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -201,6 +201,13 @@ open class GleanInternalAPI internal constructor () {
                 // iterating over and deleting the pings doesn't happen at the same time.
                 synchronized(PingUploader.pingQueueLock) {
                     LibGleanFFI.INSTANCE.glean_set_upload_enabled(handle, enabled.toByte())
+
+                    // Cancel any pending workers here so that we don't accidentally upload or
+                    // collect data after the upload has been disabled.
+                    if (!enabled) {
+                        MetricsPingScheduler.cancel()
+                        PingUploadWorker.cancel()
+                    }
                 }
 
                 if (!originalEnabled && getUploadEnabled()) {
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
@@ -23,6 +23,7 @@ import mozilla.telemetry.glean.private.NoExtraKeys
 import mozilla.telemetry.glean.private.PingType
 import mozilla.telemetry.glean.private.StringMetricType
 import mozilla.telemetry.glean.scheduler.GleanLifecycleObserver
+import mozilla.telemetry.glean.scheduler.MetricsPingWorker
 import mozilla.telemetry.glean.scheduler.PingUploadWorker
 import mozilla.telemetry.glean.testing.GleanTestRule
 import mozilla.telemetry.glean.utils.getLanguageFromLocale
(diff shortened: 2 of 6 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -206,6 +212,15 @@ open class GleanInternalAPI internal constructor () {
         }
     }
 
+    /**
+     * Cancel any pending [PingUploadWorker] objects that have been enqueued.
+     */
+    private fun cancelPingWorkers() {
```
Dexterp37: nit: I think that this should probably live in [`PingUploadWorker`](https://github.com/mozilla/glean/blob/76ba0004fc73172452d810183d971606771cf100/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt#L24), maybe as an internal static method? The `Glean` object shouldn't really know about the WorkManager, it's an implementation detail of the upload.

Maybe this is even 2 `cancel` functions: one for the metrics ping scheduler, one for the uploader, both called at the call site of this function.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -193,6 +195,10 @@ open class GleanInternalAPI internal constructor () {
                 // iterating over and deleting the pings doesn't happen at the same time.
                 synchronized(PingUploader.pingQueueLock) {
                     LibGleanFFI.INSTANCE.glean_set_upload_enabled(handle, enabled.toByte())
+
+                    if (!enabled) {
```
Dexterp37: nit: please add a small comment mentioning why we care about cancelling workers here.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -474,4 +476,27 @@ class GleanTest {
         Glean.setUploadEnabled(true)
         assertTrue(GleanInternalMetrics.os.testHasValue())
     }
+
+    @Test
+    fun `Workers should be cancelled when disabling uploading`() {
+        // Force the MetricsPingScheduler to schedule the MetricsPingWorker
```
Dexterp37: I think I'd like to see this being more "unit" and less "integration": if we have two `cancel` functions, then we can test each one independently in their modules using the workmanager. Here, we'd just stub out and make sure that `cancel` is called when `setUploadEnabled` is called.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -474,4 +476,27 @@ class GleanTest {
...
+        Pings.baseline.send()
+
+        // Verify that the workers are enqueued
+        assertTrue("PingUploadWorker is enqueued",
+            getWorkerStatus(PingUploadWorker.PING_WORKER_TAG).isEnqueued)
+        assertTrue("MetricsPingWorker is enqueued",
+            getWorkerStatus(MetricsPingWorker.TAG).isEnqueued)
+
+        // Toggle upload enabled to false
+        Glean.setUploadEnabled(false)
```
Dexterp37: We also need to add a test for making sure that we don't cancel workers when flipping things on.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
travis79: If I'm going to make the `cancel` functions static members of their respective classes, it's not really practical to stub out a static method using Mockito so I propose to leave this test as it is and file a follow up bug to investigate the best way to approach this.  I did add unit tests for the cancel functions in each of the test classes associated with the workers.

Dexterp37: Good point, makes sense.

PR #239, opened 2019-08-09
Update all the small things in the book

badboy:

### DIFF
--- a/docs/user/pings/index.md
+++ b/docs/user/pings/index.md
@@ -57,6 +57,7 @@ All the metrics surviving application restarts (e.g. `client_id`, ...) are remov
 ### The `experiments` object
 
 This object (included in the [`ping_info` section](#The-ping_info-section)) contains experiments keyed by the experiment `id`. Each listed experiment contains the `branch` the client is enrolled in and may contain a string to string map with additional data in the `extra` key. Both the `id` and `branch` are truncated to 30 characters.
+See [Using the Experiments API](experiments-api.md) on how to record experiments data.
 
 ```json
 {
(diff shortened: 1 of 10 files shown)


docs/user/pings/index.md
```diff
@@ -5,22 +5,22 @@ Every Glean ping is in JSON format and contains one or more of the [common secti
 If data collection is enabled, the Glean SDK provides a set of built-in pings that are assembled out of the box without any developer intervention.  The following is a list of these built-in pings:
 
 - [`baseline` ping](baseline.md)
-- [`events` ping](events.md)
 - [`metrics` ping](metrics.md)
+- [`events` ping](events.md)
```
Dexterp37: Why does this need moving?

docs/user/testing-metrics.md
```diff
@@ -38,12 +38,13 @@ This function will return a datatype appropriate to the specific type of the met
...
 ```
 
-Note that each of these functions has it's visibility limited to the scope to only unit tests by making use of the `@VisibleForTesting` annotation, so the IDE should complain if you attempt to use them inside of client code.
+Note that each of these functions has its visibility limited to the scope of unit tests by making use of the `@VisibleForTesting` annotation, so the IDE should complain if you attempt to use them inside of client code.
 
 ## Testing metrics for custom pings
 
 In order to test metrics where the metric is included in more than one ping, the test functions take an optional `pingName` argument.
-This is the name of the ping that the metric is being sent in, such as `events` for the "events" ping, or `metrics` for the "metrics" ping.
+This is the name of the ping that the metric is being sent in, such as `events` for the ["events" ping](pings/events.md),
```
Dexterp37: In other document files, we're using "`events` ping" rather than ""events" ping". Should we be consistent with the other usages?

docs/user/pings/index.md
badboy: To have the same order as it does in the table of contents (sidebar). (I mention this in the commit so that we also know it when we ever come back to it)

docs/user/testing-metrics.md
badboy: Yeah, the phrasing here makes it a bit difficult. We keep repeating "events" and "metrics"

badboy: I'm gonna use `"metrics"` to signal it's a string and change the other to make it consistent.

PR #241, opened 2019-08-09
[doc only] 1571744: Document the custom distribution type

mdboom:

### DIFF
--- a/docs/dev/core/internal/payload.md
+++ b/docs/dev/core/internal/payload.md
@@ -68,7 +68,7 @@ A [Timespan](../../../user/metrics/timespan.md) is represented as an object of t
 }
 ```
 
-## Timing Distribution
+### Timing Distribution
 
 A [Timing distribution](../../../user/metrics/timing_distribution.md) is represented as an object with the following fields.
 
@@ -156,6 +156,27 @@ When the application starts up again, there is no good way to determine if the d
 To get around this, on application startup, any queued events are immediately collected into pings and then cleared.
 These "startup-triggered pings" are likely to have a very short duration, as recorded in `ping_info.start_time` and `ping_info.end_time` (see [the `ping_info` section](../../../user/pings/index.md#The-ping_info-section)).
 The maximum timestamp of the events in these pings are quite likely to exceed the duration of the ping, but this is to be expected.
+
+### Custom Distribution
+
+A [Custom distribution](../../../user/metrics/custom_distribution.md) is represented as an object with the following fields.
+
+| Field name | Type | Description |
+|---|---|---|
+| `sum` | Integer | The sum of all recorded values. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All buckets [0, max) are reported, so that the histograms can be aggregated in the pipeline without the pipeline knowing anything about the distribution of the buckets. | 
+
+#### Example:
+
+```json
+{
+    "sum": 3,
+    "values": {
+        "0": 1,
+        "1": 3,
+    }
+}
+```
 
 ### Labeled metrics
 
--- a/docs/user/metrics/custom_distribution.md
+++ b/docs/user/metrics/custom_distribution.md
@@ -1,0 +1,86 @@
+# Custom Distribution
+
+Custom distributions are used to record the distribution of arbitrary values.
+
+It should be used only when direct control over how the histogram buckets are computed is required.
+Otherwise, look at the standard distribution metric types:
+
+* [Timing Distributions](timing_distribution.md))
+
+> Note: Custom distributions are currently only allowed for GeckoView metrics (the `gecko_datapoint` parameter is present).
+
+## Configuration
+
+Custom distributions have the following required parameters:
+
+  - `range_min`: (Integer) The minimum value of the first bucket
+  - `range_max`: (Integer) The minimum value of the last bucket
+  - `bucket_count`: (Integer) The number of buckets
+  - `histogram_type`: 
+    - `linear`: The buckets are evenly spaced
+    - `exponential`: The buckets follow a natural logarithmic distribution
+
+In addition, the metric should specify:
+
+  - `unit`: (String) The unit of the values in the metric. For documentation purposes only -- does not affect data collection.
+
+If you wanted to create a custom distribution to the fraction of memory that is overhead as a percentage, first you need to add an entry for it to the `metrics.yaml` file:
+
+```YAML
+graphics:
+  checkerboard_peak:
+    type: custom_distribution
+    description: >
+      Peak number of CSS pixels checkerboarded during a checkerboard event.
+    range_min: 1
+    range_max: 66355200
+    bucket_count: 50
+    histogram_type: exponential
+    unit: pixels
+    gecko_datapoint: MEMORY_HEAP_OVERHEAD_FRACTION
+    ...
+```
+
+## API
+
+Now you can use the custom distribution from the application's code.
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Graphics
+
+Graphics.checkerboardPeak.accumulateSamples([23])
+```
+
+There are test APIs available too.  For convenience, properties `sum` and `count` are exposed to facilitate validating that data was recorded correctly.
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Graphics
+
+// Was anything recorded?
+assertTrue(Graphics.checkerboardPeak.testHasValue())
+
+// Get snapshot
+val snapshot = Graphics.checkerboardPeak.testGetValue()
+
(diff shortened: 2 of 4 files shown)


docs/dev/core/internal/payload.md
```diff
@@ -157,6 +157,27 @@ To get around this, on application startup, any queued events are immediately co
...
 The maximum timestamp of the events in these pings are quite likely to exceed the duration of the ping, but this is to be expected.
 
+### Custom Distribution
+
+A [Custom distribution](../../../user/metrics/custom_distribution.md) is represented as an object with the following fields.
+
+| Field name | Type | Description |
+|---|---|---|
+| `sum` | Integer | The sum of all recorded values. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. |
```
badboy: Are custom distributions always exponential? 
Also it doesn't send the range and bucket count parameters? I thought the pipeline would be interested in all data being mostly self-contained

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+## Configuration
+
+Custom distributions have the following required parameters:
+
+  - `range_min`: The minimum value of the first bucket
+  - `range_max`: The minimum value of the last bucket
+  - `bucket_count`: The number of buckets
+  - `histogram_type`: 
+    - `linear`: The buckets are evenly spaced
+    - `exponential`: The buckets follow a natural logarithmic distribution
```
badboy: so they can be either linear or exponential, therefore we should send this information in the payload.

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+
+In addition, the metric should specify:
+
+  - `unit`: The unit of the values in the metric. For documentation purposes only -- does not affect data collection.
```
badboy: Is this a string then?

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+```
+
+## API
+
+Now you can use the custom distribution from the application's code.
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Memory
+
+Memory.heapOverheadFraction.accumulateSamples([23])
```
badboy: It doesn't have a method to just accumulate a single value?

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+// Does the sum have the expected value?
+assertEquals(11, snapshot.sum)
+
+// Usually you don't know the exact timing values, but how many should have been recorded.
+assertEquals(2L, snapshot.count())
+```
+
+## Limits
+
+* The maximum value of `bucket_count` is 100.
```
badboy: Was this chosen as none of the geckoview-metrics needs anything higher or for any other reason?

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+    - `exponential`: The buckets follow a natural logarithmic distribution
+
+In addition, the metric should specify:
+
+  - `unit`: The unit of the values in the metric. For documentation purposes only -- does not affect data collection.
+
+If you wanted to create a custom distribution to the fraction of memory that is overhead as a percentage, first you need to add an entry for it to the `metrics.yaml` file:
+
+```YAML
+memory:
```
Dexterp37: minor-nit: is it worth using a different example here? We're adding memory distributions as well and this might be a bit confusing to people (even though the unit is clearly different!). I'd suggest something that has nothing to do with memory (e.g. pixel distribution?)

docs/dev/core/internal/payload.md
Dexterp37: Can we also mention that this is reporting all buckets [0, max)? A small note somewhere about the rationale behind this would be very useful for our future selves, IMHO.!

mdboom: Per [discussion here](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/375#issuecomment-519207980) it was decided to not send these parameters in the ping, but to explicitly send a contiguous range of buckets (even if 0).  This eliminates the need for the server and the client to "match".

docs/user/metrics/custom_distribution.md
mdboom: The idea is that the server (and aggregator) don't have to care about the bucketing function.  And this has been confirmed to work with the existing pipeline implementation...

docs/user/metrics/custom_distribution.md
mdboom: Yes.  I'll add that detail.

docs/user/metrics/custom_distribution.md
mdboom: Not at present, but that's basically an artifact of the fact that this is currently GeckoView-only.  @Dexterp37: What do you think?  Should we add a single sample method (even though nothing will use it?)

docs/user/metrics/custom_distribution.md
mdboom: Desktop telemetry currently has this limit as well, unless the metric is on an allow list.

None of the slated metrics from GeckoView use more than 100 buckets, and therefore it didn't seem worth the complexity of adding an allow list and all that, so it's just a hard limit.

(We can always expand this or add an allow list later if needed).

docs/user/metrics/custom_distribution.md
mdboom: Good point.  I wanted to use a "real" example, but I agree a non-memory one might be less confusing.

mdboom: It seems all the other "real" ones are timing adjacent and will have a similar confusion with timing.  I guess I'll make one up about pixels.

docs/user/metrics/custom_distribution.md
Dexterp37: I don't think we should since, as you stated, nothing will really be using it :)

badboy:
Tests failing because `custom_distribution.md` wasn't added to SUMMARY.md and is therefore not rendered.

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+
+* [Timing Distributions](timing_distribution.md))
+
+> Note: Custom distributions are currently only allowed for GeckoView metrics (the `gecko_datapoint` parameter is present).
+
+## Configuration
+
+Custom distributions have the following required parameters:
+
+  - `range_min`: (int) The minimum value of the first bucket
```
badboy:
```suggestion
  - `range_min`: (Number) The minimum value of the first bucket
```

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+  - `range_max`: (int) The minimum value of the last bucket
```
badboy:
```suggestion
  - `range_max`: (Number) The minimum value of the last bucket
```

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+  - `bucket_count`: (int) The number of buckets
```
badboy:
```suggestion
  - `bucket_count`: (Number) The number of buckets
```

docs/user/metrics/custom_distribution.md
```diff
@@ -0,0 +1,86 @@
...
+  - `histogram_type`: 
+    - `linear`: The buckets are evenly spaced
+    - `exponential`: The buckets follow a natural logarithmic distribution
+
+In addition, the metric should specify:
+
+  - `unit`: (string) The unit of the values in the metric. For documentation purposes only -- does not affect data collection.
```
badboy:
```suggestion
  - `unit`: (String) The unit of the values in the metric. For documentation purposes only -- does not affect data collection.
```

docs/user/metrics/custom_distribution.md
mdboom: I think it's important to communicate that these are integers -- you can't set range 0-1, bucket_count of 100 and see 0.01, etc., for example.

badboy: `Integers` then?

mdboom: Sounds good to me

dbaron:
For what it's worth (since I was just reading through this documentation), the [new page](https://mozilla.github.io/glean/book/user/metrics/custom_distribution.html) seems a bit confusing since the example seems to be half about memory overhead percentages and half about pixel checkerboarding counts.

mdboom:
Thanks -- that's just an edit gone half way.  I'll file a PR to clean it up -- it should be the pixel checkerboarding example all the way through.

PR #243, opened 2019-08-13
[doc only] Document memory distribution metric type

mdboom:

### DIFF
--- a/docs/dev/core/internal/payload.md
+++ b/docs/dev/core/internal/payload.md
@@ -75,7 +75,28 @@ A [Timing distribution](../../../user/metrics/timing_distribution.md) is represe
 | Field name | Type | Description |
 |---|---|---|
 | `sum` | Integer | The sum of all recorded values. |
-| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. Buckets with no values are not reported. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All contiguous buckets between the minimum non-zero bucket and the maximum non-zero bucket are included. |
+
+#### Example:
+
+```json
+{
+    "sum": 3,
+    "values": {
+        "0": 1,
+        "1": 3,
+    }
+}
+```
+
+### Memory Distribution
+
+A [Memory distribution](../../../user/metrics/memory_distribution.md) is represented as an object with the following fields.
+
+| Field name | Type | Description |
+|---|---|---|
+| `sum` | Integer | The sum of all recorded values. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All contiguous buckets between the minimum non-zero bucket and the maximum non-zero bucket are included. |
 
 #### Example:
 
@@ -157,7 +178,7 @@ A [Custom distribution](../../../user/metrics/custom_distribution.md) is represe
 | Field name | Type | Description |
 |---|---|---|
 | `sum` | Integer | The sum of all recorded values. |
-| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All buckets [0, max) are reported, so that the histograms can be aggregated in the pipeline without the pipeline knowing anything about the distribution of the buckets. | 
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All buckets [min, max + 1] are reported, so that the histograms can be aggregated in the pipeline without the pipeline knowing anything about the distribution of the buckets. | 
 
 #### Example:
 
--- a/docs/dev/core/internal/payload.md
+++ b/docs/dev/core/internal/payload.md
@@ -75,7 +75,17 @@ A [Timing distribution](../../../user/metrics/timing_distribution.md) is represe
 | Field name | Type | Description |
 |---|---|---|
 | `sum` | Integer | The sum of all recorded values. |
-| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All contiguous buckets between the minimum non-zero bucket and the maximum non-zero bucket are included. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. |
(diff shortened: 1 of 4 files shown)


docs/user/metrics/memory_distribution.md
```diff
@@ -0,0 +1,71 @@
...
+// Does the sum have the expected value?
+assertEquals(11, snapshot.sum)
+
+// Usually you don't know the exact memory values, but how many should have been recorded.
+assertEquals(2L, snapshot.count())
+```
+
+## Limits
+
+* The maxmimum memory size that can be recorded is 1 Terabyte (2^40 bytes). Longer times will be truncated to 1 Terabyte.
```
Dexterp37: nit: Longer times -> Larger sizes?

docs/user/metrics/memory_distribution.md
```diff
@@ -0,0 +1,71 @@
...
+      The heap memory allocated
+    memory_unit: kilobyte
+    ...
+```
+
+## API
+
+Now you can use the memory distribution from the application's code.
+
+For example, to measure page load time on a number of tabs that are loading at the same time, each tab object needs to store the running timer ID.
```
Dexterp37: nit: this paragraph needs de-copy-pasting :=)

docs/user/metrics/memory_distribution.md
```diff
@@ -0,0 +1,71 @@
...
+import org.mozilla.yourApplication.GleanMetrics.Memory
+
+fun allocateMemory(nbytes: Int) {
+    Memory.heapAllocated.accumulate(nbytes / 1024)
+}
+```
+
+There are test APIs available too.  For convenience, properties `sum` and `count` are exposed to facilitate validating that data was recorded correctly.
+
+Continuing the `pageLoad` example above, at this point the metric should have a `sum == 11` and a `count == 2`:
```
badboy:
```suggestion
Continuing the `heapAllocated` example above, at this point the metric should have a `sum == 11` and a `count == 2`:
```

badboy: Plus the numbers don't match up either

docs/user/metrics/memory_distribution.md
```diff
@@ -0,0 +1,71 @@
...
+assertEquals(2L, snapshot.count())
+```
+
+## Limits
+
+* The maxmimum memory size that can be recorded is 1 Terabyte (2^40 bytes). Longer times will be truncated to 1 Terabyte.
+
+## Examples
+
+* How much memory was allocated by a given process?
```
badboy: that question reads like a total or maximum, but the distribution better answers how many different allocations of different sizes happened, doesn't it?

docs/dev/core/internal/payload.md
```diff
@@ -75,7 +75,28 @@ A [Timing distribution](../../../user/metrics/timing_distribution.md) is represe
 | Field name | Type | Description |
 |---|---|---|
 | `sum` | Integer | The sum of all recorded values. |
-| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. Buckets with no values are not reported. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. All contiguous buckets between the minimum non-zero bucket and the maximum non-zero bucket are included. |
```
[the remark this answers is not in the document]
mdboom: So, all of this has been in a bit of flux so you're right to call for some clarification.

Basically it's this:

- For functional distributions, the range is from the first bucket with a non-zero value, to the bucket after the last bucket with a non-zero value -- since you need to have one extra bucket so you know where the last bucket ends.  You don't want to start at the first bucket, since for these functional histograms it can be a lot of buckets.

- For custom distributions, the range is from the first bucket regardless of its value (defined by rangeMin) to the bucket after the last bucket with the non-zero value.

I'll try to come up with a way to explain this more clearly and update this PR.

mdboom:
@badboy: Please let me know if these new explanations are clearer.  I'll go back and update the comments in a-c as well if they are.

docs/user/metrics/memory_distribution.md
mdboom: Numbers don't match up?

docs/dev/core/internal/payload.md
```diff
@@ -75,7 +75,45 @@ A [Timing distribution](../../../user/metrics/timing_distribution.md) is represe
 | Field name | Type | Description |
 |---|---|---|
 | `sum` | Integer | The sum of all recorded values. |
-| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. Buckets with no values are not reported. |
+| `values` | Map&lt;String, Integer&gt; | The values in each bucket. The key is the minimum value for the range of that bucket. |
+
+A contiguous range of buckets is always sent, so that the server can aggregate and visualize distributions, without knowing anything about the specific bucketing function used.
+This range starts with the first bucket with a non-zero accumulation, and ends at one bucket beyond the last bucket with a non-zero accumulation (so that the upper bound on the last bucket is retained).
```
badboy: One last thing: With the current mechanism we have 313 or so buckets (based on the maximum recordable time, 10 minutes). If that last bucket is non-empty, does it mean we add a 314th (empty) bucket in the serialization?

docs/user/metrics/memory_distribution.md
```diff
@@ -0,0 +1,72 @@
...
+// Does the sum have the expected value?
+assertEquals(11, snapshot.sum)
+
+// Usually you don't know the exact memory values, but how many should have been recorded.
+assertEquals(2L, snapshot.count())
+```
+
+## Limits
+
+* The maxmimum memory size that can be recorded is 1 Terabyte (2^40 bytes). Larger sizes will be truncated to 1 Terabyte.
```
badboy: We could be fancy and rely on HTML here:

```suggestion
* The maxmimum memory size that can be recorded is 1 Terabyte (2<sup>40</sup> bytes). Larger sizes will be truncated to 1 Terabyte.
```

docs/dev/core/internal/payload.md
mdboom: That's a good question.  I think probably no, because the last bucket is basically an overflow bucket and basically "unbounded".  Will need to circle back on the implementation and make sure that's what it does...

PR #246, opened 2019-08-15
[doc only] 1568767: Document `sdk_generator.gradle` options

mdboom:

### DIFF
--- a/docs/user/android-build-configuration-options.md
+++ b/docs/user/android-build-configuration-options.md
@@ -1,0 +1,25 @@
+# Android build script configuration options
+
+This chapter describes build configuration options that control the behavior of Glean's `sdk_generator.gradle` script.
+These options are not usually required for normal use.
+
+## `allowMetricsFromAAR`
+
+Normally, Glean looks for `metrics.yaml` and `pings.yaml` files in the root directory of the Glean-using project.
+However, in some cases, these files may need to ship inside the dependencies of the project.
+For example, this is used in the `engine-gecko` component to grab the `metrics.yaml` from the `geckoview` AAR.
+
+To turn on this behavior, set a variable on the Gradle [ext](https://docs.gradle.org/current/dsl/org.gradle.api.plugins.ExtraPropertiesExtension.html) object:
+
+```groovy
+ext.allowMetricsFromAAR = true
+```
+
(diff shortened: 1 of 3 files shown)


docs/user/android-build-configuration-options.md
```diff
@@ -0,0 +1,35 @@
...
+
+## `gleanNamespace`
+
+Should be set to the Java/Kotlin namespace that the Glean API is in.  
+Normally, this should be left at the default, but it is useful to switch between different implementations of Glean.
+
+## `allowGleanInternal`
+
+This allows a library to define metrics and pings in namespaces that are reserved for the Glean library itself.
+This should never be set from an application or library other than Glean.
```
mdboom: I'm not 100% sold we should document these last two: Telling users where our "backdoors" are might not be a good thing. But including them for completeness now and I'm happy to delete.

travis79: I'm with you on this, I don't think documenting things that shouldn't be touched is a good idea.

docs/user/android-build-configuration-options.md
```diff
@@ -0,0 +1,35 @@
...
+ext.allowMetricsFromAAR = true
+```
+
+When this flag is set, every direct dependency of your library will be searched for a `metrics.yaml` file, and those metrics will be treated as the metrics as if they were defined by your library.
+That is, API wrappers accessible from your library will be generated for those metrics.
+
+The `metrics.yaml` can be added to the dependency itself by calling this on each relevant build variant:
+
+```groovy
+    variant.packageLibraryProvider.get().from("${topsrcdir}/path/metrics.yaml")
```
badboy: No indentation needed

docs/user/android-build-configuration-options.md
badboy: Given this is user-facing, I agree. Let's leave it out.

Dexterp37: I'd recommend moving them to our [internal docs](https://mozilla.github.io/glean/book/dev/core/internal/index.html), but I have no strong opinion on this :)

badboy:
Rekicked tests, as Rust Beta was failing with out-of-memory. If that keeps happening, I'm inclined to un-require them for now, until we look into that.

docs/user/android-build-configuration-options.md
mdboom: Personally, I think it's fine that they are just documented in comments for now.

PR #247, opened 2019-08-15
 Bug 1566001 - Add limits for the experiment's extra HashMap

travis79:
Adds limits to the `extras` HashMap in experiments.rs

### DIFF
--- a/docs/user/experiments-api.md
+++ b/docs/user/experiments-api.md
@@ -39,7 +39,8 @@ assertEquals(
 
 ## Limits
 
-* Fixed maximum experiment and branch id lengths: 30. Longer strings are truncated. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
+* `experimentId`, `branch`, and the keys of the 'extra' field are fixed at a maximum length of 30. Length for the values of the `extra` field is fixed at 50. Longer strings used as ids, keys, or values are truncated to their respective maximum length. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
+* `extra` map is limited to 20 entries. If passed a map which contains more elements than this, it is truncated to 20 elements.  **WARNING** Which items are truncated is nondeterministic due to the unordered nature of maps and what's left may not necessarily be the first elements added.
 
 ## Reference
 
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -597,6 +597,61 @@ mod test {
     }
 
     #[test]
+    fn limits_on_experiments_extras_are_applied_correctly() {
+        let t = tempfile::tempdir().unwrap();
+        let name = t.path().display().to_string();
+        let glean = Glean::with_options(&name, "org.mozilla.glean.tests", true).unwrap();
+
+        let experiment_id = "test-experiment_id".to_string();
+        let branch_id = "test-branch-id".to_string();
+        let mut extras = HashMap::new();
+
+        let too_long_key = "0123456789".repeat(5);
+        let too_long_value = "0123456789".repeat(6);
+
+        // Build and extras HashMap that's a little too long in every way
+        for n in 0..21 {
+            extras.insert(format!("{}-{}", n, too_long_key), too_long_value.clone());
+        }
+
+        // Mark the experiment as active.
+        glean.set_experiment_active(experiment_id.clone(), branch_id.clone(), Some(extras));
+
+        // Make sure it is active
+        assert!(
+            glean.test_is_experiment_active(experiment_id.clone()),
+            "An experiment with the truncated id should be available"
+        );
+
+        // Get the data
+        let experiment_data = glean.test_get_experiment_data_as_json(experiment_id.clone());
+        assert!(
+            !experiment_data.is_none(),
+            "Experiment data must be available"
+        );
+
+        // Parse the JSON and validate the lengths
+        let parsed_json: RecordedExperimentData =
+            ::serde_json::from_str(&experiment_data.unwrap()).unwrap();
+        assert_eq!(
+            20,
+            parsed_json.clone().extra.unwrap().len(),
+            "Experiments extra must be less than max length"
+        );
+
+        for (key, value) in parsed_json.extra.as_ref().unwrap().iter() {
+            assert!(
+                key.len() <= 30,
+                "Experiments extra key must be less than max length"
+            );
+            assert!(
+                value.len() <= 50,
+                "Experiments extra value must be less than max length"
+            );
+        }
+    }
+
+    #[test]
     fn experiments_status_is_correctly_toggled() {
         let t = tempfile::tempdir().unwrap();
         let name = t.path().display().to_string();
--- a/glean-core/src/metrics/experiment.rs
+++ b/glean-core/src/metrics/experiment.rs
@@ -18,9 +18,16 @@ use crate::Lifetime;
 // An internal ping name, not to be touched by anything else
 const INTERNAL_STORAGE: &str = "glean_internal_info";
 
-/// The maximum length of the experiment id and the branch id. Identifiers
-/// longer than this number of characters are truncated.
+/// The maximum length of the experiment id, the branch id, and the keys of the
+/// `extra` map. Identifiers longer than this number of characters are truncated.
 const MAX_EXPERIMENTS_IDS_LEN: usize = 30;
+/// The maximum length of the experiment `extra` values.  Values longer than this
+/// limit will be truncated.
+const MAX_EXPERIMENT_VALUE_LEN: usize = 50;
+/// The maximum number of extras allowed in the `extra` hash map.  Any items added
+/// beyond this limit will be dropped. Note that truncation of a hash map is
+/// nondeterministic in which items are truncated.
+const MAX_EXPERIMENTS_EXTRAS_SIZE: usize = 20;
 
 /// The data for a single experiment.
 #[derive(Debug, Clone, Deserialize, Serialize)]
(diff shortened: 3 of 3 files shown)


glean-core/src/metrics/experiment.rs
```diff
@@ -101,23 +117,38 @@ impl ExperimentMetric {
...
-            truncate_string_at_boundary(branch, MAX_EXPERIMENTS_IDS_LEN)
+            truncate_string_at_boundary(branch, MAX_EXPERIMENTS_BRANCH_LEN)
         } else {
             branch
         };
 
-        // TODO (bug 1566001): add sane limits for the extra keys and
-        // values as well.
+        // Apply limits to extras
+        let truncated_extras = if extra.is_some() {
```
mdboom: You could do:

```
let truncated_extras = extra.and_then(|extra| { ... })
```

to avoid the `extra.unwrap()` and `else { None }`.

Maybe just a personal preference because I like to avoid calling `unwrap()` in non-test code since it *could* panic (even though in this specific case it never would).

But defer to the real Rust expert @badboy on this one.

glean-core/src/metrics/experiment.rs
```diff
@@ -101,23 +117,38 @@ impl ExperimentMetric {
...
+            let mut temp_map = HashMap::new();
+            for (key, value) in extra.unwrap().into_iter() {
+                let truncated_key = truncate_string_at_boundary(key, MAX_EXPERIMENT_KEY_LEN);
+                let truncated_value = truncate_string_at_boundary(value, MAX_EXPERIMENT_VALUE_LEN);
+                temp_map.insert(truncated_key, truncated_value);
+            }
```
mdboom: You could avoid converting all values only to throw some of them away by integrating the `take` here as well.

```
extra.unwrap().into_iter().take(MAX_EXPERIMENTS_EXTRAS_LEN)
```

travis79: That's less overhead, and then I shouldn't even need to apply `take` again if the list is already truncated to the correct number of entries.

glean-core/src/metrics/experiment.rs
travis79: TIL: This sounds like a better way to handle the Optional value rather than by force unwrapping it with `unwrap`.  Chalk this up to my lack of Rust knowledge.

glean-core/src/metrics/experiment.rs
```diff
@@ -18,9 +18,25 @@ use crate::Lifetime;
...
 const INTERNAL_STORAGE: &str = "glean_internal_info";
 
+// The following values are maximums related to experiments, these were
+// chosen based on current desktop usage. See here, for reference:
+// https://searchfox.org/mozilla-central/rev/3366c3d24f1c3818df37ec0818833bf085e41a53/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#58
+
 /// The maximum length of the experiment id and the branch id. Identifiers
 /// longer than this number of characters are truncated.
-const MAX_EXPERIMENTS_IDS_LEN: usize = 30;
+const MAX_EXPERIMENTS_IDS_LEN: usize = 100;
```
badboy: Just to double-check: the pipeline schema doesn't imply any limit, correct?

docs/user/experiments-api.md
```diff
@@ -39,7 +39,8 @@ assertEquals(
 
 ## Limits
 
-* Fixed maximum experiment and branch id lengths: 30. Longer strings are truncated. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
+* `experimentId` and `branch` fields are fixed at a maximum length of 100. Fixed maximum key and value lengths for the `extra` map is 50. Longer strings used as ids, keys, or values are truncated to their respective maximum length. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
```
badboy: This limit-extension is just to match what's on Desktop? If so I'm fine with it.

glean-core/src/lib.rs
```diff
@@ -596,6 +596,61 @@ mod test {
...
+            !experiment_data.is_none(),
+            "Experiment data must be available"
+        );
+
+        // Parse the JSON and validate the lengths
+        let parsed_json: RecordedExperimentData =
+            ::serde_json::from_str(&experiment_data.unwrap()).unwrap();
+        assert_eq!(
+            20,
+            parsed_json.clone().extra.unwrap().len(),
```
badboy: We shouldn't need to clone:

```suggestion
            parsed_json.extra.as_ref().unwrap().len(),
```

glean-core/src/metrics/experiment.rs
Dexterp37: Let's not bump this limit and keep using this for experiment branch too. If that becomes a problem for the experiments library, we can discuss and bump.

glean-core/src/lib.rs
badboy: We could also take out `extra`, put it into a variable and re-use it below.

```
let extra = parsed_json.extra.unwrap();
```

glean-core/src/metrics/experiment.rs
```diff
@@ -101,23 +117,31 @@ impl ExperimentMetric {
         }
 
         // Make sure that branch id is within the expected limit.
-        let truncated_branch = if branch.len() > MAX_EXPERIMENTS_IDS_LEN {
+        let truncated_branch = if branch.len() > MAX_EXPERIMENTS_BRANCH_LEN {
```
Dexterp37: Let's keep using `MAX_EXPERIMENTS_IDS_LEN` here. This PR is specifically only about the `extra` :-)

glean-core/src/metrics/experiment.rs
```diff
@@ -18,9 +18,25 @@ use crate::Lifetime;
...
+const MAX_EXPERIMENTS_IDS_LEN: usize = 100;
+/// The maximium length of the experiment branch id.  Any branch id longer
+/// than this will be truncated.
+const MAX_EXPERIMENTS_BRANCH_LEN: usize = 100;
+/// The maximum number of extras allowed in the extras hash map.  Any items
+/// added beyond this limit will be dropped.
+const MAX_EXPERIMENTS_EXTRAS_LEN: usize = 20;
+/// The maximum length of the experiment extra keys.  Keys and values longer
+/// than this limit will be truncated.
+const MAX_EXPERIMENT_KEY_LEN: usize = 50;
```
Dexterp37: Let's use the same `MAX_EXPERIMENTS_IDS_LEN` here, for consistency? On desktop, I don't see things that need 50 (I only see `type` and `enrollmentID` being used), so 30 should be enough.

badboy:
Oh wait, one more comment: Uff, that lib.rs is getting long, I suggest we split it up into a test file (in a follow up)

glean-core/src/metrics/experiment.rs
```diff
@@ -18,9 +18,25 @@ use crate::Lifetime;
...
+/// The maximum length of the experiment extra values.  Keys and values longer
```
Dexterp37: nit: this comment is wrong: "Keys and values longer than..." should be changed to "Values longer than..."

glean-core/src/metrics/experiment.rs
```diff
@@ -101,23 +117,31 @@ impl ExperimentMetric {
...
             branch
         };
 
-        // TODO (bug 1566001): add sane limits for the extra keys and
-        // values as well.
+        // Apply limits to extras
+        let truncated_extras = extra.and_then(|extra| {
+            let mut temp_map = HashMap::new();
+            for (key, value) in extra.into_iter().take(MAX_EXPERIMENTS_EXTRAS_LEN) {
+                let truncated_key = truncate_string_at_boundary(key, MAX_EXPERIMENT_KEY_LEN);
```
Dexterp37: In this PR we should, at the very least, show a warning in logcat and not silently truncate. Moreover, I agree with mike: let's file a follow-up bug for reporting this with the Glean SDK error reporting mechanism: this will allow, in the future, to make debug builds break if limits are hit.

Once the bug is filed, let's leave a comment here with the bug number.

docs/user/experiments-api.md
Dexterp37: Let's not change current limits, please. This will introduce a different behaviour between glean-ac and glean-core. Let's limit this PR to the `extra` map.

glean-core/src/metrics/experiment.rs
```diff
@@ -101,23 +117,31 @@ impl ExperimentMetric {
...
         } else {
             branch
         };
 
-        // TODO (bug 1566001): add sane limits for the extra keys and
-        // values as well.
+        // Apply limits to extras
+        let truncated_extras = extra.and_then(|extra| {
+            let mut temp_map = HashMap::new();
+            for (key, value) in extra.into_iter().take(MAX_EXPERIMENTS_EXTRAS_LEN) {
```
Dexterp37: Let's show a warning if the `extra` has too many items.

badboy:
Rekicked tests, as Rust Beta was failing with out-of-memory. If that keeps happening, I'm inclined to un-require them for now, until we look into that.

badboy:
If you rebase this on master, the tests should (more reliably) run again (less tests that is, beta got removed for now, android is working again)

glean-core/src/metrics/experiment.rs
travis79: Sorry, I had originally intended to use the same constant for both, but then that didn't really feel right when I got done.

docs/user/experiments-api.md
travis79: Again, was just trying to match what I found for desktop, but I can put it back for now.  Knowing what they wanted to name experiments and branches already (fairly long), I'm pretty sure we will need to address this soon.

Dexterp37: Let's keep this PR focused and address this problem when needed later on. Given we're going to migrate (hopefully) soon, we don't want to diverge implementation and create confusion in docs

glean-core/src/metrics/experiment.rs
travis79: It does 😞 
[here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/glean/short_id.1.schema.json)

travis79:
> Oh wait, one more comment: Uff, that lib.rs is getting long, I suggest we split it up into a test file (in a follow up)

Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1574467).

glean-core/src/metrics/experiment.rs
travis79: Good idea.  Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1574468)

travis79:
Another follow-up bug to add limit violations to the Glean error reporting [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1574468).

docs/user/experiments-api.md
```diff
@@ -39,7 +39,8 @@ assertEquals(
 
 ## Limits
 
-* Fixed maximum experiment and branch id lengths: 30. Longer strings are truncated. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
+* `experimentId` and `branch`, and the 'extra.key' fields are fixed at a maximum length of 30. Length for the `extra.value` field is 50. Longer strings used as ids, keys, or values are truncated to their respective maximum length. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
```
Dexterp37:
```suggestion
* `experimentId`, `branch`, and the keys of the 'extra' fields are fixed at a maximum length of 30. Length for the `extra.value` field is 50. Longer strings used as ids, keys, or values are truncated to their respective maximum length. For the original Kotlin implementation of the Glean SDK, this is measured in Unicode characters. For the Rust implementation, this is measured in the number of bytes when the string is encoded in UTF-8.
```

glean-core/src/metrics/experiment.rs
```diff
@@ -18,9 +18,16 @@ use crate::Lifetime;
 // An internal ping name, not to be touched by anything else
 const INTERNAL_STORAGE: &str = "glean_internal_info";
 
-/// The maximum length of the experiment id and the branch id. Identifiers
-/// longer than this number of characters are truncated.
+/// The maximum length of the experiment id and the branch id and the key field of
```
Dexterp37:
```suggestion
/// The maximum length of the experiment id and the branch id and the keys of
```

glean-core/src/metrics/experiment.rs
```diff
@@ -18,9 +18,16 @@ use crate::Lifetime;
...
+/// the `extra` map. Identifiers longer than this number of characters are truncated.
 const MAX_EXPERIMENTS_IDS_LEN: usize = 30;
+/// The maximum length of the experiment `extra` values.  Values longer than this
+/// limit will be truncated.
+const MAX_EXPERIMENT_VALUE_LEN: usize = 50;
+/// The maximum number of extras allowed in the `extra` hash map.  Any items added
+/// beyond this limit will be dropped. Note that truncation of a hash map is
+/// nondeterministic in which items are truncated.
+const MAX_EXPERIMENTS_EXTRAS_LEN: usize = 20;
```
Dexterp37:
```suggestion
const MAX_EXPERIMENTS_EXTRAS_SIZE: usize = 20;
```

PR #249, opened 2019-08-16
HACK: Fix up missing libtinfo.so.5

badboy:
The Debian container in use is shipping libtinfo.so.6, but the Clang deployed in the NDK requires v5.
We hack around that by symlinking the new to the old version, they seem to be mostly compatible.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -59,6 +59,12 @@ commands:
           name: Install missing Android SDK
           command: |
               sdkmanager 'build-tools;21.0.0'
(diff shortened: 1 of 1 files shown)


badboy:
> r+ if this works - let's file a follow-up bug for potentially removing this hack in the future. What would we need to remove it?

A clang that's not linked against old libraries. (or the old library)

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1574419

PR #250, opened 2019-08-16
Disable beta builds

badboy:
They are currently mostly failing due to out of memory issues.
Disabling to not waste resources.
We will re-enable them once we figure out what's going on.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -270,7 +270,8 @@ workflows:
   run-tests:
     jobs:
       - Rust tests - stable
-      - Rust tests - beta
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -270,7 +270,7 @@ workflows:
   run-tests:
     jobs:
       - Rust tests - stable
-      - Rust tests - beta
+      #- Rust tests - beta
```
Dexterp37: Can you file a bug about adding them back and drop the bug number, as a comment, here?

PR #251, opened 2019-08-16
Functional histograms

badboy:
I need to double-check that the functionality matches exactly what is in AC now.


glean-core/src/histogram/exponential.rs
```diff
@@ -51,142 +46,60 @@ fn exponential_range(min: u64, max: u64, bucket_count: usize) -> Vec<u64> {
...
-/// }
-///
-/// assert_eq!(10, hist.count());
-/// assert_eq!(55, hist.sum());
-/// ```
 #[derive(Debug, Clone, Serialize, Deserialize)]
-pub struct Histogram {
-    min: u64,
-    max: u64,
+pub struct Exponential {
```
mdboom: This could be confusing since `Functional` histograms are technically `Exponential` as well.  Maybe `Functional` vs. `Precomputed` instead?

glean-core/src/histogram/functional.rs
```diff
@@ -0,0 +1,136 @@
...
+            let key = *key;
+
+            if min_key.is_none() || key < min_key.unwrap() {
+                min_key = Some(key);
+            }
+
+            if max_key.is_none() || key > max_key.unwrap() {
+                max_key = Some(key);
+            }
+        }
```
mdboom: Could we use [max](https://doc.rust-lang.org/std/iter/trait.Iterator.html#method.max)?

glean-core/src/histogram/functional.rs
```diff
@@ -0,0 +1,136 @@
...
+    pub fn functional(log_base: f64, buckets_per_magnitutde: f64) -> Histogram<Functional> {
+        Histogram {
+            values: HashMap::new(),
+            count: 0,
+            sum: 0,
+            bucketing: Functional::new(log_base, buckets_per_magnitutde),
+        }
+    }
+
+    /// Get a snapshot of _all_ values.
```
mdboom: all -> contiguous values in a range

glean-core/src/histogram/exponential.rs
badboy: But then "linear" is also a precomputed histogram, but the computation function is different.

glean-core/src/histogram/functional.rs
badboy: We could. But we would also need to use `min`. Meaning we run through the list twice.
By doing the iteration ourselves, we can save one trip through the list (I should add a comment).

mdboom: Ah good point.  (/me kinda wants to file a bug to add min_max to Rust stdlib...)

glean-core/src/histogram/exponential.rs
mdboom: Maybe `PrecomputedExponential`?

Dexterp37: `PrecomputedExponential`++

glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt
```diff
@@ -156,13 +156,15 @@ class TimingDistributionMetricTypeTest {
...
-        assertEquals(1L, snapshot.values[1])
-        // Check that the 2L fell into the second bucket
-        assertEquals(1L, snapshot.values[2])
-        // Check that the 3L fell into the third bucket
-        assertEquals(1L, snapshot.values[3])
+        assertEquals(6L * secondsToNanos, snapshot.sum)
+
+        // We should get a sample in 3 buckets.
+        // These numbers are a bit magic, but they correspond to
+        // `hist.sample_to_bucket_minimum(i * seconds_to_nanos)` for `i = 1..=3`.
```
Dexterp37: nit: I'd add "... that lives in the Rust code".

glean-core/src/histogram/functional.rs
```diff
@@ -0,0 +1,136 @@
...
+use serde::{Deserialize, Serialize};
+
+use super::{Bucketing, Histogram};
+
+#[derive(Debug, Clone, Serialize, Deserialize)]
+pub struct Functional {
+    exponent: f64,
+}
+
+impl Functional {
```
Dexterp37: Can we add some documentation here or above to describe what's this about?

glean-core/src/histogram/mod.rs
```diff
@@ -0,0 +1,83 @@
...
+    // bucket minimum -> value
+    values: HashMap<u64, u64>,
+
+    count: u64,
+    sum: u64,
+
+    bucketing: B,
+}
+
+pub trait Bucketing {
```
Dexterp37: NICE! While you already mention that documenting traits it's uncommon in the Rust community, I'd really appreciate if we could document what is this trait for :o

badboy: I never said documenting traits is uncommon! 

This absolutely needs documentation.

Dexterp37: > 
> 
> I never said documenting traits is uncommon!

My memory is definitely failing me these days. Sorry for putting words in your "fingers" :-)

PR #252, opened 2019-08-19
Memory dist

badboy:
Based on #251, needs to be rebased when that is merged.


glean-core/src/histogram/exponential.rs
```diff
@@ -51,142 +46,68 @@ fn exponential_range(min: u64, max: u64, bucket_count: usize) -> Vec<u64> {
...
+    /// Get a snapshot of _all_ values.
+    pub fn snapshot_values(&self) -> HashMap<u64, u64> {
+        let mut res = self.values.clone();
 
-    /// Check if this histogram recorded any values.
-    pub fn is_empty(&self) -> bool {
-        self.count() == 0
+        for &min_bucket in &self.bucketing.bucket_ranges {
+            // Fill in missing entries.
+            let _ = res.entry(min_bucket).or_insert(0);
```
mdboom: There's no need to fill in all values -- just from 0 to (max_non_zero_bucket + 1).  I think this is a useful optimization to reduce ping sizes.  See [here](https://github.com/mozilla/glean/blob/master/docs/dev/core/internal/payload.md#custom-distribution)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/MemoryDistributionMetricType.kt
```diff
@@ -0,0 +1,149 @@
...
+        if (!testHasValue(pingName)) {
+            throw NullPointerException()
+        }
+
+        val ptr = LibGleanFFI.INSTANCE.glean_memory_distribution_test_get_value_as_json_string(
+                Glean.handle,
+                this.handle,
+                pingName)!!
+
+        return DistributionData.fromJsonString(ptr.getAndConsumeRustString())!!
```
mdboom: There's a behavior change here vs. glean a-c -- the a-c version gives access to the "raw" values (i.e. without the contiguous values inserted).  I don't know if it matters, but it could cause some tests to fail (of course no one is using these distributions in third-party code yet).

glean-core/src/histogram/functional.rs
```diff
@@ -0,0 +1,152 @@
...
+        for key in self.values.keys() {
+            let key = *key;
+
+            if min_key.is_none() || key < min_key.unwrap() {
+                min_key = Some(key);
+            }
+
+            if max_key.is_none() || key > max_key.unwrap() {
+                max_key = Some(key);
+            }
```
mdboom: There's potential for a bug here if we're not careful.  I might just be missing some parts of the code here, so feel free to just point me at them if I'm wrong.

This *assumes* that only non-zero-valued entries are in the map in order to find the min/max.  If you ran this over values that were already serialized/deserialized from the database (and thus have an extra zero-valued entry at the end), you would end up adding an extra element to the end each time you do so.  The glean-ac implementation gets around this by using a different serialization for (a) persistence vs. (b) the ping payload.  For (a), it doesn't insert the zero values and just saves the "raw" sparse data.  For (b) it does.  Not only does that avoid that bug, it avoids doing extra work for the local persistence.  I *think* here, it's always inserting the contiguous values for both (but I could be wrong -- that serialization is more implicit here and I just might be missing it...)  If having two different serializations doesn't make sense in this implementation, then I think we at least need to fix the bug along the lines of:

```
  for (key, val) in self.values.iter() {
      if val == 0 {
          continue;
      }
      ...
  }
```

We're also going to need to do something similar for custom distributions, where we need to serialize the parameters for persistence, but we don't need to send them in the payload...

glean-core/android/src/main/java/mozilla/telemetry/glean/private/MemoryDistributionMetricType.kt
badboy: Hm, that makes it slightly more complex to test that something is empty (null or 0), but IMO us not having 2 different formats of serialization to support is better.

glean-core/src/histogram/functional.rs
badboy: We serialize the `Histogram<Functional>` into the database, that is only the hash map of stored values.
As we don't store 0-buckets, we can't run into that problem.

persistence format and payload format are already different.

PR #253, opened 2019-08-19
 Bug 1568503 - Ensure queued initial tasks are executed in order

travis79:
This ensures that tasks queued before Glean is initialized are executed in the proper order, including tasks that may have been launched after the flushing of the queue has began.


glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt
```diff
@@ -60,4 +63,50 @@ class DispatchersTest {
...
+    @Test
+    fun `queued tasks are executed in the order they are received`() {
+        var orderedString = ""
+        var curNum = 0
+
+        // Set testing mode to false as we want every call to `launch` to execute as a coroutine
+        Dispatchers.API.setTestingMode(false)
+        Dispatchers.API.setTaskQueueing(true)
+
+        // Spawn a coroutine that will add elements to the orderedString.  This will continue to add
```
badboy: Why a string? Naturally I would have just reached for a list.

travis79: > Did you check that the test really (reliably) fails without the changes to `Dispatchers.kt`?

I could get the new test to fail intermittently, and (locally) it appeared that the changes removed the failure.  Of course, it's still failing on CI so I may have to dig a little deeper.

travis79: A list should work fine. I went for a string as it was easy to compare the two side by side to see when/where it went wrong, but it looks like a list gives me the same ability.

travis79:
I modified the test to more _reliably_ fail, per your request.  In doing so I also changed the queue flushing approach to work better at executing in order.

mdboom:
This looks great.  One thing that doesn't "feel right" is the hardcoded timeout for the initial events, just because if it is hit, results are going to be weird and we might not immediately know about it.  I wonder if it would be possible to record an error (that gets sent in the ping) if that timeout times out...

travis79:
> This looks great. One thing that doesn't "feel right" is the hardcoded timeout for the initial events, just because if it is hit, results are going to be weird and we might not immediately know about it. I wonder if it would be possible to record an error (that gets sent in the ping) if that timeout times out...

Yeah, I understand.  I'm not a fan of timeouts either but in this case it felt better than having a situation where Glean would hang.  It wasn't necessary before because it wasn't using the `runBlocking` to ensure that the queued items get executed before any items that come in after the queue, so I added it as a safety measure.

I don't think that either `invalid_value` or `invalid_label` really fit with a timeout error, what do you think about filing a bug about adding an error for this case as follow-up work?

mdboom:
Yeah -- sounds good to me to file a bug to come back to this.  Thanks!

mdboom:
Oh, and I'll file that bug...

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -111,13 +116,28 @@ internal object Dispatchers {
...
+                }
+                queueInitialTasks = false
+                taskQueue.clear()
+            }
+
+            // In order to ensure that the queued tasks are executed in the proper order, we will
+            // wait up to 5 seconds for it to complete, otherwise we will reset the flag so that
+            // new tasks may continue to run.
+            runBlocking {
+                val result = withTimeoutOrNull(QUEUE_PROCESSING_TIMEOUT_MS) {
```
badboy: What is the return value of this block when it doesn't time out?

glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt
```diff
@@ -60,4 +65,52 @@ class DispatchersTest {
         assertEquals("Tasks have executed", 3, threadCanary)
         assertEquals("Task queue is cleared", 0, Dispatchers.API.taskQueue.size)
     }
+
+    @Test
+    fun `queued tasks are executed in the order they are received`() {
+        val orderedList = mutableListOf<String>()
```
badboy: Well, I guess I would have gone for a list of ints, but converting them to a string is just as fine

glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt
```diff
@@ -60,4 +65,52 @@ class DispatchersTest {
...
+        Dispatchers.API.setTaskQueueing(true)
+
+        // Spawn a coroutine that will add elements to the orderedList.  This will continue to add
+        // elements to the queue until there are at least 50 elements in the queue. At that point,
+        // the second coroutine below will flush and disable the queue.
+        val addJob = GlobalScope.launch {
+            (0..100).forEach { num ->
+                Dispatchers.API.launch {
+                    orderedList.add("$num")
+                }.also { job ->
```
badboy: Oh noes, another operator

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -111,13 +116,28 @@ internal object Dispatchers {
...
+                    job.join()
+                }
+
+                if (result == null) {
```
badboy: In case of an error it could be that tasks stick around in the `taskQueue`, right?

badboy:
Stable test is failing. Looks like we're getting OOM-killed again.
Is this rebased on latest master?

I might have an idea how to reduce the impact of that...

badboy:
Oh wait, the oom mitigation wasn't landed on master

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
travis79: It returns the coroutine back.  See [here](https://kotlin.github.io/kotlinx.coroutines/kotlinx-coroutines-core/kotlinx.coroutines/with-timeout-or-null.html)

glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt
travis79: Probably lighter weight yet to go with Ints, point taken.

glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt
travis79: While my original Kotlinism works, I could be more clear.  Let me rephrase it...

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
travis79: Yes, good catch!  The queue should probably be emptied if we error.

travis79:
> This LGTM, thanks for tackling it. Can we make sure this test is not flaky (i.e. execute it 4-5 times on CI before merging?). Moreover, does the test consistently fail without your changes?

Thanks for reviewing!  Yes, the test fails consistently now.  That was one of the things that @badboy brought up in his review and so I made sure it was maximum chance to fail.  And as far as I can tell, it's not flaky just had some issues with the CI container size causing failures last week.

PR #256, opened 2019-08-22
Use a container with more memory available on CI

badboy:


badboy:
> Seems fine to me. Does this cost us more?

I ... don't fully know, but I think not, it's just a feature that comes with our mozilla account (I'll double-check though)

mdboom:
I wouldn't worry too much about cost -- I'm sure whatever it is is cheaper than the developer cost of moving to taskcluster etc.

PR #259, opened 2019-08-27
Bug 1576741 - Fix intermittent Dispatcher test failure

travis79:
This reinstates the test and corrects the intermittent failure.


glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -117,12 +116,28 @@ internal object Dispatchers {
             // situations. If a library or thread that runs before init happens to record
             // between when the queueInitialTasks is set to false and the taskQueue finishing
             // launching, then that task could be executed out of the queued order.
+            val that = this
```
mdboom: If this is a common idiom, ignore my comment... setting `this` to `that` seems confusing.  Maybe `dispatchersObject` instead?

travis79: This was just needed because the context of `this` changes inside of the closure and doing this captures the correct reference to `this`.  I'll rename it to be more clear.

PR #261, opened 2019-09-03
Bug 1576774 - Removed android-components support-ktx dependency

travis79:
Inlining wasn't ideal in every case, so I went ahead and copied over the extension files for the JSON stuff from android-components to the /utils directory and added the extension for `Context` that detects the main process to the `Glean.kt` file.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -4,15 +4,16 @@
 
 package mozilla.telemetry.glean
 
+import android.app.ActivityManager
 import android.util.Log
 import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
+import android.os.Process
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ProcessLifecycleOwner
 import com.sun.jna.StringArray
 import kotlinx.coroutines.Job
-import mozilla.components.support.ktx.android.content.isMainProcess
 import mozilla.telemetry.glean.config.Configuration
 import mozilla.telemetry.glean.config.FfiConfiguration
 import mozilla.telemetry.glean.utils.getLocaleTag
@@ -73,6 +74,10 @@ open class GleanInternalAPI internal constructor () {
     // Keep track of ping types that have been registered before Glean is initialized.
     private val pingTypeQueue: MutableList<PingType> = mutableListOf()
 
+    // This is used to cache the process state and is used by the function `isMainProcess()`
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal var isMainProcess: Boolean? = null
+
     /**
      * Initialize the Glean SDK.
      *
@@ -96,7 +101,7 @@ open class GleanInternalAPI internal constructor () {
     ) {
         // In certain situations Glean.initialize may be called from a process other than the main
         // process.  In this case we want initialize to be a no-op and just return.
-        if (!applicationContext.isMainProcess()) {
+        if (!isMainProcess(applicationContext)) {
             Log.e(LOG_TAG, "Attempted to initialize Glean on a process other than the main process")
             return
         }
@@ -511,6 +516,8 @@ open class GleanInternalAPI internal constructor () {
             LibGleanFFI.INSTANCE.glean_test_clear_all_stores(handle)
         }
 
+        isMainProcess = null
+
         // Init Glean.
         Glean.testDestroyGleanHandle()
         Glean.setUploadEnabled(true)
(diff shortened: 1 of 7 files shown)


travis79:
> Do we want to copy over all of the JSON helper methods if we only use some of them? I'm sure Proguard will cull them, but just in terms of maintenance burden...

That's a good question and one I was going to pose as part of this PR.  I'm happy to strip them, considering we aren't using them.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -556,6 +557,25 @@ open class GleanInternalAPI internal constructor () {
     internal fun testHasPingType(pingName: String): Boolean {
         return LibGleanFFI.INSTANCE.glean_test_has_ping_type(handle, pingName).toBoolean()
     }
+
+    @VisibleForTesting
```
Dexterp37: Why is this visible for testing? Doesn't seem to be directly tested anyway.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -556,6 +557,25 @@ open class GleanInternalAPI internal constructor () {
...
+    internal var isMainProcess: Boolean? = null
```
Dexterp37: Why is this `internal`? Can't this just be `private`?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -556,6 +557,25 @@ open class GleanInternalAPI internal constructor () {
...
+
+    /**
+     *  Returns true if we are running in the main process false otherwise.
+     */
+    private fun Context.isMainProcess(): Boolean {
```
Dexterp37: Do we really need to extend `Context` with this, since we're the only ones using it, once?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringListMetricType.kt
```diff
@@ -153,6 +153,6 @@ class StringListMetricType(
         } catch (e: org.json.JSONException) {
             throw NullPointerException()
         }
-        return jsonRes.toList<String>()
+        return jsonRes.toList()
```
Dexterp37: We're really only using `toList` (and `asSequence`, used by it): let's not import other code which we won't need.

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/JSONArray.kt
```diff
@@ -0,0 +1,57 @@
...
+
+fun JSONArray.asSequence(): Sequence<Any> = asSequence { i -> get(i) }
+
+/**
+ * Convenience method to convert a JSONArray into a List
+ *
+ * @return list with the JSONArray values, or an empty list if the JSONArray was null
+ */
+@Suppress("UNCHECKED_CAST")
+fun <T> JSONArray?.toList(): List<T> {
```
Dexterp37: Let's also import the relevant test coverage for the imported functions coming from [this file](https://github.com/mozilla-mobile/android-components/blob/cd3eb4d3473db5f3fd6524decf1c8cee80cc7548/components/support/ktx/src/test/java/mozilla/components/support/ktx/android/org/json/JSONArrayTest.kt)

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/JSONObject.kt
```diff
@@ -0,0 +1,98 @@
...
+package mozilla.telemetry.glean.utils
+
+import org.json.JSONObject
+import java.util.TreeMap
+
+/**
+ * Returns the value mapped by {@code key} if it exists, and
+ * if the value returned is not null. If it's null, it returns null
+ */
+fun JSONObject.tryGet(key: String): Any? = if (isNull(key)) null else get(key)
```
Dexterp37: Are we using anything in this file at all?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: Because there is a test associated with this function and variable that I hadn't copied over from AC yet.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: Because of the test that I hadn't copied over from AC yet.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: Nope, we can change it to a regular function.  Extensions don't have any additional overhead, that I can see, in case that's a concern, and doing it inside of the GleanInternalApi class limits the scope of the extension, so it really just changes it from a statically called function to a non-statically accessed one.

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/JSONObject.kt
travis79: `tryGetLong()` is the only thing used.  I've combined the functions we use into a single file and test file so it will be easier to review once I commit this change.

PR #268, opened 2019-09-09
Custom distribution

badboy:


glean-core/src/histogram/exponential.rs
```diff
@@ -52,7 +53,13 @@ fn exponential_range(min: u64, max: u64, bucket_count: usize) -> Vec<u64> {
 /// and `bucket_count` buckets.
 #[derive(Debug, Clone, Serialize, Deserialize)]
 pub struct PrecomputedExponential {
-    bucket_ranges: Vec<u64>,
+    // Don't serialize the (potentially large) array of ranges, instead compute them on first
+    // access.
+    #[serde(skip)]
+    bucket_ranges: OnceCell<Vec<u64>>,
```
mdboom: TIL about `OnceCell`.  Cool.

glean-core/src/histogram/functional.rs
```diff
@@ -66,7 +70,11 @@ impl Histogram<Functional> {
     }
 
     /// Get a snapshot of all contiguous values.
-    pub fn snapshot_values(&self) -> HashMap<u64, u64> {
+    ///
+    /// **Caution** This is a more specific implemenation of `snapshot_values` on functional
```
mdboom:
```suggestion
    /// **Caution** This is a more specific implementation of `snapshot` on functional
```

glean-core/src/histogram/mod.rs
```diff
@@ -5,14 +5,41 @@
...
 pub use functional::Functional;
+pub use linear::PrecomputedLinear;
 
 mod exponential;
 mod functional;
+mod linear;
+
+/// Different kinds of histograms.
+#[derive(Debug, Clone, Copy, Serialize, Deserialize)]
+#[serde(rename_all = "lowercase")]
```
mdboom: We use to send this information in the payload, but it turned out to be unnecessary.  So, now that this is just an implementation detail, we don't *have to* do the rename since we aren't writing to a schema.  Using strings rather than ints is probably a good idea though for local data migration reasons, though.

badboy: Oh, well, guess we can remove this then completely.

glean-core/src/histogram/functional.rs
badboy: I keep typing the same mistake in ~~implemenation~~ implementation.

`snapshot_values` is correct though. That's the function that the other histograms use. We can't overwrite that, so we have to use a new function name, which is why I also added that comment

PR #270, opened 2019-09-09
Enable iOS CI builds

badboy:


badboy:
I think this is kinda ready now, I would just wait for the 3 iOS prs to go in, rebase it and see that it is green (therefore avoiding reformatting code that goes away anyway for the sake of green builds)

Dexterp37:
> 
> 
> I think this is kinda ready now, I would just wait for the 3 iOS prs to go in, rebase it and see that it is green (therefore avoiding reformatting code that goes away anyway for the sake of green builds)

Makes sense to me.

PR #271, opened 2019-09-09
1579886 - Add iOS lifecycle observer

travis79:

### DIFF
--- a/glean-core/ios/Glean.xcodeproj/project.pbxproj
+++ b/glean-core/ios/Glean.xcodeproj/project.pbxproj
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		1F60589723148BF800307A9F /* GleanLifecycleObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F60589623148BF800307A9F /* GleanLifecycleObserver.swift */; };
 		BF3DE39B2243A2F20018E23F /* Glean.framework in Frameworks */ = {isa = PBXBuildFile; fileRef = BF3DE3912243A2F20018E23F /* Glean.framework */; };
 		BF3DE3A02243A2F20018E23F /* GleanTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = BF3DE39F2243A2F20018E23F /* GleanTests.swift */; };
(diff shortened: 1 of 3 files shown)


glean-core/ios/Glean.xcodeproj/project.pbxproj
```diff
@@ -7,6 +7,7 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		1F60589723148BF800307A9F /* GleanLifecycleObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F60589623148BF800307A9F /* GleanLifecycleObserver.swift */; };
```
badboy: I'm already getting annoyed by the fact that every PR will change the project file now ... (not your fault of course)

glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
```diff
@@ -0,0 +1,31 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+class GleanLifecycleObserver {
```
Dexterp37: nit: should we have docs for this? Will `GleanLifecycleObserver` be part of the API surface?

travis79: I don't think it should show up as part of the public API, according to [documentation](https://docs.swift.org/swift-book/LanguageGuide/AccessControl.html) the default access in Swift is internal. You have to be explicit about public access, so it was not my intent to allow this to leak to the public API.  (Unless you see a need to, that is)

glean-core/ios/Glean.xcodeproj/project.pbxproj
travis79: Yeah, it's annoying to be sure.  It's definitely not as clean as the Android approach to this.

glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
travis79: If you think this should be public, we can file a follow up bug to this.

PR #272, opened 2019-09-09
1579887 - Add Configuration struct

travis79:

### DIFF
--- a/glean-core/ios/Glean/Config/Configuration.swift
+++ b/glean-core/ios/Glean/Config/Configuration.swift
@@ -1,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// The Configuration struct describes how to configure Glean
+///
+/// - parameters:
+///   * serverEndpoint: A `String` representing the server the pings are sent to.  This should only be changed
+///     for tests
+///   * userAgent: the user agent used when sending pings, only to be used internally.
+///   * connectionTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
+///   * readTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
+///   * maxEvents: the number of events to store before the events ping is sent.
+///   * logPings: whether to log ping contents to the console. This is only meant to be used internally by the
+///     `GleanDebugActivity`.
+///   * httpClient: The HTTP client implementation to use for uploading pings.
+///   * pingTag: String tag to be applied to headers when uploading pings for debug view. This is only meant to
+///     be used internally by the `GleanDebugActivity`.
+///   * channel: the release channel the application is on, if known. This will be sent along with all the pings,
+///     in the `client_info` section.
+struct Configuration {
+    let serverEndpoint: String = Configuration.DEFAULT_TELEMETRY_ENDPOINT
+    let userAgent: String = Configuration.DEFAULT_USER_AGENT
+    let connectionTimeout: Int = Configuration.DEFAULT_CONNECTION_TIMEOUT
+    let readTimeout: Int = Configuration.DEFAULT_READ_TIMEOUT
+    let logPings: Bool = Configuration.DEFAULT_LOG_PINGS
+    // let httpClient: HttpPingUploader = HttpPingUploader()
+    let maxEvents: Int? = nil
+    let pingTag: String? = nil
+    let channel: String? = nil
+
+    static let DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
+    static let DEFAULT_USER_AGENT = "Glean/\(getGleanVersion()) (iOS)"
+    static let DEFAULT_CONNECTION_TIMEOUT = 10000
+    static let DEFAULT_READ_TIMEOUT = 30000
+    static let DEFAULT_LOG_PINGS = false
+
+    /// Returns the current Glean version from the Bundle as a `String`
+    ///
+    /// - returns: The `String` representation of the version
+    static func getGleanVersion() -> String {
+        return (Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String) ?? ""
+    }
+}
--- a/glean-core/ios/GleanTests/Config/ConfigurationTests.swift
+++ b/glean-core/ios/GleanTests/Config/ConfigurationTests.swift
@@ -1,0 +1,59 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import XCTest
+@testable import Glean
+
+class ConfigurationTests: XCTestCase {
+    private var config: Configuration?
+
+    override func setUp() {
+        config = Configuration()
+    }
+
+    override func tearDown() {
+        config = nil
+    }
+
+    func testInit() {
+        XCTAssertEqual(
+            config?.serverEndpoint,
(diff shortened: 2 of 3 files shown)


glean-core/ios/GleanTests/Config/ConfigurationTests.swift
```diff
@@ -0,0 +1,59 @@
...
+
+    override func setUp() {
+        config = Configuration()
+    }
+
+    override func tearDown() {
+        config = nil
+    }
+
+    func testInit() {
```
badboy: Hm, these tests seem a bit superfluous. Just checking that something is a default value seems ... unnecessary to me.

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        )
+        XCTAssertEqual(
+            config?.logPings,
+            Configuration.DEFAULT_LOG_PINGS,
+            "Default log pings is set"
+        )
+        // TODO test config?.httpClient
+        XCTAssertNil(
+            config?.maxEvents,
+            "Default UserAgent is set"
```
badboy: Wrong message here

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        )
+        XCTAssertNil(
+            config?.pingTag,
+            "Default UserAgent is set"
```
badboy: Wrong message here

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
```diff
@@ -0,0 +1,59 @@
...
+        )
+        XCTAssertNil(
+            config?.channel,
+            "Default UserAgent is set"
```
badboy: Wrong message here

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,25 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+struct Configuration {
```
Dexterp37: nit: docs :)

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
travis79: I was dusting the rust off of my testing knowledge with iOS, and we actually test for these things on the Android side, which is what I was looking to for inspiration on what to test. I agree that it seems a little silly to test defaults, and I'm happy to nuke this test if we agree there is no need for it.

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
travis79: ARG!  My copy-pasta is undercooked :(

travis79:
> r+wc
> 
> Would be super nice if we end up just passing that object to FFI (though we right now need to transform it anyway because it contains more options than the FFI expects).
> 
> We should be able to make a helper function that spits out the FFI format...  then we could do something like `config.ffi()`, or even make it a computed property and do `config.ffi`.  Let me see if I can add an example of what I'm thinking...

badboy:
> Let me see if I can add an example of what I'm thinking...

I'd also be fine with seeing that implemented when it comes to it :)

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -2,6 +2,18 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/// The Configuration struct describes how to configure Glean
```
badboy: Does swift also has special handling for "doc comments" (3 slashes)?
Does it come with a nice doc renderer?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,40 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// The Configuration struct describes how to configure Glean
+///
+/// - parameters:
```
badboy: Let's go the markdown route and also use the same phrasing as for the Rust code (unless there's guidelines for docs we should adhere to?)

```suggestion
/// # Arguments
```

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,40 @@
...
+///   - serverEndpoint: A `String` representing the server the pings are sent to.  This should only be changed for tests
```
badboy:
```suggestion
///   * `serverEndpoint`: A `String` representing the server the pings are sent to. This should only be changed for tests
```

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,40 @@
...
+///   -userAgent the user agent used when sending pings, only to be used internally.
```
badboy: Spaces after `-` and actually can we use `*` as in the rust code base?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,40 @@
...
+///   -connectionTimeout the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
```
badboy: As above and below

glean-core/ios/GleanTests/Config/ConfigurationTests.swift
badboy: Let's have them for now, and remove when we feel like it.

glean-core/ios/Glean/Config/Configuration.swift
travis79: See [here](https://developer.apple.com/library/archive/documentation/Xcode/Reference/xcode_markup_formatting_ref/index.html) for more info on the markdown style handling I was using for reference.

travis79: Basically it uses standard markdown with a few special thing to denote sections like `- parameters:`, etc...

Instead of the triple slash, we can also use:
```Swift
/** 
... 
*/
```

glean-core/ios/Glean/Config/Configuration.swift
travis79: I think the "Quick Help" looks for the `- parameters:` specifically

glean-core/ios/Glean/Config/Configuration.swift
badboy: triple-slash it is!

glean-core/ios/Glean/Config/Configuration.swift
badboy: Well, then we gotta keep it :)

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -0,0 +1,44 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// The Configuration struct describes how to configure Glean
+///
+/// - parameters:
+///   * serverEndpoint: A `String` representing the server the pings are sent to.  This should only be changed
+///     for tests
+///   * userAgent the user agent used when sending pings, only to be used internally.
```
badboy: Above it has a colon, here it doesn't. One style here would be good, whatever the doc parser parsers right

travis79: Ahh good catch!

PR #273, opened 2019-09-10
1579888 - Add HttpPingUploader

travis79:
**ATTENTION** Requires landing #272 first and rebasing this PR on master after it lands.

### DIFF
--- a/glean-core/ios/Glean.xcodeproj/xcshareddata/xcschemes/Glean.xcscheme
+++ b/glean-core/ios/Glean.xcodeproj/xcshareddata/xcschemes/Glean.xcscheme
@@ -26,6 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      codeCoverageEnabled = "YES"
       shouldUseLaunchSchemeArgsEnv = "YES">
       <Testables>
          <TestableReference
--- a/glean-core/ios/Glean/Config/Configuration.swift
+++ b/glean-core/ios/Glean/Config/Configuration.swift
@@ -2,45 +2,97 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/// The Configuration struct describes how to configure Glean
-///
-/// - parameters:
-///   * serverEndpoint: A `String` representing the server the pings are sent to.  This should only be changed
-///     for tests
-///   * userAgent: the user agent used when sending pings, only to be used internally.
-///   * connectionTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
-///   * readTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
-///   * maxEvents: the number of events to store before the events ping is sent.
-///   * logPings: whether to log ping contents to the console. This is only meant to be used internally by the
-///     `GleanDebugActivity`.
-///   * httpClient: The HTTP client implementation to use for uploading pings.
-///   * pingTag: String tag to be applied to headers when uploading pings for debug view. This is only meant to
-///     be used internally by the `GleanDebugActivity`.
-///   * channel: the release channel the application is on, if known. This will be sent along with all the pings,
-///     in the `client_info` section.
+/// The Configuration struct describes how to configure Glean as well as providing convenience
+/// property for calculating the `FfiConfiguration`
 struct Configuration {
-    let serverEndpoint: String = Configuration.DEFAULT_TELEMETRY_ENDPOINT
-    let userAgent: String = Configuration.DEFAULT_USER_AGENT
-    let connectionTimeout: Int = Configuration.DEFAULT_CONNECTION_TIMEOUT
-    let readTimeout: Int = Configuration.DEFAULT_READ_TIMEOUT
-    let logPings: Bool = Configuration.DEFAULT_LOG_PINGS
-    // let httpClient: HttpPingUploader = HttpPingUploader()
-    let maxEvents: Int? = nil
-    let pingTag: String? = nil
-    let channel: String? = nil
+    let serverEndpoint: String
+    let userAgent: String
+    let connectionTimeout: Int
+    let logPings: Bool
+    let httpClient: HttpPingUploader
+    let maxEvents: Int32?
+    let pingTag: String?
+    let channel: String?
 
     // swiftlint:disable identifier_name
     static let DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
     static let DEFAULT_USER_AGENT = "Glean/\(getGleanVersion()) (iOS)"
     static let DEFAULT_CONNECTION_TIMEOUT = 10000
-    static let DEFAULT_READ_TIMEOUT = 30000
     static let DEFAULT_LOG_PINGS = false
     // swiftlint:enable identifier_name
 
-    /// Returns the current Glean version from the Bundle as a `String`
+    /// This init is for internal testing setup only.
+    ///
+    /// - parameters:
+    ///   * serverEndpoint: A `String` representing the server the pings are sent to.
+    ///     This should only be changed for tests.
+    ///   * userAgent: the user agent used when sending pings, only to be used internally.
+    ///   * connectionTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
+    ///   * logPings: whether to log ping contents to the console.
+    ///     This is only meant to be used internally by the `GleanDebugActivity`.
+    ///   * httpClient: The HTTP client implementation to use for uploading pings.
+    ///   * maxEvents: the number of events to store before the events ping is sent.
+    ///   * pingTag: String tag to be applied to headers when uploading pings for debug view.
+    ///     Used internally by the `GleanDebugActivity`.
+    ///   * channel: the release channel the application is on, if known.
+    ///     This will be sent along with all the pings, in the `client_info` section.
+    internal init(
+        serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
+        userAgent: String = DEFAULT_USER_AGENT,
+        connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
+        logPings: Bool = DEFAULT_LOG_PINGS,
+        httpClient: HttpPingUploader = HttpPingUploader(),
+        maxEvents: Int32? = nil,
+        pingTag: String? = nil,
+        channel: String? = nil
+    ) {
+        self.serverEndpoint = serverEndpoint
+        self.userAgent = userAgent
+        self.connectionTimeout = connectionTimeout
+        self.logPings = logPings
+        self.httpClient = httpClient
+        self.maxEvents = maxEvents
+        self.pingTag = pingTag
+        self.channel = channel
+    }
+
+    /// Create a new Glean `Configuration` object
+    ///
+    /// - parameters:
+    ///   * connectionTimeout the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
+    ///   * maxEvents the number of events to store before the events ping is sent.
+    ///   * httpClient The HTTP client implementation to use for uploading pings.
+    ///   * channel the release channel the application is on, if known.
+    ///     This will be sent along with all the pings, in the `client_info` section.
+    public init(
+        connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
+        maxEvents: Int32? = nil,
+        httpClient: HttpPingUploader = HttpPingUploader(),
+        channel: String? = nil
+    ) {
+        self.serverEndpoint = Configuration.DEFAULT_TELEMETRY_ENDPOINT
+        self.userAgent = Configuration.DEFAULT_USER_AGENT
+        self.connectionTimeout = connectionTimeout
+        self.logPings = Configuration.DEFAULT_LOG_PINGS
+        self.httpClient = httpClient
+        self.maxEvents = maxEvents
+        self.pingTag = nil
+        self.channel = channel
+    }
+
+    /// Returns the current Glean version as a `String`
     ///
     /// - returns: The `String` representation of the version
     static func getGleanVersion() -> String {
-        return (Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String) ?? ""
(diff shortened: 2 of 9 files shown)


glean-core/ios/Glean.xcodeproj/xcshareddata/xcschemes/Glean.xcscheme
```diff
@@ -26,6 +26,7 @@
       buildConfiguration = "Debug"
       selectedDebuggerIdentifier = "Xcode.DebuggerFoundation.Debugger.LLDB"
       selectedLauncherIdentifier = "Xcode.DebuggerFoundation.Launcher.LLDB"
+      codeCoverageEnabled = "YES"
```
badboy: What does this do? (and was it an intentional change?)

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+class HttpPingUploader {
+    let LOG_TAG = "glean/HttpPingUploader"
+
+    func logPing(path: String, data: String, config: Configuration) {
+        if config.logPings {
+            NSLog("\(LOG_TAG) : \(path)\n\(data)")
```
badboy: I will never get used to `\(...)` for string interpolation...

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
...
+        }
+    }
+
+    func upload(path: String, data: String, config: Configuration, callback: @escaping (Bool, Error?) -> Void) {
```
badboy: What's the `@escaping` part for?

badboy: Oh one thing I'd like to learn about Swift (and this is as good a place as anywhere else):

Are arguments always passed by value? Or does Swift auto-reference if needed? Or can you explicitly pass references?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
...
+                }
+            }
+            task.resume()
+        }
+    }
+
+    func buildRequest(path: String, data: String, config: Configuration) -> URLRequest? {
+        if let url = URL(string: config.serverEndpoint + path) {
+            var request = URLRequest(url: url)
+            request.addValue("application/json; charset=utf-8", forHTTPHeaderField: "Content-Type")
```
badboy: `addValue(..., forHTTPHeaderField: ...)` is a weird API o_O

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
...
+                        // Known client (4xx) errors:
+                        // 404 - not found - POST/PUT to an unknown namespace
+                        // 405 - wrong request type (anything other than POST/PUT)
+                        // 411 - missing content-length header
+                        // 413 - request body too large (Note that if we have badly-behaved clients that
+                        //       retry on 4XX, we should send back 202 on body/path too long).
+                        // 414 - request path too long (See above)
+
+                        // Something our client did is not correct. It's unlikely that the client is going
+                        // to recover from this by re-trying again, so we just log and error and report a
```
badboy:
```suggestion
                        // to recover from this by re-trying again, so we just log an error and report a
```

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
...
+                        // successful upload to the service.
+                        callback(true, error)
+                    default:
+                        // Known other errors:
+                        // 500 - internal error
+
+                        // For all other errors we log a warning an try again at a later time.
```
badboy:
```suggestion
                        // For all other errors we log a warning and try again at a later time.
```

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
```diff
@@ -0,0 +1,61 @@
...
+        userAgent: "Glean/Test 25.0.2",
+        connectionTimeout: 3050,
+        readTimeout: 7050,
+        pingTag: "Tag"
+    )
+    
+    func testUploadSuccess() {
+        class MockUploader: HttpPingUploader {
+            override func upload(path: String, data: String, config: Configuration, callback: @escaping (Bool, Error?) -> Void) {
+                super.upload(path: path, data: data, config: config, callback: callback)
```
badboy: Will this try to send an actual HTTP request during testing?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -3,16 +3,54 @@
...
+    let userAgent: String
+    let connectionTimeout: Int
+    let readTimeout: Int
+    let logPings: Bool
+    let httpClient: HttpPingUploader
+    let maxEvents: Int?
+    let pingTag: String?
+    let channel: String?
 
+    internal init(
```
Dexterp37: Can you please add a bit more docs about this? It's ok to copy stuff from the Android implementation, if needed.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,83 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+class HttpPingUploader {
```
Dexterp37: nit: class docs :)

glean-core/ios/Glean.xcodeproj/xcshareddata/xcschemes/Glean.xcscheme
travis79: This allows me to see what coverage we have for our tests, it's a handy feature to make @Dexterp37 happy ;)

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: And I will never get used to `${}` a la Kotlin

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: The `@escaping` marks this closure as escaping. If a closure is passed as an argument to a function and it is invoked after the function returns, the closure is escaping. You can also think of it like the closure argument escapes the function body.  A while back Swift changed the default to noescaping and removed the `@noescaping` so we now have to explicitly mark escaping closures with `@escaping` so the compiler knows you are aware of the implications.

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: Yeah, it's pretty consistent across Swift to have an 'unnamed' first parameter and named parameters after that.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: Not yet, I wasn't quite done with this one and needed to research how to set up a mock HTTP server like we do in Android.

glean-core/ios/Glean/Config/Configuration.swift
travis79: Will do

glean-core/ios/Glean.xcodeproj/xcshareddata/xcschemes/Glean.xcscheme
badboy: We should see to extract that on CI at some point :)

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: > Oh one thing I'd like to learn about Swift (and this is as good a place as anywhere else):
> 
> Are arguments always passed by value? Or does Swift auto-reference if needed? Or can you explicitly pass references?

Most cases, things are passed by value, including `struct` and `enum`. But **not** `class`.  Class is always treated as a reference type in Swift and is passed by reference.  If you want to force the pass by reference to a value type, you can use the `inout` keyword in the parameter list.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: I've now updated this to use `OHHTTPStub` similar to `/mozilla-mobile/telemetry-ios` to better test the server responses.  This also prompted me to add a carthage Cartfile to import the dependency as well as some project settings changes to hook in the `carthage` command.  

I also had to add a `.swiftlint.yml` file in order to exclude the output director for the `OHHTTPStub` framework so their linting issues won't affect us.

travis79: It looks like `telemetry-ios` uses a cool trick to only import the `OHHTTPStub` framework on tests, I'll have to ask Garvan about that.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -0,0 +1,121 @@
...
+    ///     * path: The URL path to append to the server address
+    ///     * data: The serialized text data to send
+    ///     * config: The Glean `Configuration` object
+    func logPing(path: String, data: String, config: Configuration) {
+        if config.logPings {
+            NSLog("\(Constants.logTag) : \(path)\n\(data)")
+        }
+    }
+
+    /// Synchronously upload a ping to Mozilla servers.
```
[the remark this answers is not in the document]
badboy: We do. Well, let's change that later.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        connectionTimeout: 3050,
+        logPings: true,
+        pingTag: "Tag"
+    )
+
+    private func setupHttpResponseStub(statusCode: Int32 = 200) {
+        stub(condition: isHost("incoming.telemetry.mozilla.org")) { data in
+            let body = (data as NSURLRequest).ohhttpStubs_HTTPBody()
+            let str = String(data: body!, encoding: .utf8) ?? ""
+            print("STUB RECEIVED: \(str) \n")
```
badboy: Stray debug print?

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
```diff
@@ -0,0 +1,167 @@
...
+            // swiftlint:disable force_try
+            let json = try! JSONSerialization.jsonObject(with: body!, options: []) as? [String: Any]
+            XCTAssert(json != nil)
+            XCTAssertEqual(json?["ping"] as? String, "test")
+
+            return OHHTTPStubsResponse(
+                jsonObject: ["foo": "bar"],
```
badboy: Why do we return a `foo: bar` here?

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
```diff
@@ -0,0 +1,167 @@
...
+        setupHttpResponseStub(statusCode: 200)
+
+        expectation = expectation(description: "Completed upload")
+
+        let httpPingUploader = HttpPingUploader()
+        httpPingUploader.upload(path: testPath, data: testPing, config: testConfig) { success, _ in
+            testValue = success
+            self.expectation?.fulfill()
+        }
+        waitForExpectations(timeout: 10.0) { error in
```
badboy: That's a rather long timeout. Is it necessary?

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
```diff
@@ -0,0 +1,167 @@
...
+            if error != nil {
+                print("Test timed out waiting for upload: \(error!)")
```
badboy: Can't we just assert with a message here?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -2,43 +2,93 @@
...
-        return (Bundle.main.infoDictionary?["CFBundleShortVersionString"] as? String) ?? ""
+        return "\(GleanVersionNumber)"
+    }
+
+    /// Helper function to retrive the application's Documents directory for persistent file storage
+    ///
+    /// - returns: `String` representation of the path to the Documents directory
+    static func getDocumentsDirectory() -> String {
+        let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+        let documentsDirectory = paths[0]
```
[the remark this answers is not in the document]
travis79: Yes, it's the standard way to retrieve a writable, persistent storage directory for the application.  I've been hoping for a better API for this for some time.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: Ooops, good catch!

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: Because it looks like the only functions that allow for this require either a file or data or a JSON object to be returned.  This is something I borrowed from the [telemetry-ios](https://github.com/mozilla-mobile/telemetry-ios/blob/d092de49d715d0bd1c6fad4bd89c85e51172d444/TelemetryTests/TelemetryTests.swift#L71) library.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: Sure, that's a good idea.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
mdboom: Can we just return an empty JSON object, i.e. `{}`?

travis79: @mdboom yes, that's probably the easiest solution since we aren't using the response body for anything.

glean-core/ios/GleanTests/HttpPingUploaderTests.swift
travis79: looks like 2 seconds is plenty locally

Dexterp37: Since we're using timeout constants to please CI elsewhere in the Glean SDK (Android..), what about using the same value everywhere? I think we went for 5s or so

PR #278, opened 2019-09-12
1580609 - Add iOS Dispatchers

travis79:

### DIFF
--- a/glean-core/ios/Glean/Dispatchers.swift
+++ b/glean-core/ios/Glean/Dispatchers.swift
@@ -1,0 +1,168 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+/// This class manages background execution of Glean tasks.
+///
+/// This class makes use of the higher level `Operation` and `OperationQueue` API's in order to allow
+/// for observable background operation with the capabilities to pause, cancel, and resume tasks
+class Dispatchers {
+    /// This is the shared singleton access to the Glean Dispatchers
+    static let shared = Dispatchers()
+
+    // This struct is used for organizational purposes to keep the class constants in a single place
+    struct Constants {
+        static let logTag = "glean/Dispatchers"
+        static let maxQueueSize = 100
+    }
+
+    // This is the task queue that all Glean background operations will be executed on.  It is currently
+    // set to be a serial queue by setting the `maxConcurrentOperationsCount` to 1, but could allow for
+    // concurrent operations by increasing the parameter.
+    lazy var operationQueue: OperationQueue = {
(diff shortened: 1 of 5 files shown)


glean-core/ios/GleanTests/DispatchersTest.swift
```diff
@@ -0,0 +1,109 @@
...
+//  GleanTests
+//
+//  Created by Travis Long on 9/12/19.
+//  Copyright © 2019 Jan-Erik Rediger. All rights reserved.
+//
+
+@testable import Glean
+import XCTest
+
+class DispatchersTest: XCTestCase {
```
Dexterp37: Let's make extract sure this isn't flaky :) Unleash your testing obscure power :-D

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -0,0 +1,166 @@
...
+    /// Dispatchers.shared.launch {
+    ///     // Do some work here
+    /// }.completionBlock = {
+    ///     print("Done!")
+    /// }
+    /// ```
+    ///
+    /// You could also use the returned `Operation` to await that specific task like this:
+    /// ```
+    /// let operation = Dispatchers.shared.launch {
```
badboy: Wait, just assigning it to a variable automatically awaits the job to be finished?

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -0,0 +1,166 @@
...
+                // If we are in testing mode, go ahead and execute the operation and wait until it
+                // is finished before continuing to give us synchronous execution of the task.
+                operation.start()
+                operation.waitUntilFinished()
+            } else {
+                operationQueue.addOperation(operation)
+            }
+        }
+
+        return operation
```
badboy: (Not critical to landing this, this is internal API so we control all execution paths)

To answer my own question from above to await something returned from `launch` you need to call `waitUntilFinished`.
But what if I call `start` on the thing that is returned? Is this returned operation a clone of what's in the queue and thus things get executed twice? Or is it the same object and the task in the queue will not be executed again later?

glean-core/ios/GleanTests/DispatchersTest.swift
travis79: I'm pretty confident in this one, there are better ways of controlling and awaiting the async bits, but I'll try and put it through as much testing as possible.

glean-core/ios/Glean/Dispatchers.swift
travis79: You shouldn't be able to call `start()` on a `finished` operation, if I understand how the `Operation` class works. It needs to be in the `isReady` state.

badboy: What happens then? An exception?

travis79: I will look it up again to be sure, but I think it's a no-op.

PR #281, opened 2019-09-12
1579889 - Add ping upload scheduling for `baseline` and `events` pings

travis79:
Requires #278 to land first and rebase on master after that.

### DIFF
--- a/glean-core/ios/Glean/Config/Configuration.swift
+++ b/glean-core/ios/Glean/Config/Configuration.swift
@@ -9,7 +9,6 @@ public struct Configuration {
     let userAgent: String
     let connectionTimeout: Int
     let logPings: Bool
-    let httpClient: HttpPingUploader
     let maxEvents: Int32?
     let pingTag: String?
     let channel: String?
@@ -41,7 +40,6 @@ public struct Configuration {
         userAgent: String = DEFAULT_USER_AGENT,
         connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
         logPings: Bool = DEFAULT_LOG_PINGS,
-        httpClient: HttpPingUploader = HttpPingUploader(),
         maxEvents: Int32? = nil,
         pingTag: String? = nil,
         channel: String? = nil
@@ -50,7 +48,6 @@ public struct Configuration {
         self.userAgent = userAgent
         self.connectionTimeout = connectionTimeout
         self.logPings = logPings
-        self.httpClient = httpClient
         self.maxEvents = maxEvents
         self.pingTag = pingTag
         self.channel = channel
@@ -67,14 +64,12 @@ public struct Configuration {
     public init(
         connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
         maxEvents: Int32? = nil,
-        httpClient: HttpPingUploader = HttpPingUploader(),
         channel: String? = nil
     ) {
         self.serverEndpoint = Configuration.DEFAULT_TELEMETRY_ENDPOINT
         self.userAgent = Configuration.DEFAULT_USER_AGENT
         self.connectionTimeout = connectionTimeout
         self.logPings = Configuration.DEFAULT_LOG_PINGS
-        self.httpClient = httpClient
         self.maxEvents = maxEvents
         self.pingTag = nil
         self.channel = channel
--- a/glean-core/ios/Glean/Config/Configuration.swift
+++ b/glean-core/ios/Glean/Config/Configuration.swift
@@ -7,18 +7,16 @@
 public struct Configuration {
     let serverEndpoint: String
     let userAgent: String
-    let connectionTimeout: Int
     let logPings: Bool
     let maxEvents: Int32?
     let pingTag: String?
     let channel: String?
 
-    // swiftlint:disable identifier_name
-    static let DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
-    static let DEFAULT_USER_AGENT = "Glean/\(getGleanVersion()) (iOS)"
-    public static let DEFAULT_CONNECTION_TIMEOUT = 10000
-    static let DEFAULT_LOG_PINGS = false
-    // swiftlint:enable identifier_name
+    struct Constants {
+        static let defaultTelemetryEndpoint = "https://incoming.telemetry.mozilla.org"
+        static let defaultUserAgent = "Glean/\(getGleanVersion()) (iOS)"
+        static let defaultLogPings = false
+    }
 
     /// This init is for internal testing setup only.
     ///
@@ -26,27 +24,23 @@ public struct Configuration {
     ///   * serverEndpoint: A `String` representing the server the pings are sent to.
     ///     This should only be changed for tests.
     ///   * userAgent: the user agent used when sending pings, only to be used internally.
-    ///   * connectionTimeout: the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
     ///   * logPings: whether to log ping contents to the console.
     ///     This is only meant to be used internally by the `GleanDebugActivity`.
-    ///   * httpClient: The HTTP client implementation to use for uploading pings.
     ///   * maxEvents: the number of events to store before the events ping is sent.
     ///   * pingTag: String tag to be applied to headers when uploading pings for debug view.
     ///     Used internally by the `GleanDebugActivity`.
     ///   * channel: the release channel the application is on, if known.
     ///     This will be sent along with all the pings, in the `client_info` section.
     internal init(
-        serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
-        userAgent: String = DEFAULT_USER_AGENT,
-        connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
-        logPings: Bool = DEFAULT_LOG_PINGS,
+        serverEndpoint: String = Constants.defaultTelemetryEndpoint,
+        userAgent: String = Constants.defaultUserAgent,
+        logPings: Bool = Constants.defaultLogPings,
         maxEvents: Int32? = nil,
         pingTag: String? = nil,
         channel: String? = nil
     ) {
         self.serverEndpoint = serverEndpoint
         self.userAgent = userAgent
-        self.connectionTimeout = connectionTimeout
         self.logPings = logPings
         self.maxEvents = maxEvents
         self.pingTag = pingTag
@@ -56,20 +50,16 @@ public struct Configuration {
     /// Create a new Glean `Configuration` object
     ///
     /// - parameters:
-    ///   * connectionTimeout the timeout, in milliseconds, to use when connecting to the `serverEndpoint`.
     ///   * maxEvents the number of events to store before the events ping is sent.
-    ///   * httpClient The HTTP client implementation to use for uploading pings.
     ///   * channel the release channel the application is on, if known.
     ///     This will be sent along with all the pings, in the `client_info` section.
     public init(
-        connectionTimeout: Int = DEFAULT_CONNECTION_TIMEOUT,
         maxEvents: Int32? = nil,
         channel: String? = nil
     ) {
-        self.serverEndpoint = Configuration.DEFAULT_TELEMETRY_ENDPOINT
-        self.userAgent = Configuration.DEFAULT_USER_AGENT
-        self.connectionTimeout = connectionTimeout
-        self.logPings = Configuration.DEFAULT_LOG_PINGS
+        self.serverEndpoint = Constants.defaultTelemetryEndpoint
+        self.userAgent = Constants.defaultUserAgent
+        self.logPings = Constants.defaultLogPings
         self.maxEvents = maxEvents
         self.pingTag = nil
         self.channel = channel
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -113,20 +113,17 @@ public class Glean {
 
     /// Handle background event and send appropriate pings
     func handleBackgroundEvent() {
-        // Perform task on a background thread
-        _ = Dispatchers.shared.launch {
-            Glean.backgroundTaskID = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
-                // End the task if time expires
-                UIApplication.shared.endBackgroundTask(Glean.backgroundTaskID)
-                Glean.backgroundTaskID = UIBackgroundTaskIdentifier.invalid
-            }
-
-            self.sendPingsByName(pingNames: ["baseline", "events"])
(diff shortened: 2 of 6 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -105,7 +111,47 @@ public class Glean {
...
-        // sendPings()
+        // Perform task on a background thread
+        _ = Dispatchers.shared.launch {
+            Glean.backgroundTaskID = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
+                // End the task if time expires
+                UIApplication.shared.endBackgroundTask(Glean.backgroundTaskID)
+                Glean.backgroundTaskID = UIBackgroundTaskIdentifier.invalid
+            }
+
+            self.sendPingsByName(pingNames: ["baseline", "events"])
```
Dexterp37: Why are you adding `events`? We don't have any event metric type. May be clearer to have this later on, once the rest is added

glean-core/ios/Glean/Glean.swift
```diff
@@ -21,8 +21,14 @@ public class Glean {
     var handle: UInt64 = 0
     private var initialized: Bool = false
     private var uploadEnabled: Bool = true
-    private var configuration: Configuration?
+    var configuration: Configuration?
```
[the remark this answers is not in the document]
badboy: Default is "internal". Further below `Glean.shared.configuration` is used.

glean-core/ios/Glean/Glean.swift
```diff
@@ -105,7 +111,47 @@ public class Glean {
 
     /// Handle background event and send appropriate pings
     func handleBackgroundEvent() {
-        // sendPings()
+        // Perform task on a background thread
+        _ = Dispatchers.shared.launch {
```
badboy: Why the dispatch here? `sendPingsByName` is already dispatched.
As I understand the `beginBackgroundTask` will make sure that the task stays around if the application is closed.

but the double-dispatch makes me uncomfortable. Would we do this for pretty much any `sendPing` triggered? e.g. also by those triggered on custom pings?

glean-core/ios/Glean/Glean.swift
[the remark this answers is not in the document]
Dexterp37: Ah darn, different language different default, of course. Too many languages :P

glean-core/ios/Glean/Glean.swift
badboy: Fun fact: if this stays the _only_ thing we need to enable events for users is the Swift wrapper.

All logic determining if events are there and need to be sent is in Rust already, so it will just start to work.

Therefore I'd be fine having it here already.

Dexterp37: Good point. Fine with me!

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -8,7 +8,10 @@
 public class HttpPingUploader {
```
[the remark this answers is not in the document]
travis79: I'm totally in agreement for this!  There aren't as many network API implementations to deal with as Android so it's unlikely that anyone would use anything other than `URLSession`.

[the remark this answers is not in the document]
travis79: Bug filed [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1581102)

glean-core/ios/Glean/Glean.swift
travis79: I read this as "Glean, it just works!"

glean-core/ios/Glean/Glean.swift
travis79: You are right, it probably doesn't need this and could just be executed at this point.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -6,13 +6,18 @@
...
     struct Constants {
+        // Since ping file names are UUIDs, this matches UUIDs for filtering purposes
+        static let filePattern = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
         static let logTag = "glean/HttpPingUploader"
+        static let pingsDir = "pings"
     }
 
-    public init() {
-        // intentionally left empty to allow public instantiation
+    public init(configuration: Configuration) {
```
badboy: Do we need the full configuration object?
Probably makes things easier to use to just pass it in fully, but what about only storing the things in this struct that we actually use?

glean-core/ios/Glean/Glean.swift
```diff
@@ -108,7 +114,46 @@ public class Glean {
...
+        Glean.backgroundTaskId = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
+            // End the task if time expires
+            UIApplication.shared.endBackgroundTask(Glean.backgroundTaskId)
+            Glean.backgroundTaskId = UIBackgroundTaskIdentifier.invalid
+        }
+
+        self.sendPingsByName(pingNames: ["baseline", "events"])
+
+        // End task assertion
+        UIApplication.shared.endBackgroundTask(Glean.backgroundTaskId)
```
badboy: As I understand it marking something with `beginBackgroundTask` ensures the app isn't killed right away before `endBackgroundTask` happens.
With `sendPingsByName` dispatching to the thread pool, it will return immeidately, likely before the task even started executing.
`endBackgroundTask` is then called while the Rust backend might just be reading data and writing the pings.

Shouldn't the begin/end task wrapping happen inside `sendPingsByName` inside the threadpool-launched task?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -21,10 +26,8 @@ public class HttpPingUploader {
     ///     * path: The URL path to append to the server address
     ///     * data: The serialized text data to send
     ///     * config: The Glean `Configuration` object
```
badboy: Doc needs update.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -122,4 +127,91 @@ public class HttpPingUploader {
...
+        let pingDirectory = HttpPingUploader.getOrCreatePingDirectory()
+
+        do {
+            let storageDirectory = try FileManager.default.contentsOfDirectory(
+                atPath: pingDirectory.absoluteString
+            )
+
+            for file in storageDirectory {
+                if file.lastPathComponent.matches(Constants.filePattern) {
+                    NSLog("\(Constants.logTag) : Processing ping: \(file)")
```
badboy: Can't we rename the `logPing` method from above and reuse it so we don't have to add the `\(Constants.logTag\` in all the strings?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -122,4 +127,91 @@ public class HttpPingUploader {
...
+                            NSLog(
+                                "\(Constants.logTag) : Error processing ping file: \(file) - \(error.debugDescription)"
+                            )
+                        } else {
+                            do {
+                                try FileManager.default.removeItem(
+                                    atPath: pingDirectory.appendingPathComponent(file).absoluteString
+                                )
+                            } catch {
+                                print("\(Constants.logTag) : Error deleting ping file: \(file)")
```
badboy: Oh a print. Guess we can all unify that into the same log call later.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -122,4 +127,91 @@ public class HttpPingUploader {
...
+                            }
+                        }
+                    }
+                } else {
+                    // Delete files that don't match the UUID filePattern regex
+                    NSLog("\(Constants.logTag) : Pattern mismatch. Deleting \(file)")
+                    try FileManager.default.removeItem(
```
badboy: Will `try` just explode if stuff fails?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -122,4 +127,91 @@ public class HttpPingUploader {
...
+                        atPath: pingDirectory.appendingPathComponent(file).absoluteString
+                    )
+                }
+            }
+        } catch {
+            NSLog("\(Constants.logTag) : Error while enumerating files in ping directory")
```
badboy: Never mind, I guess this catches all the exceptions from above's `try` calls

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -122,4 +127,91 @@ public class HttpPingUploader {
...
+
+    /// This function encapsulates processing of a single ping file
+    ///
+    /// - parameters:
+    ///   * file: The `URL` of the file to process
+    ///   * callback: Allows for an action to occur as the result of the async upload operation
+    func processFile(_ file: URL, callback: @escaping (Bool, Error?) -> Void) {
+        do {
+            let data = try String(contentsOfFile: file.absoluteString, encoding: .utf8)
+            // let data = try String(contentsOf: file, encoding: .utf8)
```
badboy: Left-over comment.

glean-core/ios/Glean/Utils.swift
```diff
@@ -11,13 +11,34 @@ extension String: Error {
     public var errorDescription: String? { return self }
 }
 
+/// Checks to see if a string matches a regex
```
badboy: This doc string should be moved down to the `matches` function and the other newly introduced `var`s should be documented as well.

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -9,7 +9,6 @@ public struct Configuration {
     let userAgent: String
     let connectionTimeout: Int
```
Dexterp37: We have now removed this from the Kotlin implementation. Both the connection and the read timeouts are implementation details of the uploader class: I'd recommend removing it from here as well (and from the docs).

glean-core/ios/Glean/Glean.swift
```diff
@@ -108,7 +114,46 @@ public class Glean {
...
+        Glean.backgroundTaskId = UIBackgroundTaskIdentifier.invalid
+    }
+
+    func sendPingsByName(pingNames: [String]) {
```
Dexterp37: nit: can we have a docstring here?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -21,10 +26,8 @@ public class HttpPingUploader {
     ///     * path: The URL path to append to the server address
     ///     * data: The serialized text data to send
     ///     * config: The Glean `Configuration` object
-    func logPing(path: String, data: String, config: Configuration) {
-        if config.logPings {
-            NSLog("\(Constants.logTag) : \(path)\n\(data)")
-        }
+    func logPing(path: String, data: String) {
+        NSLog("\(Constants.logTag) : \(path)\n\(data)")
```
Dexterp37: We're printing this message slightly differently compared to the Android implementation: we are printing a JSON indented version of the string, not the raw string there. Can you file a bug for looking up how to achieve the same result (or if it even make sense to do so, given how logs are consumed on iOS?)?

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: I think we are down to just simple types in the Configuration struct now, so along with the timeout, I'll see if there are any other things in the configuration struct that could be removed.

glean-core/ios/Glean/Glean.swift
travis79: Yes, it should...  fixing this

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: The logPing method was meant to work with the GleanDebugActivity.  I would love to have a unified logging class that would hold the tag in the instance and format the output uniformly 😄, should I file a bug?

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: I think if we made a GleanLogger class, we could then have log levels like debug, error, etc. and could wrap that around whether we want to call `print()` or `NSLog()`

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: Rather than file a bug, I think we can address this as part of work for [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1580856).

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: I think we can address this as part of work for [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1580856).

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: no the catch below handles all of the try's for the FileManager stuff in this do/catch block.

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: I'm going to drop a comment in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1580856) regarding this.  I think that the result of that bug should give us a logging class that we can use on iOS to achieve closer results to Android loggging.

glean-core/ios/Glean/Glean.swift
```diff
@@ -114,19 +114,28 @@ public class Glean {
 
     /// Handle background event and send appropriate pings
     func handleBackgroundEvent() {
+        // This asks the OS for more time when going to background to perform critical tasks.
+        // This will be signaled complete when `sendPingsByName` is done.
         Glean.backgroundTaskId = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
```
badboy: What happens when the app is moved to the background twice while the task is already running?

travis79: According to the [docs](https://developer.apple.com/documentation/uikit/uiapplication/1623031-beginbackgroundtask) it's safe to call this multiple times as long as each execution has a matching call to the endBackgroundTask.  This can be called at any time, including when not yet going to background.  

So, to answer your question, it looks like we would have the background tasks executed twice, both of them requesting additional time to complete (if needed).  I think that this is the desired behavior we are looking for as each backgrounding event should launch the required pings.

PR #282, opened 2019-09-13
Remove concept fetch.

Dexterp37:
This is a big PR, sorry about that. I recommend reviewing it commit by commit, I tried to break it down to make it more navigable.

Highlights:

- This is the port of mozilla-mobile/android-components#4381 to glean-core
- This restores the legacy Android HTTP uploader that we used to have in early Glean SDK days (see [this](https://github.com/mozilla-mobile/android-components/blob/9b37af821f70b2b053a9f0ab3a787d2d4b8991be/components/service/glean/src/main/java/mozilla/components/service/glean/net/HttpPingUploader.kt)). This was only slightly tweaked to use the passed full `url` instead of just the server endpoint. Test coverage was ported as well.
- It moves ping file processing functions from `PingUploader` to `PingUploadWorker` (what!? We don't have any test coverage for these functions!??!)
- The rest is merely importing other changes from a-c to here: the configuration changes for using the BaseUploader, etc.


Dexterp37:
I'm really surprised by the `MatchingDeclarationName` detekt error. I've not removed a `TestUtils` class :\

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
```diff
@@ -0,0 +1,112 @@
...
+
+    /**
+     * Log the contents of a ping to the console, if configured to do so in
+     * [Configuration.logPings].
+     *
+     * @param path the URL path to append to the server address
+     * @param data the serialized text data to send
+     * @param config the Glean configuration object
+     */
+    private fun logPing(path: String, data: String, config: Configuration) {
```
travis79: nit: Passing the whole configuration has always seemed like overkill here when all we need is the boolean.  In fact, maybe the check to log should be in the doUpload function and wrapped around the call to `logPing` and then this function could do exactly what it says it does.

Dexterp37: This is a great idea! I just changed that.

PR #283, opened 2019-09-13
Add CI check for GleanFfi.h consistency

badboy:


Makefile
```diff
@@ -50,6 +50,7 @@ test-ios:
 cbindgen:
 	RUSTUP_TOOLCHAIN=nightly \
 	cbindgen glean-core/ffi --lockfile Cargo.lock -o glean-core/ffi/examples/glean.h
+	cp glean-core/ffi/examples/glean.h glean-core/ios/Glean/GleanFfi.h
```
Dexterp37: nit: having the header in the `example` directory looks a bit weird (especially since we then copy it to a non example FFI location ;) ). What if we have some other location. e.g. `c/glean.h`?

badboy: We need to have it in the ios/Glean directory to be included in the iOS build.

Dexterp37: Let me clarify: I'm suggesting to keep the copying part, just renaming the directory we're copying from.

badboy: Ah! What do you think of placing the canonical copy in glean-core/ffi directly?

Dexterp37: Sounds good to me!

PR #285, opened 2019-09-13
On iOS enable logging with a level filter.

badboy:
On debug builds enable debug logging.
On release builds enable info logging.

### DIFF
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -54,11 +54,18 @@ pub extern "C" fn glean_enable_logging() {
         });
     }
 
-    // On iOS always enable Debug logging
+    // On iOS enable logging with a level filter.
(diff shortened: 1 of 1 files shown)


glean-core/ffi/src/lib.rs
```diff
@@ -54,11 +54,18 @@ pub extern "C" fn glean_enable_logging() {
         });
     }
 
-    // On iOS always enable Debug logging
+    // On iOS enable logging with a level filter.
     #[cfg(target_os = "ios")]
     {
+        // Debug logging in debug mode.
+        // (Note: `debug_assertions` is the next best thing to determine if this is a debug build)
+        #[cfg(debug_assertions)]
```
Dexterp37: Ha! Nice trick

badboy: more of a hack

PR #288, opened 2019-09-13
Re-initialize core metrics after toggling upload enabled

badboy:


glean-core/ios/Glean/Glean.swift
```diff
@@ -85,10 +95,29 @@ public class Glean {
...
     ///     * enabled: When true, enable metric collection.
     public func setUploadEnabled(_ enabled: Bool) {
-        uploadEnabled = enabled
-
         if isInitialized() {
-            glean_set_upload_enabled(handle, enabled ? 1 : 0)
+            let originalEnabled = getUploadEnabled()
+
+            _ = Dispatchers.shared.launch {
+                // TODO: Synchronize access around ping queue lock
```
badboy: On the Kotlin side we [synchronize around a shared lock for the reasons listed below](https://github.com/mozilla/glean/blob/10c682db9c610742220135061bd95572fcec74e0/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L206).

@travis79, can you help me determine if we need that here and if so, how to do it?

travis79: One way to do this would be to use a similar method to what I did with the `AtomicBoolean` in `Utils.swift`.  Whether or not this needs to be done, I do not know for sure but I would say yes, we need to synchronize access to the files so that we don't delete while the uploader is processing files.

travis79: Another thing we could do is implement a way to cancel the uploader task, which wouldn't be too difficult with the `Operation` API.  Create an instance of a `GleanOperation` that wraps it and save a reference to that in order to be able to see if it's currently running and cancel it.

badboy: Wait, we don't actually need it.
The ping uploader runs on the same Dispatcher queue and not as a completely separate task (as it does on Android).

So we have a guarantee that this new task will only run after a ping upload is finished (if it is already running or ahead of this in the queue)

Dexterp37: > 
> 
> Wait, we don't actually need it.
> The ping uploader runs on the same Dispatcher queue and not as a completely separate task (as it does on Android).

But do we really want it to be that way? If we do that, dispatching telemetry might be dependant on network when uploading.

badboy: That's the way it currently is. If we want to change that then we need to change how it interacts with the general system, yes. But it doesn't make sense to introduce any locking in here right now before we have that other mechanism.

Dexterp37: Works for me, but please do file a bug and leave a comment in the code.

badboy: 👍

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1581783

glean-core/ios/Glean/Glean.swift
```diff
@@ -84,10 +86,18 @@ public class Glean {
...
+        glean_on_ready_to_send_pings(self.handle)
+
         // Signal Dispatcher that init is complete
         Dispatchers.shared.flushQueuedInitialTasks()
+    }
 
-        self.initialized = true
+    /// Initialize the core metrics internally managed by Glean (e.g. client id).
+    private func initializeCoreMetrics() {
+        /// left empty for now, we don't have additional core metrics
```
Dexterp37: What's needed here? Is there a bug for implementing these? Can you reference it here?

glean-core/ios/Glean/Glean.swift
```diff
@@ -104,10 +114,29 @@ public class Glean {
...
-        uploadEnabled = enabled
-
         if isInitialized() {
-            glean_set_upload_enabled(handle, enabled.toByte())
+            let originalEnabled = getUploadEnabled()
+
+            _ = Dispatchers.shared.launch {
+                // glean_set_upload_enabled might delete all of the queued pings.
+                // Currently a ping uploader could be scheduled ahead of this,
+                // at which point it wil pick up scheduled pings before the setting was toggled.
```
mdboom:
```suggestion
                // at which point it will pick up scheduled pings before the setting was toggled.
```

glean-core/ios/Glean/Glean.swift
```diff
@@ -104,10 +114,29 @@ public class Glean {
     /// - parameters:
     ///     * enabled: When true, enable metric collection.
     public func setUploadEnabled(_ enabled: Bool) {
-        uploadEnabled = enabled
-
         if isInitialized() {
-            glean_set_upload_enabled(handle, enabled.toByte())
+            let originalEnabled = getUploadEnabled()
+
+            _ = Dispatchers.shared.launch {
```
travis79: Is this something that we want to execute on the serial queue?  If we don't want to worry about what else is in the Glean Dispatchers, you can also use Grand Central Dispatch to execute this on the fly on a background thread like this:
```Swift
DispatchQueue.global().async {
	// Background work
}
```

badboy: Right now, yes, as it then runs for sure after any uploader.
Once the other code lands we should re-evaluate where this runs?

PR #289, opened 2019-09-16
Register and send pings

badboy:
Based on #281, needs that to be merged first, then rebased.


glean-core/ios/Glean/Glean.swift
```diff
@@ -71,6 +75,15 @@ public class Glean {
...
         }
 
+        // If any pings were registered before initializing, do so now
+        for ping in self.pingTypeQueue {
+            self.registerPingType(ping)
+        }
+
+        if !Dispatchers.shared.testingMode {
+            self.pingTypeQueue.removeAll()
+        }
```
mdboom: Do we need any sort of synchronization here?  So we don't add and flush ping types at the same time...  (Not saying we do, just want to confirm...)

glean-core/ios/Glean/Metrics/Ping.swift
```diff
@@ -0,0 +1,43 @@
...
+public class Ping {
+    var handle: UInt64
+    let name: String
+    let includeClientId: Bool
+
+    /// The public constructor used by automatically generated metrics.
+    public init(name: String, includeClientId: Bool) {
+        self.name = name
+        self.includeClientId = includeClientId
+        self.handle = glean_new_ping_type(name, includeClientId ? 1 : 0)
```
travis79: Didn't you add an extension for this..   something like `getByte` or `toByte`?  I thought I saw that somewhere but now I don't see it.  It seems like that might be a good extension for readability if we have to do much Bool -> UInt8 conversion.

glean-core/ios/Glean/Glean.swift
```diff
@@ -29,6 +29,8 @@ public class Glean {
         static let logTag = "glean/Glean"
     }
 
+    private var pingTypeQueue: [Ping] = []
```
travis79: This is a big nit and definitely could just be biased towards a particular style, but I typically use the initializer syntax for arrays like this:
```
private var pingTypeQueue = [Ping]()
```

Your way is perfectly valid though. So again, just a style nit.

glean-core/ios/GleanTests/Metrics/PingTests.swift
```diff
@@ -0,0 +1,130 @@
...
+
+    override func setUp() {
+        lastPingJson = nil
+        expectation = nil
+
+        Glean.shared.resetGlean(clearStores: true)
+    }
+
+    private func setupHttpResponseStub(statusCode: Int32 = 200) {
+        stub(condition: isHost("incoming.telemetry.mozilla.org")) { data in
```
travis79: I'm guilty of leaving this in my test code too, but it occurred to me when looking at this that if the telemetry endpoint changed in Glean, it would break this test.  Maybe relying on the constant in `Configuration` rather than a hardcoded endpoint would be safer.

glean-core/ios/Glean/Glean.swift
badboy: We'll verify in a follow-up

glean-core/ios/Glean/Metrics/Ping.swift
badboy: I ... actually might have done that. Fixing in a followup

glean-core/ios/Glean/Glean.swift
badboy: Hm, Let's use it.

glean-core/ios/GleanTests/Metrics/PingTests.swift
badboy: `Configuration.Constants.defaultTelemetryEndpoint` contains the `https://`, making this not easily work with `isHost`.

I'll take it to a follow-up

PR #291, opened 2019-09-16
1580856 - Add iOS logging utility to make logging from Swift eas…

travis79:


travis79:
I attempted to match the format and functionality that I found we used from the Android logging.

glean-core/ios/Glean/Glean.swift
```diff
@@ -29,6 +29,8 @@ public class Glean {
         static let logTag = "glean/Glean"
     }
 
+    private let logger = Logger(tag: Constants.logTag)
```
badboy: With just that one constant left, maybe it's enough to just inline it?

glean-core/ios/Glean/Utils/Logger.swift
```diff
@@ -0,0 +1,79 @@
...
+        log(message: message, level: .error)
+    }
+
+    /// Private function that formats and outputs the log message based on the level
+    ///
+    /// - parameters:
+    ///     * message: The message to log
+    ///     * level: The `LogLevel` at which to output the message
+    private func log(message: String, level: LogLevel) {
+        let formattedMessage = "\(level.rawValue)/\(tag): \(message)"
```
badboy: TIL: enum's `rawValue`

badboy:
Heads up: I landed #293, so this needs to be rebased (I have the feeling we lost `toByte` before due to a bad rebase, so just making you aware)

glean-core/ios/Glean/Glean.swift
travis79: I'm okay with that, I think I was mainly trying to keep up appearances.

glean-core/ios/Glean/Utils/Logger.swift
travis79: Yeah, `enum`s are overpowered in Swift, the raw type can be something other than a string as well.

glean-core/ios/Glean/Glean.swift
badboy: Hm, let's keep it consistent then and keep the constants. It's all gonna be compiled away

PR #295, opened 2019-09-17
Generate files at build time

badboy:
For context:

At the moment this uses a minimal bash script, that's essentially just a wrapper around the python part (which has the same inline python script as the gradle script.
This is not intended to be the final version and it's definitely not the version users of Glean will consume.

This is also currently blocked by Glean-ios not supporting all metric types listed in the files (but it will build with the next glean_parser release because it just ignores some files)


glean-core/ios/metrics.yaml
```diff
@@ -0,0 +1,238 @@
...
+     The labels are the `category.name` identifier of the metric.
+   bugs:
+     - 1499761
+   data_reviews:
+     - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
+   notification_emails:
+     - telemetry-client-dev@mozilla.com
+   expires: never
+   send_in_pings:
+     - all_pings
```
mdboom: Can we share the metrics.yaml and pings.yaml files between android and iOS?

badboy: We might be, yes. Though that will require moving the files and linking them back (as both generators expect the files to be in the platform-specific project root).

Plus not all metrics line up, e.g. `device_manufacturer` will be a pretty boring metric on iOS, and `android_sdk_version` definitely isn't helpful on iOs.

Dexterp37: True. However, `device_manufacturer` is mandatory (required by our schema definition) so it must be there. `android_sdk_version` is optional, as it's Android specific.

Nothing prevents us from having multiple metrics.yaml files: the parsers can digest more than one registry file and will collapse things together :-) Regardless of the solution, it would be ok for me to have this as a follow-up, though.

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,45 @@
+#!/bin/bash
+
+GLEAN_PARSER_VERSION=1.6.0
+
+read -r -d '' RUN_PYTHON_SCRIPT <<'EOF'
```
Dexterp37: Mh.. I wish we had a way to share this between the gradle plugin and this.

glean-core/ios/metrics.yaml
badboy: Good to know about `device_manufacturer` (and we should double-check when we implement the core metrics to have all required fields).

Yes, the Glean-internal parser should be easily able to pick up multiple files (we can specify them).
For users of course it should default to a root-level `metrics.yaml`.

definitely a follow-up to merge those files

glean-core/ios/sdk_generator.sh
badboy: Yeah, before we land this in any form that it should be used by others we want to unify this across both Glean SDKs

mdboom: There is a related bug that the Gradle plugin will need to resolve to make this possible: https://bugzilla.mozilla.org/show_bug.cgi?id=1579541

badboy:
I think this is now in a state where we can discuss landing it.
This is still not the final form and not yet for external consumers, but it does have all the pieces to generate the internal and baseline metrics as well as the pings for Glean itself.
It would be reasonably easy to also use it in the sample app as a test.

It's not as polished as the Gradle script and not yet documented (is it worst documenting how it's internally used when it might as well change for external users?).
It also uses `pipenv` because that's apparently the only way to get a reliable, working and selected version of Python (and that's on macos 10.14 mojave, I think Catalina will be even more fun because there's not even a system python anymore).

What do you all think about landing this as is for now and then iterating on it?

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,50 @@
+#!/bin/bash
```
Dexterp37: Can you add a license header here?

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,50 @@
...
+
+set -eu  pipefail
```
Dexterp37: Can you also add a comment about what is the purpose of this file?

Dexterp37:
> It would be reasonably easy to also use it in the sample app as a test.

That might be a good testbed to understand how to make this ready for prime time: "is it easy enough to be a Glean thing"?

> It's not as polished as the Gradle script and not yet documented (is it worst documenting how it's internally used when it might as well change for external users?).

Since this is not final, I'm ok with not documenting this for now. If there's anything important to be called out, let's do that as part of code comments.

> What do you all think about landing this as is for now and then iterating on it?

I'm fine with iterating on this after landing this initial integration, as it is definitely an improvement over the checked in code pieces.

Let's just keep in mind what the final outcome of this should be: an easy way to integrate in a consumer application build system.

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,55 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Fail the script on first error
```
Dexterp37: Whoops, sorry if I was unclear here. I really meant to add a comment about what this file is used for.

samples/ios/app/metrics.yaml
```diff
@@ -0,0 +1,26 @@
...
+# automatically converted to Kotlin code at build time using the `glean_parser`
+# PyPI package.
+
+$schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
+
+sample_metrics:
+  test:
+    type: counter
+    description: >
+      A simple counter defined in a third-party library
```
Dexterp37: nit: this is not defined in a 3rd party library:P

samples/ios/app/metrics.yaml
```diff
@@ -0,0 +1,26 @@
...
+    # To see the result immediately when pressing the "send ping" button in the
```
Dexterp37: I think we can remove the commented block here

samples/ios/app/sdk_generator.sh
```diff
@@ -0,0 +1,54 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Fail the script on first error
+set -e
```
Dexterp37: eugh :o Do we *really* need to copy this over? Can't we symlink or something?

badboy: For now ... we kinda do, unless I add some of the parameters to the xcode invocation, let me try that.

samples/ios/app/metrics.yaml
badboy: ah oh :D so much for "just copying this over"

samples/ios/app/metrics.yaml
badboy: d'oh! right

samples/ios/app/sdk_generator.sh
Dexterp37: If you prefer, we can have the sample app changes as a separate PR and can move forward landing the first part, if others agree

badboy: Nope, let's do this right.

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+  INPUT_FILE=${!tmp}
+  DEFINITION_FILES="${DEFINITION_FILES} ${INPUT_FILE}"
+
+  let COUNTER=COUNTER+1
+done
+
+# If no defition files were specified, bail out and fail the build.
+if [ -z "$DEFINITION_FILES" ]; then
+  echo "Missing definition files!"
+  echo "Specify your metrics.yaml and pings.yaml as Input Files to the 'Run Script' step"
```
Dexterp37: Note that we don't fail build on Android if no registry files are provided. None is strictly required.

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+  exit 1
+fi
+
+# When the special argument "internal" is passed, don't add a namespace import and also allow all reserved items.
+NAMESPACE=Glean
```
Dexterp37: Can you please file a bug for supporting the other build time features we have? e.g autodocs and glinter

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+# It uses the `glean_parser`.
+# See https://mozilla.github.io/glean_parser/ for details.
+#
+# To use it in a Swift project, follow these steps:
+# 1. Import the `sdk_generator.sh` script into your project.
+# 2. Add your `metrics.yaml` and (optionally) `pings.yaml` to your project.
+# 3. Add a new "Run Script" build step and set the command to `bash $PWD/sdk_generator.sh`
+# 4. Add your definition files (`metrics.yaml`, `pings.yaml`) as Input Files for the "Run Script" step.
+# 5. Run the build once.
+# 6. Add the files in the `Generated` folder to your project.
```
Dexterp37: So this requires manual input to both set this up in the project *and* use the generated stuff? I wonder if there's a way to automate that..

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,100 @@
...
+    '-m',
+    module_name
+] + sys.argv[3:])
+EOF
+)
+
+# Collect the definition files as passed from xcode
+DEFINITION_FILES=""
+COUNTER=0
+while [ $COUNTER -lt $SCRIPT_INPUT_FILE_COUNT ]; do
```
Dexterp37: On Android we always look for `metrics.yaml` and `pings.yaml` in the project root, this would remove the need of step (2) in the instructions at the top of this doc.

badboy: Oh, well then, I can remove this. Seems like glean_parser is happy to just ignore if files are not there

glean-core/ios/sdk_generator.sh
badboy: Removed the first manual step (well, people will still want to add it to their project, but they don't need to define it as input files to the script anymore).

They will still need to add the generated stuff so that it gets compiled and there's no way around that short of editing the project file (... and we don't want to go _that_ way, do we?)

badboy:
Ok, I take it all back.
Seems like I need to define both the _input_ files and the _output_ files for Xcode to do reliable builds.
It kinda makes sense (it only reruns the script if the input changed, it might start the next build step before the script is even finished, because it doesn't know that it should wait for it).

glean-core/ios/sdk_generator.sh
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1585298

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,84 @@
...
+
+# When the special argument "internal" is passed, don't add a namespace import and also allow all reserved items.
+NAMESPACE=Glean
+INTERNAL=""
+if [ "$1" = "internal" ]; then
+  NAMESPACE=""
+  INTERNAL="--allow-reserved"
+fi
+
+pipenv --venv || pipenv install --three
```
mdboom: pipenv seems like overkill here, because we aren't actually using it to install any dependencies.  

Additionally, it's going to create a new virtual environment for every directory that this script is run from, i.e. one for Glean itself, and one for all the libraries and apps we have in the future that use Glean.  It's only 17MB each (so not nearly as bad as the 350MB each that we have from using Miniconda in Android land -- that's what we get for not needing to support Windows here...)

I think it might be better / more standard to:

1) Create a virtualenv in a single central location:

```
python3 -m venv ~/.local/glean/venv
```

2) Run glean parser from there...

```
~/.local/glean/venv/bin/python3 -c "$RUN_PYTHON_SCRIPT" ...
```

glean-core/ios/sdk_generator.sh
```diff
@@ -0,0 +1,84 @@
...
+        f'{module_name}=={expected_version}'
+    ])
+
+subprocess.check_call([
+    sys.executable,
+    '-m',
+    module_name
+] + sys.argv[3:])
+EOF
+)
```
mdboom: This whole thing is really a hack to get around limitations in Gradle and being unable to call `pip` in the normal way.  Since here we're just writing a bash script, we can just do (from bash, not this Python stub):

```
~/.local/glean/venv/bin/pip install --upgrade glean_parser==$GLEAN_PARSER_VERSION
~/.local/glean/venv/bin/python3 -m glean_parser translate ...
```

(And should probably put the path to the central venv in a variable, too...)

glean-core/ios/sdk_generator.sh
mdboom: I guess alternatively if you *wanted* to use `pipenv` you could include a Pipfile that sets `glean_parser==1.8.2` and have it pick that up, then we don't need to do any custom updating nonsense at all...

glean-core/ios/metrics.yaml
```diff
@@ -0,0 +1,238 @@
...
+      The user-visible version of the operating system (e.g. "1.2.3").
+    bugs:
+      - 1497894
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
+    expires: never
+
+  android_sdk_version:
```
travis79: I think that this is asking a bit much of iOS to know which Android SDK version it is using...

badboy: Woops, that is a left-over when I re-imported that file.
The whole reason it's a copy and not a symlink is that I initially removed this.

glean-core/ios/sdk_generator.sh
badboy: Uhm, that makes _a lot more_ sense.

glean-core/ios/sdk_generator.sh
badboy: Well, we do use `pip` after all.
The reason I reached for pipenv are that for some reason xcode includes some half-broken Python thingy, which is picked up first if I don't use pipenv. I can't run any python on that.

Additionally, I really don't like a project-specific script installing a global across-projects-used virtualenv. That is just calling for issues once anyone has multiple Glean-using projects, that might depend on different versions of the parser.
Plus, we can't even rely on anything outside of the project directory to exist or be writable.

I'll try again using venv directly and calling pip without the detour through a hacky script/

PR #296, opened 2019-09-17
Add concurrent operation for uploading and cancellation of background tasks

travis79:
I tried to break this down into logical commits for easier review.  I would recommend reviewing commit by commit, if possible.

### DIFF
--- a/glean-core/ios/Glean/Dispatchers.swift
+++ b/glean-core/ios/Glean/Dispatchers.swift
@@ -20,13 +20,21 @@ class Dispatchers {
 
     private let logger = Logger(tag: Constants.logTag)
 
-    // This is the task queue that all Glean background operations will be executed on.  It is currently
-    // set to be a serial queue by setting the `maxConcurrentOperationsCount` to 1, but could allow for
-    // concurrent operations by increasing the parameter.
-    lazy var operationQueue: OperationQueue = {
+    // This is a task queue for all Glean background operations that are required to be executed in order.
+    // It is currently set to be a serial queue by setting the `maxConcurrentOperationsCount` to 1
+    lazy var serialOperationQueue: OperationQueue = {
         var queue = OperationQueue()
-        queue.name = "Glean dispatch queue"
+        queue.name = "Glean serial dispatch queue"
         queue.maxConcurrentOperationCount = 1
+        return queue
+    }()
+
+    // This is a task queue for all Glean background operations that need be executed concurrently.
+    // It is currently set to use a maximum of 4 concurrent operations at a time.
+    lazy var concurrentOperationsQueue: OperationQueue = {
+        var queue = OperationQueue()
+        queue.name = "Glean concurrent dispatch queue"
+        queue.maxConcurrentOperationCount = 4
         return queue
     }()
 
@@ -34,7 +42,7 @@ class Dispatchers {
     private var queueInitialTasks = AtomicBoolean(true)
 
     // This array will hold the queued initial tasks that are launched before Glean is initialized
-    lazy var preInitOperations: [Operation] = [Operation]()
+    lazy var preInitOperations = [Operation]()
 
     // When true, jobs will be run synchronously
     var testingMode = false
@@ -43,39 +51,27 @@ class Dispatchers {
     // property
     private init() {}
 
-    /// This function launches a background `Operation`
+    /// This function launches a background `Operation` on a serially executed queue.
     ///
     /// - parameters:
     ///   * block: The block of code to execute, as a Closure that accepts no arugments and returns Void
     ///
     /// This function is used throughout Glean in order to launch tasks in the background, typically
-    /// recording of metrics, uploading of pings, etc.  This function returns an `Operation` which can
-    /// be used in conjunction with a completion handler to perform work after the task is complete.
+    /// recording of metrics and things that need to execute in order. Since this executes the tasks on
+    /// a non-concurrent (serial) queue, the tasks are executed in the order that they are launched.
     ///
-    /// For example:
-    /// ```
-    /// Dispatchers.shared.launch {
-    ///     // Do some work here
-    /// }.completionBlock = {
-    ///     print("Done!")
-    /// }
-    /// ```
+    /// If `queueInitialTasks` is enabled, then the operation will be created and added to the
+    /// `preInitOperations` array but not executed until flushed.
     ///
-    /// You could also use the returned `Operation` to await that specific task like this:
-    /// ```
-    /// let operation = Dispatchers.shared.launch {
-    ///     // Do some work here
-    /// }
-    /// operation.waitUntilFinished()
-    func launch(block: @escaping () -> Void) -> Operation {
-        let operation = GleanOperation(block)
-
+    /// If `testingMode` is enabled, then launch will await the execution of the task (unless queuing is
+    /// enabled)
+    func launch(block: @escaping () -> Void) {
         if queueInitialTasks.value {
             // If we are queuing tasks, typically before Glean has been initialized, then we should
             // just add the created Operation to the preInitOperations array, provided there are
             // less than the max queued operations stored.
             if preInitOperations.count < Constants.maxQueueSize {
-                preInitOperations.append(operation)
+                preInitOperations.append(BlockOperation(block: block))
 
                 if testingMode {
                     logger.info("Task queued for execution in test mode")
@@ -88,17 +84,66 @@ class Dispatchers {
         } else {
             // If we are not queuing initial tasks, we can go ahead and execute the Operation by
             // adding it to the `operationQueue`
+            serialOperationQueue.addOperation(block)
+
+            // If we are in testing mode, go ahead and wait until it is finished before continuing
+            // to ensure synchronous execution of the task.
             if testingMode {
-                // If we are in testing mode, go ahead and execute the operation and wait until it
-                // is finished before continuing to give us synchronous execution of the task.
-                operation.start()
-                operation.waitUntilFinished()
-            } else {
-                operationQueue.addOperation(operation)
+                serialOperationQueue.waitUntilAllOperationsAreFinished()
             }
         }
+    }
 
-        return operation
+    /// This function launches a background `Operation` on a concurrently executed queue.
+    ///
+    /// - parameters:
+    ///   * operation: The `Operation` to execute
+    ///
+    /// This function is used to execute tasks in an asynchrounous manner and still give us the ability
+    /// to cancel the tasks by creating them as `Operation`s rather than using GCD.
+    ///
+    /// This function specifically ignores the `queueInitialTasks` flag because the only tasks that
+    /// should be launched by this are the ping upload schedulers and those should run regardless of
+    /// the initialized state.
+    ///
+    /// If `testingMode` is enabled, then launch will just execute the task rather than adding it to the
+    /// concurrent queue to avoid asynchronous issues while testing
+    func launchAsync(operation: Operation) {
+        if testingMode {
+            operation.start()
+            operation.waitUntilFinished()
+        } else {
+            concurrentOperationsQueue.addOperation(operation)
+        }
+    }
+
+    /// This function launches a background block of code on a concurrently executed queue.
+    ///
+    /// - parameters:
+    ///   * block: The block of code to execute, as a Closure that accepts no arugments and returns Void
+    ///
+    /// This function is used to execute tasks in an asynchrounous manner and still give us the ability
+    /// to cancel the tasks by creating them as `Operation`s rather than using GCD.
+    ///
+    /// This function specifically ignores the `queueInitialTasks` flag because the only tasks that
+    /// should be launched by this are the ping upload schedulers and those should run regardless of
+    /// the initialized state.
+    ///
+    /// If `testingMode` is enabled, then launch will just execute the task rather than adding it to the
+    /// concurrent queue to avoid asynchronous issues while testing
+    func launchAsync(block: @escaping () -> Void) {
+        if testingMode {
+            block()
+        } else {
+            concurrentOperationsQueue.addOperation(block)
+        }
+    }
+
+    /// Cancel any pending background tasks
+    func cancelBackgroundTasks() {
+        serialOperationQueue.cancelAllOperations()
+        concurrentOperationsQueue.cancelAllOperations()
+        preInitOperations.removeAll()
     }
 
     /// Stop queuing tasks and process any tasks in the queue.
@@ -108,7 +153,7 @@ class Dispatchers {
         // `waitUntilFinished` parameter since this is a serial queue and any newly queued tasks
         // should execute after the `preInitOperations` that are being added here. For tests, we
         // need to await all of the tasks to finish execution, so we set this to true.
-        self.operationQueue.addOperations(preInitOperations, waitUntilFinished: testingMode)
+        self.serialOperationQueue.addOperations(preInitOperations, waitUntilFinished: testingMode)
 
         // Turn off queuing to allow for normal background execution mode
         queueInitialTasks.value = false
@@ -147,22 +192,5 @@ class Dispatchers {
     /// When enabled, tasks are queued until launched by calling `flushQueuedInitialTasks()`
     func setTaskQueuing(enabled: Bool) {
         queueInitialTasks.value = enabled
-    }
-
-    /// This is an internal class that represents a Glean background task
-    class GleanOperation: Operation {
-        let backgroundTask: () -> Void
-
-        init(_ block: @escaping () -> Void) {
-            backgroundTask = block
-        }
-
-        override func main() {
-            if isCancelled {
-                return
-            }
-
-            backgroundTask()
-        }
     }
 }
--- a/glean-core/ios/Glean/Dispatchers.swift
+++ b/glean-core/ios/Glean/Dispatchers.swift
@@ -34,7 +34,6 @@ class Dispatchers {
     lazy var concurrentOperationsQueue: OperationQueue = {
         var queue = OperationQueue()
         queue.name = "Glean concurrent dispatch queue"
-        queue.maxConcurrentOperationCount = 4
         return queue
     }()
 
--- a/glean-core/ios/Glean/Dispatchers.swift
+++ b/glean-core/ios/Glean/Dispatchers.swift
@@ -29,8 +29,7 @@ class Dispatchers {
         return queue
     }()
 
-    // This is a task queue for all Glean background operations that need be executed concurrently.
-    // It is currently set to use a maximum of 4 concurrent operations at a time.
+    // This is a task queue for all Glean background operations that can be executed concurrently.
     lazy var concurrentOperationsQueue: OperationQueue = {
         var queue = OperationQueue()
         queue.name = "Glean concurrent dispatch queue"
@@ -50,10 +49,10 @@ class Dispatchers {
     // property
     private init() {}
 
-    /// This function launches a background `Operation` on a serially executed queue.
+    /// This function launches an `Operation` on a serially executed queue.
     ///
     /// - parameters:
-    ///   * block: The block of code to execute, as a Closure that accepts no arugments and returns Void
+    ///   * block: The block of code to execute
     ///
     /// This function is used throughout Glean in order to launch tasks in the background, typically
     /// recording of metrics and things that need to execute in order. Since this executes the tasks on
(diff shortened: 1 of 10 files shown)


glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +84,66 @@ class Dispatchers {
...
+            // If we are in testing mode, go ahead and wait until it is finished before continuing
+            // to ensure synchronous execution of the task.
             if testingMode {
-                // If we are in testing mode, go ahead and execute the operation and wait until it
-                // is finished before continuing to give us synchronous execution of the task.
-                operation.start()
-                operation.waitUntilFinished()
-            } else {
-                operationQueue.addOperation(operation)
+                serialOperationQueue.waitUntilAllOperationsAreFinished()
```
badboy: That's a mouth-full.

glean-core/ios/Glean/Scheduler/GleanOperation.swift
```diff
@@ -0,0 +1,43 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// Base class for other Glean Operations that contains the management of
+/// the executing and finished key-value coding/observers
+class GleanOperation: Operation {
+
+    private var _executing = false {
```
badboy: Can you explain this block of code? It looks ... weird

glean-core/ios/Glean/Glean.swift
```diff
@@ -111,6 +111,11 @@ public class Glean {
         if isInitialized() {
             glean_set_upload_enabled(handle, enabled.toByte())
         }
+
+        // Cancel background metric recording and upload tasks
+        if !uploadEnabled {
```
badboy: In #288 this function is changed. We probably should land that first, then change it here to cancel background tasks before we disable stuff.

glean-core/ios/GleanTests/DispatchersTest.swift
```diff
@@ -106,4 +106,76 @@ class DispatchersTest: XCTestCase {
...
+        // Now cancel the tasks
+        Dispatchers.shared.cancelBackgroundTasks()
+
+        // Assert the task was removed from the queue
+        XCTAssertEqual(Dispatchers.shared.preInitOperations.count, 0, "Task must be removed")
+
+        // Set up our test conditions for normal execution by setting queuing to false
+        Dispatchers.shared.setTaskQueuing(enabled: false)
+
+        // Create some counters to use to determine if the tasks were actually cancelled
```
badboy: Can you split this into 2 tests? First testing that we clear preInit operations, second that we clear others.

(Maybe even split into preInit/sync/async?)

glean-core/ios/GleanTests/DispatchersTest.swift
```diff
@@ -106,4 +106,76 @@ class DispatchersTest: XCTestCase {
...
+        let concurrentOperation = BlockOperation()
+        concurrentOperation.addExecutionBlock {
+            while !concurrentOperation.isCancelled {
+                asyncCounter += 1
+            }
+        }
+        Dispatchers.shared.launchAsync(operation: concurrentOperation)
+
+        // Let the tasks run for 1 second
+        sleep(1)
```
badboy: Hmmm, not a fan, but I guess this is what we have to do.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +84,66 @@ class Dispatchers {
...
+    /// This function is used to execute tasks in an asynchrounous manner and still give us the ability
+    /// to cancel the tasks by creating them as `Operation`s rather than using GCD.
+    ///
+    /// This function specifically ignores the `queueInitialTasks` flag because the only tasks that
+    /// should be launched by this are the ping upload schedulers and those should run regardless of
+    /// the initialized state.
+    ///
+    /// If `testingMode` is enabled, then launch will just execute the task rather than adding it to the
+    /// concurrent queue to avoid asynchronous issues while testing
+    func launchAsync(block: @escaping () -> Void) {
```
Dexterp37: Looks like the logic here is the same of the other `launchAsync`. Can we call the other instead of duplicating, here? Looks like the only difference is in the `block`

glean-core/ios/Glean/Glean.swift
```diff
@@ -152,7 +152,10 @@ public class Glean {
     /// If the ping currently contains no content, it will not be assembled and
     /// queued for sending.
     func sendPingsByName(pingNames: [String]) {
-        _ = Dispatchers.shared.launch {
+        // This runs in the background to allow for processing of pings off of the
+        // main thread. Please note that the ping uploader will spawn other async
+        // operations if there are pings to upload.
+        Dispatchers.shared.launchAsync {
```
Dexterp37: nit: maybe this could be just `launch`?

glean-core/ios/Glean/Scheduler/GleanOperation.swift
Dexterp37: I second this

glean-core/ios/Glean/Scheduler/PingUploadOperation.swift
```diff
@@ -0,0 +1,69 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+class PingUploadOperation: GleanOperation {
```
Dexterp37: nit: can you please add docs for this?

glean-core/ios/GleanTests/DispatchersTest.swift
```diff
@@ -106,4 +106,76 @@ class DispatchersTest: XCTestCase {
...
+
+        let concurrentOperation = BlockOperation()
+        concurrentOperation.addExecutionBlock {
+            while !concurrentOperation.isCancelled {
+                asyncCounter += 1
+            }
+        }
+        Dispatchers.shared.launchAsync(operation: concurrentOperation)
+
+        // Let the tasks run for 1 second
```
Dexterp37: Why is this needed? Can you document the rationale?

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -20,21 +20,29 @@ class Dispatchers {
...
         queue.maxConcurrentOperationCount = 1
         return queue
     }()
 
+    // This is a task queue for all Glean background operations that need be executed concurrently.
+    // It is currently set to use a maximum of 4 concurrent operations at a time.
+    lazy var concurrentOperationsQueue: OperationQueue = {
+        var queue = OperationQueue()
+        queue.name = "Glean concurrent dispatch queue"
+        queue.maxConcurrentOperationCount = 4
```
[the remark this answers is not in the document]
travis79: That's fine, it was an arbitrary choice of number and I didn't realize that the default was determined by the OS.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -43,39 +51,27 @@ class Dispatchers {
...
-
+    /// recording of metrics and things that need to execute in order. Since this executes the tasks on
+    /// a non-concurrent (serial) queue, the tasks are executed in the order that they are launched.
+    ///
+    /// If `queueInitialTasks` is enabled, then the operation will be created and added to the
+    /// `preInitOperations` array but not executed until flushed.
+    ///
+    /// If `testingMode` is enabled, then launch will await the execution of the task (unless queuing is
+    /// enabled)
+    func launch(block: @escaping () -> Void) {
```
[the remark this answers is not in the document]
travis79: It's async (as in it happens on a background thread), but the execution of that particular thread queue is serial rather than concurrent so only one task launched by this function can be executed at a time.  I think the naming of the `launchAsync` functions I added is probably the problem here, they should be `launchConcurrent`.  I intended to have something that we could use that would be like running a task in a coroutine.

glean-core/ios/Glean/Dispatchers.swift
travis79: Yeah, Swift sometimes likes to read like a sentence.

glean-core/ios/Glean/Dispatchers.swift
travis79: I think we can do that, there is no way to convert an operation into a closure but we can do the other way around, so this function can call the one that accepts an operation.

glean-core/ios/Glean/Dispatchers.swift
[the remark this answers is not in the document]
Dexterp37: Ok, I thought so. I think we should really hint that this is used for the API. We don't want other stuff going on this queue. I'd also drop the async suffix, as everything is async here

glean-core/ios/Glean/Glean.swift
travis79: I can leave this out for now

glean-core/ios/Glean/Glean.swift
travis79: The bug was to move this off of the serial queue.  If we call this with `launch` and it takes some time to perform the network upload, then the serial execution queue is blocked.

travis79: The `launchAsync` puts this on a different execution queue that is concurrently executed so as to not block the execution of other tasks.

glean-core/ios/Glean/Scheduler/GleanOperation.swift
travis79: This is something I just learned about `Operation`.  The `isExecuting` and `isFinished` and other member properties use [KVC](https://developer.apple.com/library/archive/documentation/Cocoa/Conceptual/KeyValueCoding/index.html) (key-value coding) which provides indirect access to properties and [KVO](https://developer.apple.com/documentation/swift/cocoa_design_patterns/using_key-value_observing_in_swift) (key-value observing) to allow for event driven observation of these properties. 

This is how that signalling looks to trigger events that are hooked into this. A more in-depth explanation can be found in the docs for [`Operation`](https://developer.apple.com/documentation/foundation/operation)

glean-core/ios/GleanTests/DispatchersTest.swift
travis79: It's needed just so that the background tasks have some time to run.  I can us uSleep for a shorter delay or attempt to find some other way to guarantee that the tasks are running.

glean-core/ios/GleanTests/DispatchersTest.swift
travis79: I don't like sleeps either, but in this case it's just a delay and not waiting for something to occur.  Since both you and @Dexterp37 had issue with this line, let me see if I can find another way to accomplish this.

glean-core/ios/GleanTests/DispatchersTest.swift
Dexterp37: No no, that's fine, just extend the comment to mention that.

glean-core/ios/Glean/Dispatchers.swift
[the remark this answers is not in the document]
travis79: I'll rename these to be more descriptive.  `launchAPI` for the serial queue, and I can rename the other to `launchConcurrent` or similar.

glean-core/ios/Glean/Scheduler/GleanOperation.swift
travis79: I added some comments to help clarify this

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -20,21 +20,28 @@ class Dispatchers {
...
+    lazy var serialOperationQueue: OperationQueue = {
         var queue = OperationQueue()
-        queue.name = "Glean dispatch queue"
+        queue.name = "Glean serial dispatch queue"
         queue.maxConcurrentOperationCount = 1
         return queue
     }()
 
+    // This is a task queue for all Glean background operations that need be executed concurrently.
+    // It is currently set to use a maximum of 4 concurrent operations at a time.
```
Dexterp37: nit: this is not set to 4 anymore, but you can see that the OS determines how many concurrent operations :)

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +83,62 @@ class Dispatchers {
...
+    /// concurrent queue to avoid asynchronous issues while testing
+    func launchConcurrent(block: @escaping () -> Void) {
+        launchConcurrent(operation: BlockOperation(block: block))
+    }
 
-        return operation
+    /// Cancel any pending background tasks
+    func cancelBackgroundTasks() {
+        serialOperationQueue.cancelAllOperations()
+        concurrentOperationsQueue.cancelAllOperations()
```
Dexterp37: Note: this will also cancel `sendPingsByName`. Could this result in partial pings or corrupted data?

glean-core/ios/Glean/Dispatchers.swift
travis79: Good catch, thanks!

glean-core/ios/Glean/Dispatchers.swift
travis79: Since this cancels the network request, I would expect this to not perform a partial upload.  Looking at the [docs](https://developer.apple.com/documentation/foundation/urlsessiontask/1411591-cancel) it looks like it notifies the task so that it can cancel gracefully.

Dexterp37: This is doing ping persistence to disk in addition to the upload. If we cancel (kill the request?) while writing to disk then we result in partial pings being written. Upload doesn't concern me too much

travis79: Ahh I see what you are asking.  The answer is, I'm pretty sure it's safe, and here is why:  The way that the `sendPingsByName` function launches on the concurrent queue is by passing a closure block to be executed.  Since this block doesn't do anything to handle the cancel state, it would run to completion.  Cancel in this case does not mean that it is aborting the thread, it means that it signals the operation to cancel.  See the test I added and how I used the isCancelled in the small counter operations I used.  If I didn't do that, then they would run forever (As I found out the first time I tried the test without it).

travis79: Want me try to write a test around this?

Dexterp37: Ah, I see now. That makes sense, thanks for explaining. Can you add a small comment above the `concurrentOperationsQueue.cancelAllOperations()` line to explain why this is safe (basically, what you just told me). Writing a test for this would be a bit tricky, I'd say not a blocker for now, but file a bug for it if you feel like :)

travis79: I'll add the comment, and I filed a bug to follow up on making the operation that is calling the function that serializes pings obey the cancel state gracefully [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1582464).

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -20,21 +20,27 @@ class Dispatchers {
...
+    // It is currently set to be a serial queue by setting the `maxConcurrentOperationsCount` to 1
+    lazy var serialOperationQueue: OperationQueue = {
         var queue = OperationQueue()
-        queue.name = "Glean dispatch queue"
+        queue.name = "Glean serial dispatch queue"
         queue.maxConcurrentOperationCount = 1
         return queue
     }()
 
+    // This is a task queue for all Glean background operations that need be executed concurrently.
```
badboy: Actually, we don't guarantee concurrency, it's just a possibility.

```suggestion
    // This is a task queue for all Glean background operations that can be executed concurrently.
```

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -43,39 +49,27 @@ class Dispatchers {
     // property
     private init() {}
 
-    /// This function launches a background `Operation`
+    /// This function launches a background `Operation` on a serially executed queue.
     ///
     /// - parameters:
     ///   * block: The block of code to execute, as a Closure that accepts no arugments and returns Void
```
badboy: Now that I re-read this I think it's weird that we explain the closure. That's what the type system is for.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -43,39 +49,27 @@ class Dispatchers {
...
-    func launch(block: @escaping () -> Void) -> Operation {
-        let operation = GleanOperation(block)
-
+    /// recording of metrics and things that need to execute in order. Since this executes the tasks on
+    /// a non-concurrent (serial) queue, the tasks are executed in the order that they are launched.
+    ///
+    /// If `queueInitialTasks` is enabled, then the operation will be created and added to the
+    /// `preInitOperations` array but not executed until flushed.
+    ///
+    /// If `testingMode` is enabled, then launch will await the execution of the task (unless queuing is
```
badboy:
```suggestion
    /// If `testingMode` is enabled, then `launchAPI` will await the execution of the task (unless queuing is
```

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +82,72 @@ class Dispatchers {
...
+    ///   * operation: The `Operation` to execute
+    ///
+    /// This function is used to execute tasks in an asynchrounous manner and still give us the ability
+    /// to cancel the tasks by creating them as `Operation`s rather than using GCD.
+    ///
+    /// This function specifically ignores the `queueInitialTasks` flag because the only tasks that
+    /// should be launched by this are the ping upload schedulers and those should run regardless of
+    /// the initialized state.
+    ///
+    /// If `testingMode` is enabled, then launch will just execute the task rather than adding it to the
```
badboy:
```suggestion
    /// If `testingMode` is enabled, then `launchConcurrent` will just execute the task rather than adding it to the
```

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +82,72 @@ class Dispatchers {
...
-                operation.start()
-                operation.waitUntilFinished()
-            } else {
-                operationQueue.addOperation(operation)
+                serialOperationQueue.waitUntilAllOperationsAreFinished()
             }
         }
+    }
+
+    /// This function launches a background `Operation` on a concurrently executed queue.
```
badboy: The operation isn't "background". IMO leaving it out is just as clear.

```suggestion
    /// This function launches an `Operation` on a concurrently executed queue.
```

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +82,72 @@ class Dispatchers {
...
+        if testingMode {
+            operation.start()
+            operation.waitUntilFinished()
+        } else {
+            concurrentOperationsQueue.addOperation(operation)
+        }
+    }
 
-        return operation
+    /// This function launches a background block of code on a concurrently executed queue.
```
badboy: The blcok isn't "background". IMO leaving it out is just as clear.
```suggestion
    /// This function launches a block of code on a concurrently executed queue.
```

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -88,17 +82,72 @@ class Dispatchers {
...
+        // This will remove all queued operations prior to Glean.initialize() so that they won't be
+        // executed if flushQueuedInitialTasks() is called
+        preInitOperations.removeAll()
+
+        // This will cancel operations in the serially executed queue. This includes most of the
+        // metrics recording and things that need to execute in order. This would not stop the currently
+        // executing operation, but would prevent all remaining operations from executing.
+        serialOperationQueue.cancelAllOperations()
+
+        // This will cancel operations that are executing concurrently. This doesn't abort the thread
```
badboy:
```suggestion
        // This will cancel operations that are executing concurrently. 
        // This doesn't abort running operations immediately
```

glean-core/ios/Glean/Metrics/StringMetric.swift
```diff
@@ -47,7 +47,7 @@ public class StringMetricType {
     public func set(_ value: String) {
         guard !self.disabled else { return }
 
-        _ = Dispatchers.shared.launch {
+        _ = Dispatchers.shared.launchAPI {
```
badboy: We don't need the assignment anymore, do we?

glean-core/ios/Glean/Metrics/TimespanMetric.swift
```diff
@@ -56,7 +56,7 @@ public class TimespanMetricType {
 
         let startTime = timestampNanos()
 
-        _ = Dispatchers.shared.launch {
+        _ = Dispatchers.shared.launchAPI {
```
badboy: As above and below a couple of times. no `_ =` needed

travis79: There were several places this still existed, I squashed them all.

PR #297, opened 2019-09-18
Implement String metric type

badboy:


glean-core/ios/Glean/Metrics/StringMetric.swift
```diff
@@ -0,0 +1,96 @@
...
+        Dispatchers.shared.assertInTestingMode()
+
+        let pingName = pingName ?? self.sendInPings[0]
+        return glean_string_test_has_value(Glean.shared.handle, self.handle, pingName) != 0
+    }
+
+    /// Returns the stored value for testing purposes only. This function will attempt to await the
+    /// last task (if any) writing to the the metric's storage engine before returning a value.
+    ///
+    /// Throws a `String` exception if no value is stored
```
Dexterp37: wow, what's a `String` exception!?

glean-core/ios/Glean/Metrics/StringMetric.swift
```diff
@@ -0,0 +1,96 @@
...
+import Foundation
+
+/// This implements the developer facing API for recording string metrics.
+///
+/// Instances of this class type are automatically generated by the parsers at build time,
+/// allowing developers to record values that were previously registered in the metrics.yaml file.
+///
+/// The string API only exposes the `StringMetricType.set` method, which takes care of validating the input
+/// data and making sure that limits are enforced.
+public class StringMetricType {
```
Dexterp37: Is there a bug for adding usage examples [to the docs](https://mozilla.github.io/glean/book/user/metrics/string.html)?

badboy: I should probably just add the docs here.,

glean-core/ios/Glean/Metrics/StringMetric.swift
badboy: Just a String as an exception. On Kotlin we throw a NullPointerException I think, let me check if we can throw something similar

Dexterp37: No no, I think it's fine :) I was merely asking for curiosity. As a separate thing, did you check if docs show up in XCode when doing autocompletion?

badboy: auto completion on a metric

The first few items are internal and won't pop up when used from the outside (this is in the Glean tests, so it can see everything)

glean-core/ios/Glean/Metrics/StringMetric.swift
mdboom: I also created [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1581586) yesterday for this in general.

docs/user/metrics/string.md
```diff
@@ -43,6 +45,27 @@ assertTrue(SearchDefault.name.testHasValue())
 assertEquals("wikipedia", SearchDefault.name.testGetValue())
 ```
 
+### Swift
+
+```Swift
+// Record a value into the metric.
+SearchDefault.name.set("duck duck go")
```
Dexterp37: Ah, so we're not generating the metrics in a Glean namespace? The reason why we had one was to prevent naming collision in the code.

badboy: There's no namespaces in Swift :(

badboy: I mean ... we're (mis)using empty enums as namespaces, but you then can't do "import GleanMetrics" to use just `SearchDefault.name`

Dexterp37: Mh, I see. I would see, for example, `SearchDefault` easily clashing with something else in a browser application :( Maybe we could add a `Glean` prefix to categories? @mdboom , @travis79  thoughts?

badboy: That's independent of the string metric though, I'll open a bug for discussion as a follow-up

Dexterp37: Yes, good point. Thanks!

mdboom: Yeah, I like the idea of a `Glean` prefix (given the constraints of the language).

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1582133

PR #298, opened 2019-09-18
Enable looking up labels by their index in LabeledMetricType

Dexterp37:
This is needed in order to accumulate to `labeled_counter`(s) when required by Gecko categorical histograms.

I went for the easiest solution here: we pay the cost of storing the list of strings on the Kotlin side instead of having a new FFI function for that. Would that be ok, as a short term solution for feature parity? I'm happy to file a bug and take on this after.

This is a port of mozilla-mobile/android-components#4439


Dexterp37:
> 
> 
> I agree this is fine as a stop gap. Let's file a bug to come back and move `get(index)` to Rust with its own FFI function.

Thanks. Done: https://bugzilla.mozilla.org/show_bug.cgi?id=1582121

PR #299, opened 2019-09-18
Document Swift API for counters

badboy:


docs/user/metrics/counter.md
```diff
@@ -35,6 +37,24 @@ assertTrue(Controls.refreshPressed.testHasValue())
...
+
+There are test APIs available too:
+
+```Swift
+@testable import Glean
+
+// Was anything recorded?
+XCTAssert(Controls.refreshPressed.testHasValue())
+// Does the counter have the expected value?
+XCTAssertEqual(6, Controls.refreshPressed.testGetValue())
```
travis79: This probably needs to be:
```Swift
XCTAssertEqual(6, try Controls.refreshPressed.testGetValue())
```

badboy: Woops, true.

PR #300, opened 2019-09-18
Implement timespan metric type

badboy:


docs/user/metrics/timespan.md
```diff
@@ -66,22 +66,63 @@ assertTrue(Auth.loginTime.testHasValue())
...
+
+There are test APIs available too:
+
+```Swift
+@testable import Glean
+
+// Was anything recorded?
+XCTAssert(Auth.loginTime.testHasValue())
+// Does the timer have the expected value
+XCTAssert(Auth.loginTime.testGetValue() > 0)
```
travis79: This is missing `try`

glean-core/ios/GleanTests/Metrics/TimespanMetricTests.swift
```diff
@@ -0,0 +1,178 @@
...
+            category: "telemetry",
+            name: "timespan_metric",
+            sendInPings: ["store1"],
+            lifetime: .application,
+            disabled: false,
+            timeUnit: .millisecond
+        )
+
+        do {
+            _ = try metric.testGetValue()
```
travis79: There is also an [`XCTAssertThrowsError`](https://developer.apple.com/documentation/xctest/1500795-xctassertthrowserror) function that may be useful here

badboy: for _some_ reason I didn't find that before oO

PR #305, opened 2019-09-19
Set core and baseline metric

badboy:
TODO:

* [x] Check that the core metrics are correct on an actual device


glean-core/ios/Glean/Generated/GleanBaseline.swift
```diff
@@ -0,0 +1,29 @@
+// -*- mode: Swift -*-
+
+// AUTOGENERATED BY glean_parser.  DO NOT EDIT.
+// MANUALLY UPDATED. WILL BE AUTO-GENERATED IN THE FUTURE.
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+internal enum GleanBaseline {
```
Dexterp37: This should not be checked in, right?

glean-core/ios/Glean/Generated/GleanInternalMetrics.swift
```diff
@@ -0,0 +1,113 @@
+// -*- mode: Swift -*-
+
+// AUTOGENERATED BY glean_parser.  DO NOT EDIT.
+// MANUALLY UPDATED. WILL BE AUTO-GENERATED IN THE FUTURE.
+
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+internal enum GleanInternalMetrics {
```
Dexterp37: As above, this should not be checked in, right?

glean-core/ios/Glean/Glean.swift
```diff
@@ -67,7 +67,7 @@ public class Glean {
 
         self.handle = withFfiConfiguration(
             dataDir: getDocumentsDirectory().absoluteString,
-            packageName: Bundle.main.bundleIdentifier!,
+            packageName: AppInfo.name,
```
Dexterp37: Mh, I think this should be the `bundeldentifier`, as far as I can tell [from here](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html)

glean-core/ios/Glean/Utils/Sysctl.swift
```diff
@@ -0,0 +1,158 @@
...
+
+// swiftlint:disable identifier_name
+// REASON: short names are ok
+
+// swiftlint:disable line_length
+// REASON: URLs and doc strings
+
+/// Original: https://github.com/mattgallagher/CwlUtils/blob/0e08b0194bf95861e5aac27e8857a972983315d7/Sources/CwlUtils/CwlSysctl.swift
+/// Copyright © 2016 Matt Gallagher ( https://www.cocoawithlove.com ). All rights reserved.
+/// License: ISC (https://github.com/mattgallagher/CwlUtils/blob/0e08b0194bf95861e5aac27e8857a972983315d7/LICENSE.txt)
```
Dexterp37: Mh. can we even do that? Are licenses compatible? Should we hear legal?

glean-core/ios/Glean/Generated/GleanBaseline.swift
badboy:
```
 // MANUALLY UPDATED. WILL BE AUTO-GENERATED IN THE FUTURE.
```

It needs to right now, because we're not auto-generating them yet.

glean-core/ios/Glean/Glean.swift
badboy: Yeah, I needed to commit it so I can build the sample application and see what the values actually are.

glean-core/ios/Glean/Utils/Sysctl.swift
badboy: Should be, I double-check

badboy: We're fine: 

> The following licences are compatible with the MPL, and code under these licenses can be integrated into Mozilla products.
> [...]
> The ISC license

(from our Licensing Runbook, also https://www.gnu.org/licenses/license-list.en.html#ISC)

glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -141,3 +141,36 @@ func timestampNanos() -> UInt64 {
...
+    if NSLocale.current.languageCode == nil {
+        return "??"
+    } else {
+        if NSLocale.current.regionCode == nil {
+            return NSLocale.current.languageCode!
+        } else {
+            return "\(NSLocale.current.languageCode!)-\(NSLocale.current.regionCode!)"
+        }
+    }
+}
```
mdboom: Just to confirm -- I assume iOS doesn't have any of the out-of-date locale stuff that we have to workaround [here](https://github.com/mozilla-mobile/android-components/blob/master/components/service/glean/src/main/java/mozilla/components/service/glean/utils/LocaleUtils.kt#L38)?

glean-core/ios/Glean/Generated/GleanBaseline.swift
travis79: internal is redundant here as it is the default

glean-core/ios/Glean/Utils/Utils.swift
travis79: I'm pretty sure Apple has this right, but good point

glean-core/ios/Glean/Generated/GleanBaseline.swift
badboy: Oh true. I actually took the output of the glean_parser, so I guess ... I need to fix t hat.

glean-core/ios/Glean/Utils/Utils.swift
badboy: We're fine.

glean-core/ios/Glean/Generated/GleanBaseline.swift
travis79: Don't block on this for this, I know it's a glean_parser thing.

glean-core/ios/Glean/Glean.swift
badboy: Oh wait, I just moved `Bundle.main.bundleIdentifier!` into the `AppInfo` struct. All good here.

PR #307, opened 2019-09-19
Allow easy migration of pending pings from glean-ac

Dexterp37:
This changes the location glean-core stores pending pings to. It's changing from `pings` to `pending_pings`, which is the same location being used by [glean-ac](https://github.com/mozilla-mobile/android-components/blob/914091cebc92046dbc71742ddc8aefa7d7880371/components/service/glean/src/main/java/mozilla/components/service/glean/storages/PingStorageEngine.kt#L50).

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -214,7 +214,7 @@ impl PingMaker {
     /// The directory will be created inside the `data_path`.
     /// The `pings` directory (and its parents) is created if it does not exist.
     fn get_pings_dir(&self, data_path: &Path) -> std::io::Result<PathBuf> {
-        let pings_dir = data_path.join("pings");
+        let pings_dir = data_path.join("pending_pings");
         create_dir_all(&pings_dir)?;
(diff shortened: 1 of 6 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -214,7 +214,7 @@ impl PingMaker {
     /// The directory will be created inside the `data_path`.
     /// The `pings` directory (and its parents) is created if it does not exist.
     fn get_pings_dir(&self, data_path: &Path) -> std::io::Result<PathBuf> {
-        let pings_dir = data_path.join("pings");
+        let pings_dir = data_path.join("pending_pings");
```
mdboom: I would be nice to have a way to share constants between Rust and Kotlin/Swift etc.  Maybe there's a cbindgen config for that?  Anyway, not necessary here at all...

Dexterp37: I [just filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1582448) about this!

PR #308, opened 2019-09-19
Initialize and use Glean in sample app

badboy:
This makes the sample app compile.
It correctly generates a baseline ping on background now.

For some reason the data is not sent (the data task is launched, no request ever goes out) -> Followup https://bugzilla.mozilla.org/show_bug.cgi?id=1582447

Depends on #305, needs a rebase once that is merged.


glean-core/ios/Glean/Glean.swift
```diff
@@ -66,7 +67,7 @@ public class Glean {
         self.configuration = configuration
 
         self.handle = withFfiConfiguration(
-            dataDir: getDocumentsDirectory().absoluteString,
+            dataDir: getDocumentsDirectory().relativePath,
```
Dexterp37: Why does this need to be relative?

badboy: Explained in the commit message:

> We handle file URLs only. These have a 'file://' prefix.
> We don't want the 'file://' prefix.

I don't really have an idea why it worked before.

Dexterp37: We handle as in.. *iOS* requires us to?

badboy: The iOS API we use (`FileManager.default.urls`) gives us URLs.
But we need to pass these through to glean-core, so turn them into strings.
And then the low-level Rust APIs expect proper path strings, not file URLs.

Dexterp37: I see, thanks for clarifying. Can you add a comment with what you just wrote before this line?

PR #309, opened 2019-09-19
Migrate sequence numbers

Dexterp37:
This PR enables the cross-platform Glean SDK to migrate sequence numbers from glean-ac.

This additionally adds initial test coverage for the migration process.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -26,6 +26,7 @@ import mozilla.telemetry.glean.rust.toByte
 import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
 import mozilla.telemetry.glean.GleanMetrics.GleanInternalMetrics
 import mozilla.telemetry.glean.GleanMetrics.Pings
+import mozilla.telemetry.glean.acmigration.GleanACDataMigrator
 import mozilla.telemetry.glean.net.BaseUploader
 import mozilla.telemetry.glean.private.PingType
 import mozilla.telemetry.glean.private.RecordedExperimentData
@@ -130,12 +131,34 @@ open class GleanInternalAPI internal constructor () {
             maxEvents = this.configuration.maxEvents
         )
 
-        handle = LibGleanFFI.INSTANCE.glean_initialize(cfg)
+        // Start the migration from glean-ac, if needed.
+        val migrator = GleanACDataMigrator(applicationContext)
+
+        // TODO: like we do in glean-ac, we should check if we already have a client id.
+        // If we don't, then we don't need to go through migration and this is likely a new
+        // client.
+        val newSequenceNums = if (migrator.shouldMigrateData()) {
+            migrator.getACMetadata().toFfi()
+        } else {
+            Triple(null, null, 0)
+        }
+
+        handle = LibGleanFFI.INSTANCE.glean_initialize_migration(
+            cfg,
+            newSequenceNums.first,
+            newSequenceNums.second,
+            newSequenceNums.third
+        )
 
         // If initialization of Glean fails we bail out and don't initialize further.
         if (handle == 0L) {
             return
         }
+
+        // TODO:
+        // - mark as migrated.
+        // - land a patch on a-c soon-ish to set wasMigrated to false, so that if we
+        //   need to roll back, we can restart the migration again.
 
         // If any pings were registered before initializing, do so now
         this.pingTypeQueue.forEach { this.registerPingType(it) }
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -155,10 +155,9 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
-        // TODO:
-        // - mark as migrated.
-        // - land a patch on a-c soon-ish to set wasMigrated to false, so that if we
-        //   need to roll back, we can restart the migration again.
+        // TODO: uncomment the line below once all migration dependencies have landed.
+        // See bug 1583454.
+        // migrator.markAsMigrated()
 
         // If any pings were registered before initializing, do so now
         this.pingTypeQueue.forEach { this.registerPingType(it) }
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
@@ -1,0 +1,164 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+* License, v. 2.0. If a copy of the MPL was not distributed with this
+* file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+package mozilla.telemetry.glean.acmigration
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.annotation.VisibleForTesting
+import com.sun.jna.StringArray
+import java.lang.NullPointerException
+
+/**
+ * A class encapsulating the code used for migrating data from glean-ac
+ * to glean-core. This class, along with all the migration code, should be removed
+ * from this codebase 6 months after the last application was migrated to
+ * glean-core.
+ */
+internal class GleanACDataMigrator(
+    private val applicationContext: Context
+) {
+    companion object {
+        internal const val SEQUENCE_NUMBERS_FILENAME = "mozilla.components.service.glean.ping.PingMaker"
+        internal const val MIGRATION_PREFS_FILE = "mozilla.components.service.glean.GleanACDataMigrator"
+    }
+
+    internal val migrationPrefs: SharedPreferences? by lazy {
+        applicationContext.getSharedPreferences(
+            MIGRATION_PREFS_FILE,
+            Context.MODE_PRIVATE
+        )
+    }
+
+    /**
+     * A data class representing the metadata from glean-ac.
+     *
+     * @param alreadyMigrated whether or not migration was already performed; if true
+     *        the remaining metadata is not loaded and will be `null`.
+     * @param sequenceNumbers the mapping between AC storage names and their sequence
+     *        numbers.
+     * @return a [Triple] containing the keys and the values representing the sequence
+     *         numbers map, with the addition of the number of elements in the map.
+     */
+    internal data class ACMetadata(
+        val alreadyMigrated: Boolean,
+        val sequenceNumbers: Map<String, Int>
+    ) {
+        fun toFfi(): Triple<StringArray?, IntArray?, Int> {
+            // The Map is sent over FFI as a pair of arrays, one containing the
+            // keys, and the other containing the values.
+            // In Kotlin, Map.keys and Map.values are not guaranteed to return the entries
+            // in any particular order. Therefore, we iterate over the pairs together and
+            // create the keys and values arrays step-by-step.
+            val len = sequenceNumbers.size
+
+            if (len == 0) {
+                return Triple(null, null, 0)
+            }
+
+            val seqList = sequenceNumbers.toList()
+            val keys = StringArray(Array<String>(sequenceNumbers.size) { seqList[it].first }, "utf-8")
+            val values = IntArray(sequenceNumbers.size) { seqList[it].second }
+
+            return Triple(keys, values, len)
+        }
+    }
+
+    /**
+     * Get the metadata from glean-ac.
+     *
+     * @return an instance of [ACMetadata] with, at least, [ACMetadata.alreadyMigrated].
+     */
+    fun getACMetadata(): ACMetadata {
+        if (wasMigrated()) {
+            return ACMetadata(true, emptyMap())
+        }
+
+        return ACMetadata(
+            alreadyMigrated = false,
+            sequenceNumbers = getSequenceNumbers()
+        )
+    }
+
+    /**
+     * Load the sequence numbers that were stored in glean-ac [SharedPreferences].
+     *
+     * @return a `Map<String, Int>` in which entries are (`<ping_name>_seq`, nextSequenceNumber).
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun getSequenceNumbers(): Map<String, Int> {
+        val seqPrefs = applicationContext.getSharedPreferences(
+            SEQUENCE_NUMBERS_FILENAME,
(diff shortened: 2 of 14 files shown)


Dexterp37:
@travis79 @badboy @mdboom thoughts about the overall design of this?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -130,7 +131,22 @@ open class GleanInternalAPI internal constructor () {
...
+        val migrator = GleanACDataMigrator(applicationContext)
+        val acMetadata = migrator.getACMetadata()
+
+        handle = if (acMetadata.alreadyMigrated) {
+            LibGleanFFI.INSTANCE.glean_initialize(cfg)
+        } else {
+            val ffiAcData = acMetadata.toFfi()
+            LibGleanFFI.INSTANCE.glean_initialize_migration(cfg, ffiAcData)
+
+            // TODO:
```
Dexterp37: These are other todos.

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -0,0 +1,163 @@
...
+import androidx.annotation.VisibleForTesting
+import com.sun.jna.StringArray
+import com.sun.jna.Structure
+import com.sun.jna.ptr.IntByReference
+import java.lang.NullPointerException
+
+/**
+ * Define the order of fields as laid out in memory.
+ * **CAUTION**: This must match _exactly_ the definition on the Rust side.
+ *  If this side is changed, the Rust side need to be changed, too.
```
travis79: nit:
```suggestion
 * If this side is changed, the Rust side needs to be changed, too.
```

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -130,13 +131,33 @@ open class GleanInternalAPI internal constructor () {
...
         )
 
-        handle = LibGleanFFI.INSTANCE.glean_initialize(cfg)
+        // Start the migration from glean-ac, if needed.
+        val migrator = GleanACDataMigrator(applicationContext)
+
+        // TODO: like we do in glean-ac, we should check if we already have a client id.
+        // If we don't, then we don't need to go through migration and this is likely a new
+        // client.
+        val newSequenceNums = if (migrator.shouldMigrateData()) {
```
badboy: Can we wrap the `else` branch in `{}`? I find it being inline a bit hard to read.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -130,13 +131,33 @@ open class GleanInternalAPI internal constructor () {
...
+            migrator.getACMetadata().toFfi()
+        } else Pair(null, null)
+
+        handle = LibGleanFFI.INSTANCE.glean_initialize(
+            cfg,
+            newSequenceNums.first,
+            newSequenceNums.second,
+            if (newSequenceNums.first != null) newSequenceNums.first!!.size().toInt() else 0
```
badboy: Can we pull this out into a (then self-documenting) variable with a good name?

glean-core/ffi/src/lib.rs
```diff
@@ -131,7 +136,22 @@ pub unsafe extern "C" fn glean_initialize(cfg: *const FfiConfiguration) -> u64 {
...
         let glean_cfg = glean_core::Configuration::try_from(&*cfg)?;
-        let glean = Glean::new(glean_cfg)?;
+
+        let seq_nums = from_raw_string_array_and_int_array(
+            seq_num_keys,
+            seq_num_values,
+            seq_num_len
+        );
+
+        let glean = if (&seq_nums).as_ref().ok().is_none() {
```
badboy: That's a mouthful.
Can we swap around the if?

```rust
if let Ok(seq_nums) = seq_nums { 
	Glean::with_sequence_numbers(glean_cfg, seq_nums)
} else { 
	Glean::new(...) 
}

File: glean-core/src/ac_migration/mod.rs [hunk 6]
```diff
@@ -0,0 +1,108 @@
...
+/// Stores the sequence numbers from glean-ac in glean-core.
+pub(super) fn migrate_sequence_numbers(
+    glean: &Glean,
+    seq_numbers: HashMap<String, i32>
+) {
+    let ping_maker = PingMaker::new();
+
+    for (store_name_with_suffix, next_seq) in seq_numbers.iter() {
+        // Note: glean-ac stores the sequence numbers as '<ping_name>_seq',
+        // glean-core requires '<ping_name>#sequence'.
```
badboy: I guess we could change that to `_seq` as well if we want?
We'd still need a check anyway I guess, so just keeping it is just as good.

glean-core/src/ac_migration/mod.rs
```diff
@@ -0,0 +1,108 @@
...
+}
+
+/// Stores the sequence numbers from glean-ac in glean-core.
+pub(super) fn migrate_sequence_numbers(
+    glean: &Glean,
+    seq_numbers: HashMap<String, i32>
+) {
+    let ping_maker = PingMaker::new();
+
+    for (store_name_with_suffix, next_seq) in seq_numbers.iter() {
```
badboy: We own `seq_numbers`, we can iterate and consume it with `.into_iter()` and save us the `&0` below.

glean-core/src/database/mod.rs
```diff
@@ -298,7 +298,7 @@ mod test {
 
     #[test]
     fn test_panicks_if_fails_dir_creation() {
-        assert!(Database::new("").is_err());
+        assert!(Database::new("!#\"'@#°ç").is_err());
```
badboy: > Alessio shouting: "!#\"'@#°ç!"

(sorry, nothing to see here, please move on)

glean-core/src/lib.rs
```diff
@@ -142,6 +145,32 @@ impl Glean {
         Ok(glean)
     }
 
+    /// Create and initialize a new Glean object.
```
badboy: Needs updated docs do describe  what the additional parameter is for

glean-core/src/ac_migration/mod.rs
```diff
@@ -0,0 +1,108 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! A module containing glean-core code for supporting data migration
+//! from glean-ac.
```
badboy: IMO we should document what it migrates and also call out that this is definitely temporary while we transition Glean AC and planned to be removed in 2020

glean-core/src/ping/mod.rs
```diff
@@ -48,7 +48,31 @@ impl PingMaker {
...
+            lifetime: Lifetime::User,
+            ..Default::default()
+        });
+
+        // It's safe to add `next_seq` because the glean-ac migration code
+        // clears the glean-core database before the migration starts.
+        seq.add(glean, next_seq);
+    }
+
+    /// This is pub(super) exclusively for enabling the migration tests.
```
badboy: IMO the first line should still document the method.

I'd also change the string to "This is crate-internal exclusively for enabling the migration tests." or something along the lines.

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -0,0 +1,163 @@
...
+
+import android.content.Context
+import android.content.SharedPreferences
+import androidx.annotation.VisibleForTesting
+import com.sun.jna.StringArray
+import java.lang.NullPointerException
+
+/**
+ * A class encapsulating the code used for migrating data from glean-ac
+ * to glean-core. This class, along all the migration code, should be removed
```
badboy:
```suggestion
 * to glean-core. This class, along with all the migration code, should be removed
```

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -0,0 +1,163 @@
...
+     *
+     * @return true if glean-ac data was found and needs to be moved over,
+     *         false otherwise.
+     */
+    internal fun shouldMigrateData(): Boolean {
+        if (wasMigrated()) {
+            return false
+        }
+
+        // TODO: check if we have any previous a-c data (i.e. client_id). If so, migrate.
```
badboy: Just as a check: will those todos land as follow-ups? Does it make sense to add their bug numbers?

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -67,7 +67,12 @@ internal interface LibGleanFFI : Library {
 
     // Glean top-level API
 
-    fun glean_initialize(cfg: FfiConfiguration): Long
+    fun glean_initialize(
```
badboy: This will affect iOS. It will need to explicitly pass `nil, nil, 0`.
Would it make sense to have a duplicate of this method for use by Kotlin instead?

glean-core/src/ac_migration/mod.rs
Dexterp37: Yes, I thought about doing that, but then I realized we'd still need a check. So I stayed with this. Happy to change though, depending on other's feedback.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
mdboom: Could we have an FFI with this explicitly for migration (e.g. `glean_initialize_migration`), so it would be clear that we can remove it and revert back to the single argument version for Kotlin once we don't need migration code anymore?

glean-core/ffi/src/from_raw.rs
```diff
@@ -110,6 +110,38 @@ pub fn from_raw_string_array_and_string_array(
...
+/// * We check the array pointer for validity (non-null).
+/// * FfiStr checks each individual char pointer for validity (non-null).
+/// * We discard invalid char pointers (null pointer).
+/// * Invalid UTF-8 in any string will return an error from this function.
+pub fn from_raw_string_array_and_int_array(
+    keys: RawStringArray,
+    values: RawIntArray,
+    len: i32,
+) -> glean_core::Result<Option<HashMap<String, i32>>> {
+    unsafe {
```
mdboom: Can we put the `unsafe` around just the minimal parts that we know require `unsafe`?  For example, I think the first clause here is actually safe, etc.

glean-core/src/ac_migration/mod.rs
mdboom: I prefer the `#sequence` going forward, since `#` feels like something we reserve for an internal detail, whereas underscores appear in ping names routinely...

glean-core/src/database/mod.rs
mdboom: Do we know these characters are invalid on all major platforms?  I think ext4 on Linux (Android) probably allows all of these.  Maybe include a `/`?  That's sure to fail everywhere, I think.

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
Dexterp37: This was separated from this PR to make things easier to review. The client-id check will come with the 'generic engine' migration code. It will come right after this PR, otherwise nothing will work. No need to put a bug number here, I guess :)

glean-core/src/lib.rs
```diff
@@ -142,6 +145,32 @@ impl Glean {
...
+    ///
+    /// This will create the necessary directories and files in `data_path`.
```
mdboom: And it clears any existing database, right?  That seems potentially surprising and should probably be called out.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
Dexterp37: Mh, Good point. Yes, it does.

glean-core/ffi/src/from_raw.rs
Dexterp37: I think we touched this topic with @badboy a while ago and we agreed on wrapping the whole function there (I've copy pasted it, basically).

glean-core/src/ac_migration/mod.rs
```diff
@@ -0,0 +1,107 @@
...
+use std::collections::HashMap;
+
+use super::Glean;
+use super::PingMaker;
+
+const GLEAN_AC_SEQUENCE_SUFFIX: &str = "_seq";
+
+/// The Glean A-C metadata migration object.
+#[derive(Debug, Clone)]
+pub struct ACMetadata {
```
badboy: Is this actually used anywhere?

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -0,0 +1,164 @@
...
+     * @param sequenceNumbers the mapping between AC storage names and their sequence
+     *        numbers.
+     * @return a [Triple] containing the keys and the values representing the sequence
+     *         numbers map, with the addition of the number of elements in the map.
+     */
+    internal data class ACMetadata(
+        val alreadyMigrated: Boolean,
+        val sequenceNumbers: Map<String, Int>
+    ) {
+        fun toFfi(): Triple<StringArray?, IntArray?, Int> {
```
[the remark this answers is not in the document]
Dexterp37: Yes, I thought about that too. Given that it's one-off and `Triple` exists, I would not bother too much.

glean-core/src/ac_migration/mod.rs
Dexterp37: Ah! Good catch.

glean-core/src/database/mod.rs
```diff
@@ -37,11 +37,11 @@ impl Database {
 
     /// Creates the storage directories and inits rkv.
     fn open_rkv(path: &str) -> Result<Rkv> {
-        let path = std::path::Path::new(path);
+        let path = std::path::Path::new(path).join("db");
```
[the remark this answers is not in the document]
Dexterp37: Thanks for filing!

PR #310, opened 2019-09-19
1581556: Register lifecycle observers on the main thread

mdboom:
This is a port of https://github.com/mozilla-mobile/android-components/pull/4477

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -10,6 +10,7 @@ import android.content.Context
 import android.content.pm.PackageManager
 import android.os.Build
 import android.os.Process
+import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ProcessLifecycleOwner
 import com.sun.jna.StringArray
@@ -33,6 +34,7 @@ import mozilla.telemetry.glean.private.RecordedExperimentData
 import mozilla.telemetry.glean.scheduler.GleanLifecycleObserver
 import mozilla.telemetry.glean.scheduler.PingUploadWorker
 import mozilla.telemetry.glean.scheduler.MetricsPingScheduler
+import mozilla.telemetry.glean.utils.ThreadUtils
 import org.json.JSONObject
 
 /**
@@ -94,6 +96,8 @@ open class GleanInternalAPI internal constructor () {
      * A LifecycleObserver will be added to send pings when the application goes
      * into the background.
      *
+     * This method must be called from the main thread.
+     *
      * @param applicationContext [Context] to access application features, such
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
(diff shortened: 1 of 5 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -162,8 +164,10 @@ open class GleanInternalAPI internal constructor () {
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.flushQueuedInitialTasks()
 
-        // At this point, all metrics and events can be recorded.
-        ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+        MainScope().launch {
```
Dexterp37: **Copied from a-c**

By looking at the docs and the [rationale for this](https://github.com/Kotlin/kotlinx.coroutines/issues/829), looks like we should declare `private val mainScope = MainScope()` in the `GleanInternalAPI` class and use `mainScope.launch` here. This follows the same pattern we've used for `CoroutineScope` in the dispatcher.

I don't know if that's *strictly* required, but I wouldn't want this to have surprising behaviours due to its short lived existence.

travis79: This is used just like this in many places throughout Android Components, it's one of the reasons I saw and suggested it.

travis79: I'm really glad we discovered this interesting little tidbit about lifecycle observers early rather than later.

travis79: Actually @Dexterp37 could you explain what exactly makes you worried that the MainScope() will be short lived?

Dexterp37: `MainScope` is a `CoroutineScope` and, as such, might do something on its setup and teardown. `MainScope`, used like that, probably lives long enough to just call `launch`. What happens to the children coroutines (the one triggered by `launch`), when `MainScope()` is GC'd? I don't know if that's a problem or not, but docs use it differently. Anyway, none it this matters as there's on-going discussion on the a-c side: this might need to change.

travis79: Thanks!  Although I would hope that the scope returned by the `MainScope()` returned a scope that didn't go "out of scope", considering recent lessons learned about Android, it's probably better to be safe than sorry.

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/ThreadUtils.kt
```diff
@@ -0,0 +1,26 @@
...
+
+import android.os.Looper
+
+object ThreadUtils {
+    private val uiThread = Looper.getMainLooper().thread
+
+    /**
+     * Assert that this code is run on the main (UI) thread.
+     */
+    fun assertOnUiThread() {
```
mdboom: This is ported from `android-components`...

PR #312, opened 2019-09-23
1583334 - Add UUID metric type for iOS

travis79:


glean-core/ios/Glean/Metrics/UuidMetric.swift
```diff
@@ -0,0 +1,104 @@
...
+        Dispatchers.shared.assertInTestingMode()
+
+        let pingName = pingName ?? self.sendInPings[0]
+        return glean_uuid_test_has_value(Glean.shared.handle, self.handle, pingName) != 0
+    }
+
+    /// Returns the stored value for testing purposes only. This function will attempt to await the
+    /// last task (if any) writing to the the metric's storage engine before returning a value.
+    ///
+    /// Throws a `String` exception if no value is stored
```
Dexterp37: Does it still throw a `String` exception?

badboy: In `Timespan` I changed that sentence to:

```
/// Throws a "Missing value" exception if no value is stored
```

PR #313, opened 2019-09-24
Implement events for iOS

badboy:
also fixes a bug in the Android part in the commit [Enqueue a new upload worker if a ping was generated after initialize](https://github.com/mozilla/glean/commit/7fe7a54e0a7bdbc57af97c72b1273342f410ee75).


glean-core/ios/Glean/Metrics/EventMetric.swift
```diff
@@ -0,0 +1,224 @@
...
+        return 0
+    }
+}
+
+/// This implements the developer facing API for recording events.
+///
+/// Instances of this class type are automatically generated by the parsers at built time,
+/// allowing developers to record events that were previously registered in the metrics.yaml file.
+///
+/// The Events API only exposes the `EventMetricType.record(_:)` method, which takes care of validating the input
```
Dexterp37: It only exposes the testing methdos :P

glean-core/ios/Glean/Metrics/EventMetric.swift
```diff
@@ -0,0 +1,224 @@
...
+            extra = extraObj
+        }
+
+        return RecordedEventData(category: category, name: name, timestamp: timestamp, extra: extra)
+    }
+
+    /// Returns the stored value for testing purposes only. This function will attempt to await the
+    /// last task (if any) writing to the the metric's storage engine before returning a value.
+    ///
+    /// Throws a `String` exception if no value is stored
```
Dexterp37: nit: throws a `Missing value`

Dexterp37: This also throws `Invalid JSON` and `Empty JSON`. I think this is throwing a lot of different exceptions. Do we really need that many exceptions communicated to the user? Maybe we can throw always the same exception (`Missing value`) and get away with Glean internal logging.

badboy: Good point, it's also just testing and invalid json is essentially just missing a value (like ... the only way this happens is memory corruption?)

glean-core/ios/Glean/Metrics/EventMetric.swift
```diff
@@ -0,0 +1,224 @@
...
+        let res = String(
+            freeingRustString: glean_event_test_get_value_as_json_string(Glean.shared.handle, self.handle, pingName)
+        )
+
+        do {
+            let data = res.data(using: .utf8)!
+            // swiftlint:disable force_cast
+            let jsonRes = try JSONSerialization.jsonObject(with: data, options: []) as! [Any]
+            if jsonRes.isEmpty {
+                throw "Empty JSON array"
```
Dexterp37: Still throwing `Empty JSON array` here :)

badboy: woops

PR #316, opened 2019-09-25
Migrate metrics ping scheduler

Dexterp37:
See the description in the individual commits.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
@@ -10,6 +10,10 @@ import androidx.annotation.VisibleForTesting
 import com.sun.jna.StringArray
 import java.lang.NullPointerException
 
+// The name of the Glean AC package, used to build the name of the files
+// of the preferences files that contain data to be migrated.
+private const val GLEAN_AC_PACKAGE_NAME = "mozilla.components.service.glean"
+
 /**
  * A class encapsulating the code used for migrating data from glean-ac
  * to glean-core. This class, along with all the migration code, should be removed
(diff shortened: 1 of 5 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/acmigration/GleanDataMigrationTest.kt
```diff
@@ -48,33 +87,51 @@ class GleanDataMigrationTest {
...
+            val request = pingServer.takeRequest(20L, TimeUnit.SECONDS)
+            val docType = request.path.split("/")[3]
+            requests[docType] = request.body.readUtf8()
+        }
+
+        // Check that we received the expected sequence number for the baseline ping.
+        val baselineJson = JSONObject(requests["baseline"]!!)
+        assertEquals("baseline", baselineJson.getJSONObject("ping_info")["ping_type"])
+        assertEquals(37, baselineJson.getJSONObject("ping_info")["seq"])
+/*
```
travis79: Just curious, but did you mean to leave this commented out code here?

Dexterp37: No, sorry about that. It was me trying to write an integration test for this. It's not simple and requires messing with lots of internals, so I left it out.

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -10,6 +10,10 @@ import androidx.annotation.VisibleForTesting
 import com.sun.jna.StringArray
 import java.lang.NullPointerException
 
+// The name of the Glean AC package, used to build the name of the files
+// of the preferences files that contain data to be migrated.
+private const val GLEAN_AC_PACKAGE_NAME = "mozilla.components.service.glean"
```
badboy: Do we need this constant anywhere outside or should we just move it down into the companion object?

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -38,13 +44,19 @@ internal class GleanACDataMigrator(
...
+        val sequenceNumbers: Map<String, Int>,
+        val metricsPingLastSentDate: String?
     ) {
+        /**
+         * Get a [Triple] containing data to be passed to the FFI layer.
+         *
+         * @return a [Triple] containing the keys and the values representing the sequence
+         *         numbers map, with the addition of the number of elements in the map.
+         */
         fun toFfi(): Triple<StringArray?, IntArray?, Int> {
```
badboy: With now another thing stored in this class, couldn't we just create and store the Ffi data in the constructor and not bother with another function call and an opaque triple?

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
Dexterp37: No, we don't. Let me make it private inside the companion object

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
Dexterp37: This is the last bit of data going there, _I promise_! Also note that `toFfi` is only called if we really need to migrate in `Glean.kt`, so I'd rather leave this as it is.

PR #317, opened 2019-09-25
1583772 - iOS: add Boolean metric type

travis79:


badboy:
Managed to resolve the merge conflict through GitHub's UI, but ... turns out the new Rust compiler brings us a new clippy and with it new errors :/

badboy:
Should be fixed in #325, only needs a rebase now

PR #318, opened 2019-09-25
1583750 - iOS: Add Datetime metric

travis79:


glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -18,6 +18,15 @@ extension String: Error {
     public var errorDescription: String? { return self }
 }
 
+/// Convenience function to convert ISO8901 string to a Date
+extension Date {
+    static func fromISO8901String(dateString: String) -> Date? {
```
Dexterp37: Please add unit tests for this.

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
```diff
@@ -0,0 +1,144 @@
...
+                self.handle,
+                pingName
+            )
+        )
+    }
+
+    /// Returns the stored value for testing purposes only. This function will attempt to await the
+    /// last task (if any) writing to the the metric's storage engine before returning a value.
+    ///
+    /// Throws a `Missing value` exception if no value is stored
```
Dexterp37: Does this throw `Missing value`?

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
```diff
@@ -0,0 +1,144 @@
...
+    ///
+    /// - parameters:
+    ///     * pingName: represents the name of the ping to retrieve the metric for.
+    ///                 Defaults to the first value in `sendInPings`.
+    ///
+    /// - returns:  value of the stored metric
+    func testGetValue(_: String? = nil) throws -> Date {
+        return Date.fromISO8901String(dateString: try testGetValueAsString())!
```
Dexterp37: Is `pingName` being passed to `testGetValueAsString` here? If not, then maybe we should add a test case for this.

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
```diff
@@ -0,0 +1,144 @@
...
+import Foundation
+
+/// This implements the developer facing API for recording datetime metrics.
+///
+/// Instances of this class type are automatically generated by the parsers at build time,
+/// allowing developers to record values that were previously registered in the metrics.yaml file.
+///
+/// The datetime API only exposes the `DatetimeMetricType.set` method, which takes care of validating the input
+/// data and making sure that limits are enforced.
+public class DatetimeMetricType {
```
Dexterp37: Can you please update the docs [here](https://mozilla.github.io/glean/book/user/metrics/datetime.html) to add the Swift usage?

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
mdboom: iso8901 -> iso8601?

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
travis79: Yes because it doesn't catch the exception from the underlying call to `testGetValueAsString`.  There is a test that specifically tests this, actually.

glean-core/ios/Glean/Metrics/DatetimeMetric.swift
travis79: > Is `pingName` being passed to `testGetValueAsString` here? If not, then maybe we should add a test case for this.

Good catch, it wasn't being passed through.



> iso8901 -> iso8601?

Oops..  good eye!

travis79:
I had to disable `trailingCommas` in the `.swiftformat` config file.  Apparently swiftformat expects there to be a trailing comma in a list, and swiftlint disagrees. So I had to disable the conflicting rule in one or the other...

badboy:
> I had to disable `trailingCommas` in the `.swiftformat` config file. Apparently swiftformat expects there to be a trailing comma in a list, and swiftlint disagrees. So I had to disable the conflicting rule in one or the other...

a-s does the same: https://github.com/mozilla/application-services/blob/master/.swiftlint.yml#L21-L22

PR #319, opened 2019-09-26
Add (empty) metrics enum to be extended by user code

badboy:


travis79:
Or are you just taking advantage of the fact that `enum can't be instantiated?

badboy:
> Or are you just taking advantage of the fact that `enum can't be instantiated?

Yup, empty enums have nothing to instantiate, they simply don't exist in memory.
For a struct we would need to privatize the constructor.

PR #323, opened 2019-09-27
Workaround for Carthage "dependency cycle" bug

badboy:
> Having a Cartfile and pulling this project from another project using carthage, will cause any Cartfile here to be run, and there is currently a carthage bug whereby pulling OHHTTPStubs reports a circular dependancy error (incorrectly).

We move this file out of the way (tests-only-Cartfile) and link it when necessary.
Users of Glean won't need the OHTTPStubs (at least not until we provide a way to test ping sending and receiving in the stub http server).
The sample app is changed accordingly (and documented).
This was tested against a build of [Lockwise for iOS](https://github.com/mozilla-lockwise/lockwise-ios/compare/master...badboy:lockwise-update) already.


badboy:
> Nice work! I saw that telemetry-ios did something similar to this.

Yeah, back then you already copied their `Cartfile` with the same comment, I just opted to ignore that comment and still do all the stuff to fully integrate it ... until I ran into the problem in lockwise...

PR #324, opened 2019-09-27
Build Swift documentation on CI

badboy:
I had to duplicate the README to fix some links, because jazzy takes it verbatim and the relative links then break it.


bin/build-swift-docs.sh
```diff
@@ -0,0 +1,16 @@
+#!/bin/bash
```
mdboom: Should we add a task in the `Makefile` for this? e.g. `make docs-ios`?

.circleci/config.yml
```diff
@@ -70,14 +70,6 @@ commands:
           command: |
               sudo ln -s /lib/x86_64-linux-gnu/libtinfo.so.6 /lib/x86_64-linux-gnu/libtinfo.so.5
 
-  install-xcpretty:
```
travis79: I guess this conflicts with the doc generation somehow?

README.ios.md
```diff
@@ -0,0 +1,43 @@
...
+**Note: The Glean SDK requires at least [Rust 1.34.2](https://blog.rust-lang.org/2019/05/14/Rust-1.34.2.html). Older versions are untested.**
+
+## Contact
+
+To contact us you can:
+
+* Find us on the Mozilla Slack in *#glean*, on [Mozilla IRC][mozirc] in *#telemetry*.
+* To report issues or request changes, file a bug in [Bugzilla in Data Platform & Tools :: Glean: SDK][newbugzilla].
+* Send an email to *glean-team@mozilla.com*.
+* The Glean Core team is: *:dexter*, *:janerik*, *:mdroettboom*, *:gfritzsche*
```
travis79: Me too?

README.ios.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Overview
+
+Refer to the documentation for [using and developing the Glean SDK][book].
+
+For an overview of Glean beyond just the SDK, see the [section in the Firefox data docs](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+
+**Note: The Glean SDK requires at least [Rust 1.34.2](https://blog.rust-lang.org/2019/05/14/Rust-1.34.2.html). Older versions are untested.**
+
+## Contact
```
Dexterp37: Instead of repeating this, can we link to the relevant section of the [main readme file](https://github.com/mozilla/glean/blob/master/README.md)?

README.ios.md
```diff
@@ -0,0 +1,43 @@
...
+
+## Documentation
+
+All documentation is available online:
+
+## [The Glean SDK Book][book]
+
+## Overview
+
+Refer to the documentation for [using and developing the Glean SDK][book].
```
Dexterp37: Should this overview be iOS specific?

README.ios.md
Dexterp37: Yes! This should be fixed in the main readme as well :) I think this might need to be changed to the whole team in Q4, given project FOG.

.circleci/config.yml
badboy: Ah, should have left a comment.
Turns out xcpretty is installed by default on the container on CircleCI and ... the one we installed wasn't used because it landed in a different path anyway.

README.ios.md
badboy: This readme is rendered as the main page for the swift-specific API documentation.
I'm considering replacing it completely with a big link to the main page instead.

badboy:
image

PR #326, opened 2019-09-30
Migrate the UUIDs and DateTime storage engines

Dexterp37:
This is the initial work for migrating the `client_id` and `first_run_date` to glean-core. To do this, it ports the `UuidStorageEngine` and the `DatetimesStoageEngine`. A few notable things:

- after talking to @fbertsch, we decided that any precision fixup needs to happen on the SDK; otherwise we'd end up with strings having different formats in the database, which is undesirable; that's the reason why there's the lookup map in `DatetimesStorageEngine`.
- there was some serious madness going on with the test! For example `2019-09-24+02:00` was getting parsed by `parseISOTimeString` , which was applying DST. On my timezone, this meant having `2019-09-24` and 1hr timezone offset and 1hr DST offset. For this reason, the test was failing as it was reporting  `2019-09-24+01:00` (only the timezone offset).
- we were not affected by this problem before since we were only using the parsing function to make sure the string [was _parseable_](https://github.com/mozilla-mobile/android-components/blob/842a84264a74c4049b80a6438d67ff2a2295c1cb/components/service/glean/src/main/java/mozilla/components/service/glean/storages/DatetimesStorageEngine.kt#L40-L41), while still using the raw string
- Is the `MetricsPingScheduler` affected by the date parsing problem as well?


glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/engines/DatetimesStorageEngine.kt
```diff
@@ -0,0 +1,94 @@
...
+    init {
+        this.applicationContext = applicationContext
+    }
+
+    companion object {
+        // Unfortunately we have no way to tell, at run time, what is the precision
+        // with which metrics were recorded in glean-ac. To work around this problem
+        // we generate the following map from all the metrics.yaml files listed in
+        // https://github.com/mozilla/probe-scraper/blob/master/repositories.yaml
+        val datePrecisionMap: Map<String, TimeUnit> = mapOf(
```
badboy: This looks like less of a hack than I thought. Kudos to that!

glean-core/android/src/test/java/mozilla/telemetry/glean/acmigration/GleanACDataMigratorTest.kt
```diff
@@ -199,4 +199,50 @@ class GleanACDataMigratorTest {
...
+
+        val migrator =
+            spy<GleanACDataMigrator>(GleanACDataMigrator(ApplicationProvider.getApplicationContext()))
+        `when`(migrator.wasMigrated()).thenReturn(false)
+
+        assertTrue(migrator.shouldMigrateData())
+    }
+
+    @Test
+    fun `shouldMigrate must return false if not migrated and client id is a knonw value`() {
```
badboy:
```suggestion
    fun `shouldMigrate must return false if not migrated and client id is a known value`() {
```

badboy: Actually, can we change "known value" in the name to "canary value"? Known value sounds to harmless.

glean-core/android/src/test/java/mozilla/telemetry/glean/acmigration/GleanDataMigrationTest.kt
```diff
@@ -34,6 +50,31 @@ class GleanDataMigrationTest {
         prefs?.edit()?.putInt("${pingName}_seq", number)?.apply()
     }
 
+    private fun setInitianlDataToMigrate(context: Context) {
```
badboy:
```suggestion
    private fun setInitialDataToMigrate(context: Context) {
```

glean-core/android/src/test/java/mozilla/telemetry/glean/acmigration/GleanDataMigrationTest.kt
```diff
@@ -53,7 +94,7 @@ class GleanDataMigrationTest {
         migrator.testResetMigrationStatus()
 
         // Set a fake sequence number for the baseline ping.
-        setFakeSequenceNumber(context, "baseline", 37)
+        setInitianlDataToMigrate(context)
```
badboy:
```suggestion
        setInitialDataToMigrate(context)
```

glean-core/android/src/main/java/mozilla/telemetry/glean/acmigration/GleanACDataMigrator.kt
```diff
@@ -157,6 +172,40 @@ internal class GleanACDataMigrator(
...
+    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+    internal fun getStoredClientId(): UUID? {
+        val clientId = uuidStorageEngine
+            .getSnapshot("glean_client_info", false)
+            ?.get("client_id")
+
+        return if (clientId == KNOWN_CLIENT_ID) {
+            // Having a KNOWN_CLIENT_ID is like having no client id for
+            // migration purposes.
+            null
```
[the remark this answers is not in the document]
Dexterp37: I think you're right. Just to confirm, was this the scenario you were thinking about?

- User disables telemetry (client id is set to c0ffee);
- Migration happens (we skip porting over the client id);
- We have a bug that sends telemetry, we send a legit client id as we didn't port over the canary one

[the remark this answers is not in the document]
mdboom: Exactly my worry (though I don't know what the "bug" might be, but I guess that's the whole point of the canary...)

PR #327, opened 2019-09-30
Fix the docs by building the project first

badboy:

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -203,6 +203,26 @@ jobs:
       - run:
           name: Run XCode tests
           command: bash bin/run-ios-tests.sh
+      - restore_cache:
+          name: Restore rubygems cache
+          key: swift-docs-gems-v1
+      - run:
+          name: Install jazzy
+          command: gem install jazzy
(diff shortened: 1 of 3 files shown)


mdboom:
Oh, dear.  If you install `jazzy` on Linux, it installs some pre-compiled binary blob as a helper:

```
$ file ~/.gem/ruby/gems/jazzy-0.11.2/bin/sourcekitten                                               
/home/mdboom/.gem/ruby/gems/jazzy-0.11.2/bin/sourcekitten: Mach-O 64-bit x86_64 executable, flags:<NOUNDEFS|DYLDLINK|TWOLEVEL|PIE>
```

:clap: :apple:

badboy:
> Does this mean non-Mac platforms can't build the docs? That's kind of a bummer, but if it's unavoidable, it's unavoidable...

That's always the case. Jazzy documents that. There are ways to get a sourcekitten to work on Linux, but it seems like that's a lot of hassle.
The same applies to swiftlint/swiftfmt, as one of them (or both?) also rely on sourcekitten.

PR #331, opened 2019-10-01
1585335 - Add CountersStorageEngine to AC data migration

travis79:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt
```diff
@@ -78,6 +78,26 @@ class CounterMetricType internal constructor(
...
 
+    /**
+     * Add to counter value synchronously.
+     *
+     * This is only to be used for the glean-ac to glean-core data migration.
+     *
+     * @param amount This is the amount to increment the counter by, defaulting to 1 if called
+     * without parameters.
+     */
+    internal fun addSync(amount: Int = 1) {
```
Dexterp37: nit: let's remove the default value here, we don't expect it to ever be called with a default value. Let's also update the docs accordingly

PR #335, opened 2019-10-02
Upload Swift code coverage

badboy:


badboy:
Let's see if that works.
I'm considering gathering all the data and move the upload into an external step just so that uploading coverage won't fail the build all the time.

.circleci/config.yml
```diff
@@ -226,6 +226,13 @@ jobs:
       - persist_to_workspace:
           root: build/
           paths: docs/swift
+      - run:
+          name: Create and upload code coverage
+          command: |
+              curl -L https://github.com/mozilla/grcov/releases/latest/download/grcov-linux-x86_64.tar.bz2 | tar jxf -
```
Dexterp37: Do we really want to stick to `latest`? Maybe we should use the current version and upgrade if needed? Thoughts?

badboy: Actually it's not needed for the xcode part (copy-pasta...)

(but even for the Rust stuff I would answer yes; it's mozilla-controlled, it's "just" code coverage, grcov has infrequent releases, ...)

PR #336, opened 2019-10-02
Remove all explicit names in project files

badboy:


badboy:
It does. This is mostly just the project name and that is definitely configured per project (it just defaults to the dev's local name).
It's also mostly just the sample app (Glean itself is already under a mozilla namespace)

samples/ios/app/base.xcconfig
```diff
@@ -1,7 +1,7 @@
 #include "../../../xcconfig/common.xcconfig"
 
 INFOPLIST_FILE = glean-sample-app/Info.plist
-PRODUCT_BUNDLE_IDENTIFIER = net.rediger.glean-sample-app
+PRODUCT_BUNDLE_IDENTIFIER = org.mozilla.glean-sample-app
```
travis79: I'm not 100% sure that this is still a problem but the bundle identifier used to get tied to the team/developer account through the signing certificate. This may cause problems if we get added to the actual Mozilla Apple developer account and want to use the correct identifier.  It may just mean that a cert needs to be transferred or deleted, or it may not be an issue at all since this will never get added to the AppStore.

travis79: Unfortunately having a sample application in the Apple ecosystem is not as straightforward and we may need to document what it will take for an individual to build and run the sample app.

badboy: Hmpf, ... well, let's see and wait if someone complains?
And if we do all this through a Mozilla Apple dev team/account, then we can figure it out.

badboy: (I mean ... I was able to compile lockwise-ios, building it for my own device only failed because some capabilities can only be done with a paid account, so it might just all be possible?)

travis79: I'm hoping that it's better than it was when I last worked with it, either that or maybe Lockwise has a trick 😄

PR #338, opened 2019-10-02
1585021: Fix ordering of events at startup to prevent negative timestamps

mdboom:
This is basically a port of https://github.com/mozilla-mobile/android-components/pull/4595 to glean-core.

glean-core didn't actually have the same underlying bug, however, it was loading queued pings from disk on the main thread.  (Oh noes, I/O on the main thread!!!)  That has been fixed so it does this on an "early" (non-queued) task just as on glean-ac.

The other changes to glean-ac were not required, since everything is already internally synchronous on the Rust side when `sendPings` is called.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
@@ -177,8 +177,13 @@ internal object Dispatchers {
         /**
          * Helper function to execute the task as either an synchronous or asynchronous operation,
          * depending on whether [testingMode] is true.
+         *
+         * WARNING: THIS SHOULD ALMOST NEVER BE USED. IF IN DOUBT, USE [launch] INSTEAD.
+         *
+         * This has internal visibility only so that it can be called directly to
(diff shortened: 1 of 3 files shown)


mdboom:
Also filed https://bugzilla.mozilla.org/show_bug.cgi?id=1585665 to port this to iOS.

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -177,8 +177,13 @@ internal object Dispatchers {
         /**
          * Helper function to execute the task as either an synchronous or asynchronous operation,
          * depending on whether [testingMode] is true.
+         *
+         * WARNING: THIS SHOULD ALMOST NEVER BE USED. IF IN DOUBT, USE [launch] INSTEAD.
```
badboy: Can we document here why that is? What makes it so dangerous?

mdboom: Sure.  It shouldn't be used for anything that might be called before initialization that needs to actually happen after initialization.  I'll add that to the comment here.

PR #339, opened 2019-10-02
Migrate labeled stuff

Dexterp37:
This merely adds test coverage for the migration code implemented by @travis79 that also magically migrates labeled things.

This works, under the hood, because we store things the same way across glean-ac and glean-core and use the same wrapping mechanism for labeled types.


Dexterp37:
Merging this, the failure seems to be unrelated (connections to codevo.io :) ).

PR #342, opened 2019-10-03
Add documentation for iOS development & review

badboy:


docs/contributing.md
```diff
@@ -38,14 +38,37 @@ cargo test --all
...
+- For changes to Rust code
+  - `cargo test --all` produces no test failures
+  - `cargo clippy --all --all-targets --all-features -- -D warnings` runs without emitting any warnings or errors.
+  - `cargo fmt` does not produce any changes to the code.
+- For changes to Kotlin code
+  - `./gradlew ktlint detekt` runs without emitting any warnings.
+  - `./gradlew test` runs without emitting any warnings or errors.
+- For changes to Swift code
+  - `swiftformat --swiftversion 5 glean-core/ios samples/ios && swiftlint` runs without emitting any warnings or producing changes.
+  - `bin/run-ios-tests.sh` (or running tests in Xcode) runs without emitting any warnings or errors.
```
travis79: Does `make test` deserve a note here?  Or is that not what we want contributors using now? (is it still something that I should be using?)

mdboom: Given that it's likely not every developer will have all of the tools to build for every platform, maybe it makes sense to change this to `test-android`, `test-ios` etc.  I, personally, like the convenience of having things go through `make` just for ease of remembering things.

docs/dev/ios/setup-ios-build-environment.md
```diff
@@ -0,0 +1,41 @@
+# Setup the iOS Build Environment
+
+## Prepare your build environment
+
+1. Install Xcode 11.0
+2. Install [Carthage](https://github.com/Carthage/Carthage): `brew install carthage`
+3. Ensure you have Python 3 installed: `brew install python`
```
mdboom: Maybe add a `brew install swiftlint swiftformat`, since these are required to meet the linting requirements?

docs/dev/ios/setup-ios-build-environment.md
```diff
@@ -0,0 +1,41 @@
...
+4. Run `bin/bootstrap.sh` to download dependencies.
+5. (Optional, only required for building on the CLI) Install [xcpretty](https://github.com/xcpretty/xcpretty): `gem install xcpretty`
+
```
mdboom: Also, to build the docs, you need to install `jazzy`, but the headers for building native Ruby extensions don't come with the system Ruby.

I had to do:

```
brew install ruby
```

add `export PATH=/usr/local/opt/ruby/bin:$PATH` so the homebrew ruby takes precedence.  (`brew link` doesn't work for this)

`gem install jazzy`

But maybe there's an easier way?

mdboom: It's also worth doing all this *before* installing `xcpretty` so everything gets into the same homebrew ruby install.

badboy: Ugh, ruby on macos is so complicated.,

Makefile
```diff
@@ -1,72 +1,101 @@
+default: test
+
+# All builds
+
+build: build-rust build-kotlin
```
mdboom: I find it a bit confusing that `build` does these two but not `build-swift`.  I understand why, of course, since we may not be on a Mac, but I think it's equally likely that a developer would be on a Mac and doesn't care about android and therefore doesn't have the Android SDK installed.  I'd suggest either (1) removing `build` altogether or (2) just making it an alias for `build-rust` which is the only thing we can be certain the user would have the deps for.

mdboom: And probably ditto for `test` and `lint`...

Makefile
```diff
@@ -1,72 +1,101 @@
...
-.PHONY: docs
 
-docs-swift:
+swift-docs:
 	bin/build-swift-docs.sh
-.PHONY: docs
+
+.PHONY: docs rust-docs kotlin-docs swift-docs
 
 linkcheck: docs
```
mdboom: Strictly speaking, I guess the only way to make this work is to also build `swift-docs`, but I think that's only iOS.  (That's kind of a pain -- we might have to modify the linkchecker config so it thinks that any links into the iOS docs are always valid...)

Makefile
```diff
@@ -1,72 +1,101 @@
...
 	# Requires https://wummel.github.io/linkchecker/
 	linkchecker --ignore-url javadoc --ignore-url docs/glean_core build/docs
 .PHONY: linkcheck
 
+# Utilities
+
+android-emulator:
```
mdboom: :+1:  CLI FTW.

docs/contributing.md
```diff
@@ -32,29 +32,61 @@ You can check that everything compiles by running the following from the
...
 ```
 
 If you plan to work on the Android component bindings, you should also review
 the instructions for [setting up an Android build environment](https://github.com/mozilla/glean/blob/master/docs/dev/setup-android-build-environment.md)
 
+To run all Kotlin tests:
+
+```
+make test-kotlin
+```
```
mdboom: "or run tests in Android Studio."  (I'm usually a CLI person as well, but the Android Studio test runner is really convenient for debugging a failing test...)

Makefile
```diff
@@ -1,72 +1,101 @@
...
-lint: clippy ktlint
-.PHONY: lint
 
 swiftlint:
 	swiftlint --strict
-	swiftformat glean-core/ios samples/ios --swiftversion 5 --verbose --lint
-.PHONY: swiftlint
 
-fmt:
+.PHONY: lint clippy ktlint swiftlint
```
mdboom: For consistency with `build`/`test`/`docs`, maybe name these `lint-kotlin`, `lint-swift`, `lint-rust` etc.?

Makefile
badboy: Sounds good!

Makefile
badboy: I thought about it ... but stuck with the old names because they are so conveniently named (and so ingrained in my brain to type `cargo clippy`, `swiftlint`, ...)

Makefile
badboy: *secretly adds a header saying you have to use macos to do glean*

badboy: I can't run the linkchecker locally anyway (because ... reasons), so ... I trust CI to have it all.

mdboom: Yeah -- at the end of the day, as long as it works on CI, that's all that really matters.  This isn't a complex thing like a compiler where it's critical to be able to test things locally.

Makefile
mdboom: Ok.  Fair enough :)

PR #344, opened 2019-10-03
Support calling the Specific Metrics API from Java

Dexterp37:
This is for feature parity with mozilla-mobile/android-components#4617


Dexterp37:
The failure seems to be unrelated, coming from 'first_run_date is consistent across timezones when migrated ' -  We might just have found a flaky test.

PR #345, opened 2019-10-03
1585665: Port 1585021 to iOS: Make sure startup events are in order

mdboom:
This is a port of #338 to iOS.

### DIFF
--- a/glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
+++ b/glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
@@ -237,4 +237,52 @@ class EventMetricTypeTests: XCTestCase {
         XCTAssertNotNil(events)
         XCTAssertEqual(1, events?.count)
     }
+
+    func testFlushQueuedEventsOnStartupDroppingPreinitEvents() {
+        setupHttpResponseStub()
+        expectation = expectation(description: "Completed upload")
+
+        let event = EventMetricType<SomeExtraKeys>(
+            category: "telemetry",
+            name: "test_event",
+            sendInPings: ["events"],
+            lifetime: .ping,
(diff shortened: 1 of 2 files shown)


glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
```diff
@@ -237,4 +237,62 @@ class EventMetricTypeTests: XCTestCase {
...
+
+        let events2 = lastPingJson?["events"] as? [Any]
+        XCTAssertNotNil(events2)
+        XCTAssertEqual(2, events2?.count)
+        XCTAssertEqual(
+            "pre-init",
+            (
+                (events2?[0] as? [String: Any])?["extra"] as? [String: Any]
+            )?["some_extra"] as? String
+        )
```
mdboom: I feel like there's got to be a better way here that my novice swiftness just couldn't figure out...

travis79: I think I would have created a helper function something like this:
```Swift
// Will throw if anything goes wrong
private func getExtraValue(from event: Any?, for key: String) -> String {
    let event = event! as! [String: Any]
    let extras = event["extra"] as! [String: Any]
    return extras[key] as! String
}
```

Then you could call it from the `XCTAssertEqual` tests like this:

```Swift
XCTAssertEqual("run1", getExtraValue(from: events![0], for: "some_extra"))
// or
XCTAssertEqual("pre-init", getExtraValue(from: events2![0], for: "some_extra"))
XCTAssertEqual("post-init", getExtraValue(from: events2![1], for: "some_extra"))
```

That at least puts most of the nasty force-unwrapping of optionals into a function.

PR #347, opened 2019-10-03
1582447 - Fix delayed ping uploading

travis79:
This fixes the delayed uploading of pings.  Due to a combination of factors, including using the wrong URLSessionTask and not using the extended background operation properly, the pings would sometimes not be sent when going to background.

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -26,7 +26,6 @@ public class Glean {
     private var uploadEnabled: Bool = true
     private var configuration: Configuration?
     private var observer: GleanLifecycleObserver?
-    static var backgroundTaskId = UIBackgroundTaskIdentifier.invalid
 
     // This struct is used for organizational purposes to keep the class constants in a single place
     struct Constants {
@@ -189,14 +188,6 @@ public class Glean {
 
     /// Handle background event and send appropriate pings
     func handleBackgroundEvent() {
-        // This asks the OS for more time when going to background to perform critical tasks.
-        // This will be signaled complete when `sendPingsByName` is done.
-        Glean.backgroundTaskId = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
-            // End the task if time expires
-            UIApplication.shared.endBackgroundTask(Glean.backgroundTaskId)
-            Glean.backgroundTaskId = UIBackgroundTaskIdentifier.invalid
-        }
-
         self.sendPingsByName(pingNames: ["baseline", "events"])
     }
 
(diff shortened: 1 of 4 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -189,14 +188,6 @@ public class Glean {
 
     /// Handle background event and send appropriate pings
     func handleBackgroundEvent() {
```
Dexterp37: We removed that from the Android application, because we need to remove this API endpoint. Can you file a bug for removing this here too?

glean-core/ios/Glean/Scheduler/PingUploadOperation.swift
```diff
@@ -34,8 +38,23 @@ class PingUploadOperation: GleanOperation {
...
-        dataTask = URLSession.shared.dataTask(with: request) { _, response, error in
+        // Build a URLSession with no-caching suitable for uploading our pings
+        let config = URLSessionConfiguration.default
+        config.requestCachePolicy = NSURLRequest.CachePolicy.reloadIgnoringLocalCacheData
+        config.urlCache = nil
+        let session = URLSession(configuration: config)
+
+        // This asks the OS for more time when going to background in order to allow for background
+        // uploading of the pings.
+        backgroundTaskId = UIApplication.shared.beginBackgroundTask(withName: "Glean Upload Task") {
```
Dexterp37: Probably a stupid question: is it common/are we ok to have a background task name with spaces in it?

glean-core/ios/Glean/Glean.swift
travis79: API endpoint?  As in it was leaking into the public API?

glean-core/ios/Glean/Scheduler/PingUploadOperation.swift
travis79: It's really just a string "tag" and so the spaces are harmless. I'd also be okay renaming it to whatever you feel is more appropriate.

glean-core/ios/Glean/Glean.swift
travis79: Also, do you have a link to the Android bug you are referring to so I can reference it in the new bug?

Dexterp37: Yes, sorry if I was unclear, my English-fu is weak today :) I meant "into the public API". The related bug is bug 1574470.

glean-core/ios/Glean/Scheduler/PingUploadOperation.swift
Dexterp37: No problem, it's fine!

glean-core/ios/Glean/Glean.swift
travis79: Properties and functions are internal by default in iOS, so we shouldn't have to do anything.

Dexterp37: Ah! Right, I got confused again :) thanks for clarifying and sorry for the noise!

PR #348, opened 2019-10-04
Upgrade `WorkManager` to version 2.2.0

Dexterp37:
This was a bit less trivial than expected: the new version required a version upgrade of some other dependencies as well, so I ended up upgrading them all. We are now matching the [same version that Android Components is using](https://github.com/mozilla-mobile/android-components/blob/master/buildSrc/src/main/java/Dependencies.kt).

This allowed to simplify the code in a few places: see the 'BONUS' commits.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -33,7 +33,7 @@ buildscript {
         androidx_junit: '1.1.1',
         androidx_lifecycle_extensions: '2.1.0',
         androidx_test: '1.2.0',
-        androidx_work: '2.0.1',
+        androidx_work: '2.2.0',
     ]
 
     ext.build = [
--- a/build.gradle
+++ b/build.gradle
@@ -31,6 +31,7 @@ buildscript {
         androidx_core: '1.0.0',
         androidx_espresso: '3.2.0',
         androidx_junit: '1.1.1',
+        androidx_lifecycle_extensions: '2.1.0',
         androidx_test: '1.2.0',
         androidx_work: '2.0.1',
     ]
--- a/build.gradle
+++ b/build.gradle
@@ -11,16 +11,16 @@ buildscript {
     // changing them. Please note that, for using in Android-Components, the
     // versions below must match the ones in that repository.
     ext.versions = [
-        android_components: '5.0.0',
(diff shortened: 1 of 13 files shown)


build.gradle
```diff
@@ -11,28 +11,29 @@ buildscript {
     // changing them. Please note that, for using in Android-Components, the
     // versions below must match the ones in that repository.
     ext.versions = [
-        android_components: '5.0.0',
+        android_components: '15.0.0',
```
travis79: Whoa! That was a big jump.

build.gradle
```diff
@@ -11,28 +11,29 @@ buildscript {
...
+        androidx_annotation: '1.1.0',
+        androidx_appcompat: '1.1.0',
+        androidx_browser: '1.2.0-alpha07',
+        androidx_core: '1.1.0',
         androidx_espresso: '3.2.0',
         androidx_junit: '1.1.1',
+        androidx_lifecycle_extensions: '2.1.0',
         androidx_test: '1.2.0',
-        androidx_work: '2.0.1',
+        androidx_work: '2.2.0',
```
travis79: Welcome to WorkManager 2.2, where we fixed all the old bugs and introduced new ones!

build.gradle
Dexterp37: Ehehhe, yeah :) We're just using this for tests, so it should not be too pressing to update it :D

build.gradle
Dexterp37: Ahahaha, right :D Let's hope more was fixed than broken :P

Dexterp37:
> 
> 
> This touched a lot more than I expected but it looks good! I had one question that came up while looking through this and that was where we use a mocked/spy context, I assume that this doesn't affect the WorkManager?

Nope, it doesn't seem to affect it!

travis79:
> > This touched a lot more than I expected but it looks good! I had one question that came up while looking through this and that was where we use a mocked/spy context, I assume that this doesn't affect the WorkManager?
> 
> Nope, it doesn't seem to affect it!

Sweet!

PR #350, opened 2019-10-04
1585953: Use bug URLs rather than numbers

mdboom:


mdboom:
I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1586367 for the Rust code coverage test failure, which seems unrelated.

PR #351, opened 2019-10-04
1586002: Fix intermittent test failure related lifecycle owner

mdboom:


mdboom:
> I take it that not holding onto an instance of the lifecycleOwner is the source of the problem?

It's the most likely culprit, but it's an interkitten...

PR #352, opened 2019-10-07
Use fixed version of grcov (and fix arguments)

badboy:
If that unbreaks CI, it needs to land before the other failing PRs.


badboy:
(code coverage is not blocking, so landing other stuff before that is ok)

PR #353, opened 2019-10-07
Include a `forUnitTests` package in the release

badboy:
This will only include a Linux-compatible build for now.
Windows (the easier part) and macOS (the harder part) will follow eventually.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -303,8 +303,14 @@ jobs:
                echo "=================================================="
                exit 1
             fi
-            ARTIFACTS="./build/maven/org/mozilla/telemetry/glean/${RAWVERSION}/"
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -279,16 +282,24 @@ jobs:
 
   android-release:
     docker:
-      - image: circleci/golang:1.13
+      - image: circleci/rust:latest
```
Dexterp37: Are we confident that this would not break anytime soon? Should we pin a version here?

badboy: Yes, we use that image everywhere else as well. In this task we really don't mind using any other image, we don't require any additional language support, I just picked our most used image.

PR #354, opened 2019-10-07
Generate glean autodocs

Dexterp37:
This changes the build script to allow generating docs for Glean internals and from AAR files. This is needed on the a-c side in order to generate docs for GeckoView metrics.


docs/user/collected/metrics.md
```diff
@@ -0,0 +1,63 @@
...
+This is a built-in ping that is assembled out of the box by the Glean SDK.
+See the Glean SDK documentation for the [`baseline` ping](https://mozilla.github.io/glean/book/user/pings/baseline.html).
+The following metrics are added to the ping:
+
+| Name | Type | Description | Data reviews | Extras | Expiration |
+| --- | --- | --- | --- | --- | --- |
+| glean.baseline.duration |[timespan](https://mozilla.github.io/glean/book/user/metrics/timespan.html) |The duration of the last foreground session.  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
+| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES").  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
+
+## glean_client_info
```
Dexterp37: This section was unexpected.. but it's nice :)

glean-core/android/sdk_generator.gradle
```diff
@@ -21,7 +21,7 @@ import org.gradle.api.internal.artifacts.ArtifactAttributes
 // so that it will be shared between all libraries that use Glean.  This is
 // important because it is approximately 300MB in installed size.
 
-String GLEAN_PARSER_VERSION = "1.7.0"
+String GLEAN_PARSER_VERSION = "1.8.3"
```
travis79: We are up to 1.9.0 now...  is that important here to be current?

Dexterp37: Good point. Yes, let me bump this.

docs/user/collected/metrics.md
```diff
@@ -0,0 +1,63 @@
+<!-- AUTOGENERATED BY glean_parser.  DO NOT EDIT. -->
```
badboy: I know this is all auto-generated. Should we still make it clear that these are Android-specific metrics?

docs/user/collected/metrics.md
```diff
@@ -0,0 +1,63 @@
...
+
+# Metrics
+This document enumerates the metrics collected by this project.
+This project may depend on other projects which also collect metrics.
+This means you might have to go searching through the dependency tree to get a full picture of everything collected by this project.
+Sorry about that.
```
badboy: I kinda feel like this "Sorry about that" sounds weird here (I know, it's included in the Glean template, so I should take the discussion there)

docs/user/collected/metrics.md
badboy: This file needs to be added to the SUMMARY.md to get rendered into the final documentation.
I'm also not sure I like the "collected" name for the directory.

I'd be okay with having this as a sub-chapter, so we can make it clear that we have some different metrics on Android on iOS instead of modifying the auto-generated file.

badboy:
(Test failure was fixed in #352, but it's not blocking merging here)

docs/user/collected/metrics.md
Dexterp37: Yes, it's a bit weird to have the directory called `collected`. I could not come up with a better name :( Gladly taking suggestions!

I did not link it to the main docs, because I wasn't sure how we want to handle docs, given that it might have platform differences. This directly boils down to how we want to handle the `metrics.yaml`. If we have one file shared across all the platforms (we don't expect to have many platform specific metrics, don't we?), then this will be the only file.

Do we already have iOS only metrics?

docs/user/collected/metrics.md
Dexterp37: I can file a bug about this.

Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1586730

docs/user/collected/metrics.md
badboy: We have android-only metrics (The Android SDK version)

badboy: `collected-data`, `collected-metrcis`, `data-collection`?

Dexterp37: > 
> 
> We have android-only metrics (The Android SDK version)

Yes, but they are marked as optional for this reason. If there's any iOS specific metric, we could mark it as optional as well, and mention that it is iOS specific. What do you think?

> 
> 
> `collected-data`, `collected-metrcis`, `data-collection`?

`collected-metrics` sounds good.

docs/user/collected-metrics/metrics.md
```diff
@@ -0,0 +1,63 @@
+<!-- AUTOGENERATED BY glean_parser.  DO NOT EDIT. -->
+
+# Metrics
+This document enumerates the metrics collected by this project.
```
mdboom: It would be nice if we could be more specific than "this project".  I would love for this to say "Glean itself".  But that's probably a separate bug for another time...

Dexterp37: We have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1578347) already filed for making that better ;)

Dexterp37:
> 
> 
> I'm not sure we want to include the generated `metrics.md` here for Glean. Most of it is already manually documented [here](https://mozilla.github.io/glean/book/user/pings/index.html), and because they occur in special locations, I think the generated docs are confusing. (We wouldn't see `metrics/glean_client_info/...` for example). Maybe a middle ground here is to update the `glean_parser` doc generator to not include the special categories beginning with `glean_` -- I think in that case the generated docs would no longer be confusing / redundant.

Sounds good to me. I will address this as soon as I have some time!

mdboom:
> Sounds good to me. I will address this as soon as I have some time!

Or file a bug and assign it to me ;)

Dexterp37:
> 
> 
> > Sounds good to me. I will address this as soon as I have some time!
> 
> Or file a bug and assign it to me ;)

Ha! You don't have to ask me twice :-P https://bugzilla.mozilla.org/show_bug.cgi?id=1586749  (thank you, Mike!)

Dexterp37:
> curl: (16) Error in the HTTP2 framing layer

Looks like iOS is failing for unrelated reasons..

PR #355, opened 2019-10-07
Build for Windows

badboy:
This creates a glean_ffi.dll for inclusion in the `forUnitTests` target,
that can then be used in unit tests on Windows machines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -101,6 +101,7 @@ jobs:
     resource_class: "medium+"
     steps:
       - rust-tests
+
   Rust tests - beta:
     docker:
       - image: circleci/rust:latest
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -268,9 +269,31 @@ jobs:
               $CODECOV_CMD || (sleep 5 && $CODECOV_CMD) || (sleep 5 && $CODECOV_CMD)
       - store_artifacts:
           path: glean-core/android/build/reports/tests
+      - run:
+          name: Install packaging dependencies
+          command: |
+            sudo apt update
+            sudo apt install -y gcc-mingw-w64
```
mdboom: Every time I see this I wish one could do something similar for Apple targets...

badboy:
> I take it this just builds the rust FFI functions into a windows DLL? If so, that's pretty much all we would need to be able to glue .Net languages onto it... rust's compile target tooling is very nice.

Yes, it's using the mingw target though (the msvc is recommended I think). It's good enough for our test target.

> This feels kinda magical, in a good way.

Yeah, turns out cross-compiling is not that hard if you have the rest of the toolchain installed. In theory doing the same for mac is not hard either (yey clang!), but _distributing_ the necessary toolchain files is ... not exactly allowed by the license.

badboy:
> For future work, we'll probably also want to make sure the DLL works by linking it from a test app and running it with wine...

Don't give me ideas.

PR #363, opened 2019-10-09
Keep track of all registered pings in test mode

badboy:
In test mode we still need to keep track of pings, so they get re-registered after a reset.
We also only need to have them in that list once.
A list where everything is tracked once is a set. So we use that.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -79,7 +79,7 @@ open class GleanInternalAPI internal constructor () {
     internal lateinit var metricsPingScheduler: MetricsPingScheduler
 
     // Keep track of ping types that have been registered before Glean is initialized.
-    private val pingTypeQueue: MutableList<PingType> = mutableListOf()
+    private val pingTypeQueue: MutableSet<PingType> = mutableSetOf()
 
     // This is used to cache the process state and is used by the function `isMainProcess()`
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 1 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -79,7 +79,7 @@ open class GleanInternalAPI internal constructor () {
     internal lateinit var metricsPingScheduler: MetricsPingScheduler
 
     // Keep track of ping types that have been registered before Glean is initialized.
-    private val pingTypeQueue: MutableList<PingType> = mutableListOf()
+    private val pingTypeQueue: MutableSet<PingType> = mutableSetOf()
```
Dexterp37: Do we need to clear this in `resetGlean`?

badboy: Nooooo, we exactly _don't_ want to clear it.
Consider this test scenario (before this patch is applied), which has the GleanTestRule:

1. `GleanTestRule` enables Glean, baseline/metrics/events ping are registered (builtin).
2. `test1` runs and sends a custom ping: `Pings.customPing.send()`.
  * `Pings.customPing` gets instantiated and registers itself with Glean, it also gets a handle (it calls `glean_new_ping_type`)
  * Test passes, all good
3. `test2` runs, `GleanTestRule` is invoked and resets Glean, deleting all data and other state it has (including registered pings)
4. `test2` sends a custom ping: `Pings.customPing.send()`.
  * The Kotlin object `Pings.customPing` already exists in memory (global object and all), so it doesn't register again in Glean
  * Test fails, because it can't send an unknown ping


Now with this patch we keep a set of all registered pings and can re-register them after a reset.
This is only relevant during testing. For a normal use the Glean object can't be reset and therefore it can't loose track of registered pings (which is why we clear the `pingTypeQueue` in `initialize` if we don't need it anymore)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -605,6 +605,11 @@ open class GleanInternalAPI internal constructor () {
                 handle,
                 pingType.handle
             )
+            // In test mode we still need to keep track of pings, so they get re-registered after a reset
```
Dexterp37: Can you add what you said in your other comment  here?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: Cool :) Let's turn this into a code comment, below :D

PR #374, opened 2019-10-14
Ignore dynamically stored labels if Glean is not initialized 

badboy:


glean-core/src/metrics/labeled.rs
```diff
@@ -177,10 +184,27 @@ where
...
-    pub fn get(&mut self, glean: &Glean, label: &str) -> T {
-        let label = match self.labels {
-            Some(_) => self.static_label(label),
-            None => self.dynamic_label(glean, label),
+    pub fn get<'a, G: Into<Option<&'a Glean>>>(&mut self, glean: G, label: &str) -> T {
+        // We have 3 scenarios to consider:
+        // * Static labels. No database access needed. We just look at what is in memory.
+        // * Dynamic labels, initialized Glean. We look up in the database all previously stored
+        //   labels in order to keep a maximum of allowed labels.
+        // * Dynamic labels, uninitialized Glean. We ignore potentially previously stored labels
```
Dexterp37: Can you reference [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1588451) here?

glean-core/android/src/test/java/mozilla/telemetry/glean/private/AccumulationsBeforeGleanInitTest.kt
```diff
@@ -0,0 +1,89 @@
...
+        )
+
+        labeledCounterMetric["label1"].add(1)
+
+        forceInitGlean()
+
+        assertEquals(1, labeledCounterMetric["label1"].testGetValue())
+    }
+
+    @Ignore
```
Dexterp37: Can you add a note here? `@Ignore("Bug 1588452 will enable it back")` or something like that

glean-core/android/src/test/java/mozilla/telemetry/glean/private/AccumulationsBeforeGleanInitTest.kt
```diff
@@ -33,6 +34,8 @@ class AccumulationsBeforeGleanInitTest {
     @Before
     fun cleanup() {
         Glean.testDestroyGleanHandle()
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.setTaskQueueing(true)
```
Dexterp37: nit: can you add a comment (the one from the commit message is ok) before this line?

This should probably happen in `Glean.resetGlean` as well for other tests. Good catch. Would you mind filing a bug for it?

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1588463

mdboom: We may also want to clear the queue to be extra sure...

badboy: Yes, but blindly clearing breaks other tests. We should take that in a follow-up, see the filed bug for it.

mdboom: "Fun"

PR #369, opened 2019-10-11
Implement timing distribution in Swift

badboy:

PR #375, opened 2019-10-14
Implement memory distribution in Swift

badboy:
Depends on #369 to be merged first, then rebased.


docs/user/metrics/memory_distribution.md
```diff
@@ -51,9 +55,44 @@ val snapshot = Memory.heapAllocated.testGetValue()
...
+XCTAssertEqual(11, snapshot.sum)
+
+// Usually you don't know the exact memory values, but how many should have been recorded.
+XCTAssertEqual(2, snapshot.count)
+```
+
+</div>
+
+{{#include ../../tab_footer.md}}
+
```
mdboom: We should start adding links to the Swift API docs for the class below in "Reference".  (And [this bug] https://bugzilla.mozilla.org/show_bug.cgi?id=1584990) can be used to do the backfill on that.

badboy: yeah, let's do that in one swipe in the backfill

PR #377, opened 2019-10-15
Add instrumented tests to sample app

brizental:
Related to [Bug 1570252](https://bugzilla.mozilla.org/show_bug.cgi?id=1570252)

I basically just copy pasted the tests from the sample app in android components, so they should do the same thing.

Unfortunately (or fortunately depending on how you look at it) this tests are failing, because while adding them to the project we found two bugs in the actual glean app.

@badboy reported one of them ([Bug 1588717](https://bugzilla.mozilla.org/show_bug.cgi?id=1588717)) and is still loking into the other one.

The one bug that is already reported is also happening for the instrumented tests in android-components. Which means that they are not probably run on CI there, as well.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -34,6 +34,7 @@ buildscript {
         androidx_lifecycle_extensions: '2.1.0',
         androidx_test: '1.2.0',
         androidx_work: '2.2.0',
+        androidx_uiautomator: '2.2.0',
     ]
 
     ext.build = [
--- a/samples/android/app/README.md
+++ b/samples/android/app/README.md
@@ -2,6 +2,14 @@
 
 A simple app showcasing the Glean SDK.
 
+## Running tests
+
+1. [Setup the Android Build Environment](https://mozilla.github.io/glean/book/dev/android/setup-android-build-environment.html?highlight=android#setup-the-android-build-environment)
+2. Run `./gradlew :glean-sample-app:connectedAndroidTest`
+
+Since the tests are instrumented they need a connected device or emulator
+present to complete.
+
 ## License
 
     This Source Code Form is subject to the terms of the Mozilla Public
--- a/samples/android/app/build.gradle
+++ b/samples/android/app/build.gradle
@@ -20,7 +20,7 @@ android {
         versionCode 1
         versionName "1.0"
 
-        testInstrumentationRunner "android.support.test.runner.AndroidJUnitRunner"
+        testInstrumentationRunner "org.mozilla.samples.gleancore.GleanTestRunner"
     }
 
     buildTypes {
@@ -38,6 +38,15 @@ dependencies {
 
     implementation "androidx.appcompat:appcompat:$rootProject.versions.androidx_appcompat"
     implementation "androidx.browser:browser:$rootProject.versions.androidx_browser"
+
+    androidTestImplementation "androidx.test:core-ktx:$rootProject.versions.androidx_test"
+    androidTestImplementation "androidx.test:runner:$rootProject.versions.androidx_test"
+    androidTestImplementation "androidx.test:rules:$rootProject.versions.androidx_test"
+    androidTestImplementation "androidx.test.ext:junit:$rootProject.versions.androidx_junit"
+    androidTestImplementation "androidx.test.uiautomator:uiautomator:$rootProject.versions.androidx_uiautomator"
+    androidTestImplementation "androidx.test.espresso:espresso-core:$rootProject.versions.androidx_espresso"
+    androidTestImplementation "androidx.work:work-testing:$rootProject.versions.androidx_work"
+    androidTestImplementation "com.squareup.okhttp3:mockwebserver:$rootProject.versions.mockwebserver"
 }
 
 ext.gleanNamespace = "mozilla.telemetry.glean"
--- a/samples/android/app/metrics.yaml
+++ b/samples/android/app/metrics.yaml
@@ -8,20 +8,92 @@
 
 $schema: moz://mozilla.org/schemas/glean/metrics/1-0-0
 
+browser.engagement:
+  click:
+    type: event
+    description: >
+      Just testing events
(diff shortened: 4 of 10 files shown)


badboy:
Ooooh, this is now failing the lint tests for the generated files.
Should we fix that or do we just want to ignore generated files completely, @Dexterp37?

badboy:
While we're at it: we should probably document how to run them (`./gradlew connectedAndroidTest`) and/or add a make task for it.

Dexterp37:
> 
> 
> Ooooh, this is now failing the lint tests for the generated files.
> Should we fix that or do we just want to ignore generated files completely, @Dexterp37?

I think this should be fixed: at the parsers level?

brizental:
> we should probably document how to run them
@badboy I just added a commit adding some instructions on how to run the instrumented tests. Nothing super detailed... Let me know if it lookgs good :)

Dexterp37:
@badboy @mdboom and @brizental what if we changed [this line](https://github.com/mozilla/glean/blob/3031fecaf4652d39e9a6539905c64156b97b61f7/glean-core/android/src/main/java/mozilla/telemetry/glean/utils/ThreadUtils.kt#L10) to make it visible from tests/make it overridable? Then we could just set that to the main thread value in the instrumented tests and be done with it :)

badboy:
Seems hacky, but doable. I'd still like to find _any_ documentation whether this behavior is expected, but I was unable to find anything so far :(

mdboom:
> @badboy @mdboom and @brizental what if we changed this line to make it visible from tests/make it overridable? Then we could just set that to the main thread value in the instrumented tests and be done with it :)

Yeah, I guess that would work.  I share @badboy's desire for documentation, though, but maybe it just doesn't exist.

Dexterp37:
> 
> 
> @badboy @mdboom and @brizental what if we changed [this line](https://github.com/mozilla/glean/blob/3031fecaf4652d39e9a6539905c64156b97b61f7/glean-core/android/src/main/java/mozilla/telemetry/glean/utils/ThreadUtils.kt#L10) to make it visible from tests/make it overridable? Then we could just set that to the main thread value in the instrumented tests and be done with it :)

Ah, scrape that. This would work in mozilla/glean, but not in a-c. How would we fix it there?

Dexterp37:
> 
> 
> > Ah, scrape that. This would work in mozilla/glean, but not in a-c. How would we fix it there?
> 
> Could we skip the assertion check in `Glean.initialize` if `Dispatchers.testingMode` is true?

Integration tests should not be in test mode, I think

mdboom:
> Integration tests should not be in test mode, I think

Ah -- it's quite possible I just don't understand the moving parts.  I assumed since these tests use the `GleanTestRule` that `testingMode` would be set...

Dexterp37:
> 
> 
> > Ah -- it's quite possible I just don't understand the moving parts. I assumed since these tests use the `GleanTestRule` that `testingMode` would be set...
> 
> I ... somehow didn't realize that.
> But I also think the application is launched once first before the test runner activates the rule and therefore that already fails?
> 
> I need to double check that again.

Garn, you're both right and I should have taken the full day off. We're in test mode. Sorry for the noise @mdboom and @badboy!

badboy:
> Garn, you're both right and I should have taken the full day off. We're in test mode. 

No, absolutely helpful to raise this, because ... I wrote and merged #378 exactly because something was loaded before test mode.

Dexterp37:
I've slightly worked around this in mozilla-mobile/android-components#4786 -  The solution there, if accepted, should be ported here as well (for the newly added rule).

brizental:
> Is there a bug for running these automatically on CircleCI later?

I thought there was but I couldn't find it :/

Dexterp37:
> 
> 
> This looks great! One thing that we may want to file a follow up bug for is we now have a lot of metrics that expire on `2100-01-01`, which may promote some bad habits. @Dexterp37 pointed this out to me last week on something I was working on so let's see what he says here.

Hey @travis79 that's true, good catch. However, I'm afraid in this specific case there isn't much we could do: this is an instrumented test that runs on real metrics (and potentially real devices). So we either set the date to a closer one and live with the fact that this might break soon-ish due to expiration, or just set it as a long distant future date and be done with it :)

I think I suggested to change that in your article, which serves a different purpose: that's a narrative for creating Glean SDK based stuff.

travis79:
> > This looks great! One thing that we may want to file a follow up bug for is we now have a lot of metrics that expire on `2100-01-01`, which may promote some bad habits. @Dexterp37 pointed this out to me last week on something I was working on so let's see what he says here.
> 
> Hey @travis79 that's true, good catch. However, I'm afraid in this specific case there isn't much we could do: this is an instrumented test that runs on real metrics (and potentially real devices). So we either set the date to a closer one and live with the fact that this might break soon-ish due to expiration, or just set it as a long distant future date and be done with it :)
> 
> I think I suggested to change that in your article, which serves a different purpose: that's a narrative for creating Glean SDK based stuff.

No worries, 80 years just seemed like an optimistic lifespan of the Glean library.  

With this being an example that people may look at to figure out how to integrate Glean, I thought it might have the same visibility as my article, which is what led me to point it out.  

I definitely don't want to block on this, but if we added this to the probe-scraper dictionary we could get notifications when a more realistic date was about to expire, even if that was 5 years in the future, rather than 80.

Dexterp37:
> With this being an example that people may look at to figure out how to integrate Glean, I thought it might have the same visibility as my article, which is what led me to point it out.

I think (and hope!) that your article will be viewed much more than these tests :)

> I definitely don't want to block on this, but if we added this to the probe-scraper dictionary we could get notifications when a more realistic date was about to expire, even if that was 5 years in the future, rather than 80.

I'd honestly be a bit hesitant to add this example to the probe-scraper (which gleans the data for the probe-dictionary). It's basically a test.

PR #378, opened 2019-10-15
Keep track of pings in all modes

badboy:
We need to keep track of pings, so they get re-registered after a reset.
This state is kept across Glean resets, which should only ever happen in test mode.
Or by the instrumentation tests (`connectedAndroidTest`), which relaunches the application activity,
but not the whole process, meaning globals, such as the ping types, still exist from the old run.

I'm not super happy with keeping this state around, but given that the
queue is actually a set and it's unlikely to have hundreds and hundreds
of custom pings it should not have much of an impact.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -605,7 +605,7 @@ open class GleanInternalAPI internal constructor () {
         // This state is kept across Glean resets, which should only ever happen in test mode.
         // Or by the instrumentation tests (`connectedAndroidTest`), which relaunches the application activity,
         // but not the whole process, meaning globals, such as the ping types, still exist from the old run.
-        // It's a set and keeping them around forver should not have much of an impact.
+        // It's a set and keeping them around forever should not have much of an impact.
         pingTypeQueue.add(pingType)
     }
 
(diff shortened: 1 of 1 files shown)


badboy:
cc @brizental

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -598,23 +594,19 @@ open class GleanInternalAPI internal constructor () {
...
-            if (Dispatchers.API.testingMode) {
-                pingTypeQueue.add(pingType)
-            }
         }
+
+        // We need to keep track of pings, so they get re-registered after a reset.
+        // This state is kept across Glean resets, which should only ever happen in test mode.
+        // Or by the instrumentation tests (`connectedAndroidTest`), which relaunches the application activity,
+        // but not the whole process, meaning globals, such as the ping types, still exist from the old run.
+        // It's a set and keeping them around forver should not have much of an impact.
```
Dexterp37: Would it make sense to try to reset in the `GleanTestRunner`?

mdboom:
```suggestion
        // It's a set and keeping them around forever should not have much of an impact.
```

mdboom: Yeah, I'm too am curious if we can fix this in a testing-only way...

badboy: I considered it, but decided against it for reasons:

* Low overhead even in production
* It's yet another configuration/state the GleanTestRunner runs in compared to unit tests and production mode.
* I'd like the tests to run as close to production as possible, otherwise we might miss other problems
* It was easier to remove code than to add :P

PR #381, opened 2019-10-16
Use glean-core specific full path for debug activity in the docs

badboy:
Probably shouldn't land before we land glean-core in a-c.

BREAKING CHANGE:

When used through Android Components,
previously users had to use the `mozilla.components.service.glean.debug.GleanDebugActivity` name.

It's now required to use `mozilla.telemetry.glean.debug.GleanDebugActivity``


badboy:
@mdboom what do you think about the note?

PR #384, opened 2019-10-16
1588451, 1530733: Handle dynamic labels in coroutine tasks

mdboom:
At present, the determination of (a) whether there are too many dynamic labels
on a metric, or (b) the label is invalid, are handled synchronously in the
`get()` operator (in Kotlin, the bracket operator).

This requires the Glean object to already be initialized at metric recording
time -- both to see how many labels have already been used, and also to record
any errors.  This both means that labeled metrics can't be recorded prior to
initialization (even though most metrics now can), but more importantly,
blocking I/O is performed to determine how many labels have already been
recorded for the metric.

This PR changes how things work so the *requested* dynamic label is stored on
the specific labeled metric as a new field. This dynamic label is then converted
to its *actual* value later as part of the coroutine task, at which point we are
guaranteed to have an initialized Glean object.

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -92,7 +92,7 @@ impl CommonMetricData {
     ///
     /// If `category` is empty, it's ommitted.
     /// Otherwise, it's the combination of the metric's `category` and `name`.
-    pub fn base_identifier(&self) -> String {
+    pub(crate) fn base_identifier(&self) -> String {
         if self.category.is_empty() {
             self.name.clone()
         } else {
@@ -104,7 +104,7 @@ impl CommonMetricData {
     ///
     /// If `category` is empty, it's ommitted.
     /// Otherwise, it's the combination of the metric's `category`, `name` and `label`.
-    pub fn identifier(&self, glean: &Glean) -> String {
+    pub(crate) fn identifier(&self, glean: &Glean) -> String {
         let base_identifier = self.base_identifier();
 
         if let Some(label) = &self.dynamic_label {
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -66,8 +66,10 @@ pub struct CommonMetricData {
     ///
     /// Disabled metrics are never recorded.
     pub disabled: bool,
-    /// Dynamic label. These need to be looked up in the database to make sure
-    /// we aren't storing too many. (Static labels are stored as part of name).
+    /// Dynamic label.
+    /// When a LabeledMetric<T> factory creates the specific metric to be
+    /// recorded to, dynamic labels are stored in the specific label so that we
+    /// can validate them when the Glean singleton is available.
     pub dynamic_label: Option<String>,
 }
 
@@ -106,11 +108,7 @@ impl CommonMetricData {
         let base_identifier = self.base_identifier();
 
         if let Some(label) = &self.dynamic_label {
-            format!(
-                "{}/{}",
-                base_identifier,
-                dynamic_label(glean, self, &base_identifier, label)
-            )
+            dynamic_label(glean, self, &base_identifier, label)
         } else {
             base_identifier
         }
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -5,6 +5,8 @@
 use std::convert::TryFrom;
 
 use crate::error::{Error, ErrorKind};
+use crate::metrics::dynamic_label;
+use crate::Glean;
 
 /// The supported metrics' lifetimes.
 ///
@@ -64,6 +66,9 @@ pub struct CommonMetricData {
     ///
     /// Disabled metrics are never recorded.
     pub disabled: bool,
+    /// Dynamic label. These need to be looked up in the database to make sure
+    /// we aren't storing too many. (Static labels are stored as part of name).
+    pub dynamic_label: Option<String>,
 }
 
 impl CommonMetricData {
@@ -81,15 +86,33 @@ impl CommonMetricData {
         }
     }
 
-    /// The metric's unique identifier.
+    /// The metric's base identifier, including the category and name, but not the label.
     ///
-    /// If `category` is empty, it's just the name.
+    /// If `category` is empty, it's ommitted.
     /// Otherwise, it's the combination of the metric's `category` and `name`.
-    pub fn identifier(&self) -> String {
+    pub fn base_identifier(&self) -> String {
         if self.category.is_empty() {
             self.name.clone()
         } else {
             format!("{}.{}", self.category, self.name)
+        }
+    }
+
+    /// The metric's unique identifier, including the category, name and label.
+    ///
+    /// If `category` is empty, it's ommitted.
+    /// Otherwise, it's the combination of the metric's `category`, `name` and `label`.
+    pub fn identifier(&self, glean: &Glean) -> String {
+        let base_identifier = self.base_identifier();
+
+        if let Some(label) = &self.dynamic_label {
+            format!(
+                "{}/{}",
+                base_identifier,
+                dynamic_label(glean, self, &base_identifier, label)
+            )
+        } else {
+            base_identifier
         }
     }
 
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -11,6 +11,7 @@ use rkv::{Rkv, SingleStore, StoreOptions};
 
 use crate::metrics::Metric;
 use crate::CommonMetricData;
+use crate::Glean;
 use crate::Lifetime;
 use crate::Result;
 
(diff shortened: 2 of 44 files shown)


glean-core/src/metrics/labeled.rs
```diff
@@ -184,32 +101,82 @@ where
...
             }
-        };
-        let label = format!("{}/{}", self.submetric.meta().name, label);
+            None => self.new_metric_with_dynamic_label(label.to_string()),
+        }
+    }
+}
 
-        self.new_metric_with_name(label)
+/// Correct a dynamic label
```
Dexterp37: nit: "Validates a dynamic label"?

glean-core/src/common_metric_data.rs
```diff
@@ -64,6 +66,9 @@ pub struct CommonMetricData {
     ///
     /// Disabled metrics are never recorded.
     pub disabled: bool,
+    /// Dynamic label. These need to be looked up in the database to make sure
+    /// we aren't storing too many. (Static labels are stored as part of name).
+    pub dynamic_label: Option<String>,
```
Dexterp37: Why can't this live in the `LabeledMetricType`?

glean-core/src/common_metric_data.rs
```diff
@@ -81,18 +86,36 @@ impl CommonMetricData {
...
         } else {
             format!("{}.{}", self.category, self.name)
         }
     }
 
+    /// The metric's unique identifier, including the category, name and label.
+    ///
+    /// If `category` is empty, it's ommitted.
+    /// Otherwise, it's the combination of the metric's `category`, `name` and `label`.
+    pub fn identifier(&self, glean: &Glean) -> String {
```
Dexterp37: Instead of changing this to require `glean` for labels, given that it has repercussions on the whole codebase, what about introducing a new `identifier_with_label` and only require the Glean object for that function?

glean-core/src/metrics/labeled.rs
```diff
@@ -2,15 +2,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use std::collections::HashSet;
-
+//! Module documentation
```
Dexterp37: nit: is this a placeholder?

glean-core/src/metrics/labeled.rs
```diff
@@ -184,32 +101,82 @@ where
...
+/// Returns the valid label if within the limitations,
+/// or `OTHER_LABEL` on any errors.
+/// The errors are logged.
+pub fn dynamic_label<'a>(
+    glean: &Glean,
+    meta: &CommonMetricData,
+    base_identifier: &'a str,
+    label: &'a str,
+) -> &'a str {
+    let key = format!("{}/{}", base_identifier, label);
```
Dexterp37: We can use the function we have to generate this identifier

glean-core/src/metrics/labeled.rs
mdboom: I changed the word because it really changes it, not just reporting when it's wrong.

Maybe "Validates a dynamic label, changing it to __other__ if it's invalid." ??

glean-core/src/common_metric_data.rs
mdboom: It needs to exist in every metric type that can be labeled.  So on `counter`, not `labeled_counter`.

`LabeledMetricType` is basically a factory for a regular metric type, if that helps.

The `labeled_counter` creates a specific copy of a `counter` with the label value stored so that later when the actual recording happens in a coroutine we can do the validation and correction of that label.

glean-core/src/metrics/labeled.rs
mdboom: Yes -- in an earlier version I made this whole module public which then required a docstring.  I rolled that back so this should just be removed.

glean-core/src/common_metric_data.rs
mdboom: I thought about this. Strictly speaking, you only need to call this version on metrics that can be labeled.  But I thought that would just introduce confusion about when one should be used over another and lead to unexpected bugs...  I'm happy to change this, though, if you feel strongly.

glean-core/src/metrics/labeled.rs
mdboom: Not without creating infinite recursion :)  But I can probable refactor this a bit better.

glean-core/src/metrics/labeled.rs
Dexterp37: Yes, good idea, I like your new phrasing more!

glean-core/src/metrics/labeled.rs
Dexterp37: Gotcha :) Forget about that. Would you kindly add a comment before that line hinting at the infinite recursion? :)

glean-core/src/common_metric_data.rs
Dexterp37: I see. Much clearer now, thanks for bearing with me on this. Can you integrate the comment above this line with what you said here?

glean-core/src/metrics/labeled.rs
mdboom: I refactored this so that the combination of base identifiers and labels all happens in a single function (which I think is better for maintenance anyway).  So no comment about infinite recursion is necessary...

glean-core/src/common_metric_data.rs
```diff
@@ -81,18 +88,32 @@ impl CommonMetricData {
...
```
badboy:
```suggestion
    pub(crate) fn identifier(&self, glean: &Glean) -> String {
```

glean-core/src/database/mod.rs
```diff
@@ -68,11 +69,16 @@ impl Database {
...
     /// Iterates with the provided transaction function on the data from
     /// the given storage.
-    pub fn iter_store_from<F>(&self, lifetime: Lifetime, storage_name: &str, mut transaction_fn: F)
-    where
+    pub fn iter_store_from<F>(
+        &self,
+        lifetime: Lifetime,
+        storage_name: &str,
+        metric_key: Option<&str>,
+        mut transaction_fn: F,
```
badboy: We're getting lots of arguments here.
Can we document what they mean?
Right now I know what `metric_key` is. I won't know when I look at this again on Monday.

glean-core/src/database/mod.rs
```diff
@@ -109,6 +115,24 @@ impl Database {
         }
     }
 
+    /// Returns `True` if the storage has the given metric.
```
badboy: I usually phrase the documentation on things like these slightly different (the true/false is implied and denoted by the return type):

> Determine if the storage has the given metric

badboy: (then again Rust's own libstd documentation uses the same way as you did here, I would just prefer `true` [lower case], we're not Python 🐍 here :P)

glean-core/src/metrics/labeled.rs
```diff
@@ -184,32 +100,98 @@ where
...
+///
+/// * `label` - The requested label
+///
+/// ## Return value
+///
+/// Returns the entire identifier for the metric, including the base identifier and the
+/// corrected label.
+/// The errors are logged.
+pub fn dynamic_label(
+    glean: &Glean,
```
badboy: Looking here and in the database implementation again: Do we really need `Glean`? Or do we just need `&Database`? This way the database can pass `self` and everyone else passed `glean.storage()` and `.record` doesn't need to pass a new thing at all

glean-core/src/database/mod.rs
```diff
@@ -126,8 +150,8 @@ impl Database {
     }
 
     /// Records a metric in the underlying storage system.
-    pub fn record(&self, data: &CommonMetricData, value: &Metric) {
-        let name = data.identifier();
+    pub fn record(&self, glean: &Glean, data: &CommonMetricData, value: &Metric) {
+        let name = data.identifier(glean);
```
badboy: As mentioned in the other part, if identifier requires `&Database` instead of Glean, we could pass `self` here and don't take the `glean: &Glean` argument.

glean-core/src/common_metric_data.rs
Dexterp37: Fine by me then, with the changes by @badboy

glean-core/src/metrics/labeled.rs
Dexterp37: That's a killer suggestion! I like it! And would definitely make it easier to digest touching so many files :)

glean-core/src/database/mod.rs
mdboom: I changed it as you suggested in the first place to sidestep the whole issue.

glean-core/src/metrics/labeled.rs
mdboom: We need `glean` to record errors.  (Unless we refactored the error recording code to only require a database, but that would mean we couldn't use "real" labeled counters for error recording, we'd have to sort of duplicate that code directly on top of the database API).

badboy: *goes off to rework the whole error reporting system* Hmmmmmm

mdboom: So consensus is to refactor the error recording code to make this possible?

badboy: uhm, I'd rather not. Not before merging to a-c. That would be a significant change (and I don't even know if it's possible).

One other thing would be to return a `Result<String, (String, String)>` all the way up and do the error recording in another place, but that also seems ugly and will need to touch a couple of places to make it work.

badboy: I think for now we don't have a better way than to keep it like you did, file a bug and take a look at it again in a week or two once we're more open for bigger changes again.

badboy: (This is all under the premise that we want to get this in before a first release next week, as one of the last big changes, this kinda affects the current QA process)

mdboom: Ok.  Fair enough.  Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1589675)

PR #385, opened 2019-10-17
Update the Glean ping schema revision to a recent version

Dexterp37:
I noticed that we were using a quite old version in our tests. I updated the version to the latest release.


mdboom:
Cool.  We should do the same thing for android-components.  Maybe now's a good time before the migration?

Dexterp37:
> 
> 
> Cool. We should do the same thing for android-components. Maybe now's a good time before the migration?

Good point. On it!

PR #389, opened 2019-10-17
Update documentation to reflect both supported platforms everywh…

badboy:
[This is bug 1584990](https://bugzilla.mozilla.org/show_bug.cgi?id=1584990)

This PR does multiple things:

* Adds links to the Swift API reference for Glean on all metric types
* Splits the getting started chapter into an Android and an iOS part
	* Sub-chapters because they are vastly different and quite long
* Ensures that examples for testing metrics and other parts contain both Kotlin and Swift
	* Inline with the language switcher as they are mostly similar

It currently does not have the full docs on how to do the metric generation at build time for iOS yet, mostly because I need to write it and check against the sample app, but also because it's not in its final state just yet.

I think this can already get a look though.
If we want to land this right now, we might want to add a big notice on top of the "Getting started" page for iOS making it clear that it's not final yet.

### DIFF
--- a/docs/user/adding-glean-to-your-project.md
+++ b/docs/user/adding-glean-to-your-project.md
@@ -13,6 +13,10 @@ Products using the Glean SDK to collect telemetry **must**:
 
 ### Integrating with your project
 
+{{#include ../tab_header.md}}
+
+<div data-lang="Kotlin" class="tab">
+
 #### Setting up the dependency
 
 Use Gradle to download the library from [maven.mozilla.org](https://maven.mozilla.org/) ([Setup repository](https://github.com/mozilla-mobile/android-components/blob/master/README.md#maven-repository)) by adding the following to your Gradle configuration:
@@ -58,20 +62,43 @@ apply from: 'https://raw.githubusercontent.com/mozilla-mobile/android-components
 
 There are [additional parameters](android-build-configuration-options.md) that can be set to control the behavior of the `sdk_generator.gradle` script, but they are rarely used in normal use.
 
+</div>
+
+<div data-lang="Swift" class="tab">
+
+#### Setting up the dependency
+
+Glean can be consumed through [Carthage](https://github.com/Carthage/Carthage), a dependency manager for macOS and iOS.
+For consuming the latest version of Glean, add the following line to your `Cartfile`::
+
+```
+github "mozilla/glean" "master"
+```
+
+#### Integrating with the build system
+
+> **Note:** This is not yet documented, as the exact mechanism is not finalized. See [Bug 1589383](https://bugzilla.mozilla.org/show_bug.cgi?id=1589383) for more information.
+
+</div>
+
+{{#include ../tab_footer.md}}
+
 ### Adding new metrics
 
 All metrics that your project collects must be defined in a `metrics.yaml` file.
-This file should be at the root of the module (the same directory as the `build.gradle` file you updated).
+Add this file to your project and define it as an input file for the `sdk_generator.sh` script in the `Run Script` step defined before.
 The format of that file is documented [with `glean_parser`](https://mozilla.github.io/glean_parser/metrics-yaml.html).
 To learn more, see [adding new metrics](adding-new-metrics.md).
 
-> **Important**: as stated [above](#before-using-glean), any new data collection requires documentation and data-review. This is also required for any new metric automatically collected by the Glean SDK.
+> **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
+> This is also required for any new metric automatically collected by the Glean SDK.
 
 ### Adding custom pings
 
 Please refer to the [custom pings documentation](pings/custom.md).
 
-> **Important**: as stated [above](#before-using-glean), any new data collection, including new custom pings, requires documentation and data-review. This is also required for any new ping automatically collected by the Glean SDK.
+> **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
+> This is also required for any new metric automatically collected by the Glean SDK.
 
 ### Testing metrics
 
@@ -87,14 +114,17 @@ These specific steps are described in [the `probe_scraper` documentation](https:
 
 The following steps are required for applications using the Glean SDK, but not libraries.
 
+{{#include ../tab_header.md}}
+
+<div data-lang="Kotlin" class="tab">
+
 ### Initializing the Glean SDK
 
 The Glean SDK should only be initialized from the main application, not individual libraries.  If you are adding Glean support to a library, you can safely skip this section.
 Please also note that the Glean SDK does not support use across multiple processes, and must only be initialized on the application's main process. Initializing in other processes is a no-op.
 Additionally, Glean must be initialized on the main (UI) thread of the applications main process. Failure to do so will throw an `IllegalThreadStateException`.
 
-Before any data collection can take place, the Glean SDK **must** be initialized from the application.
-An excellent place to perform this operation is within the `onCreate` method of the class that extends Android's `Application` class.
+An excellent place to initialize Glean is within the `onCreate` method of the class that extends Android's `Application` class.
 
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Pings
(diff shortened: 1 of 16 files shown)


docs/user/adding-new-metrics.md
```diff
@@ -1,7 +1,7 @@
 # Adding new metrics
 
-All metrics that your project collects must be defined in a `metrics.yaml` file. 
-This file should be at the root of the application or library module (the same directory as the `build.gradle` file you updated). 
+All metrics that your project collects must be defined in a `metrics.yaml` file.
+This file should be at the root of the application or library module (the same directory as the `build.gradle` file you updated).
```
Dexterp37: This is Android specific, as there's no `build.gradle` for iOS. I think I'm fine with removing the sentence between the parentheses completely.

docs/user/adding-new-metrics.md
```diff
@@ -48,7 +48,7 @@ category2.subcategory:  # Categories can contain subcategories
 
 The details of the metric parameters are described in [metric parameters](metric-parameters.md).
 
-The `metrics.yaml` file is used to generate `Kotlin` code that becomes the public API to access your application's metrics.
+The `metrics.yaml` file is used to generate `Kotlin` or `Swift` code that becomes the public API to access your application's metrics.
```
Dexterp37: Since we'll be adding more bindings soon-ish, I'll make this a bit more general. What about `... is used to generate code in the target language (e.g. Kotlin, Swift, ...) that becomes ...`

docs/user/get-started/ios.md
```diff
@@ -0,0 +1,105 @@
...
+
+```
+github "mozilla/glean" "master"
+```
+
+#### Integrating with the build system
+
+In order for the Glean SDK to generate an API for your metrics, a Python environment must be accessible at build time.
+
+**TODO: Fill in how to add it to the build system**
```
Dexterp37: Do we have a bug for this? If so, can we reference it here?

docs/user/adding-new-metrics.md
travis79: The `build.gradle` directory instructions seems to be very Android specific, while this part of the docs seems to be referring to all platforms.

docs/user/get-started/ios.md
```diff
@@ -0,0 +1,105 @@
...
+> **Important**: as stated [before](../adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
+> This is also required for any new metric automatically collected by the Glean SDK.
+
+### Testing metrics
+
+In order to make testing metrics easier 'out of the box', all metrics include a set of test API functions in order to facilitate unit testing.  These include functions to test whether a value has been stored, and functions to retrieve the stored value for validation.  For more information, please refer to [Unit testing Glean metrics](../testing-metrics.md).
+
+### Adding metadata about your project to the pipeline
+
+In order for data to be collected from your project, metadata must be added to `probe_scraper`.
```
Dexterp37: This is out of scope for this bug, but do you think we should sync up this section with https://docs.telemetry.mozilla.org/concepts/glean/glean.html#how-to-use-glean ? We require people to file data-engineering bugs for setting up these things, I think.

@fbertsch, how do we want this process to be? When adding the SDK to an app, do we want people to file bugs and we take care of the rest?

docs/user/get-started/ios.md
travis79: File a bug and reference here so we don't forget it, please!

docs/user/adding-new-metrics.md
badboy: Ah, I changes this somewhere else. Will change it here.

docs/user/get-started/ios.md
badboy: I plan to fill this in a bit before landing, but I'll also file a bug for expanding.

docs/user/get-started/ios.md
travis79: This makes the documentation very _Mozilla_ specific.  Should we be careful of this if Glean is going to be used by third-parties?

docs/user/get-started/ios.md
```diff
@@ -0,0 +1,105 @@
...
+
+The following steps are required for applications using the Glean SDK, but not libraries.
+
+### Initializing the Glean SDK
+
+The Glean SDK should only be initialized from the main application, not individual libraries.
+If you are adding Glean support to a library, you can safely skip this section.
+Please also note that the Glean SDK does not support use across multiple processes, and must only be initialized on the application's main process.
+
+Before any data collection can take place, the Glean SDK **must** be initialized from the application.
```
travis79: Maybe out of scope for this but is this _really_ true?  I know we would like them to initialize Glean as early as possible but we can still queue data collection tasks before init.

docs/user/get-started/android.md
```diff
@@ -0,0 +1,137 @@
...
+
+The following steps are required for applications using the Glean SDK, but not libraries.
+
+### Initializing the Glean SDK
+
+The Glean SDK should only be initialized from the main application, not individual libraries.  If you are adding Glean support to a library, you can safely skip this section.
+Please also note that the Glean SDK does not support use across multiple processes, and must only be initialized on the application's main process. Initializing in other processes is a no-op.
+Additionally, Glean must be initialized on the main (UI) thread of the applications main process. Failure to do so will throw an `IllegalThreadStateException`.
+
+Before any data collection can take place, the Glean SDK **must** be initialized from the application.
```
travis79: Maybe out of scope for this bug, but is this _really_ true?  Considering we have the queuing for pre-init tasks, is it worth specifically calling out that we can't do something that we sort of can?

docs/user/testing-metrics.md
```diff
@@ -39,7 +47,58 @@ This function will return a datatype appropriate to the specific type of the met
...
+    // ...
+}
+```
+
+This will ensure that metrics are done recording when the other test functions are used.
+
+To check if a value exists (i.e. it has been recorded), there is a `testHasValue()` function on each of the metric instances:
+
+```Swift
+XCTAsserTrue(GleanMetrics.Search.defaultSearchEngineUrl.testHasValue())
```
travis79:
```suggestion
XCTAssertTrue(GleanMetrics.Search.defaultSearchEngineUrl.testHasValue())
```

docs/user/testing-metrics.md
```diff
@@ -84,3 +148,34 @@ assertEquals(3, events.size)
...
+```Swift
+// Record a metric value with extra to validate against
+GleanMetrics.BrowserEngagement.click.record([.font: "Courier"])
+
+// Record more events without extras attached
+BrowserEngagement.click.record()
+BrowserEngagement.click.record()
+
+// Check if we collected any events into the 'click' metric
+XCTAsserTrue(BrowserEngagement.click.testHasValue())
```
travis79:
```suggestion
XCTAssertTrue(BrowserEngagement.click.testHasValue())
```

docs/user/get-started/ios.md
Dexterp37: Good catch! This is indeed not true anymore, thanks to your work.

docs/user/get-started/ios.md
Dexterp37: This is a good point. However, right now, our docs are heavily *Mozilla* specific. For example, all of our docs assume we're being consumed through `mozilla.components.service.glean` (so through our A-C shim). We might want to document that in the future. I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1589392) for it.

docs/user/adding-new-metrics.md
```diff
@@ -107,5 +114,30 @@ is accessible in Kotlin as:
...
+```YAML
+views:
+  login_opened:
+    ...
+```
+
+is accessible in Kotlin as:
+
+```Swift
+GleanMetrics.Views.loginOpened...
```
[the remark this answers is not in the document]
travis79: Shouldn't need to import it as long as it's in the same application.

docs/user/adding-glean-to-your-project.md
```diff
@@ -11,132 +11,8 @@ Products using the Glean SDK to collect telemetry **must**:
 
 ## Usage
 
-### Integrating with your project
+Glean is supported on Android an iOS.
+The setup differs between these two platforms. Check the corresponding chapters to get started.
```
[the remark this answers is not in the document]
badboy: Now using language tabs on a single page

PR #390, opened 2019-10-17
1589504: Fix logic handling timeout of queued tasks

mdboom:


glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -144,12 +144,14 @@ internal object Dispatchers {
             // wait up to 5 seconds for it to complete, otherwise we will reset the flag so that
             // new tasks may continue to run.
             runBlocking {
-                withTimeoutOrNull(QUEUE_PROCESSING_TIMEOUT_MS) {
+                if (withTimeoutOrNull(QUEUE_PROCESSING_TIMEOUT_MS) {
                     job.join()
-                }?.let {
+                } == null) {
```
travis79: you could also do something like this:

```Kotlin
withTimeoutOrNull(QUEUE_PROCESSING_TIMEOUT_MS) {
    ...
} ?: {
	// Executes if the left hand of the elvis operator is null
}
```

mdboom: I tried the elvis operator, but that only works for expressions.  This doesn't return a result so the compiler complains.

PR #395, opened 2019-10-21
Relax naming of enums and classes as generated by glean_parser

badboy:
This might be a short-term fix until we suppress those in the
glean_parser itself.


badboy:
@mdboom We decided to have this right now in Glean, so that we can merge #377 and have CI green.
We have  mozilla/glean_parser#129  as well as [Bug 1590080](https://bugzilla.mozilla.org/show_bug.cgi?id=1590080) to decide the way forward, so that it reaches all users.

Neither a-c nor Fenix currently fail on this, because they already apply some rules (that probably just ignore those files? Not 100% sure, I didn't understand the config).

Under these conditions, would you be fine in merging here for now?

mdboom:
It looks like a-c just [skips all build products](https://github.com/mozilla-mobile/android-components/blob/ad4af6e34171e5c00ed08f0b5f9f2b4f6edd3c85/build.gradle#L243).  Can't quite figure out why it works on Fenix, though.

I suppose this is fine -- was just hoping we could dogfood the errors coming from glean_parser here rather than suppressing them.

PR #401, opened 2019-10-21
Expose testGetNumRecordedErrors

mdboom:
This is an (incomplete) approach to exposing `testGetNumRecordedErrors` across the FFI.

As you can see, it's going to require quite a bit of busy work across all of the metric types (at least the ones that can record errors).

An alternative approach would be to dynamically upcast to the `MetricType` trait and get a `CommonMetricData` object to pass that to the underlying `test_get_num_recorded_errors` function in `error_recording.rs`.  I think in that event, we'd still need to write an FFI function for every metric type to get a handle to a `CommonMetricData` instance, and all of the ownership complications of having a handle to a part of an object...

Another alternative would be to use a single ConcurrentHandleMap for all metrics that stores `MetricType`, and then dynamically downcast that when making type-specific calls...

Thoughts?  All of these options seem awfully big for the feature...

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
@@ -13,6 +13,7 @@ import mozilla.telemetry.glean.rust.LibGleanFFI
 import mozilla.telemetry.glean.rust.getAndConsumeRustString
 import mozilla.telemetry.glean.rust.toBoolean
 import mozilla.telemetry.glean.rust.toByte
+import mozilla.telemetry.glean.testing.ErrorType
 import org.json.JSONArray
 import org.json.JSONObject
 
@@ -229,5 +230,24 @@ class EventMetricType<ExtraKeysEnum : Enum<ExtraKeysEnum>> internal constructor(
             result.add(deserializeEvent(jsonRes.getJSONObject(i)))
         }
         return result
+    }
+
+    /**
+     * Returns the number of errors recorded for the given metric.
+     *
+     * @param errorType The type of the error recorded.
+     * @param pingName represents the name of the ping to retrieve the metric for.
+     *                 Defaults to the first value in `sendInPings`.
+     * @return the number of errors recorded for the metric.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    @JvmOverloads
+    fun testGetNumRecordedErrors(errorType: ErrorType, pingName: String = sendInPings.first()): Int {
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.assertInTestingMode()
+
+        return LibGleanFFI.INSTANCE.glean_event_test_get_num_recorded_errors(
+            Glean.handle, this.handle, errorType.ordinal, pingName
+        )
     }
 }
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
@@ -13,25 +13,26 @@ package mozilla.telemetry.glean.private
 import android.os.SystemClock
 import androidx.test.core.app.ApplicationProvider
 import androidx.test.ext.junit.runners.AndroidJUnit4
+import java.lang.NullPointerException
+import java.util.concurrent.TimeUnit
 import mozilla.telemetry.glean.Dispatchers
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.checkPingSchema
 import mozilla.telemetry.glean.getContextWithMockedInfo
 import mozilla.telemetry.glean.getMockWebServer
 import mozilla.telemetry.glean.resetGlean
+import mozilla.telemetry.glean.testing.ErrorType
 import mozilla.telemetry.glean.testing.GleanTestRule
 import mozilla.telemetry.glean.triggerWorkManager
 import org.json.JSONObject
 import org.junit.Assert.assertEquals
-import org.junit.Assert.assertTrue
 import org.junit.Assert.assertFalse
 import org.junit.Assert.assertNotNull
+import org.junit.Assert.assertTrue
 import org.junit.Assert.fail
-import org.junit.Test
 import org.junit.Rule
+import org.junit.Test
 import org.junit.runner.RunWith
-import java.lang.NullPointerException
-import java.util.concurrent.TimeUnit
 
 // Declared here, since Kotlin can not declare nested enum classes.
 enum class clickKeys {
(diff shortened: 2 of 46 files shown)


glean-core/ffi/src/macros.rs
```diff
@@ -79,6 +81,29 @@ macro_rules! define_metric {
...
+            error_type: i32,
+            storage_name: FfiStr
+        ) -> i32 {
+            crate::HandleMapExtension::call_infallible(&*crate::GLEAN, glean_handle, |glean| {
+                crate::HandleMapExtension::call_infallible(&*$metric_map, metric_id, |metric| {
+                    let error_type = std::convert::TryFrom::try_from(error_type).unwrap();
+                    let storage_name = crate::FallibleToString::to_string_fallible(&storage_name).unwrap();
+                    test_get_num_recorded_errors(
+                        glean,
+                        (metric as &dyn metrics::MetricType).meta(),
```
badboy: No need for dynamic dispatch here.

```suggestion
                        glean_core::metrics::MetricType::meta(metric),
```

mdboom:
Sure -- I just wanted to throw this up for discussion.  I don't have any real objections to this approach other than it requiring a lot of new code and busy work.  But it does feel a little bit like we need to do that either way we choose, just in different ways...

Anyway, no urgency on this.  We can brainstorm alternatives when we have time...

mdboom:
@badboy: Any thoughts on this?  I'm happy to crank out the rest of the mechanical changes here if this is the best approach...

badboy:
So ... all in all we won't get around to have that bit more boilerplate to implement it per type.
One other option I just came up with, which would reduce the number of exposes FFI functions, but couple the layers more tightly:

* Assign an ID to every metric type (e.g. in the order of their definition in `glean-core/src/metrics.rs`
* Have the platform call a "generic" function `test_get_num_recorded_errors(metric_type_id: u32, glean: u64, metric: u64)`
* This function matches on the `metric_type_id`, gets the meta data and returns the number of errors (see example code below)

```rust
fn test_get_num_recorded_errors(metric_type_id: u32, glean: u64, metric: u64) -> u64 { 
  	let meta = match metric_type_id {
		0 => BOOLEAN_METRICS.get(metric).meta(),
		1 => COUNTER_METRICS.get(metric).meta(),
		// ...
  	};
  	test_get_num_recorded_errors(
    	glean,
        meta,
        error_type,
        Some(&storage_name)
    ).unwrap_or(0)
}
```

and on the kotlin side:

```kotlin
class CounterMetricType {
    fun testGetNumRecordedErrors(errorType: ErrorType, pingName: String = sendInPings.first()): Int {
        Dispatchers.API.assertInTestingMode()

        return LibGleanFFI.INSTANCE.glean_test_get_num_recorded_errors(
            1, Glean.handle, this.handle, errorType.ordinal, pingName
        )
    }
}
```

But as you can see this means we're coupling everything through IDs, which we can't properly type-check anymore (we could of course wrap those in an enum on the Kotlin side to make it a bit nicer).

But now new metric types mean we need to change this enum/id-mapping as well.
Plus the compiler won't tell us if we also forget to update the match in `test_get_num_recorded_errors` on the Rust side.


All in all I think the implementation as done in this PR is preferable.
Maybe at some point in the future we can use code generation for the Kotlin/Swift side so we don't have to manually implement that `testGetNumRecordedErrors` all the time

mdboom:
This all makes sense.  Certainly build-time type safety is a good thing to maintain.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -15,6 +15,7 @@ import mozilla.telemetry.glean.rust.toBoolean
 import mozilla.telemetry.glean.rust.toByte
 import org.json.JSONArray
 import org.json.JSONObject
+import mozilla.telemetry.glean.testing.ErrorType
```
badboy: Our linter doesn't complain about unordered import statements here?
Consider me the linter: This line should be moved a bit higher to the other `mozilla.telemetry.glean` imports

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -361,4 +362,24 @@ class EventMetricTypeTest {
             pingJson.getJSONArray("events").getJSONObject(1).getJSONObject("extra").getString("someExtra")
         )
     }
+
+    @Test
+    fun `Long extra values record an error`() {
+
```
badboy: Needless extra line

glean-core/ffi/src/macros.rs
```diff
@@ -79,6 +81,29 @@ macro_rules! define_metric {
...
+            glean_handle: u64,
+            metric_id: u64,
+            error_type: i32,
+            storage_name: FfiStr
+        ) -> i32 {
+            crate::HandleMapExtension::call_infallible(&*crate::GLEAN, glean_handle, |glean| {
+                crate::HandleMapExtension::call_infallible(&*$metric_map, metric_id, |metric| {
+                    let error_type = std::convert::TryFrom::try_from(error_type).unwrap();
+                    let storage_name = crate::FallibleToString::to_string_fallible(&storage_name).unwrap();
+                    test_get_num_recorded_errors(
```
badboy:
```suggestion
                    glean_core::test_get_num_recorded_errors(
```

glean-core/ffi/src/lib.rs
```diff
@@ -8,7 +8,7 @@ use std::os::raw::c_char;
 use ffi_support::{define_string_destructor, ConcurrentHandleMap, FfiStr};
 use lazy_static::lazy_static;
 
-use glean_core::Glean;
+use glean_core::{test_get_num_recorded_errors, Glean};
```
badboy: Let's use the absolute path in the macro, then this `use` should not be needed.

glean-core/ffi/src/event.rs
```diff
@@ -13,11 +13,12 @@ use glean_core::{CommonMetricData, Lifetime};
 use crate::ffi_string_ext::FallibleToString;
 use crate::handlemap_ext::HandleMapExtension;
 use crate::{
-    define_metric, from_raw_int_array_and_string_array, from_raw_string_array, RawIntArray,
-    RawStringArray, GLEAN,
+    define_metric, from_raw_int_array_and_string_array, from_raw_string_array,
+    test_get_num_recorded_errors, RawIntArray, RawStringArray, GLEAN,
```
badboy: Let's use the absolute path in the macro, then this `use` should not be needed.

glean-core/ffi/src/datetime.rs
```diff
@@ -6,10 +6,13 @@ use std::os::raw::c_char;
 
 use ffi_support::FfiStr;
 
-use crate::{define_metric, handlemap_ext::HandleMapExtension, GLEAN};
+use crate::{
+    define_metric, handlemap_ext::HandleMapExtension, test_get_num_recorded_errors, GLEAN,
```
badboy: Let's use the absolute path in the macro, then this `use` should not be needed.

(from here on I'm not gonna repeat this)

mdboom:
I fixed the unnecessary Rust imports.

I sorted all of the imports in the Kotlin files this touches.

Also added a CHANGELOG entry -- not a breaking change, but big enough to merit noting.

There are two follow on bugs: [add this to labeled metrics](https://bugzilla.mozilla.org/show_bug.cgi?id=1593055) and [add this to Swift](https://bugzilla.mozilla.org/show_bug.cgi?id=1593056).

badboy:
> Also added a CHANGELOG entry -- not a breaking change, but big enough to merit noting.
Oh, see, I actually wanted to ask for that. Thanks!

> There are two follow on bugs: [add this to labeled metrics](https://bugzilla.mozilla.org/show_bug.cgi?id=1593055) and [add this to Swift](https://bugzilla.mozilla.org/show_bug.cgi?id=1593056).

Perfect, we can then tackle that in time.

PR #404, opened 2019-10-22
Update glean_parser to v1.9.5

brizental:


brizental:
@mdboom thanks for the review, Mike. I just realized that I had updated the glean_parser version incorrectly the previous times. Lesson learned :)

Also, mozilla-mobile/android-components#4820. The android-components PR.

badboy:
> And (sorry), I think this is still missing one place in `/samples/ios/app/sdk_generator.sh`.

That file is a symlink to the other sdk_generator.sh

badboy:
and we got a bug for making it consistent: https://bugzilla.mozilla.org/show_bug.cgi?id=1590448

mdboom:
Oh, sorry about that.  Grep showed that as separate hits for me and didn't
think to look that it was a symlink...

PR #407, opened 2019-10-23
1589743: Don't log pings twice

mdboom:
This now matches the behavior of glean-ac (logging from the uploader)


mdboom:
> But I guess keeping it in the uploader makes more sense, otherwise we might miss some that are triggered from Rust internally where we don't have the state to log these, right?

Yes, I think so.

All of this will probably need to be revisited with the planned architecture refactor to have the Rust core more "in charge" of uploading.

PR #408, opened 2019-10-24
Prepare for next release cycle

badboy:
For local development and easy testing in a-c and Fenix we should have a version number bigger than the released one.

The gradle/maven way is to `major+1` + `-SNAPSHOT`.
The Rust/semver way is to `major+1` + `-alpha.0` (less relevant, as we don't push to crates.io just yet).

I would totally expect that we go `19.1.0` next first, but that's for the release to decide.

### DIFF
--- a/docs/dev/cut-a-new-release.md
+++ b/docs/dev/cut-a-new-release.md
@@ -20,7 +20,10 @@ These are the steps needed to cut a new release from latest master.
 
     2. Add a new `# Unreleased Changes` on top
 
-2. Bump `libraryVersion` in the top-level [.buildconfig.yml](https://github.com/mozilla/glean/blob/master/.buildconfig.yml) file. Be sure you're following semver, and if in doubt, ask.
+2. Bump the versions
+    * Rust crates: Bump `version` in [`glean-core/Cargo.toml`](https://github.com/mozilla/glean/blob/master/glean-core/Cargo.toml) and [`glean-core/Cargo.toml`](https://github.com/mozilla/glean/blob/master/glean-core/ffi/Cargo.toml).
+    * Kotlin package: Bump `libraryVersion` in the top-level [.buildconfig.yml](https://github.com/mozilla/glean/blob/master/.buildconfig.yml) file.
+    * Be sure you're following semver, and if in doubt, ask.
 3. Land the commits that perform the steps above. This takes a PR, typically, because of branch protection on master.
 4. Cut the actual release.
     1. Click "Releases", and then "Draft a New Release" in the github UI.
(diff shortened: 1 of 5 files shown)


glean-core/ffi/Cargo.toml
```diff
@@ -1,8 +1,12 @@
 [package]
 name = "glean-ffi"
-edition = "2018"
-version = "19.0.0"
+version = "20.0.0-alpha.0"
```
Dexterp37: Is there any way this could be inherited from the top level `Cargo.toml`?

glean-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "glean-core"
-version = "19.0.0"
+version = "20.0.0-alpha.0"
```
Dexterp37: Uh, can we please add this [here](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html) as well?

glean-core/ffi/Cargo.toml
badboy: No, the two crates are independently versioned.
It's a bit of a bummer that there's no better support in cargo itself just yet, but I plan to paper over that with a bit more tooling.

glean-core/Cargo.toml
badboy: Good point

docs/dev/cut-a-new-release.md
```diff
@@ -20,11 +20,17 @@ These are the steps needed to cut a new release from latest master.
...
 3. Land the commits that perform the steps above. This takes a PR, typically, because of branch protection on master.
 4. Cut the actual release.
     1. Click "Releases", and then "Draft a New Release" in the github UI.
     2. Enter `v<myversion>` as the tag. In the example above it would be `v0.1.0`. It's important this is the same as the tags you put in the links in the changelog.
     3. Under the description, paste the contents of the release notes from `CHANGELOG.md`.
     4. Note that the release is not available until the CI build completes for that tag.
         - You can check [on CircleCI for the running build](https://circleci.com/gh/mozilla/glean).
+5. After the release, bump the versions to the next pre-release:
+    * Rust crates: Use the next major version and append `-alpha.1`
+    * Kotlin package: Use the next major version and append `-SNAPSHOT`
```
mdboom: It would be great to automate some of this stuff.

I really like [bumpversion](https://pypi.org/project/bumpversion/), and it is probably configurable enough to handle these use cases.  But folks had trouble using it on Windows for glean_parser, so it was ultimately dropped.  Still there might be alternative tools of that type that would work well here.

(All is to say, let's file a bug for release automation...  This PR is fine as-is).

badboy: There's also [cargo-release](https://github.com/sunng87/cargo-release) which has hooks for file modification and just executing scripts :)

I'm a happy user for pure-Rust crates.

mdboom: Sure -- I actually looked at it but didn't see arbitrary file modification (which we would need) in the summary...  I agree something in Rust is the way to go -- everyone working on this already has a working `cargo`...

badboy:
> LGTM :) Let's file a bug about automatic the version quirks either via bumpversion (as suggested by @mdboom ) or via the Rust-only solution.

https://bugzilla.mozilla.org/show_bug.cgi?id=1591070

docs/dev/cut-a-new-release.md
badboy: Well, it can execute scripts.

PR #409, opened 2019-10-24
1586736 - Add Glean-iOS debug functionality

travis79:
This adds CLI handling for the `sendPing`, `tagPings`, and `logPings` commands by using the UserDefaults API to receive and parse launch arguments.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -40,3 +40,4 @@ samples/ios/app/.venv/
 .mypy_cache/
 .coverage
 htmlcov/
+*.log
--- a/docs/user/debugging.md
+++ b/docs/user/debugging.md
@@ -1,4 +1,24 @@
-# Debugging products using the Glean SDK for Android
+# Debugging products using the Glean SDK
+
+## Contents
+
+1. [Debugging the Glean SDK in Android](#debugging-the-glean-sdk-in-android)
+2. [Debugging the Glean SDK in iOS](#debugging-the-glean-sdk-in-ios)
+
+### Important considerations when using Glean SDK debug tools
+
+- Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+
+- There are a couple different ways in which to send pings using the Glean SDK debug tools.
+    1. You can tag pings using the debug tools and trigger them manually using the UI.  This should always produce a ping with all required fields.
+    2. You can tag _and_ send pings using the debug tools.  This has the side effect of potentially sending a ping which does not include all fields because `sendPings` triggers pings to be sent before certain application behaviors can occur which would record that information.  For example, `duration` is not calculated or included in a baseline ping sent with `sendPing` because it forces the ping to be sent before the `duration` metric has been recorded.  Keep in mind that there may be nothing to send, in which case no ping is generated.
+    3. You can trigger a command while the instrumented application is still running.  This is useful for toggling commands or for triggering pings that have schedules that are difficult to trigger manually.  This is especially useful if you need to trigger a ping submission after some activity within the application, such as with the metrics ping.
+
+### Glean Log messages
+
+Glean logs warnings and errors through the platform-specific logging frameworks.  See the platform-specific instructions for information on how to view the logs.
+
+## Debugging the Glean SDK in Android
 
 The Glean SDK exports the `GleanDebugActivity` that can be used to toggle debugging features on or off.
 Users can invoke this special activity, at run-time, using the following [`adb`](https://developer.android.com/studio/command-line/adb) command:
@@ -25,7 +45,7 @@ In the above:
 
 For example, to direct a release build of the Glean sample application to (1) dump pings to logcat, (2) tag the ping with the `test-metrics-ping` tag, and (3) send the "metrics" ping immediately, the following command can be used:
 
-```
+```shell
 adb shell am start -n org.mozilla.samples.glean/mozilla.telemetry.glean.debug.GleanDebugActivity \
   --ez logPings true \
   --es sendPing metrics \
@@ -35,27 +55,19 @@ adb shell am start -n org.mozilla.samples.glean/mozilla.telemetry.glean.debug.Gl
 > **Note:** In previous versions of Glean the activity was available with the name `mozilla.components.service.glean.debug.GleanDebugActivity`.
 >
 > If you're debugging an old build, try running:
-> ```
+>
+> ```shell
 > adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity \
 >   --ez logPings true \
 >   --es sendPing metrics \
 >   --es tagPings test-metrics-ping
 > ```
 
-### Important GleanDebugActivity notes!
+### Glean Log messages
 
-- Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
-
-- There are a couple different ways in which to send pings through the GleanDebugActivity.
-    1. You can use the `GleanDebugActivity` in order to tag pings and trigger them manually using the UI.  This should always produce a ping with all required fields.
-    2. You can use the `GleanDebugActivity` to tag _and_ send pings.  This has the side effect of potentially sending a ping which does not include all fields because `sendPings` triggers pings to be sent before certain application behaviors can occur which would record that information.  For example, `duration` is not calculated or included in a baseline ping sent with `sendPing` because it forces the ping to be sent before the `duration` metric has been recorded.
-
-## Glean Log messages
-
-Glean logs warnings and errors through the Android logging framework.
 When running a Glean-powered app in the Android emulator or on a device connected to your computer via cable, there are several ways to read the log output.
 
-### Android Studio
+#### Android Studio
 
 Android Studio can show the logs of a connected emulator or device.
 To display the log messages for an app:
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
 
 [View Logs with Logcat]: https://developer.android.com/studio/debug/am-logcat
 
-### Command line
+#### Command line
 
 On the command line you can show all of the log output using:
 
-```
+```shell
 adb logcat
 ```
 
 This is the unfiltered output of all log messages.
 You can match for `glean` using grep:
 
-```
+```shell
 adb logcat | grep -i glean
 ```
 
 A simple way to filter for only the application that is being debugged is by using [pidcat][], a wrapper around `adb`, which adds colors and proper filtering by application ID and log level.
 Run it like this to filter for an application:
 
-```
+```shell
 pidcat [applicationId]
 ```
 
 In the above `[applicationId]` is the product's application id as defined in the manifest file and/or build script. For the Glean sample application, this is `org.mozilla.samples.glean` for a release build and `org.mozilla.samples.glean.debug` for a debug build.
 
 [pidcat]: https://github.com/JakeWharton/pidcat
+
+## Debugging the Glean SDK in iOS
+
+For debugging and validation purposes on iOS, Glean makes use of a custom URL scheme which is implemented _within the application_ that is consuming Glean.  Glean provides some convenience functions to facilitate this, but it's up to the consuming application to enable this functionality.  Applications that enable this Glean SDK feature will be able to launch the application from a URL with the Glean debug commands embedded in the URL itself.
+
+### Available commands and query format
+
+There are 3 available commands that you can use with the Glean SDK debug tools
+
+- `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log
+- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
+- `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
+
+The structure of the custom URL uses the following format:
+
+`<protocol>://glean?<command 1>=<paramter 1>&<command 2>=<parameter 2> ...`
+
+Where:
+
+- `<protocol>` is the "Url Scheme" that has been added for your app (see Instrumenting the application below), such as `glean-sample-app`.
+- This is followed by `://` and then `glean` which is required for the Glean SDK to recognize the command is meant for it to process.
+- Following standard URL query format, the next character after `glean` is the `?` indicating the beginning of the query.
+- This is followed by one or more queries in the form of `<command>=<parameter>`, where the command is one of the commands listed above, followed by an `=` and then the value or parameter to be used with the command.
+
+There are a few things to consider when creating the custom URL:  
+
+- Invalid commands will log an error and cause the entire url to be ignored.
+- Not all commands are required to be encoded in the URL, you can mix and match the commands that you need.
+- Multiple instances of commands are not allowed in the same url and, if present, will cause the entire url to be ignored.
+
+### Instrumenting the application for Glean debug functionality
+
+In order to enable the debugging features in a Glean SDK consuming iOS application, it is necessary to add some information to the application's `Info.plist`, and add a line and possibly an override for a function in the `AppDelegate.swift`.
+
+#### Register custom URL scheme in `Info.plist`
+
+> **Note:** If your application already has a custom URL scheme implemented, there is no need to implement a second scheme, you can simply use that and skip to the next section about adding the convenience method.  If the app doesn't have a custom URL scheme implemented, then you will need to perform the following instructions to register your app to recieve custom URLs.
+
+Find and open the application's `Info.plist` and right click any blank area and select `Add Row` to create a new key.
+
+You will be prompted to select a key from a drop-down menu, scroll down to and select `URL types`.  This creates an array item, which can be expanded by clicking the triangle disclosure icon.
+
+Select `Item 0`, click on it and click the disclosure icon to expand it and show the `URL identifier` line.  Double-click the value field and fill in your identifier, typically the same as the bundle ID.
+
+Right-click on `Item 0` and select `Add Row` from the context menu.  In the dropdown menu, select `URL Schemes` to add the item.
+
+Click on the disclosure icon of `URL Schemes` to expand the item, double-click the value field of `Item 0` and key in the value for your application's custom scheme.  For instance, the Glean sample app uses `glean-sample-app`, which allows for custom URL's to be crafted using that as a protocol, for example: `glean-sample-app://glean?logPings=true`
+
+#### Add the `Glean.handleCustomUrl()` convenience function and necessary overrides
+
+In order to handle the incoming Glean SDK debug commands, it is necessary to implement the override in the application's `AppDelegate.swift` file.  Within that function, you can make use of the convenience function provided in Glean `handleCustomUrl(url: URL)`.  
+
+An example of a simple implementation of this would look like this:
+
+```Swift
+func application(_: UIApplication,
+                 open url: URL,
+                 options _: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+    // ...
+
+    // This does nothing if the url isn't meant for Glean.
+    Glean.shared.handleCustomUrl(url: url)
+
+    // ...
+
+    return true
+}
+```
+
+If you need additional help setting up a custom URL scheme in your application, please refer to [Apple's documentation](https://developer.apple.com/documentation/uikit/inter-process_communication/allowing_apps_and_websites_to_link_to_your_content/defining_a_custom_url_scheme_for_your_app).
+
+### Invoking the Glean-iOS debug commands
+
+Now that the app has the Glean SDK debug functionality enabled, there are a few ways in which we can invoke the debug commands.
+
+#### Using a web browser
+
+Perhaps the simplest way to invoke the Glean SDK debug functionality is to open a web browser and type/paste the custom URL into the address bar.  This is especially useful on an actual device because there isn't a good way to launch from the command line and process the URL for an actual device.
+
+Using the glean-sample-app as an example: to activate ping logging, tag the pings to go to the Glean Debug View, and force the `events` ping to be sent, enter the following URL in a web browser on the iOS device:
+
+```shell
+glean-sample-app://glean?logPings=true&tagPings=My-ping-tag&sendPing=events
+```
+
+This should cause iOS to prompt you with a dialog asking if you want to open the URL in the Glean Sample App, and if you select "Okay" then it will launch (or resume if it's already running) the application with the indicated commands and parameters and immediately force the collection and submission of the events ping.
+
+> **Note:** This method does not work if the browser you are using to input the command is the same application you are attempting to pass the Glean debug commands to.  So, you couldn't use Firefox for iOS to trigger commands within Firefox for iOS.
+
+It is also possible to encode the URL into a 2D barcode or QR code and launch the app via the camera app.  After scanning the encoded URL, the dialog prompting to launch the app should appear as if the URL were entered into the browser address bar.
+
+#### Using the command line
+
+This method is useful for testing via the Simulator, which typically requires a Mac with Xcode installed, including the Xcode command line tools.  In order to perform the same command as above with using the browser to input the URL, you can use the following command in the command line terminal of the Mac:
+
+```shell
+xcrun simctl openurl booted "glean-sample-app://glean?logPings=true&tagPings=My-ping-tag&sendPing=events"
+```
+
+This will launch the simulator and again prompt the user with a dialog box asking if you want to open the URL in the Glean Sample App (or whichever app you are instrumenting and testing).
+
+### Glean log messages
+
+If debugging in the simulator, the logging messages can be seen in the console window within Xcode.
+
+When running a Glean-powered app in the iOS Simulator or on a device connected to your computer via cable, the following steps will help in getting the logs:
+
+1. Plug in the device and open Xcode
+2. Choose Window -> Devices from the menu bar
+3. Under the DEVICES section in the left column, choose the device
+4. To see the device console, click the up-triangle at the bottom left of the right hand panel
+
+> **Note:** It is helpful to filter the console window by adding the keyword "glean" to the filter box.  This will allow you to see only messages generated by the Glean SDK.
+
+You can refer to [this Apple technical Q&A](https://developer.apple.com/library/archive/qa/qa1747/_index.html) for more information about debugging deployed iOS apps.
--- a/glean-core/ios/Glean.xcodeproj/project.pbxproj
+++ b/glean-core/ios/Glean.xcodeproj/project.pbxproj
@@ -7,6 +7,8 @@
 	objects = {
 
 /* Begin PBXBuildFile section */
+		1F39E7B0239F0505009B13B3 /* GleanDebugTools.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F39E7AF239F0505009B13B3 /* GleanDebugTools.swift */; };
+		1F39E7B3239F0777009B13B3 /* GleanDebugUtilityTests.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F39E7B2239F0777009B13B3 /* GleanDebugUtilityTests.swift */; };
 		1F6058932314863400307A9F /* Configuration.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F6058922314863400307A9F /* Configuration.swift */; };
 		1F605895231489AB00307A9F /* HttpPingUploader.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F605894231489AB00307A9F /* HttpPingUploader.swift */; };
 		1F60589723148BF800307A9F /* GleanLifecycleObserver.swift in Sources */ = {isa = PBXBuildFile; fileRef = 1F60589623148BF800307A9F /* GleanLifecycleObserver.swift */; };
@@ -75,6 +77,8 @@
 /* End PBXContainerItemProxy section */
 
 /* Begin PBXFileReference section */
+		1F39E7AF239F0505009B13B3 /* GleanDebugTools.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GleanDebugTools.swift; sourceTree = "<group>"; };
+		1F39E7B2239F0777009B13B3 /* GleanDebugUtilityTests.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GleanDebugUtilityTests.swift; sourceTree = "<group>"; };
 		1F6058922314863400307A9F /* Configuration.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = Configuration.swift; sourceTree = "<group>"; };
 		1F605894231489AB00307A9F /* HttpPingUploader.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = HttpPingUploader.swift; sourceTree = "<group>"; };
 		1F60589623148BF800307A9F /* GleanLifecycleObserver.swift */ = {isa = PBXFileReference; lastKnownFileType = sourcecode.swift; path = GleanLifecycleObserver.swift; sourceTree = "<group>"; };
@@ -159,6 +163,22 @@
 /* End PBXFrameworksBuildPhase section */
 
 /* Begin PBXGroup section */
+		1F39E7AE239F04D3009B13B3 /* Debug */ = {
+			isa = PBXGroup;
+			children = (
+				1F39E7AF239F0505009B13B3 /* GleanDebugTools.swift */,
+			);
+			path = Debug;
+			sourceTree = "<group>";
+		};
+		1F39E7B1239F0741009B13B3 /* Debug */ = {
+			isa = PBXGroup;
+			children = (
+				1F39E7B2239F0777009B13B3 /* GleanDebugUtilityTests.swift */,
+			);
+			path = Debug;
+			sourceTree = "<group>";
+		};
 		1F60588D231483D600307A9F /* Scheduler */ = {
 			isa = PBXGroup;
 			children = (
@@ -249,6 +269,7 @@
 		BF3DE3932243A2F20018E23F /* Glean */ = {
 			isa = PBXGroup;
 			children = (
+				1F39E7AE239F04D3009B13B3 /* Debug */,
 				97C5C1A223708C1E00B79C93 /* Testing */,
 				1F6058912314861B00307A9F /* Config */,
 				BF3227D723336CF800CD0111 /* Generated */,
@@ -269,6 +290,7 @@
 		BF3DE39E2243A2F20018E23F /* GleanTests */ = {
 			isa = PBXGroup;
 			children = (
+				1F39E7B1239F0741009B13B3 /* Debug */,
 				1FB8F8392326EBA500618E47 /* Config */,
 				BF43A8CB232A613100545310 /* Metrics */,
 				BF80AA5923992FFB00A8B172 /* Net */,
@@ -572,6 +594,7 @@
 				BF30FDC4233260B500840607 /* TimespanMetric.swift in Sources */,
 				BFE1CDCE233B989A0019EE47 /* GleanMetrics.swift in Sources */,
 				BF6C53B2232F870C00E3B43A /* Ping.swift in Sources */,
+				1F39E7B0239F0505009B13B3 /* GleanDebugTools.swift in Sources */,
 				BFB59A9F2342A24000F40CA8 /* GleanInternalMetrics.swift in Sources */,
 				97C5C1A423708C6700B79C93 /* ErrorType.swift in Sources */,
 				BF10007E23548AF400064051 /* MemoryUnit.swift in Sources */,
@@ -585,6 +608,7 @@
 			isa = PBXSourcesBuildPhase;
 			buildActionMask = 2147483647;
 			files = (
+				1F39E7B3239F0777009B13B3 /* GleanDebugUtilityTests.swift in Sources */,
 				BFAED50A2369752400DF293D /* StringListMetricTests.swift in Sources */,
 				BF890561232BC227003CA2BA /* StringMetricTests.swift in Sources */,
 				1FD4527723395EEB00F4C7E8 /* UuidMetricTests.swift in Sources */,
--- a/glean-core/ios/Glean/Debug/GleanDebugTools.swift
+++ b/glean-core/ios/Glean/Debug/GleanDebugTools.swift
@@ -1,0 +1,145 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+import Foundation
+
+/// This class is meant to encapsulate the Glean debug functionality.
+class GleanDebugUtility {
+    // This struct is used for organizational purposes to keep the class constants in a single place
+    struct Constants {
+        static let logTag = "glean/GleanDebugUtility"
+        // Since ping tags are transmitted via HTTP in a header, and since they are displayed to the
+        // user on the Glean Debug View, we need to constrain the tag to a safe character set and
+        // limit the length to prevent possible attack vectors.
+        static let pingTagRegexPattern = "^[a-zA-Z0-9-]{1,20}$"
+    }
+
+    private static let logger = Logger(tag: Constants.logTag)
+
+    /// When applications are launched using the custom URL scheme, this helper function will process
+    /// the URL and parse the debug commands
+    ///
+    /// - parameters:
+    ///     * url: A `URL` object containing the Glean debug commands as query parameters
+    ///
+    /// There are 3 available commands that you can use with the Glean SDK debug tools
+    ///
+    /// - `logPings`: If "true", will cause pings that are submitted to also be echoed to the device's log
+    /// - `tagPings`:  This command expects a string to tag the pings with and redirects them to the Glean Debug View
+    /// - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
+    ///
+    /// The structure of the custom URL uses the following format:
+    ///
+    /// `<protocol>://glean?<command 1>=<paramter 1>&<command 2>=<parameter 2> ...`
+    ///
+    /// Where:
+    ///
+    /// - `<protocol>://` is the "Url Scheme" that has been added for the app and doesn't matter to Glean.
+    /// - `glean` is required for the Glean SDK to recognize the command is meant for it to process.
+    /// - `?` indicating the beginning of the query.
+    /// - `<command>=<parameter>` are instances of the commands listed above  separated by `&`.
+    ///
+    /// There are a few things to consider when creating the custom URL:
+    ///
+    /// - Invalid commands will trigger an error and be ignored.
+    /// - Not all commands are requred to be encoded in the URL, you can mix and match the commands that you need.
+    /// - Special characters should be properly URL encoded and escaped since this needs to represent a valid URL.
+    static func handleCustomUrl(url: URL) {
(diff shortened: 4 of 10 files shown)


badboy:
How do you pass these arguments using xcode?

travis79:
> How do you pass these arguments using xcode?

In Xcode, select Product -> Scheme -> Edit Scheme to open the edit scheme dialog.  From there you should see a tab for "Arguments".  There is also a way using [ios-deploy]()https://github.com/ios-control/ios-deploy).  I'm still verifying that this all works and will document the usage as soon as I'm comfortable with the process.  That's why this is just a draft PR yet.

Dexterp37:
> 
>  I'm still verifying that this all works and will document the usage as soon as I'm comfortable with the process. That's why this is just a draft PR yet.

Cool :-) I have a few questions about this:

- How can someone trigger a baseline ping from a released binary?
- Does using this require XCode?
- Can logs from a released binary be captured in some way as we do with "adb logcat"?
- Can other applications access logs from this? For example, can a malicious application enable "log pings" and read the content of the pings from the logs?

travis79:
These are all great questions!

>     * How can someone trigger a baseline ping from a released binary?
Due to how launch arguments are passed, this can only be done during launch of the application and can't be done arbitrarily while the app is running.  Later, if we can implement the custom URL scheme, this would give us more capabilities to do things like this.


>     * Does using this require XCode?
Yes and no.  The most straightforward way to do this is through the IDE.  You need Xcode (or at least the command line tools) in order to have the build tools.  There is another tools ios-deploy, that I mentioned in the comment above that can deploy an app to a device and pass it command line args, and there are CLI tools for doing the same in a simulator.  I hope to document several methods of passing args to the app.

>     * Can logs from a released binary be captured in some way as we do with "adb logcat"?
yes, if Xcode (or lldb actually) is attached, you can see the logs.  I'm not as clear on this so let me research it more and get back to you.

>     * Can other applications access logs from this? For example, can a malicious application enable "log pings" and read the content of the pings from the logs?
I'm also not as clear on this so let me research it more and get back to you, my gut says "no", especially knowing how Apple likes to enforce the sandbox.

badboy:
> yes, if Xcode (or lldb actually) is attached, you can see the logs. I'm not as clear on this so let me research it more and get back to you.

If an iPhone is connected over USB to a macbook the "Console" application should show its lgos, and that should include at least the errors Glean logs.

Dexterp37:
> 
> 
> These are all great questions!
> 
> > ```
> > * How can someone trigger a baseline ping from a released binary?
> > ```
> 
> Due to how launch arguments are passed, this can only be done during launch of the application and can't be done arbitrarily while the app is running.

I'm afraid this would prevent QA from covering some specific scenarios that can currently be covered on Glean SDK for Android:

- start the application & mark pings as debug pings;
- perform some action;
- trigger pings that have otherwise complex schedule (e.g. `metrics` or custom pings);

travis79:
> > These are all great questions!
> > > ```
> > > * How can someone trigger a baseline ping from a released binary?
> > > ```
> > 
> > 
> > Due to how launch arguments are passed, this can only be done during launch of the application and can't be done arbitrarily while the app is running.
> 
> I'm afraid this would prevent QA from covering some specific scenarios that can currently be covered on Glean SDK for Android:
> 
>     * start the application & mark pings as debug pings;
> 
>     * perform some action;
> 
>     * trigger pings that have otherwise complex schedule (e.g. `metrics` or custom pings);

Then perhaps it's time to evaluate the custom URL scheme now in order to give QA that functionality for iOS.  I was also planning on discussing this with the iOS team(s) to see how they handle this situation for QA puppeting, so maybe they can provide additional suggestions.

Dexterp37:
> 
> 
> > > These are all great questions!
> > > > ```
> > > > * How can someone trigger a baseline ping from a released binary?
> > > > ```
> > > 
> > > 
> > > Due to how launch arguments are passed, this can only be done during launch of the application and can't be done arbitrarily while the app is running.
> > 
> > 
> > I'm afraid this would prevent QA from covering some specific scenarios that can currently be covered on Glean SDK for Android:
> > ```
> > * start the application & mark pings as debug pings;
> > 
> > * perform some action;
> > 
> > * trigger pings that have otherwise complex schedule (e.g. `metrics` or custom pings);
> > ```
> 
> Then perhaps it's time to evaluate the custom URL scheme now in order to give QA that functionality for iOS. I was also planning on discussing this with the iOS team(s) to see how they handle this situation for QA puppeting, so maybe they can provide additional suggestions.

@travis79 I'd recommend writing a mini proposal for this specific feature: enumerate what's currently supported on Android (I mean, what can QA do with the debug activity?) and check how to support the same set of things on iOS.

I'd list all the available options on the doc (URL scheme, command line arguments). @badboy , thoughts?

badboy:
Maybe we should ask how QA even works on iOS?
Do they run builds through Xcode? Do they use pre-release versions through TestFlight? Do they only test on app store-released versions?

Dexterp37:
> 
> 
> Maybe we should ask how QA even works on iOS?
> Do they run builds through Xcode? Do they use pre-release versions through TestFlight? Do they only test on app store-released versions?

Regardless of how it _currently_ works, we know that doing QA on released binaries it's a feature that we need.

It's ok to as QA about their current practices, but we should target feature parity with what can be done on Android.

garvankeeley:
> Maybe we should ask how QA even works on iOS?
> Do they run builds through Xcode? Do they use pre-release versions through TestFlight? Do they only test on app store-released versions?

On Firefox iOS, all three are used

glean-core/ios/Glean.xcodeproj/project.pbxproj
```diff
@@ -735,6 +735,7 @@
 				);
 				INFOPLIST_FILE = Glean/Info.plist;
 				INSTALL_PATH = "$(LOCAL_LIBRARY_DIR)/Frameworks";
+				IPHONEOS_DEPLOYMENT_TARGET = 11.0;
```
badboy: What does this do? Does it force a minimum of iOS 11?

samples/ios/app/glean-sample-app/Info.plist
```diff
@@ -2,6 +2,17 @@
 <!DOCTYPE plist PUBLIC "-//Apple//DTD PLIST 1.0//EN" "http://www.apple.com/DTDs/PropertyList-1.0.dtd">
 <plist version="1.0">
 <dict>
+	<key>CFBundleURLTypes</key>
```
badboy: So this becomes something that each app needs to implement, right? (which makes sense, because each app needs an individual scheme).

Dexterp37:
@travis79 @badboy can we please create a document about the different techniques and all the pros/cons and how that compares to what's currently possible for Android? I personally find it a bit hard otherwise to follow :)

glean-core/ios/Glean.xcodeproj/project.pbxproj
travis79: No it's just a target deployment, like the target version of Android doesn't have to be the minimum supported SDK version.

samples/ios/app/glean-sample-app/Info.plist
travis79: Yes, it's looking like this will have to be implemented at the application level as it's not possible to receive the ApplicationDelegate from the Glean framework/library.  But this is all just experiments yet, I wasn't ready for further review yet so I haven't had time to test out what all is possible yet.

travis79:
> @travis79 @badboy can we please create a document about the different techniques and all the pros/cons and how that compares to what's currently possible for Android? I personally find it a bit hard otherwise to follow :)

Certainly, I will do the same thing I did for the metrics ping scheduler, but I'm not done researching what is possible and what is correct, which is why this is still a draft PR.  I committed this (and probably shouldn't have pushed it as it drew your attention) before I was going to be AFK to save where I was working so I could pick it up on Monday.  As soon as I know what our options are to get the same functionality as Android (if that's even possible), I'll report the findings to the rest of the team so we can discuss the best way forward.

samples/ios/app/Cartfile
```diff
@@ -1,2 +1,2 @@
 github "httpswift/swifter" ~> 1.4.7
-github "mozilla/glean" "master"
+github "travis79/glean" "ios-debug-tools"
```
travis79: This is just for testing and needs to be reverted before merging.

badboy: Hmpf, I really dislike that it has to build against master.

samples/ios/app/glean-sample-app/AppDelegate.swift
```diff
@@ -39,6 +38,17 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
         return true
     }
 
+    func application(_: UIApplication,
+                     open url: URL,
+                     options _: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+        let host = url.host?.removingPercentEncoding
```
Dexterp37: nit: what about moving these checks within the `handleCustomUrl` so that the app would only be required to do a function call?

travis79: I thought about that, and can do that, it just felt a little weird if the app was doing something for itself as it would have a 'host' that would be specific to it.  It would basically be a no-op to call it if the 'host' wasn't "glean", so let me make that change.  I still have tests to add as well...

samples/ios/app/Cartfile
travis79: I can pin it to a version when I revert this...

travis79:
Okay, this has been updated to now use a custom URL scheme, including setting up the glean-sample-app as an example of how to use this in an app.  I still need to write the docs for this method, including what an integrating app will need to do in order to make use of this functionality.

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,64 @@ public class Glean {
...
+                    }
+
+                } else {
+                    logger.error("No value passed to tagPings")
+                }
+            case "logPings":
+                logPings = param.value == "true" ? true : false
+                logger.debug("Log pings set to: \(logPings!)")
+            case "sendPing":
+                sendPing = param.value
```
badboy: Should we enforce a reasonable value here?
Afaik this just gets passed through, but that means we end up copying it over to Rust and I would hate people passing in very long strings again and again.

glean-core/ios/Glean/Glean.swift
```diff
@@ -25,7 +25,7 @@ public class Glean {
 
     var handle: UInt64 = 0
     private var uploadEnabled: Bool = true
-    private var configuration: Configuration?
+    var configuration: Configuration?
```
badboy: Why does this need to become non-private?

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,64 @@ public class Glean {
...
+        }
+
+        configuration = Configuration(
+            serverEndpoint: configuration?.serverEndpoint ?? Configuration.Constants.defaultTelemetryEndpoint,
+            userAgent: configuration?.userAgent ?? Configuration.Constants.defaultUserAgent,
+            logPings: logPings ?? configuration?.logPings ?? Configuration.Constants.defaultLogPings,
+            maxEvents: configuration?.maxEvents,
+            pingTag: tagName ?? configuration?.pingTag,
+            channel: configuration?.channel
+        )
```
badboy: Do we really need to re-set the whole configuration object again and again? I find the whole inline `??` in this block unreadable.

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,64 @@ public class Glean {
...
+                logPings = param.value == "true" ? true : false
+                logger.debug("Log pings set to: \(logPings!)")
+            case "sendPing":
+                sendPing = param.value
+            default:
+                logger.error("Unknown parameter passed to handleCustomUrl")
+            }
+        }
+
+        configuration = Configuration(
```
badboy: We disabled `redundantSelf`, thus `self.configuration` would IMO be clearer (but see comment blow)

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,64 @@ public class Glean {
         }
     }
 
+    /// When applications are launched using the custom URL scheme, this helper function will process
+    /// the URL and parse the debug commands
```
badboy: Document what a proper URL will look like and what query parameters it parses out.

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,64 @@ public class Glean {
...
+                } else {
+                    logger.error("No value passed to tagPings")
+                }
+            case "logPings":
+                logPings = param.value == "true" ? true : false
+                logger.debug("Log pings set to: \(logPings!)")
+            case "sendPing":
+                sendPing = param.value
+            default:
+                logger.error("Unknown parameter passed to handleCustomUrl")
```
badboy: Quite honestly, should we error-out if we encounter an unknown query parameter?
Right now this wouldn't be an attack vector, but maybe cancelling early on unexpected values is still good

samples/ios/app/Cartfile
badboy: Nah, I mean more the fact that we can't just point to "this directory two layers above".
With this we always need to land changes to master first, then change the tests afterwards to use new master features.
Not really something we can fix here (but maybe we can point it to the integration branch of bors once we have it, at least for CI)

glean-core/ios/Glean/Glean.swift
travis79: Sure, we don't enforce any sort of limitations on the Android side of things is the only reason I didn't do any checking here.  It probably shouldn't be too difficult to check this against the collection of registered pings, let me look into it and I'll file a bug to port the same restrictions to Android.

glean-core/ios/Glean/Glean.swift
travis79: I needed to change this to `internal` so that I can check values in a test to verify that the `handleCustomUrl` function is properly setting the `pingTag` and `logPings` values.  

In relaxing this to `internal` access, this makes it in line with the Android configuration instance in the Glean object, which is also listed as internal.

glean-core/ios/Glean/Glean.swift
travis79: Well, because the configuration is filled with `let` constants, yes, I need to rebuild the entire configuration object.  If we relaxed some of the constant values to variables, or even if we made a copy constructor, it would simplify this.  Let me explore what I can do to simplify this...

glean-core/ios/Glean/Glean.swift
travis79: Sure!

glean-core/ios/Glean/Glean.swift
travis79: In order to match the functionality of the `GleanDebugActivity`, this totally makes sense. We bail out early in the Android tools if there is a bad command so I should probably do that here as well.

glean-core/ios/Glean/Glean.swift
travis79: Bug for adding the same sort of validation to the `GleanDebugActivity` [filed here](https://bugzilla.mozilla.org/show_bug.cgi?id=1599483)

samples/ios/app/Cartfile
travis79: > Nah, I mean more the fact that we can't just point to "this directory two layers above".
> With this we always need to land changes to master first, then change the tests afterwards to use new master features.
> Not really something we can fix here (but maybe we can point it to the integration branch of bors once we have it, at least for CI)

Unfortunately we can't use relative URL's in a File URI, which is what is expected when we use the `git` command in the Cartfile.  I tried this several different ways but just couldn't get it to find it unless I put an absolute path in the URL (i.e "file:///Users/tlong/Documents/glean" works, but "file://../../.." doesn't).

travis79: I'll file a bug to change this to point at the integration branch or something once bors is enabled.

travis79: [Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1599531)

badboy: We could, for now, change CI to run these tests only on master, thus at least making them not fail on the PR itself.

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
 pidcat [applicationId]
 ```
 
 In the above `[applicationId]` is the product's application id as defined in the manifest file and/or build script. For the Glean sample application, this is `org.mozilla.samples.glean` for a release build and `org.mozilla.samples.glean.debug` for a debug build.
 
 [pidcat]: https://github.com/JakeWharton/pidcat
+
+## Debugging the Glean SDK in iOS
+
+Unlike Android, we aren't able to export a View or similar to the `GleanDebugActivity` in Android that can process debugging commands.  Instead, iOS allows for implementation of a custom URL scheme _within the application_.  Instead of launching the application through a proxy like `GleanDebugActivity`, we can launch the application from a URL with the debugging commands embedded in the URL itself.
```
Dexterp37:
```suggestion
iOS allows for implementation of a custom URL scheme _within the application_, so that users can launch the application from a URL with the debugging commands embedded in the URL itself.
```

Dexterp37: I'd personally like to keep the comparison between platforms out: in the log run, we might want a script for enabling consistent testing flows across platforms. See [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1568656)

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+### Available commands and query format
+
+There are 3 available commands that you can use with the Glean SDK debug tools
+
+- `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log
```
Dexterp37: What do you think of restructuring the doc a bit? One thing we could do is only report the options at the beginning of the page, before the platform specific bits. Then in each platform we could just add how these options translate to the platform specific way of working.

This doesn't need to happen as part of this PR. You can file a bug and take it after it if you want!

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+- `tagPings`: This command expects a string with upper and lower case letters, numbers and dashes, with a max length of 20 characters that will be used to tag the ping and cause it to appear in the Glean Debug View.
+- `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
+
+The structure of the custom URL uses the following format:
+
+`<protocol>://glean?<command 1>=<paramter 1>&<command 2>=<parameter 2> ...`
+
+Where:
+
+- `<protocol>` is the "Url Scheme" that has been added for your app (see Instrumenting the application below), such as `glean-sample-app`
```
Dexterp37: Do we have any recommendation for the protocol? What if the application already has a protocol there? Can we use the existing protocol? If so, can you add that as a note?

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+Right-click on `Item 0` and select `Add Row` from the context menu.  In the dropdown menu, select `URL Schemes` to add the item.
+
+Click on the disclosure icon of `URL Schemes` to expand the item, double-click the value field of `Item 0` and key in the value for your application's custom scheme.  For instance, the Glean sample app uses `glean-sample-app`, which allows for custom URL's to be crafted using that as a protocol, for example: `glean-sample-app://glean?logPings=true`
+
+#### Add the `Glean.handleCustomUrl()` convenience function and necessary overrides
+
+In order to handle the incoming Glean SDK debug commands, it is necessary to implement the override in the application's `AppDelegate.swift` file:
+
+```Swift
+func application(_: UIApplication,
```
Dexterp37: I think it might be better to condense this with the one below.

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+                  options _: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool
+```
+
+Within that function, you can make use of the convenience function provided in Glean `handleCustomUrl(url: URL)`.  An example of a simple implementation of this would look like this:
+
+```Swift
+func application(_: UIApplication,
+                  open url: URL,
+                  options _: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
+    // This does nothing if the url isn't meant for Glean.
```
Dexterp37:
```suggestion
    // ...
    // This does nothing if the url isn't meant for Glean.
```

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+    Glean.shared.handleCustomUrl(url: url)
```
Dexterp37:
```suggestion
    Glean.shared.handleCustomUrl(url: url)
    // ...
```

Dexterp37: This is just to hint that there _can_ be existing code there :)

docs/user/debugging.md
```diff
@@ -35,27 +42,28 @@ adb shell am start -n org.mozilla.samples.glean/mozilla.telemetry.glean.debug.Gl
...
+>
+> ```shell
 > adb shell am start -n org.mozilla.samples.glean/mozilla.components.service.glean.debug.GleanDebugActivity \
 >   --ez logPings true \
 >   --es sendPing metrics \
 >   --es tagPings test-metrics-ping
 > ```
 
-### Important GleanDebugActivity notes!
+### Important GleanDebugActivity notes
```
Dexterp37: This part applies to both the ios and the android part, as it depicts the two different workflows and it's not specifically tied to Android.

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+This method is useful for testing via the Simulator, which typically requires a Mac with Xcode installed, including the Xcode command line tools.  In order to perform the same command as above with using the browser to input the URL, you can use the following command in the command line terminal of the Mac:
+
+```shell
+xcrun simctl openurl booted "glean-sample-app://glean?logPings=true&tagPings=My-ping-tag&sendPing=events"
+```
+
+This will launch the simulator and again prompt the user with a dialog box asking if you want to open the URL in the Glean Sample App (or whichever app you are instrumenting and testing).
+
+#### Triggering commands if the app is already running
```
Dexterp37: I think this paragraph can be safely removed if we restructure this document, since this applies to all platforms and it's outlining the workflows.

docs/user/debugging.md
```diff
@@ -71,28 +79,132 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+This should cause iOS to prompt you with a dialog asking if you want to open the URL in the Glean Sample App, and if you select "Okay" then it will launch (or resume if it's already running) the application with the indicated commands and parameters and immediately force the collection and submission of the events ping.
+
+> **Note:** This method does not work if the browser you are using to input the command is the same application you are attempting to pass the Glean debug commands to.  So, you couldn't use Firefox for iOS to trigger commands within Firefox for iOS.
+
+It is also possible to encode the URL into a 2d barcode and launch the app via the camera app.  After scanning the encoded URL, the dialog prompting to launch the app should appear as if the URL were entered into the browser address bar.
+
+#### Using the command line
+
+This method is useful for testing via the Simulator, which typically requires a Mac with Xcode installed, including the Xcode command line tools.  In order to perform the same command as above with using the browser to input the URL, you can use the following command in the command line terminal of the Mac:
```
Dexterp37: Is there any way to _only_ install the command line tools needed to run this?

docs/user/debugging.md
travis79: Not a problem, my first pass at this was more conversational and I had to go back and refactor it, this is an artifact of that.  I can remove the comparisons to Android functionality.

docs/user/debugging.md
travis79: Most (if not all) of our iOS apps already implement a custom URL scheme like this based on the application name, so it is actually preferable to use the existing one.  We are really only adding a command processor to the existing mechanism, so it makes sense to recommend using an existing one.  Other than that, I don't know what other recommendations we could make here.

Dexterp37: This makes sense to me! Let's call out that existing protocols can be used.

docs/user/debugging.md
travis79: Yes, you can install the command line tools but not the simulator, it requires a lot of stuff within Xcode and there is not an easy way to get it installed without Xcode.

travis79:
> I'm only reviewing the docs of this PR :) Leaving the rest to @badboy . @travis79 is there a bug filed for the security review? Please not that it should probably block this from being merged.

This is an interesting thing to consider... just having this in the Glean SDK doesn't enable it by default in an application, it requires explicitly enabling it in the application.  Do we then want to have a blanket security review associated with the Glean SDK iOS tools, or should it be secreview'd when it is enabled in an application?  Technically an app could use Glean without enabling this functionality.

Dexterp37:
> 
> 
> > I'm only reviewing the docs of this PR :) Leaving the rest to @badboy . @travis79 is there a bug filed for the security review? Please not that it should probably block this from being merged.
> 
> This is an interesting thing to consider... just having this in the Glean SDK doesn't enable it by default in an application, it requires explicitly enabling it in the application. Do we then want to have a blanket security review associated with the Glean SDK iOS tools, or should it be secreview'd when it is enabled in an application? Technically an app could use Glean without enabling this functionality.

That's true, but we should do our due diligence before merging code that could be potentially risky :) The sec review is _for our feature_. Products using the feature should not require the sec review, unless the sec review on the feature state that as a requirement.

docs/user/debugging.md
travis79: [Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1599804)

docs/user/debugging.md
travis79: I added this information to the beginning of the section that describes how to implement this in the app, stating that if the app already has a custom URL scheme, then you can skip to adding the `Glean.handleCustomUrl()` section.

travis79:
I've started a [secreview? bug here](https://bugzilla.mozilla.org/show_bug.cgi?id=1599834).  I still need to fill out the form and attach it to the bug but I hope to finish that today.

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+- Not all commands are requred to be encoded in the URL, you can mix and match the commands that you need.  
+- Special characters should be properly URL encoded and escaped since this needs to represent a valid URL.
+
+### Instrumenting the application for Glean debug functionality
+
+In order to enable the debugging features in a Glean SDK consuming iOS application, it is necessary to add some information to the application's `Info.plist`, and add a line and possibly an override for a function in the `AppDelegate.swift`.
+
+#### Register custom URL scheme in `Info.plist`
+
+If your application already has a custom URL scheme implemented, there is no need to implement a second scheme, you can simply use that and skip to the next section about adding the convenience method.  If the app doesn't have a custom URL scheme implemented, then you will need to perform the following instructions to register your app to recieve custom URLs.
```
Dexterp37:
```suggestion
> **Note:** If your application already has a custom URL scheme implemented, there is no need to implement a second scheme, you can simply use that and skip to the next section about adding the convenience method.  If the app doesn't have a custom URL scheme implemented, then you will need to perform the following instructions to register your app to recieve custom URLs.
```

Dexterp37: What do you think of the above? It will render as an "important" text box, to highlight it.

.gitignore
```diff
@@ -39,4 +39,5 @@ glean_app*
 samples/ios/app/.venv/
 .mypy_cache/
 .coverage
-htmlcov/
\ No newline at end of file
+htmlcov/
+raw_sample_xcodetest.log
```
badboy: We produce multiple different of these log files.
(raw_sample_xcodetest.log, raw_xcodetest.log, raw_xcodebuild.log, raw_sample_xcodebuild.log)
Let's ignore them all.

```suggestion
*.log
```

docs/user/debugging.md
```diff
@@ -1,4 +1,24 @@
...
+1. [Debugging the Glean SDK in Android](#debugging-the-glean-sdk-in-android)
+2. [Debugging the Glean SDK in iOS](#debugging-the-glean-sdk-in-ios)
+
+### Important considerations when using Glean SDK debug tools
+
+- Options that are set using the adb flags are not immediately reset and will persist until the application is closed or manually reset.
+
+- There are a couple different ways in which to send pings using the Glean SDK debug tools.
+    1. You can tag pings using the debug tools and trigger them manually using the UI.  This should always produce a ping with all required fields.
+    2. You can tag _and_ send pings using the debug tools.  This has the side effect of potentially sending a ping which does not include all fields because `sendPings` triggers pings to be sent before certain application behaviors can occur which would record that information.  For example, `duration` is not calculated or included in a baseline ping sent with `sendPing` because it forces the ping to be sent before the `duration` metric has been recorded.
```
badboy: Worth mentioning that it might not send the ping if it is empty

docs/user/debugging.md
```diff
@@ -1,4 +1,24 @@
...
+    3. You can trigger a command while the instrumented application is still running.  This is useful for toggling commands or for triggering pings that have schedules that are difficult to trigger manually.  This is especially useful if you need to trigger a ping submission after some activity within the application, such as with the metrics ping.
+
+### Glean Log messages
+
+Glean logs warnings and errors through the platform specific logging frameworks.  See the platform specific instructions for information on how to view the logs.
```
badboy:
```suggestion
Glean logs warnings and errors through the platform-specific logging frameworks.  See the platform-specific instructions for information on how to view the logs.
```

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+## Debugging the Glean SDK in iOS
+
+For debugging and validation purposes on iOS, Glean makes use of a custom URL scheme which is implemented _within the application_ that is consuming Glean.  Glean provides some convenience functions to facilitate this, but it's up to the consuming application to enable this functionality.  Applications that enable this Glean SDK feature will be able to launch the application from a URL with the Glean debug commands embedded in the URL itself.
+
+### Available commands and query format
+
+There are 3 available commands that you can use with the Glean SDK debug tools
+
+- `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log
+- `tagPings`: This command expects a string with upper and lower case letters, numbers and dashes, with a max length of 20 characters that will be used to tag the ping and cause it to appear in the Glean Debug View.
```
badboy: I would rephrase that to first tell what it does, then mention the restrictions.

```suggestion
- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
```

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+Where:
+
+- `<protocol>` is the "Url Scheme" that has been added for your app (see Instrumenting the application below), such as `glean-sample-app`.
+- This is followed by `://` and then `glean` which is required for the Glean SDK to recognize the command is meant for it to process.
+- Following standard URL query format, the next character after `glean` is the `?` indicating the beginning of the query.
+- This is followed by one or more queries in the form of `<command>=<parameter>`, where the command is one of the commands listed above, followed by an `=` and then the value or parameter to be used with the command.
+
+There are a few things to consider when creating the custom URL:  
+
+- Invalid commands will trigger an error and be ignored.
```
badboy:
```suggestion
- Invalid commands will log an error and be ignored.
```

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+#### Add the `Glean.handleCustomUrl()` convenience function and necessary overrides
+
+In order to handle the incoming Glean SDK debug commands, it is necessary to implement the override in the application's `AppDelegate.swift` file.  Within that function, you can make use of the convenience function provided in Glean `handleCustomUrl(url: URL)`.  
+
+An example of a simple implementation of this would look like this:
+
+```Swift
+func application(_: UIApplication,
+                  open url: URL,
+                  options _: [UIApplication.OpenURLOptionsKey: Any] = [:]) -> Bool {
```
badboy: nit. Should this be aligned with the above `_`?

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+Using the glean-sample-app as an example: to activate ping logging, tag the pings to go to the Glean Debug View, and force the `events` ping to be sent, enter the following URL in a web browser on the iOS device:
+
+```shell
+glean-sample-app://glean?logPings=true&tagPings=My-ping-tag&sendPing=events
+```
+
+This should cause iOS to prompt you with a dialog asking if you want to open the URL in the Glean Sample App, and if you select "Okay" then it will launch (or resume if it's already running) the application with the indicated commands and parameters and immediately force the collection and submission of the events ping.
+
+> **Note:** This method does not work if the browser you are using to input the command is the same application you are attempting to pass the Glean debug commands to.  So, you couldn't use Firefox for iOS to trigger commands within Firefox for iOS.
```
badboy: Uh, interesting.

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+It is also possible to encode the URL into a 2d barcode and launch the app via the camera app.  After scanning the encoded URL, the dialog prompting to launch the app should appear as if the URL were entered into the browser address bar.
```
badboy:
```suggestion
It is also possible to encode the URL into a 2D barcode or QR code and launch the app via the camera app.  After scanning the encoded URL, the dialog prompting to launch the app should appear as if the URL were entered into the browser address bar.
```

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,77 @@ public class Glean {
...
+    /// - `glean` is required for the Glean SDK to recognize the command is meant for it to process.
+    /// - `?` indicating the beginning of the query.
+    /// - `<command>=<parameter>` are instances of the commands listed above  separated by `&`.
+    ///
+    /// There are a few things to consider when creating the custom URL:
+    ///
+    /// - Invalid commands will trigger an error and be ignored.
+    /// - Not all commands are requred to be encoded in the URL, you can mix and match the commands that you need.
+    /// - Special characters should be properly URL encoded and escaped since this needs to represent a valid URL.
+    public func handleCustomUrl(url: URL) {
```
badboy: I wonder if we should return a value indicating whether _any_ command was run, e.g. for the application to know about it? Not sure if that is really useful, it's just an idea I had when I read the above documentation

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,77 @@ public class Glean {
...
+            let params = components.queryItems else {
+            logger.error("Error parsing query parameters")
+            return
+        }
+
+        for param in params {
+            switch param.name {
+            case "tagPings":
+                if let pingTag = param.value,
+                    let pingTagPattern = try? NSRegularExpression(pattern: "^[a-zA-Z0-9-]{1,20}$") {
```
badboy: Can we note this in words somewhere? While the regex is simple, I'm afraid of us losing the context why we do this.

glean-core/ios/Glean/Glean.swift
```diff
@@ -263,6 +263,77 @@ public class Glean {
...
+                    logger.debug("No value passed to tagPings, clearing current value")
+                }
+            case "logPings":
+                configuration?.logPings = param.value == "true" ? true : false
+                logger.debug("Log pings set to: \(configuration!.logPings)")
+            case "sendPing":
+                if let pingName = param.value {
+                    // Check to see if the passed ping name matches a registered ping name
+                    // before sending the ping
+                    if pingTypeQueue.contains(where: { $0.name == pingName }) {
```
badboy: Do we need this? Does Android do this?
Glean-core should log an error if you're trying to send an unknown ping (which might be very helpful for debugging)

badboy: Oh! We should also defer the sending until after we looped through all parameters.
Otherwise we might first send a ping and only then parse the tag and later set it.

glean-core/ios/GleanTests/GleanTests.swift
```diff
@@ -3,22 +3,139 @@
...
+                jsonObject: [],
+                statusCode: statusCode,
+                headers: ["Content-Type": "application/json"]
+            )
+        }
+    }
+
     func testInitializeGlean() {
-        let glean = Glean.shared
+        Glean.shared.initialize()
```
badboy: With the above `setUp` method we already initialize Glean, making this the second initialize, which shouldn't be a problem, but it will log an error and not re-initialize.

That doesn't make the test wrong, but we could also remove that line and add a comment

samples/ios/app/Cartfile
badboy: With #531 landed you should be able to keep this as is and CI will do the right thing.

badboy:
Woops, forgot to add a comment.

Overall this looks good. Have some small things that should be changed, but nothing drastic.
this is blocked on the secreview for now as well.

docs/user/debugging.md
travis79: Yeah, you can't pass yourself URL's to process, apparently :/

glean-core/ios/Glean/Glean.swift
travis79: Let's wait until someone needs it.  It certainly sounds interesting, but I can't think of a use case aside from debugging at the moment.

glean-core/ios/Glean/Glean.swift
travis79: > Do we need this?

I don't know, [you are the one that asked for it 😉 ](https://github.com/mozilla/glean/pull/409#discussion_r349500378).  This was just the simplest way I found to accomplish it.

> Does Android do this?

I also [filed a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1599483) to replicate this behavior in Android.

badboy: Now I asked this question in both directions! (: 
Guess we keep it, it's debugging so I don't even care much about performance.

glean-core/ios/Glean/Glean.swift
travis79: That's actually a **_really_** good question.  I went looking for why we match against this particular pattern and couldn't find any restrictions in Glean or the glean_parser that required it.  I couldn't even find anything that looked like it directly required it for the Glean Debug View.  @Dexterp37 or @akkomar can refresh my memory as to why this restriction is in place.

Side note, glean_parser appears to use ping names for tests that [violate the pattern](https://github.com/mozilla/glean_parser/blob/967c6bafd35b66a41c9d2a0e06aaa31d59cb5eaf/tests/test_parser.py#L44).

Dexterp37: > We reference the pattern in our docs, but we don't provide any explanation or justification as to why.

The [prototype design document](https://docs.google.com/document/d/1w0aJK6AqqGAo_wr-iXMyBsmLYQTMKc5BXuSVkBS-MxU/edit?ts=5c6c335a#heading=h.xo415plei88v) states:

> Behind the scenes, glean will take care of marking the HTTP submissions of tagged pings by using a custom HTTP header, X-Debug-ID. The value for this header is the user-provided tag name, which needs to be constrained to a safe character set: [a-zA-Z0-9-]{1,20}.

This has nothing to do with the ping itself, it's only for our own sanity when communicating within http headers.

travis79: Thanks!  I had lost the context that that document contained!.  I'll add that statement as both a comment in the code and a line in the docs.

travis79:
Unblocked as the functionality was cleared in secreview.

travis79:
Just a note: I'd like to squash and organize the commits on this PR a little before merging.

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+
+- `<protocol>` is the "Url Scheme" that has been added for your app (see Instrumenting the application below), such as `glean-sample-app`.
+- This is followed by `://` and then `glean` which is required for the Glean SDK to recognize the command is meant for it to process.
+- Following standard URL query format, the next character after `glean` is the `?` indicating the beginning of the query.
+- This is followed by one or more queries in the form of `<command>=<parameter>`, where the command is one of the commands listed above, followed by an `=` and then the value or parameter to be used with the command.
+
+There are a few things to consider when creating the custom URL:  
+
+- Invalid commands will log an error and be ignored.
+- Not all commands are requred to be encoded in the URL, you can mix and match the commands that you need.
```
badboy:
```suggestion
- Not all commands are required to be encoded in the URL, you can mix and match the commands that you need.  
```

docs/user/debugging.md
```diff
@@ -71,28 +83,143 @@ More information can be found in the [View Logs with Logcat][] help article.
...
+- `<protocol>` is the "Url Scheme" that has been added for your app (see Instrumenting the application below), such as `glean-sample-app`.
+- This is followed by `://` and then `glean` which is required for the Glean SDK to recognize the command is meant for it to process.
+- Following standard URL query format, the next character after `glean` is the `?` indicating the beginning of the query.
+- This is followed by one or more queries in the form of `<command>=<parameter>`, where the command is one of the commands listed above, followed by an `=` and then the value or parameter to be used with the command.
+
+There are a few things to consider when creating the custom URL:  
+
+- Invalid commands will log an error and be ignored.
+- Not all commands are requred to be encoded in the URL, you can mix and match the commands that you need.  
+- Special characters should be properly URL encoded and escaped since this needs to represent a valid URL.
```
badboy: Afaik none of the commands should even accept anything outside of [a-zA-Z0-9-], thus not really needing encoding, right?

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -0,0 +1,141 @@
...
+        let logPings: Bool?
+        let pingNameToSend: String?
+    }
+
+    /// Helper function to process parameters passed as a query to the custom URL processed by `handleCustomUrl`
+    ///
+    /// - parameters:
+    ///     * urlQueryItems: A `URLQueryItem` representing the commands passed to Glean
+    ///
+    /// - returns: A tuple containing the values of the parameters `pingTag`, `logPings`, `pingNamesToSend` or nil.
```
badboy: Can we document when it returns nil? That is on invalid values or duplicated commands

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -0,0 +1,141 @@
...
+    private static func processCustomUrlQuery(urlQueryItems: [URLQueryItem]) -> ParsedQueryCommands? {
```
badboy: Yey! I do like this refactoring into its own function! :)
Much better then what I had in mind doing that inline :D

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -0,0 +1,141 @@
...
+        }
+
+        if let parsedCommands = processCustomUrlQuery(urlQueryItems: params) {
+            if let tag = parsedCommands.pingTag {
+                Glean.shared.configuration?.pingTag = tag
+                logger.debug("Pings tagged with: \(tag)")
+            }
+
+            if let logPings = parsedCommands.logPings {
+                Glean.shared.configuration?.logPings = logPings //
```
badboy:
```suggestion
                Glean.shared.configuration?.logPings = logPings
```

docs/user/debugging.md
travis79: I think ping names can contain underscores, but that's not usually a problem with URL encoding.  I'll remove this for now since we shouldn't allow characters that require encoding.

PR #410, opened 2019-10-24
1575848: Record an error if preinit tasks timeout

mdboom:


mdboom:
@janerik, @travis: Should we remove the swiftlint for TODOs?

glean-core/android/metrics.yaml
```diff
@@ -208,39 +208,52 @@ glean.internal.metrics:
...
+      - https://bugzilla.mozilla.org/1499761
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
+    expires: never
+    send_in_pings:
+      - all_pings
+    no_lint:
+      - COMMON_PREFIX
```
mdboom: This just fixes the indentation.

badboy: huh, wonder why these were all wrong to begin with.

travis79:
> @janerik, @travis: Should we remove the swiftlint for TODOs?

You want to do this so that we can catch the TODO's on build?

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -158,6 +158,9 @@ class Dispatchers {
         // `waitUntilFinished` parameter since this is a serial queue and any newly queued tasks
         // should execute after the `preInitOperations` that are being added here. For tests, we
         // need to await all of the tasks to finish execution, so we set this to true.
+
+        // TODO 1591094: Android has a timeout of 5 seconds for running preinit tasks.
```
badboy: swiftfmt formats this to `// TODO: 1591094: ...`

A better format would be: `// TODO(bug 1591094)` and then swiftfmt is happy as well

mdboom: Ah -- I thought swiftlint was just complaining about having a `TODO` at all.  (The output in CircleCI isn't terribly helpful -- I wonder if there's something we can do there...)

glean-core/android/metrics.yaml
mdboom: /me goes to add yamllint (if that even exists) to CI...

glean-core/ios/Glean/Dispatchers.swift
badboy: Unfortunately the output is awful indeed and it is `swiftformat` that is complaining.
I'll whip something up.

glean-core/android/metrics.yaml
```diff
@@ -208,39 +208,52 @@ glean.internal.metrics:
...
+    description:
+      Set to true if the tasks that are queued prior to Glean initialization
+      time out.
+    bugs:
+      - https://bugzilla.mozilla.org/1575848
+    data_reviews:
+      - TBD
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
+    expires: never
```
Dexterp37: Are we adding this to investigate a specific problem? If so, do we want to retain this really `forever`?

mdboom: Not to investigate a specific problem.  Just that tasks timing out would lead to data loss, and it would be good to know about it if it ever happens.  I thought `never` because I sort of see this in the same category as the other error reporting -- a general telemetry health metric...

mdboom:
Marking as WIP, since it needs data review for the new metric.

glean-core/android/metrics.yaml
Dexterp37: That's perfect, I wanted to make sure I wasn't missing something :)

PR #411, opened 2019-10-24
CLEANUP: Remove test for preinit timing distribution recording

mdboom:
I missed this as part of #400.  It is now fully covered by the tests added
there.

Also \o/ last `@Ignore`d test!


mdboom:
I guess we could just remove the ignore - but the two added [here](https://github.com/mozilla/glean/blob/master/glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt#L178) are more thorough...

badboy:
> I guess we could just remove the ignore - but the two added [here](https://github.com/mozilla/glean/blob/master/glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt#L178) [](/mozilla/glean/blob/HEAD@{2019-10-24T15:23:00Z}/glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt#L178) are more thorough...

works for me.

PR #416, opened 2019-10-28
Update dependencies

badboy:
[cargo-outdated](https://crates.io/crates/cargo-outdated) is a useful tool


glean-core/Cargo.toml
```diff
@@ -9,12 +9,12 @@ license = "MPL-2.0"
 edition = "2018"
 
 [dependencies]
-serde = { version = "1.0.90", features = ["derive"] }
-serde_json = "1.0.39"
-rkv = "0.9.4"
+serde = { version = "1.0.102", features = ["derive"] }
+serde_json = "1.0.41"
+rkv = "0.10.2"
```
Dexterp37: Can this potentially break loading data from an older version? This shouldn't given that it's no major version upgrade, but can you check please?

badboy: Nope, it's the same underlying lmdb, it's just bug fixes in rkv and lmdb-sys

PR #417, opened 2019-10-28
Truncate strings in a list at string boundaries

badboy:


mdboom:
:+1: yeah -- that's a "panic bomb" averted.

PR #424, opened 2019-10-29
Release version 19.1.0

badboy:


.buildconfig.yml
```diff
@@ -1,4 +1,4 @@
-libraryVersion: 20.0.0-SNAPSHOT
```
Dexterp37: Mh, this looks weird. Are we switching back to `20.0.0-SNAPSHOT` after this PR?

Dexterp37: Sounds good to me. Let's merge this, take it back to 20, and then discuss this in the next Glean SDK meeting

PR #432, opened 2019-11-01
1593055: Add testGetNumRecordedErrors to labeled metrics

mdboom:
#401 didn't include labeled metrics (in an attempt to not make that PR any bigger than it already was).

### DIFF
--- a/glean-core/src/metrics/labeled.rs
+++ b/glean-core/src/metrics/labeled.rs
@@ -125,6 +125,14 @@ where
             None => self.new_metric_with_dynamic_label(label.to_string()),
         }
     }
+
+    /// Get the template submetric.
+    ///
+    /// The template submetric is the actual metric that is cloned and modified
+    /// to record for a specific label.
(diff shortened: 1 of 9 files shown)


glean-core/src/metrics/labeled.rs
```diff
@@ -125,6 +125,14 @@ where
             None => self.new_metric_with_dynamic_label(label.to_string()),
         }
     }
+
+    /// Get the template submetric.
+    ///
+    /// The template submetric is the actual metric that is cloned and modified
+    /// to record for a specific label.
+    pub fn get_submetric(&self) -> &T {
```
mdboom: I'm not crazy on the access specifier here, but it's needed for use in the FFI layer, so `pub(crate)` isn't good enough.  @badboy: Any thoughts of alternatives?

badboy: We have the same problem for the `test_*` functions already. They are exposed.

We could think of hiding them from docs, just to make it clear that they are not actually public API

PR #433, opened 2019-11-04
Prepare for crates.io releases

badboy:
This adds a bit of metadata, makes the list of files to package explicit
and includes READMEs and license files per crate (because `cargo`
refuses to just pick the top-level README).

After merging what I need to do:

* Fetch master
* Reset the version numbers to 19.1.0
* Publish both crates

Later I'm also gonna provide configurations to use `cargo-release` and document that in the overall release process.

### DIFF
--- a/glean-core/README.md
+++ b/glean-core/README.md
@@ -1,0 +1,54 @@
+# Glean SDK
+
+The `Glean SDK` is a modern approach for a Telemetry library and is part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+
+## `glean-core`
+
+This library provides the core functionality of the Glean SDK, including implementations of all metric types, the ping serializer and the storage layer.
+It's used in all platform-specific wrappers.
+
+It's not intended to be used by users directly.
+Each supported platform has a specific Glean package with a nicer API.
+A nice Rust API will be provided by the [Glean](https://crates.io/crates/glean) crate.
+
+## Documentation
+
+All documentation is available online:
+
+* [The Glean SDK Book][book]
+* [API documentation][apidocs]
+
+[book]: https://mozilla.github.io/glean/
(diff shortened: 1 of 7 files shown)


glean-core/ffi/LICENSE
```diff
@@ -0,0 +1,373 @@
+Mozilla Public License Version 2.0
```
travis79: How unfortunate that we need multiple copies of our license at different directory levels.

badboy: oh well, at least this file doesn't change :)

glean-core/README.md
```diff
@@ -0,0 +1,50 @@
+# Glean SDK
+
+The `Glean SDK` is a modern approach for a Telemetry library and is part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+
+## `glean-core`
+
+This library provides the core functionality of the Glean SDK, including implementations of all metric types, the ping serializer and the storage layer.
+It's used in all platform-specific wrappers.
```
Dexterp37: Can we call out that this is not meant to be used directly by users? This is meant to be consumed by other Glean packages?

glean-core/ffi/README.md
```diff
@@ -0,0 +1,24 @@
+# Glean SDK
+
+The `Glean SDK` is a modern approach for a Telemetry library and is part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+
+## `glean-ffi`
+
+This library provides the FFI of the Glean SDK.
```
Dexterp37: nit: can we expand FFI and explain a bit more what's this about?

glean-core/README.md
badboy: Well, it can :D

PR #435, opened 2019-11-04
Cargo release

badboy:


mdboom:
Note before merging: This should be extended to also update the Glean version in `GleanGradlePlugin.groovy`.

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
-5. After the release, bump the versions to the next pre-release:
-    * Rust crates: Use the next major version and append `-alpha.1`
-    * Kotlin package: Use the next major version and append `-SNAPSHOT`
+    5. Release the Rust crates:
+
+       ```
+       cd glean-core
+       cargo publish --verbose
+       cd ffi
+       cargo publish --verbose
```
mdboom: Can we do this from CI?  (Could be a follow-on to this PR...)

bin/prepare-release.sh
```diff
@@ -0,0 +1,150 @@
...
+echo
+echo "Create tag v${NEW_VERSION} now? [y/N]"
+read -r RESP
+echo
+if [ "$RESP" != "y" ] && [ "$RESP" != "Y" ]; then
+    echo "No new tag. Proceed manually."
+    exit 0
+fi
+
+run git tag "v${NEW_VERSION}"
```
mdboom: If you add a tag here, and then follow the instructions below to make a pull request, merge it, and then tag using the Github UI, this tag won't really have any effect, right, since it's on the user fork (and will probably just cause confusion because there will be a tag in the user fork pointing to a different commit than that same tag on master).

Generally, I prefer to have scripts automate as much as possible, so it would be nice to just remove the manual github UI step described below to tag the release -- but I don't know how that would work with branch protection.

mdboom: Oh, and I guess this tag won't have the changelog in the commit message like it does using the instructions below.  (It could with some serious sed-fu, probably...)

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
+       ```
+
+> **Important:**
+> The release process produces release artifacts for the Kotlin library.
+> Artifacts for iOS wil be [generated on release soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1598276).
```
mdboom:
```suggestion
> Artifacts for iOS will be [generated on release soon](https://bugzilla.mozilla.org/show_bug.cgi?id=1598276).
```

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
+
+       ```
+       cd glean-core
+       cargo publish --verbose
+       cd ffi
+       cargo publish --verbose
+       ```
+
+> **Important:**
+> The release process produces release artifacts for the Kotlin library.
```
mdboom: It also publishes wheels to PyPI (as of yesterday).  There is a follow-on [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1598021) to also push those wheels to github releases, just so everything will also be in one place.

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
+3. Make a new branch with any fixes to be included in the release, *remembering not to make any breaking API changes.*.
+   This may involve cherry-picking fixes from master, or developing a new fix directly against the branch.
+   Example:
+
+   ```
+   git checkout -b fixes-for-v1.19.1 release-v19.1
+   git cherry-pick 37d35304a4d1d285c8f6f3ce3df3c412fcd2d6c6
+   git push -u origin fixes-for-v1.19.1
+   ```
+4. Follow the above steps for cuting a new release, except that:
```
mdboom:
```suggestion
4. Follow the above steps for cutting a new release, except that:
```

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
+    * When opening a PR to land the commits, target the `release-vX.Y` branch rather than master.
+    * When cutting the new release via github's UI, target the `release-vX.Y` branch rather than master.
+
+> **Note:** Point-releases for older versions should be rarely required.
+> We support the last released version and fixes should go there whenver possible.
```
mdboom:
```suggestion
> We support the last released version and fixes should go there whenever possible.
```

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
+   ```
+   git checkout -b fixes-for-v1.19.1 release-v19.1
+   git cherry-pick 37d35304a4d1d285c8f6f3ce3df3c412fcd2d6c6
+   git push -u origin fixes-for-v1.19.1
+   ```
+4. Follow the above steps for cuting a new release, except that:
+    * When opening a PR to land the commits, target the `release-vX.Y` branch rather than master.
+    * When cutting the new release via github's UI, target the `release-vX.Y` branch rather than master.
+
+> **Note:** Point-releases for older versions should be rarely required.
```
mdboom: Yeah, I agree with this.  If the project gets so high-velocity that we need bugfix branches on a regular basis, we can always revisit this.

bin/prepare-release.sh
```diff
@@ -0,0 +1,150 @@
+#!/usr/bin/env bash
```
mdboom: I wish we could write this in something other than bash, but I guess it's still our best option for "cross-platform and always available".  Not a problem with this PR, I guess I just wish the world were different :)

We should probably have @Dexterp37 test this on Windows and/or WSL.  When we were using `bumpversion` for `glean_parser` (which basically serves the exact same purpose as this), it choked on the `git` integration because it would use the `git` from WSL (expecting UNIX line endings) with a checkout from the native Windows `git` (that assumed DOS line endings).  I don't know if there's a general solution to that problem, but maybe there's a way we can help it use the correct `git` or provide a user override.

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,58 @@
...
 4. Cut the actual release.
     1. Click "Releases", and then "Draft a New Release" in the github UI.
-    2. Enter `v<myversion>` as the tag. In the example above it would be `v0.1.0`. It's important this is the same as the tags you put in the links in the changelog.
+    2. Enter `v<myversion>` as the tag. It's important this is the same as the tags you put in the links in the changelog.
     3. Under the description, paste the contents of the release notes from `CHANGELOG.md`.
     4. Note that the release is not available until the CI build completes for that tag.
         - You can check [on CircleCI for the running build](https://circleci.com/gh/mozilla/glean).
-5. After the release, bump the versions to the next pre-release:
-    * Rust crates: Use the next major version and append `-alpha.1`
-    * Kotlin package: Use the next major version and append `-SNAPSHOT`
```
mdboom: Just to confirm: we no longer need this step (to go back to alpha/SNAPSHOT)?  I agree it would be nice to skip it.  I guess it only matters if someone wanted to manually publish a release between tagged releases.  That should be *really* rare and you'd have to really know what you are doing anyway.

bin/prepare-release.sh
Dexterp37: Maybe a good solution here is that I simply set my repo to

```
git config --global core.autocrlf input
git config --global core.eol lf
```

?

docs/dev/cut-a-new-release.md
Dexterp37: @badboy does it make sense to keep this manual docs somewhere in the `internal` docs section, in case automation goes crazy?

docs/dev/cut-a-new-release.md
badboy: Probably, but yes, in a followup (filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1598291)

bin/prepare-release.sh
badboy: Ah, damn, you are right. Let's not create the tag here then for now.

bin/prepare-release.sh
badboy: I could always write it in Rust :)

docs/dev/cut-a-new-release.md
badboy: a "manual release between tagged releases" would be the workflow described further below.
I don't want these to happen off master anymore.

Not sure why we would need this anywhere internal. Having a "-SNAPSHOT" version should not be needed anymore on master at all

docs/dev/cut-a-new-release.md
```diff
@@ -1,38 +1,61 @@
...
-            - Note that we try to provide PR or issue numbers (and links) for each change. Please add these if they are missing.
-
-    2. Add a new `# Unreleased Changes` on top
-
-2. Bump the versions
-    * Rust crates: Bump `version` in [`glean-core/Cargo.toml`](https://github.com/mozilla/glean/blob/master/glean-core/Cargo.toml) and [`glean-core/ffi/Cargo.toml`](https://github.com/mozilla/glean/blob/master/glean-core/ffi/Cargo.toml).  **NOTE:** There are two versions to update in `glean-core/ffi/Cargo.toml`.
-    * Kotlin package: Bump `libraryVersion` in the top-level [.buildconfig.yml](https://github.com/mozilla/glean/blob/master/.buildconfig.yml) file.
-    * Gradle plugin: Bump `project.ext.gleanVersion` in [`GleanGradlePlugin.groovy`](https://github.com/mozilla/glean/blob/master/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy).
-    * Be sure you're following semver, and if in doubt, ask.
-    * Run `make build` to make sure the `Cargo.lock` file is updated.
```
mdboom: Just remembered: We'll also need to automate this step.

badboy: I saw that, removed it and then forgot to add it to the script

badboy:
This was used in #520, so I guess it's good to merge now

PR #438, opened 2019-11-04
1593056: Implement testGetNumRecordedErrors for iOS

mdboom:


glean-core/ios/Glean/Metrics/TimingDistributionMetric.swift
```diff
@@ -149,4 +149,25 @@ public class TimingDistributionMetricType {
 
         return DistributionData(fromJson: json)!
     }
+
+    /// Returns the number of errors recorded for the given metric.
```
travis79: I'm pretty sure there's a way to reduce code duplication using `protocol` (similar to Android `interface`), but that's just tech debt and is out of scope for this PR.  

Maybe we should file a bug to reduce duplication in iOS specifically?  This could even be a "Good First Bug" or something.

glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
```diff
@@ -115,4 +115,19 @@ class TimingDistributionTypeTests: XCTestCase {
...
+            name: "timing_distribution",
+            sendInPings: ["store1"],
+            lifetime: .ping,
+            disabled: false,
+            timeUnit: .millisecond
+        )
+
+        metric.stopAndAccumulate(-1)
+
+        XCTAssertEquals(1, metric.testGetNumRecordedErrors(.invalidValue))
```
travis79:
```suggestion
        XCTAssertEqual(1, metric.testGetNumRecordedErrors(.invalidValue))
```

glean-core/ios/Glean/Metrics/TimingDistributionMetric.swift
badboy: We have the same issue on Android. I'd rather see us come up with an idea for both platforms (and future ones).

Unfortunately a protocol/interface alone is not the solution. We still need to call individually different functions per metric type.

docs/user/metrics/string_list.md
```diff
@@ -70,14 +70,16 @@ Search.engines.set(engines)
 
 There are test APIs available too:
 
-```Kotlin
+```Swift
```
badboy: Woops

glean-core/ios/Glean/Metrics/TimingDistributionMetric.swift
travis79: This was just something that occurred to me as I was looking at the same function added so many times.  

I agree, if we can find a solution for all platforms, that would be ideal.

PR #439, opened 2019-11-04
Implement counter metric type for Python bindings.

mdboom:
It's pretty helpful to have type annotations for the documentation, if nothing
else, so additionally, this adds support for mypy type annotations.


glean-core/android/src/test/java/mozilla/telemetry/glean/private/CounterMetricTypeTest.kt
```diff
@@ -56,25 +56,6 @@ class CounterMetricTypeTest {
...
-            name = "counter_metric",
-            sendInPings = listOf("store1")
-        )
-
-        // Attempt to increment the counter
-        counterMetric.add(1)
-        // Check that nothing was recorded.
-        assertFalse("Counters must not be recorded if they are disabled",
-            counterMetric.testHasValue())
-    }
```
mdboom: This test is *exactly* the same as the one below it (except the name and the comment), so I just went ahead and removed it.  Save a few cycles per test ;)

.circleci/config.yml
```diff
@@ -492,14 +492,14 @@ jobs:
             npm install -g --silent gh-pages@2.0.1
             gh-pages --dotfiles --message "[skip ci] Updates" --dist build/docs
 
-  Lint Python with flake8 and black:
+  Lint Python with flake8 and black and mypy:
```
badboy: This becomes unhandy. Should we just name it `Lint Python` and not bother to change it with every new tool we add?

.flake8
```diff
@@ -0,0 +1,2 @@
+[flake8]
+max-line-length = 88
```
badboy: :/
Any reason we don't go straight to 100?

glean-core/python/glean/_ffi.py
```diff
@@ -64,4 +66,25 @@ def make_config(data_dir, package_name):
...
+    values.append(ffi.NULL)
+
+    result = ffi.new("char *[]", values)
+
+    _global_weakkeydict[result] = values
+
+    return result
+
+
+__all__ = ["ffi", "ffi_string", "ffi_vec_string", "lib", "make_config"]
```
badboy: What does `__all__` do? Just mark what functions are exported?

.flake8
Dexterp37: How was 88 picked? Does this come from a PEP or something?

FWIW Mozilla has [these guidelines](https://developer.mozilla.org/en-US/docs/Mozilla/Developer_guide/Coding_Style#Python_Practices) for Python.

glean-core/python/glean/metrics/counter.py
```diff
@@ -0,0 +1,130 @@
...
+        Add to counter value.
+
+        Args:
+            amount (int): (default: 1) This is the amount to increment the
+                counter by.
+        """
+        if self._disabled:
+            return
+
+        # TODO: Run on the dispatcher
```
Dexterp37: nit: can we reference the bug number, here?

PR #441, opened 2019-11-05
[WIP] Automatically substitute glean's version in forUnitTest dependencies

Dexterp37:
This attempts to find the version of the Glean SDK currently being used in projects using the gradle plugin and to add that version as a constraint to any `forUnitTest` glean dependency available.

**Note** this is currently lacking the constraint part

Todo:

- [ ] Add a [dependency constraint](https://docs.gradle.org/current/javadoc/org/gradle/api/artifacts/dsl/DependencyConstraintHandler.html) to enable using `forUnitTest` dependencies without a version number
- [ ] Remove the changes to `.buildconfig.yml`
- [ ] Add to the changelog

### DIFF
--- a/.buildconfig.yml
+++ b/.buildconfig.yml
@@ -1,4 +1,4 @@
-libraryVersion: 20.0.0-SNAPSHOT
+libraryVersion: 20.0.0-TEST6
 groupId: org.mozilla.telemetry
 projects:
   glean-core:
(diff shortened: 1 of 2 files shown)


mdboom:
I might be missing something, but I think we have a chicken-and-egg problem here.  User libraries would include the Glean plugin by specifying a Glean version.  The plugin can then provide a Glean version, but they need to know it in the first place.

(Or maybe I'm just missing something about how this would work in practice...)

PR #443, opened 2019-11-05
Add a simple baseline test for the sample app

badboy:
Same test as the android sample app

* [x] Wait for #442 to land
* [x] Enable on CI


samples/ios/app/glean-sample-app/Info.plist
```diff
@@ -41,5 +41,12 @@
 		<string>UIInterfaceOrientationLandscapeLeft</string>
 		<string>UIInterfaceOrientationLandscapeRight</string>
 	</array>
+	<key>NSAppTransportSecurity</key>
```
Dexterp37: nit: can we add a comment here mentioning why these permissions are required? (i.e. sending pings to a local server on UI tests)

samples/ios/app/glean-sample-app/AppDelegate.swift
```diff
@@ -13,7 +13,16 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
     func application(_: UIApplication, didFinishLaunchingWithOptions _: [UIApplication.LaunchOptionsKey: Any]?) -> Bool {
         let glean = Glean.shared
         glean.setUploadEnabled(true)
-        glean.initialize()
+
+        if ProcessInfo.processInfo.arguments.contains("USE_MOCK_SERVER") {
+            let address = "http://localhost:9080"
+            print("using a mock server, setting address: \(address)")
```
Dexterp37: Is this and the one below meant to be a `printf`? Does Apple use any other logging mechanism that we should use here?

samples/ios/app/glean-sample-appUITests/BaselinePingTest.swift
```diff
@@ -0,0 +1,93 @@
...
+        // In UI tests it is usually best to stop immediately when a failure occurs.
+        continueAfterFailure = false
+
+        // UI tests must launch the application that they test.
+        // Doing this in setup will make sure it happens for each test method.
+        app = XCUIApplication()
+    }
+
+    override func tearDown() {
+        server?.stop()
```
Dexterp37: Is this being called after each test? If so, should we call `setupServer` in `setUp`? (or comment on why this is not being done)

bin/run-ios-sample-app-build.sh
```diff
@@ -0,0 +1,13 @@
+#!/usr/bin/env bash
```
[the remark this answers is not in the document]
Dexterp37: (it's fine to do this as a follow-up or file a bug to add it to all bash scripts)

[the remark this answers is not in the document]
badboy: *assigns it to the public domain* ... but actually I'm filing a bug to figure this out

samples/ios/app/glean-sample-app/Info.plist
badboy: This file is edited through the GUI. I'm not sure how nice it plays with adding comments :/

samples/ios/app/glean-sample-app/AppDelegate.swift
badboy: `print` it is! (or `NSLog`, which is slower).  We use the same in the Logger class in Glean.

As this is just the sample app I didn't want to add more sophisticated logging.

samples/ios/app/glean-sample-appUITests/BaselinePingTest.swift
badboy: This is by now mostly a left-over of my previous code.
I'm refactoring the server function in #444, would you okay with letting this slip here and I move around the `server.stop` over in the follow up?

Dexterp37: Yup! That's fine, thanks!

samples/ios/app/glean-sample-app/Info.plist
Dexterp37: Oh, eugh. Ok! No problem then!

badboy: Just tested it. Xcode will removed the comment if someone opens and saves the Info.plist inside Xcode.

PR #446, opened 2019-11-05
1594033, 1591188: Python: Implement core API and Dispatcher

mdboom:


glean-core/python/glean/config.py
```diff
@@ -3,9 +3,49 @@
...
+# The default server pings are sent to
+DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
+
+
+# The default UserAgent string
+def get_default_user_agent():
+    # Need to use a function here to avoid a cyclical import problem
+    from glean import __version__
+
+    return f"Glean/{__version__} (Python)"
```
Dexterp37: This is interesting. For other platforms, we used `Android` or `ios`. Here we're using a language, `Python`. This can indeed run on any platform, including `Android`, so it's probably the best choice.

Feels a bit odd, though :)

glean-core/python/glean/config.py
```diff
@@ -3,9 +3,49 @@
...
+
+
+# The default number of events to store before sending
+DEFAULT_MAX_EVENTS = 500
```
Dexterp37: Does it make sense to move all the default/constants at the beginning of the file?

glean-core/python/glean/metrics/counter.py
```diff
@@ -51,6 +52,7 @@ def __del__(self):
         if getattr(self, "_handle", 0) != 0:
             _ffi.lib.glean_destroy_counter_metric(self._handle)
 
+    @Dispatcher.launch
```
Dexterp37: We don't really want the whole function to be dispatched, we only want the [content after the disable check](https://github.com/mozilla/glean/blob/1ee40df622a8084914c8d68210fb4532ffcbea8e/glean-core/android/src/main/java/mozilla/telemetry/glean/private/CounterMetricType.kt#L74) to be dispatched. While this doesn't matter much _right now_, this might create inconsistencies in the behaviour when we add asynchrony or new metric types.

glean-core/python/glean/config.py
mdboom: We could do both.  What about `Glean/{__version__} (Python on {sys.platform})`.  sys.platform gives: `linux`, `darwin` or `win32`.

glean-core/python/glean/metrics/counter.py
mdboom: Good point -- had to refactor how the dispatchers work to make this clean, but it should work.

glean-core/python/glean/config.py
mdboom: Technically, these were all "constants", just one had to have deferred evaluation.  But I had to refactor this anyway to make the documentation generator happy :)

PR #450, opened 2019-11-06
1591188: Implement metrics.yaml/pings.yaml loader for Python bindings

mdboom:


glean-core/python/glean/_loader.py
```diff
@@ -0,0 +1,146 @@
...
+    "memory_unit",
+    "name",
+    "range_max",
+    "range_min",
+    "send_in_pings",
+    "time_unit",
+]
+
+
+class UnsupportedMetricType:
```
Dexterp37: Are we planning on removing this once we reach feature parity? If so, can you please file a bug about removing it and add the bug number as a TODO comment here?

glean-core/python/glean/glean.py
```diff
@@ -189,12 +202,17 @@ def _initialize_core_metrics(cls):
         """
         Set a few metrics that will be sent as part of every ping.
         """
+        from . import _builtins
+
         # TODO: 1594184
-        pass
+        _builtins.metrics.glean
```
Dexterp37: Is this triggering the load of builtin metrics from registry files? If so, any chance you could add a small comment about this?

glean-core/python/glean/glean.py
```diff
@@ -6,14 +6,18 @@
...
-from typing import Optional
+from typing import Optional, Set, TYPE_CHECKING
 
 
 from .config import Configuration
 from ._dispatcher import Dispatcher
 from . import _ffi
 
 
+if TYPE_CHECKING:
```
Dexterp37: nit: can we add a small comment here about what's the purpose of this?

glean-core/python/glean/_loader.py
```diff
@@ -0,0 +1,146 @@
...
+    """
+    Load metrics from a `metrics.yaml` file.
+
+    Args:
+        filepath: The path to the file, or a list of paths, to load.
+        config (dict): A dictionary of options that change parsing behavior.
+            These are documented in glean_parser:
+            https://mozilla.github.io/glean_parser/glean_parser.html#glean_parser.parser.parse_objects
+    Returns:
+        metrics: A tree of metrics, as defined in the `metrics.yaml` file.
```
Dexterp37: When you say "tree of metrics", what do you mean? What's the underlying type? A dictionary?

glean-core/python/glean/metrics/ping.py
```diff
@@ -0,0 +1,33 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+from ..glean import Glean
+from .. import _ffi
+
+
+class PingType:
```
Dexterp37: Mh, should we update the [docs here](https://mozilla.github.io/glean/book/user/pings/custom.html) too? Can be a follow-up bug.

glean-core/python/glean/_loader.py
mdboom: We'll probably never remove this -- we should have some way to handle metric types the Python bindings don't support (maybe "CustomDistributions") without needing to throw out the entire metrics.yaml file.

glean-core/python/glean/_loader.py
mdboom: They are just `objects`.  I've added an example that perhaps explains better.

glean-core/python/glean/_loader.py
Dexterp37: I see, ok!

PR #453, opened 2019-11-07
Implement ping sending in the Python bindings

mdboom:


glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt
```diff
@@ -90,7 +90,7 @@ class HttpURLConnectionUploader : PingUploader {
                     // Known other errors:
                     // 500 - internal error
 
-                    // For all other errors we log a warning an try again at a later time.
```
travis79: If you keep finding and fixing a couple of these with each python PR, you the Kotlin is going to be quite tidy :D

glean-core/python/glean/glean.py
```diff
@@ -50,12 +55,16 @@ class Glean:
...
+    # The ping uploader implementation
+    _ping_uploader: BaseUploader = HttpClientUploader()
+
     @classmethod
     def initialize(
         cls,
         configuration: Optional[Configuration] = None,
         application_id: Optional[str] = None,
         data_dir: Optional[Path] = None,
+        ping_uploader: Optional[PingUploader] = None,
```
Dexterp37: Let's be consistent with what [Kotlin does](https://github.com/mozilla/glean/blob/0b2f2ceea6171a3c6551345f0075bf51031a053a/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt#L94) and add that as part of the configuration object.

glean-core/python/glean/glean.py
```diff
@@ -85,12 +97,12 @@ def initialize(
         if data_dir is None:
             data_dir = Path(tempfile.TemporaryDirectory().name)
 
+        if ping_uploader is None:
+            ping_uploader = HttpClientUploader()
```
Dexterp37: Isn't this already set on line 59?

glean-core/python/glean/glean.py
```diff
@@ -219,5 +240,60 @@ def get_data_dir(cls) -> Path:
...
+    def test_collect(cls, ping: "PingType") -> str:
+        """
+        Collect a ping and return as a string.
+        """
+        return _ffi.ffi_decode_string(
+            _ffi.lib.glean_ping_collect(cls._handle, ping._handle)
+        )
+
+    @classmethod
+    def send_pings(cls, pings: List["PingType"]):
```
Dexterp37: Is there any way we could flag/make this private or only accessible from within this package?

glean-core/python/glean/glean.py
```diff
@@ -219,5 +240,60 @@ def get_data_dir(cls) -> Path:
...
+        Args:
+            pings (list of PingType): List of pings to send.
+        """
+        ping_names = [ping.name for ping in pings]
+
+        cls.send_pings_by_name(ping_names)
+
+    @classmethod
+    @Dispatcher.task
+    def send_pings_by_name(cls, ping_names: List[str]):
```
Dexterp37: As above, any chance we could make this visible only from within this package?

glean-core/python/glean/glean.py
mdboom: You can't really make anything private in Python, but we can flag it with a `_` prefix.  I missed these were internal in the Kotlin implementation.

PR #455, opened 2019-11-08
Implement core metrics in Python bindings

mdboom:


glean-core/python/glean/glean.py
```diff
@@ -55,11 +57,18 @@ class Glean:
...
+
+    # The version of the application sending Glean data.
+    _application_version: Optional[str] = None
+
     @classmethod
     def initialize(
         cls,
         configuration: Optional[Configuration] = None,
         application_id: Optional[str] = None,
+        application_version: Optional[str] = None,
```
Dexterp37: Can we move this and the above in the configuration object? This is where the `packageName` (from which the app id is derived) lives in [Kotlin](https://github.com/mozilla/glean/blob/e27fc8abe79741da7a45d0602babf8c334db1372/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt#L23). We should also remove the user agent property from there (sorry, I missed that during my review) since we don't allow customizing it in the Glean SDK.

glean-core/python/glean/glean.py
```diff
@@ -55,11 +57,18 @@ class Glean:
     # and saved between test runs.
     _ping_type_queue: Set["PingType"] = set()
 
+    # The application id to send in the ping.
+    _application_id: str = "glean-python"
```
Dexterp37: Let's not provide defaults for this, if possible. We should rather make the configuration object required, if needed: otherwise if anyone else starts using these bindings without paying attention to the docs, we'd end up inflating this app id. Let's be explicit!

glean-core/python/glean/util.py
```diff
@@ -0,0 +1,24 @@
...
+    Get a Gecko-compatible locale string (e.g. "es-ES", instead of the "es_ES")
+    for the currently set locale.
+
+    Returns:
+        locale (str): The locale string.
+    """
+    value = locale.getlocale()[0]
+
+    # The format of the locale string is platform depedent. At least on Linux,
+    # often an understore is used between language and country, which is not
```
Dexterp37:
```suggestion
    # often an underscore is used between language and country, which is not
```

glean-core/python/glean/util.py
```diff
@@ -0,0 +1,24 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+
+import locale
+
+
+def get_locale_tag() -> str:
```
Dexterp37: Just to double check: the `baseline` ping field this function is populating represents `application locale`, not the system locale. Does this function returns the same? Is there any difference in Python  between app locale and system locale?

mdboom: Python follows POSIX on this, where this is a system locale, and the user can override their locale, and the locale can be overridden anywhere in the process tree.  The Python locale API only gives access to the context in which it is running, so I guess that's basically analogous to "application locale"...

glean-core/python/glean/glean.py
mdboom: I'd like to think on this one.  As this library is primarily being used for writing tests, requiring a configuration object means that we can no longer provide a testing decorator to reset Glean for each test (since those don't take arguments).  We'd need a way to set it globally in a way that would survive Glean resets (or else make the API for testing far less convenient and error-prone by requiring users to reset Glean themselves).

Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1595486)

glean-core/python/glean/glean.py
mdboom: Well, looking closer, I don't think that `Configuration.packageName` is used anywhere in Kotlin.  Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1595487) about that.

Since the Kotlin version pulls this "out of thin air" from the `applicationContext` object, and the same is true of `data_dir`, I thought it was more consistent with Kotlin to have this be part of the `Glean` object.  My thinking is "the Configuration object should be pretty consistent across platforms" and the Glean init params are for platform-specific things.  But I guess I don't feel too strongly about it.  Alternatively, we could put everything on the configuration object (but then we should probably also move `data_dir` to the configuration object, too).

Dexterp37: Fair point and good catch. Let's leave this as it is for now.

glean-core/python/glean/glean.py
```diff
@@ -220,12 +233,25 @@ def _initialize_core_metrics(cls):
...
         """
-        from . import _builtins
+        from ._builtins import metrics
+
+        metrics.glean.baseline.locale.set(util.get_locale_tag())
+        metrics.glean.internal.metrics.os.set(platform.system())
+        metrics.glean.internal.metrics.os_version.set(platform.release())
+        metrics.glean.internal.metrics.architecture.set(platform.machine())
+
+        # device_model and device_manufacturer aren't easily obtainable in this
```
Dexterp37: Are these required in the Glean schema?

mdboom: Indeed they are.  So I guess we'll have to put something here...

mdboom: I'm putting "unknown" in here for now, so it passes the schema.  I don't think there's really a cross-platform way to get the manufacturer/model of the desktop computer.  We can get CPU architecture and such, and OS version things (which are already reported in other fields).

We could adjust the schema to not require these...

Dexterp37: > 
> 
> I'm putting "unknown" in here for now, so it passes the schema. I don't think there's really a cross-platform way to get the manufacturer/model of the desktop computer. We can get CPU architecture and such, and OS version things (which are already reported in other fields).

Sounds good to me for the MVP!

> We could adjust the schema to not require these...

Yeah, I suspect this might be a schema-v2 thing.. Any chance you could file a bug, tag it with schema-v2, add a comment to this part of the code with the bug number?

Dexterp37:
That was a r+ with the comment addressed 👍

glean-core/python/glean/glean.py
mdboom: I did a little poking around after I wrote this yesterday.  It turns out the information exists in the hardware BIOS.  There just doesn't seem to be a cross-platform way to get at it, so we would probably have to build it.  I agree we don't need any of this for MVP.

But I reworked the bug to be [add device info to Python bindings](https://bugzilla.mozilla.org/show_bug.cgi?id=1595751), rather than removing them from the schema.

PR #457, opened 2019-11-08
1594711: Correctly define event timestamps as in milliseconds

mdboom:


glean-core/src/metrics/event.rs
```diff
@@ -50,7 +50,7 @@ impl EventMetric {
     /// ## Arguments
     ///
     /// * `glean` - The Glean instance this metric belongs to.
-    /// * `timestamp` - A monotonically increasing timestamp, in nanoseconds.
+    /// * `timestamp` - A monotonically increasing timestamp, in milliseconds.
```
Dexterp37: There's also [another entry](https://github.com/mozilla/glean/blob/6880581aaf066fb7b2b55333e1c6def1a0e9669f/glean-core/src/event_database/mod.rs#L164) that needs to be fixed.

Dexterp37:
Ps: the Rust errors will go away if you rebase, I fixed them this morning :)

PR #458, opened 2019-11-08
1594712: Sort events by timestamp on their way out.

mdboom:


mdboom:
> We should make sure this does not become a performance bottleneck

Yeah, I agree.  Hopefully it won't.  It's always limited to at most 500 events, and they should be "mostly already sorted" which should help as well.

PR #459, opened 2019-11-08
1587136: Document difference between duration and start/end_time

mdboom:


docs/user/pings/baseline.md
```diff
@@ -18,6 +18,18 @@ The baseline ping includes the following fields:
 | `duration` | Timespan | The duration, in seconds, of the last foreground session |
 | `locale` | String | The locale of the application |
 
+A quick note about the `duration` metric: it measures specifically the time that the user spent in the last foreground session.
+This is distinct from the `start_time` and `end_time` values (on every ping), which measure the ping lifetime.  `start_time` is always set to be the time the previous ping of the same type was sent, so the span from `start_time` to `end_time` is normally longer than the span measured by `duration`.
```
Dexterp37: Does it make sense to move this text within the above table? It makes it more discoverable. I'm not sure if the same could be done with the diagram below: that would be great, but not required :)

mdboom: It looked weird to me to have so much content in the table (and it doesn't seem you can put paragraph breaks there -- or at least I couldn't figure it out).  What about using a footnote link (as I just updated this to do)?

Dexterp37: Does that linkify automatically in the rendered docs? If so, I'm happy with this!

mdboom: Yes.  It adds a little superscript `1` that links to the bit of text below.

PR #463, opened 2019-11-11
Move to 21.0.0-alpha.1 (SNAPSHOT)

mdboom:


glean-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "glean-core"
-version = "20.0.0"
+version = "21.0.0-alpha0"
```
Dexterp37: Step 5 [here](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html) says this should be `-alpha.1`. Does it make any difference?

mdboom: I don't know if it matters -- I had actually looked at what it was prior to 20.0.0 here and missed the dot.  Let's follow the docs here...  I'll update.

PR #467, opened 2019-11-11
Labeled metric in Python

mdboom:
~Depends on #455.~


glean-core/python/glean/metrics/labeled.py
```diff
@@ -0,0 +1,138 @@
...
+
+        return self._test_get_num_recorded_errors_ffi(
+            Glean._handle,
+            self._handle,
+            error_type.value,
+            _ffi.ffi_encode_string(ping_name),
+        )
+
+
+class LabeledCounterMetricType(LabeledMetricBase):
```
Dexterp37: That's really neat.

PR #468, opened 2019-11-11
Bug 1595211 - Makes GleanDebugActivity launch as a `singleInstance` Activity.

travis79:
This makes the `GleanDebugActivity` launch as a [`singleInstance` Activity](https://developer.android.com/guide/topics/manifest/activity-element.html#lmode).  This causes the `GleanDebugActivity` to correctly handle additional `Intents` if the application was launched with a `GleanDebugActivity` `Intent` using the `adb` command.

### DIFF
--- a/glean-core/android/src/main/AndroidManifest.xml
+++ b/glean-core/android/src/main/AndroidManifest.xml
@@ -2,6 +2,7 @@
     package="mozilla.telemetry.glean" >
     <application>
         <activity android:name=".debug.GleanDebugActivity"
+            android:launchMode="singleInstance"
             android:exported="true" />
     </application>
 </manifest>


glean-core/android/src/main/AndroidManifest.xml
```diff
@@ -2,6 +2,7 @@
     package="mozilla.telemetry.glean" >
     <application>
         <activity android:name=".debug.GleanDebugActivity"
+            android:launchMode="singleInstance"
```
Dexterp37: Good catch on this, this is very interesting. I have a couple of questions: have you already tried to add test coverage for this (without this line, can we make any test that fails?)? The [`singleInstance` docs](https://developer.android.com/guide/topics/manifest/activity-element.html#lmode) say that `Same as "singleTask", except that the system doesn't launch any other activities into the task holding the instance`. The `singleTask` docs mention that `If an instance of the activity already exists at the top of the target task, the system routes the intent to that instance through a call to its onNewIntent() method, rather than creating a new instance of the activity.`. Does it mean we should also implement `onNewIntent`?

Can we also have an Android expert double check this once you've answered the questions above?

I'm a bit worried that this might end up breaking other intent-based interactions for Fenix, due to the `The other modes — singleTask and singleInstance — are not appropriate for most applications, since they result in an interaction model that is likely to be unfamiliar to users and is very different from most other applications. ` statement at the end of the docs.

travis79: We don't need to implement `onNewIntent`, if you look closely, I've used `singleInstance` and not `singleTask`.  Basically, this forces the application that is launched by the `GleanDebugActivity` to be launched in it's own task stack rather than on the same one as the GleanDebugActivity, allowing the call to `finish()` to dispose of it as we would expect.

This doesn't affect any application's intent-based actions as it only affects the `GleanDebugActivity` and not Activities launched from it.  We happen to fall into one of those special cases of launching ("not appropriate for most applications") because the `GleanDebugActivity` is a headless activity that merely launches another Activity that we want to be the root Activity for the new task stack.

travis79: As for creating a test?  I can try but I imagine that this will be a lot of effort for a little gain.  I'll spend some time on that today and if it looks like it's going to be more difficult than I can put together in a day, I'll re-evaluate the situation with you.

travis79:
Follow up bug here: https://bugzilla.mozilla.org/show_bug.cgi?id=1595826

PR #469, opened 2019-11-12
1595486: Make app_id and app_version required

mdboom:


glean-core/python/glean/_dispatcher.py
```diff
@@ -7,23 +7,31 @@
 from typing import Callable, List, Tuple
 
 
-class DispatcherInternal:
+class Dispatcher:
```
mdboom: This is to change how the singleton of this is managed.  If you do something like `FooSingleton = Foo()` in Python, and then want to recreate (and thus reset the singleton), that doesn't affect all of the places that have already imported the singleton, so this was causing strange issues.  Moving this to a class with only `@classmethod` (static methods) seems like a safer pattern for singletons in Python.  (The `Glean` singleton already works this way).

Dexterp37: Oh. TIL.

glean-core/python/tests/metrics/test_labeled.py
```diff
@@ -146,7 +142,7 @@ def test_other_label_without_predefined_labels_before_glean_init():
         labeled_counter_metric[f"label_{i}"].add(1)
     labeled_counter_metric["label_0"].add(1)
 
-    Glean.initialize()
+    Glean.initialize("glean-python", glean_version)
```
Dexterp37: Since this is a test , can we have some id like `glean-python-tests`?

glean-core/python/tests/test_glean.py
```diff
@@ -15,21 +15,20 @@
...
-from glean import testing
+from glean import util
 from glean._dispatcher import Dispatcher
 from glean.metrics import CounterMetricType, Lifetime, PingType
-from glean import util
 
 
-def setup_function():
-    testing.reset_glean()
+GLEAN_APP_ID = "glean-python"
```
Dexterp37: As above, can we have a clear test app id?

PR #474, opened 2019-11-13
Allow to use the `GleanTimerId` from Java

Dexterp37:
This also has the nice side effect of adding type safety to the `GleanTimerId` for Kotlin calls, making
it an opaque type for end users.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -41,7 +41,7 @@ import org.json.JSONObject
  * Public exported type identifying individual timers for
  * [TimingDistributionMetricType][mozilla.telemetry.glean.private.TimingDistributionMetricType].
  */
-typealias GleanTimerId = Long
+data class GleanTimerId internal constructor(internal val id: Long)
 
 /**
  * The main Glean API.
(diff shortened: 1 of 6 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -41,7 +41,7 @@ import org.json.JSONObject
  * Public exported type identifying individual timers for
  * [TimingDistributionMetricType][mozilla.telemetry.glean.private.TimingDistributionMetricType].
  */
-typealias GleanTimerId = Long
+data class GleanTimerId internal constructor(internal val id: Long)
```
travis79: Nice!  This is smart!

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -354,16 +353,16 @@ internal interface LibGleanFFI : Library {
 
     fun glean_destroy_timing_distribution_metric(handle: Long)
 
-    fun glean_timing_distribution_set_start(metric_id: Long, start_time: Long): GleanTimerId
+    fun glean_timing_distribution_set_start(metric_id: Long, start_time: Long): Long
```
travis79: I guess this isn't terrible to just be a `Long` here since it should only be our platform glue code calling it and converting it to the GleanTimerId, although I wonder if there isn't some way to use Kotlin magic to keep the return type?  Not a blocker, just thinking out loud...

Dexterp37: You're right! I tried to think about that, but nothing came to mind :(

PR #481, opened 2019-11-14
WIP: Use DefaultLifecycleObserver interface, rather than lifecycle annotations

mdboom:
The official docs say that using the `androidx.lifecycle` API [requires adding](https://developer.android.com/jetpack/androidx/releases/lifecycle#declaring_dependencies) an `annotationProcessor` (Java) or `kapt` processor (Kotlin) in order to correctly handle the `@OnLifecycleEvent` annotation.

Some random components in a-c that use lifecycles do this, fwiw, but glean, both here and in a-c, does not.

It's not clear from the docs what happens if you don't do that, however [this StackOverflow question](https://stackoverflow.com/questions/49602410/when-should-we-use-android-arch-lifecyclecompiler-or-android-arch-lifecycleco) in the last, non-upvoted answer (I know how that must sound) says:

```
Lifecycle runtime will do one of 2 things:

    - Use reflection to look up the annotated methods,
    - or use generated adapters if you enabled the lifecycle-compiler annotation processor.
```

So, if we don't use the processor, it will use reflection.  We know reflection is flaky and we should avoid it, particularly with proguard.  It seems like if reflection were to fail, it would fail always, but :meh:.

I tried adding the processor as instructed, however, that doesn't work on our code base.  It raises a bunch of build errors of the form:

```
e: /home/mdboom/Work/builds/glean/glean.rs/glean-core/android/build/tmp/kapt3/stubs/debug/mozilla/telemetry/glean/GleanInternalAPI.java:24: error: <identifier> expected                      
    private final java.util.Set<mozilla.telemetry.glean.private.PingType> pingTypeQueue = null;                                                                                              
                                                        ^                                                                                                                                     
e: /home/mdboom/Work/builds/glean/glean.rs/glean-core/android/build/tmp/kapt3/stubs/debug/mozilla/telemetry/glean/GleanInternalAPI.java:24: error: <identifier> expected                      
    private final java.util.Set<mozilla.telemetry.glean.private.PingType> pingTypeQueue = null; 
                                                                        ^                                                                                                                     
e: /home/mdboom/Work/builds/glean/glean.rs/glean-core/android/build/tmp/kapt3/stubs/debug/mozilla/telemetry/glean/GleanInternalAPI.java:24: error: <identifier> expected
    private final java.util.Set<mozilla.telemetry.glean.private.PingType> pingTypeQueue = null;                                                                                               
```

As far as I can tell, this is generating Java code from the Kotlin code as part of this processing step, but since it contains the keyword `private`, it is not valid Java.  Resolving this probably involves refactoring a good chunk of our code base to rename `private` to something else, which would require `glean_parser` changes and the whole lot of it.  (There might be some clever way to "escape" this keyword, but I couldn't find it.)

Therefore, I propose this solution, which is to override the [DefaultLifecycleObserver](https://developer.android.com/reference/androidx/lifecycle/DefaultLifecycleObserver.html) interface.  This feels good because it just uses interfaces that have been in Java forever, and doesn't use any magical features like reflection or annotation processing.

I'm not 100% certain this will solve the baseline issue (since we can't reproduce locally and it only seems to happen "at scale"), but it seems like a worthwhile thing to avoid reflection in any event.

### DIFF
--- a/glean-core/android/build.gradle
+++ b/glean-core/android/build.gradle
@@ -175,6 +175,7 @@ dependencies {
 
     implementation "androidx.annotation:annotation:$versions.androidx_annotation"
     implementation "androidx.lifecycle:lifecycle-extensions:$versions.androidx_lifecycle_extensions"
+    implementation "androidx.lifecycle:lifecycle-common-java8:$versions.androidx_lifecycle_extensions"
     implementation "androidx.work:work-runtime-ktx:$versions.androidx_work"
 
     // We need a compileOnly dependency on the following block of testing
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
@@ -4,9 +4,8 @@
 
 package mozilla.telemetry.glean.scheduler
 
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleObserver
-import androidx.lifecycle.OnLifecycleEvent
+import androidx.lifecycle.DefaultLifecycleObserver
+import androidx.lifecycle.LifecycleOwner
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
 
@@ -14,12 +13,11 @@ import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
  * Connects process lifecycle events from Android to Glean's handleEvent
  * functionality (where the actual work of sending pings is done).
  */
-internal class GleanLifecycleObserver : LifecycleObserver {
+internal class GleanLifecycleObserver : DefaultLifecycleObserver {
     /**
      * Calls the "background" event when entering the background.
      */
-    @OnLifecycleEvent(Lifecycle.Event.ON_STOP)
-    fun onEnterBackground() {
+    override fun onStop(owner: LifecycleOwner) {
         // We're going to background, so store how much time we spent
         // on foreground.
         GleanBaseline.duration.stop()
@@ -34,8 +32,7 @@ internal class GleanLifecycleObserver : LifecycleObserver {
      *
      * https://developer.android.com/reference/android/app/Activity.html#onStart()
      */
-    @OnLifecycleEvent(Lifecycle.Event.ON_START)
-    fun onEnterForeground() {
+    override fun onStart(owner: LifecycleOwner) {
         // Note that this is sending the length of the last foreground session
         // because it belongs to the baseline ping and that ping is sent every
         // time the app goes to background.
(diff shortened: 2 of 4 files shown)


glean-core/android/build.gradle
```diff
@@ -175,6 +175,7 @@ dependencies {
 
     implementation "androidx.annotation:annotation:$versions.androidx_annotation"
     implementation "androidx.lifecycle:lifecycle-extensions:$versions.androidx_lifecycle_extensions"
+    implementation "androidx.lifecycle:lifecycle-common-java8:$versions.androidx_lifecycle_extensions"
```
Dexterp37: Mh, `java8`. Can we use that on older SDK levels? @pocmo, clues?

mdboom: This is the library where the `DefaultLifecycleObserver` class lives.  It's called `java8`, I believe, that class is provided for Java 8 users who can't use the newer annotation approaches.  But it should all be forward compatible with newer Javas and Kotlin.

Dexterp37: But is it backwards compatible with older versions of Java? And do we need to care?

mdboom: Good question.  I don't know.

mdboom:
I've marked this as "WIP", since we don't want to merge it just yet.

1. We definitely want to do this (or something like it that removes reflection from the equation) so that our end users don't have to be as careful about optimizing things away.

2. However, Fenix just merged [these proguard rules](https://github.com/mozilla-mobile/fenix/pulls?q=is%3Apr+is%3Aclosed) as another way to work around the problem, and it would be good to measure those effects in isolation first.

glean-core/android/build.gradle
pocmo: I assume this is fine. We have Java 8 compatibility enabled for all our AC modules:
- https://developer.android.com/studio/write/java8-support
- https://github.com/mozilla-mobile/android-components/blob/master/build.gradle#L126-L129

The Android build toolchain will perform bytecode transformations to make supported Java 8 features work on older versions of Android. However there are *some* Java 8 API features that are supported and cannot be backported and therefore are only available on API 24+.

I assume/hope that this artifact does not use any of those features. Looking at the code and configuration this does not seem to be the case. In fact the empty `DefaultLifecycleObserver` interface is the only class in it. The only thing it relies on are default interface methods and those are backwards compatible.
* https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle/lifecycle-common-java8
* https://android.googlesource.com/platform/frameworks/support/+/androidx-master-dev/lifecycle/lifecycle-common-java8/src/main/java/androidx/lifecycle/DefaultLifecycleObserver.java

Dexterp37: Thank you very much @pocmo! @mdboom , looks like we're good to go!

mdboom: Thanks for the feedback, @pocmo!  Let's still hold off on merging this until we've seen the effects of the proguard rules in isolation...

PR #483, opened 2019-11-15
Bug 1596762 - Add public header to glean-ios

travis79:


mdboom:
This is the same `Glean.h` that contains the FFI interface, or something else?

travis79:
> This is the same `Glean.h` that contains the FFI interface, or something else?

It's really a bridging header that allows the glean-ios build to be used with Swift or Objective-C, the FFI header is another file.  This is more about exposing the public Swift functions to Objective-C.

PR #486, opened 2019-11-18
Split iOS build into normal test and integration test

badboy:
The integration tests take much longer, so run them separate in parallel
to the others.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -234,17 +234,6 @@ jobs:
       - store_artifacts:
           path: raw_xcodetest.log
           destination: log
-      - run:
-          name: Build sample app and run tests
-          command: |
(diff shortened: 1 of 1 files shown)


badboy:
> 👍
> 
> We could also build once and pass the artifacts along to separate jobs for testing. (Though I guess that doesn't gain us much in parallel, it would be less overall work). (And definitely not for this PR...)

We cannot do that easily unfortunately. Thanks to how carthage (the iOS dependency manager) works.

When Glean is build as the dependency of the sample app it is:
1. Pulled in from Glean master
2. Build in release mode

Both things are unfortunately hard to change (I tried depending on a local binary package of Glean, but that was a hack and not even faster because it still needs to build in release mode).

(This btw also means we don't actually run the integration test against changes in the PR :/)

PR #491, opened 2019-11-19
Document all the Python things

mdboom:
@hackebrot: Everything documented here should be working (including the PyPI package).  I'd appreciate your feedback on this PR as a user especially.

### DIFF
--- a/docs/user/adding-glean-to-your-project.md
+++ b/docs/user/adding-glean-to-your-project.md
@@ -111,17 +111,72 @@ github "mozilla/glean" "master"
 
 </div>
 
+<div data-lang="Python" class="tab">
+
+It is recommended that you use a virtual environment for your work to isolate the dependencies for your project. There are many popular abstractions on top of virtual environments in the Python ecosystem which should help manage your project dependencies.
+
+The Python Glean bindings currently have prebuilt wheels on PyPI for x86_64 Linux only.
+
+If you're running that platform and have your virtual environment set up and activated, you can install Glean into it using:
+
+```bash
+$ python -m pip install glean_sdk
+```
+
+If you are not on x86_64 Linux, you will need to build the Glean Python bindings from source using [these instructions](../dev/python/setting-up-python-build-environment.html).
+
+</div>
+
 {{#include ../tab_footer.md}}
 
 ### Adding new metrics
 
 All metrics that your project collects must be defined in a `metrics.yaml` file.
-Add this file to your project and define it as an input file for the `sdk_generator.sh` script in the `Run Script` step defined before.
+
 The format of that file is documented [with `glean_parser`](https://mozilla.github.io/glean_parser/metrics-yaml.html).
 To learn more, see [adding new metrics](adding-new-metrics.md).
 
 > **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
 > This is also required for any new metric automatically collected by the Glean SDK.
+
+{{#include ../tab_header.md}}
+
+<div data-lang="Swift" class="tab">
+
+On iOS, add this file to your project and define it as an input file for the `sdk_generator.sh` script in the `Run Script` step defined before.
+
+</div>
+
+<div data-lang="Python" class="tab">
+
+For Python, the `metrics.yaml` file must be available and loaded at runtime.
+
+If your project is a script (i.e. just Python files in a directory), you can load the `metrics.yaml` using:
+
+```Python
+from glean import load_metrics
+
+metrics = load_metrics("metrics.yaml")
+
+# Use a metric on the returned object
+metrics.your_category.your_metric.set("value")
+```
+
+If your project is an distributable Python package, you need to include the `metrics.yaml` file using [one of the myriad ways to include data in a Python package](https://setuptools.readthedocs.io/en/latest/setuptools.html#including-data-files) and then use [`package_resources.resource_filename()`](https://setuptools.readthedocs.io/en/latest/pkg_resources.html#resource-extraction) to get the filename at runtime.
+
+```Python
+from glean import load_metrics
+from package_resources import resource_filename
+
+metrics = load_metrics(resource_filename(__name__, "metrics.yaml"))
+
+# Use a metric on the returned object
+metrics.your_category.your_metric.set("value")
+```
+
+</div>
+
+{{#include ../tab_footer.md}}
 
 ### Adding custom pings
 
@@ -144,13 +199,15 @@ These specific steps are described in [the `probe_scraper` documentation](https:
 
 The following steps are required for applications using the Glean SDK, but not libraries.
 
+The Glean SDK should only be initialized from the main application, not individual libraries.
+If you are adding Glean support to a library, you can safely skip this section.
+
 {{#include ../tab_header.md}}
 
 <div data-lang="Kotlin" class="tab">
 
 ### Initializing the Glean SDK
 
-The Glean SDK should only be initialized from the main application, not individual libraries.  If you are adding Glean support to a library, you can safely skip this section.
 Please also note that the Glean SDK does not support use across multiple processes, and must only be initialized on the application's main process. Initializing in other processes is a no-op.
 Additionally, Glean must be initialized on the main (UI) thread of the applications main process. Failure to do so will throw an `IllegalThreadStateException`.
 
@@ -207,8 +264,6 @@ When re-enabling, core Glean metrics will be recomputed at that time.
 
 ### Initializing the Glean SDK
 
-The Glean SDK should only be initialized from the main application, not individual libraries.
-If you are adding Glean support to a library, you can safely skip this section.
 Please also note that the Glean SDK does not support use across multiple processes, and must only be initialized on the application's main process.
 
 An excellent place to initialize Glean is within the `application(_:)` method of the class that extends the `UIApplicationDelegate` class.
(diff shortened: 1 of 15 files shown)


docs/user/pings/baseline.md
```diff
@@ -5,6 +5,8 @@
 This ping is intended to provide metrics that are managed by the library itself, and not explicitly set by the application or included in the application's `metrics.yaml` file.
 If the application crashes no `baseline` ping is sent, no additional ping is generated with the data from before the crash.
 
+As the `baseline` ping was specifically designed for mobile operating systems, it is not sent when using the Glean Python bindings.
```
Dexterp37: nit: should we highlight this somehow?

docs/user/pings/events.md
```diff
@@ -4,6 +4,8 @@
 The events ping's purpose is to transport all of the event metric information.
 If the application crashes, an `events` ping is generated next time the application starts with events that were not sent before the crash.
 
+The `events` ping is not sent when using the Glean Python bindings.
```
Dexterp37: Mh, why is that? Is it because it's not implemented?

docs/user/pings/metrics.md
```diff
@@ -7,6 +7,8 @@ Ideally, this window is expected to be about 24 hours, given that the collection
 Data in the [`ping_info`](index.md#the-ping_info-section) section of the ping can be used to infer the length of this window.
 If the application crashes, unsent recorded metrics are sent along with the next `metrics` ping.
 
+As the `metrics` ping was specifically designed for mobile operating systems, it is not sent when using the Glean Python bindings.
```
Dexterp37: This should be highlighted.

docs/user/adding-glean-to-your-project.md
```diff
@@ -111,18 +111,73 @@ github "mozilla/glean" "master"
 
 </div>
 
+<div data-lang="Python" class="tab">
+
+It is recommended that you use a virtual environment for your work to isolate the dependencies for your project. There are many popular abstractions on top of virtual environments in the Python ecosystem which should help manage your project dependencies.
+
+The Python Glean bindings currently have prebuilt wheels on PyPI for x86_64 Linux only.
```
hackebrot: Can you please add a link to the SDK on PyPI?

docs/user/adding-glean-to-your-project.md
```diff
@@ -111,18 +111,73 @@ github "mozilla/glean" "master"
...
+
 The format of that file is documented [with `glean_parser`](https://mozilla.github.io/glean_parser/metrics-yaml.html).
 To learn more, see [adding new metrics](adding-new-metrics.md).
 
 > **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
 > This is also required for any new metric automatically collected by the Glean SDK.
 
+{{#include ../tab_header.md}}
+
+<div data-lang="Swift" class="tab">
```
hackebrot: In this related to the Python bindings?

docs/user/adding-glean-to-your-project.md
```diff
@@ -262,16 +317,42 @@ When re-enabling, core Glean metrics will be recomputed at that time.
...
-import glean
-cfg = glean.Configuration()
-glean.Glean.initialize(cfg, "/path/to/datadir")
+from glean import Glean
+
+# Call Glean.set_upload_enabled first, since Glean.initialize might send pings
+# if there are any metrics queued up from a previous run.
+Glean.set_upload_enabled(True)
+
+Glean.initialize(
```
hackebrot: What do you think of making the argument keyword argument only to be extra explicit?

docs/user/adding-glean-to-your-project.md
```diff
@@ -262,16 +317,42 @@ When re-enabling, core Glean metrics will be recomputed at that time.
...
+See the [custom pings documentation](pings/custom.md) about adding custom pings and sending them.
+
+### Enabling and disabling metrics
+
+`Glean.set_upload_enabled()` should be called in response to the user enabling or disabling telemetry.
+This method should also be called at least once prior to calling `Glean.initialize()`.
+
+The application should provide some form of user interface to call this method.
+
+When going from enabled to disabled, all pending events, metrics and pings are cleared, except for `first_run_date`.
```
hackebrot: Can we link to the `first_run_date` ping in the glean book? 📚

docs/user/experiments-api.md
```diff
@@ -49,6 +47,44 @@ assertEquals(
...
+from glean import Glean
+
+# Annotate Glean pings with experiments data.
+Glean.set_experiment_active(
+  experiment_id = "blue-button-effective",
+  branch = "branch-with-blue-button",
+  extra: {
+    "buttonLabel": "test"
+  }
+)
```
hackebrot: The `extra` kwarg would produce a `SyntaxError` because of the `:` char:

```python
Glean.set_experiment_active(
    experiment_id="blue-button-effective",
    branch="branch-with-blue-button",
    extra={"buttonLabel": "test"},
)
```

docs/user/experiments-api.md
```diff
@@ -49,6 +47,44 @@ assertEquals(
...
+# can be removed.
+Glean.set_experiment_inactive("blue-button-effective")
+```
+
+> **Important**: Experiment IDs and branch don't need to be pre-defined in the Glean SDK registry files.
+Please also note that the `extra` dict is non-nested arbitrary `str` to `str` mapping.
+
+There are test APIs available too:
+
+```Kotlin
```
hackebrot:
```suggestion
```Python
```

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,64 @@ Note that each of these functions is marked as `internal`, you need to import `G
 
 </div>
 
+<div data-lang="Python" class="tab">
+
+It is generally a good practice to "reset" Glean prior to every unit test that uses Glean, to prevent side effects of one unit test impacting others.
+Glean contains a helper function `glean.testing.reset_glean()` for this purpose.
+It has two required arguments: the application id, and the application version.
```
hackebrot:
```suggestion
It has two required arguments: the application ID, and the application version.
```

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,64 @@ Note that each of these functions is marked as `internal`, you need to import `G
...
+Each reset of Glean will create a new temporary directory for Glean to store its data in.
+This temporary directory is automatically cleaned up the next time Glean is reset or when the testing framework finishes.
+
+The instructions below assume using `py.test` as the test runner.
```
hackebrot:
```suggestion
The instructions below assume using [pytest](https://pypi.org/project/pytest/) as the test runner.
```

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,64 @@ Note that each of these functions is marked as `internal`, you need to import `G
...
+Other test-running libraries have similar features, but are different in the details.
+
+To have this function run before every unit test in a `py.test` module:
```
hackebrot:
```suggestion
To have this function run before every unit test in a `pytest` module:
```

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,64 @@ Note that each of these functions is marked as `internal`, you need to import `G
...
+def setup_module():
+    testing.reset_glean("my-app-id", "0.1.0")
+```
+
+To have this function run before every test in an entire test suite, add the following to the `conftest.py` file:
+
+```python
+from glean import testing
+
+def pytest_runtest_module():
```
hackebrot: I'm not familiar with this API and can't find anything in the pytest source or documentation.

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,64 @@ Note that each of these functions is marked as `internal`, you need to import `G
...
+    testing.reset_glean("my-app-id", "0.1.0")
+```
```
hackebrot:
```python
import pytest
import glean


@pytest.fixture(name="reset_glean", auto=True)
def fixture_reset_glean():
    """Reset Glean before every test item."""
    testing.reset_glean("my-app-id", "0.1.0")
```

docs/user/pings/events.md
mdboom: Well, yes, right now we don't have events.  But even once that's done, it's sent when going to background (which doesn't make sense in non-mobile), and there's no way to send built-in pings.

docs/user/adding-glean-to-your-project.md
mdboom: No, I just noticed reading through this that there was a note in the general section that pertained only to iOS, so I moved it to an iOS-specific section (and then added Python-specific content below it).

docs/user/adding-glean-to-your-project.md
mdboom: That's a great idea.  I filed a [follow-on bug for that](https://bugzilla.mozilla.org/show_bug.cgi?id=1597703) -- just to keep this one more-or-less docs only.

docs/user/testing-metrics.md
mdboom: Sorry -- this is a typo on my end.  It's `pytest_runtest_setup`.  And I'll add a link to the pytest docs here.

docs/user/testing-metrics.md
mdboom: Doesn't using a fixture require specifying the fixture in every test that uses Glean?  e.g.:

```
def test_thing(fixture_reset_glean):
    pass
```

Or is that what the `auto` flag does (which I can't find the docs for)?

docs/user/adding-glean-to-your-project.md
```diff
@@ -111,18 +111,73 @@ github "mozilla/glean" "master"
 
 </div>
 
+<div data-lang="Python" class="tab">
+
+It is recommended that you use a virtual environment for your work to isolate the dependencies for your project. There are many popular abstractions on top of virtual environments in the Python ecosystem which should help manage your project dependencies.
```
mdboom: @hackebrot : Curious what you think of this.  I don't want to write a tutorial on virtual environments (and all the related tools) here -- but should we recommend anything in particular?

docs/user/pings/events.md
Dexterp37: They're also sent at startup and whenever we reach the cap (500 events), but I agree the 'background' part doesn't make sense. Since this is more about defining a new behaviour for this ping, I'd recommend filing a bug and mentioning here that "send on background" is not supported when using Python bindings.

docs/user/testing-metrics.md
hackebrot: Sorry, my bad. It's `autouse`.

See https://docs.pytest.org/en/latest/reference.html#pytest-fixture 📝 

I would say the idiomatic way to implemented the described tasks with pytest would be:

```python
import pytest
import glean


@pytest.fixture(name="run_for_every_test", scope="function", autouse=True)
def fixture_run_for_every_test():
    pass

@pytest.fixture(name="run_once_for_session", scope="session", autouse=True)
def fixture_once_for_session():
    pass
```

docs/user/pings/events.md
mdboom: That makes sense.

docs/user/testing-metrics.md
mdboom: Great.  I'll update the docs here.

Dexterp37:
Android tests failures are unrelated: I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1597917).

docs/user/testing-metrics.md
```diff
@@ -98,11 +98,57 @@ Note that each of these functions is marked as `internal`, you need to import `G
...
+Other test-running libraries have similar features, but are different in the details.
+
+Create a file `conftest.py` at the root of your test directory, and add the following to reset Glean at the start of every test in your suite:
+
+```python
+import pytest
+from glean import testing
+
+@pytest.fixture(name="reset_glean", scope="function", autouse=True)
+def reset_glean():
```
hackebrot:
```suggestion
def fixture_reset_glean():
```

PR #495, opened 2019-11-20
Reset timer when disabled, so that we don't record timespans acr…

badboy:


badboy:
absolutely

mdboom:
So, if you want to permute start and stop a couple of times with set_upload_enabled inserted, you get 143 permutations, which becomes a little hard to define the expected behavior.  So ignore my idea to try to do this methodically.

Let's add a test before merging this, though.

glean-core/tests/timespan.rs
```diff
@@ -287,3 +287,40 @@ fn set_raw_time_does_nothing_when_timer_running() {
...
+    metric.set_stop(&glean, 40);
+
+    // App code eventually starts the timer again.
+    // Upload is disabled, so this should not have any effect.
+    metric.set_start(&glean, 100);
+    // User enables telemetry upload again.
+    glean.set_upload_enabled(true);
+    // App code eventually stops the timer.
+    // None should be running.
+    metric.set_stop(&glean, 200);
```
Dexterp37: Should this also record an error? Let's check for that too

PR #496, opened 2019-11-20
Fix broken tests

Dexterp37:


mdboom:
Thanks, @Dexterp37.  This probably wouldn't have happened if we had the requirement to "test on current upstream master" like a-c has.

Dexterp37:
> 
> 
> Thanks, @Dexterp37. This probably wouldn't have happened if we had the requirement to "test on current upstream master" like a-c has.

No worries! Let's see, if this class of problems appear again, it might be a good opportunity to have such a requirement ourselves as well

PR #498, opened 2019-11-20
1597980: Add metrics to record counts of sending other pings

mdboom:


glean-core/android/metrics.yaml
```diff
@@ -55,3 +55,34 @@ glean.core.migration:
       - glean-team@mozilla.com
       - aplacitelli@mozilla.com
     expires: 2020-03-31
+
+glean.validation:
+  baseline_ping_count:
+    type: counter
+    description:
+      The number of baseline pings sent during the lifetime of this metrics
```
Dexterp37: nit: this is only for Android. Let's call it out here and below for posterity.

glean-core/android/metrics.yaml
```diff
@@ -55,3 +55,34 @@ glean.core.migration:
...
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597980#c3
+    notification_emails:
+      - glean-team@mozilla.com
+      - mdroettboom@mozilla.com
+    expires: 2019-12-20
+
+  metrics_ping_count:
+    type: counter
+    description:
+      The number of metrics pings sent during the lifetime of this baseline
```
Dexterp37: nit: this is only for Android. Let's call it out here and below for posterity.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -519,6 +520,14 @@ open class GleanInternalAPI internal constructor () {
             return@launch
         }
 
+        if (pingNames.contains("metrics")) {
```
Dexterp37: Since we also support custom pings that might contain the `metrics` word (or baseline), can we make this an equality?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -519,6 +520,14 @@ open class GleanInternalAPI internal constructor () {
...
+            GleanValidation.metricsPingCount.add()
+        }
+
+        if (pingNames.contains("baseline")) {
```
Dexterp37: Since we also support custom pings that might contain the `metrics` word (or baseline), can we make this an equality?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: `pingNames` is an array of strings, so checking if it contains `"metrics"` is already an equality check (that is: it would fail for `pingNames = ["my-metrics-things"]`)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: pingNames is a collection, so this is checking as an equality.  It would only match if there was an element "baseline" in the collection.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: `pingNames` is a collection so using `.contains()` is checking for an element that exactly matches "metrics"

Dexterp37: My bad!

PR #502, opened 2019-11-20
Properly handle errors in the database module

badboy:


badboy:
> Just curious -- is there a tool used to find possible sources of panics, or do you just grep for "unwrap"?

I had a test failure due to one of the `unwraps` panicking.
I then grepped for `unwrap\(` and `expect`, seeing >30 of these in `database/mod.rs` and considering those the most criticial.

I then went through all the code (and its callers) and thought about the implications (especially given that I knew how the mentioned test failure caused it).
Nothing automated about that other than experience.

Unfortunately it's kinda hard to write test cases for any of these, as I would need to make the database randomly fail.

PR #512, opened 2019-11-21
Build dependencies in Debug mode for test

badboy:
Looks like this might speed it up from previously 12 minutes total time to now 8 minutes.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -261,6 +261,10 @@ jobs:
       - run:
           name: Build sample app
           command: |
+            # Build in Debug mode to speed it all up
+            pushd samples/ios/app
+            carthage bootstrap --platform iOS --no-use-binaries --cache-builds --configuration Debug --verbose
+            popd
             bash bin/run-ios-sample-app-build.sh
       - store_artifacts:
           path: raw_sample_xcodebuild.log


.circleci/config.yml
```diff
@@ -261,6 +261,10 @@ jobs:
       - run:
           name: Build sample app
           command: |
+            # Build in Debug mode to speed it all up
```
Dexterp37: Why is debug mode faster?

badboy: Because it builds the Rust code in debug mode, which means no heavy optimizations that run.
Results in a bigger and less optimized binary, but we don't care for the integration tests.

Dexterp37: Wouldn't this hide a class of errors that only appears in Release? (I don't know how common that is in Rust)

badboy: Other than timing errors (which are unpredictable and could happen in whatever configuration we choose) I don't see what could be different.

If at all, we might catch _more_ errors because for example integer overflows will panic in debug mode, but not in release.

(We could re-enable to use a release build once we run the integration tests only on merge with bors, but for now I consider having faster feedback here essential)

Dexterp37: > (We could re-enable to use a release build once we run the integration tests only on merge with bors, but for now I consider having faster feedback here essential)

Sounds like a good plan! Would you consider filing a bug for this?

badboy: Done!

PR #513, opened 2019-11-21
Enable bors on the repository

badboy:
This is the configuration enabling bors (https://bors.tech) to work on
this repository.

It will require enabling GitHub checks and removing the
review-required-on-PRs protection (which will now be enforced by bors
instead).

Currently known downsides:

* "Build succeeded/failed" messages will not contain a link to CI anymore.
It's a known issue and might get fixed: https://github.com/bors-ng/bors-ng/issues/569
* `bors try` works and runs, but for some reason bors does not get the notification that something is done.

---

Here's where I played around with it: https://github.com/badboy/glean-test-tc/pull/2

Steps to land this:

* [ ] Set up GitHub checks for this repository
* [ ] Enable this repository on https://bors.tech

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -603,7 +603,7 @@ workflows:
       - Lint Python
       - Rust FFI header check
       - Lint YAML with yamllint
-  ci:
+  test:
     jobs:
       - Rust tests - stable
       - Rust code coverage
@@ -631,7 +631,7 @@ workflows:
             - docs-linkcheck
           filters:
             branches:
-              only: master
+              only: staging
 
   release:
     jobs:
--- a/bors.toml
+++ b/bors.toml
@@ -1,0 +1,18 @@
+status = [
+  "check-formatting",
+  "lint",
+  "test"
+]
+block_labels = [
+  "needs:data-review",
+  "blocked",
+  "work in progress"
+]
+required_approvals = 1
+use_codeowners = true
+delete_merged_branches = true
+cut_body_after = "---"
+
+[committer]
+name = "glean-bors"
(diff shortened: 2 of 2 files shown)


.circleci/config.yml
```diff
@@ -631,7 +631,7 @@ workflows:
             - docs-linkcheck
           filters:
             branches:
-              only: master
+              only: staging
```
mdboom: Can you elaborate on this?  I don't even see that we have a `staging` branch...

badboy: That's the branch bors uses for merging
(It takes "current master" and a PR, merges that together in "staging" and lets CI run, if that is green it gets merged into master).

In theory we don't need to let tests run on master at all anymore, though disabling that makes the config really bloated (I need to add a filter to _every single job_).
Running the deploy on staging has the advantage that if it fails, bors will complain and not merge, thus allowing a user to catch it and retry (failures on master have the tendency to get lost, because the only way to see it is by the small checkmark on the top of the page or the badge in the readme)

mdboom: Thanks for explaining.  That's cool.

(Someday we might only want to deploy docs from tagged commits, but that's an entirely separate discussion).

badboy: That one can be done.
Maybe even have docs from both master and tagged commits, e.g. https://mozilla.github.io/glean/master/book/index.html and https://mozilla.github.io/glean/v21.1.1/book/index.html

and some redirect stuff (though that gets a bit hard as we deploy on github pages)

badboy:
Closing, will need a proper rebase if we ever get to it anyway.

PR #517, opened 2019-11-21
1595548: Fix truncation of string list items

mdboom:
Oops.  ¯\_(ツ)_/¯

This will affect analysis against previously collected data, since the old values will be truncated, but new values will not.  Thoughts about that?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
   * The experiments API is no longer ignored before the Glean SDK initialized. Calls are
     recorded and played back once the Glean SDK is initialized.
 
+  * String list items were being truncated to 20, rather than 50, bytes when using
+    `.set()` (rather than `.add()`). This has been corrected, but it may result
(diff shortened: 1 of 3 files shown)


travis79:
> Oops. ¯_(ツ)_/¯
> 
> This will affect analysis against previously collected data, since the old values will be truncated, but new values will not. Thoughts about that?

This could theoretically be corrected with a script.

mdboom:
> > Oops. ¯_(ツ)_/¯
> > This will affect analysis against previously collected data, since the old values will be truncated, but new values will not. Thoughts about that?
> 
> This could theoretically be corrected with a script.

Unfortunately, not 100%.  There are product ids that are only different after the first 20 chars:

```
        REFERENCE_BROWSER("org.mozilla.reference.browser"),
        REFERENCE_BROWSER_DEBUG("org.mozilla.reference.browser.debug"),
```

PR #524, opened 2019-11-22
On tagged released build and upload an iOS binary artifact

badboy:


badboy:
Note: It should work, but I can't really test it right now as my personal CircleCI doesn't have mac builds.
I will run a test version later

PR #526, opened 2019-11-25
Implement the deletion request ping in Glean

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1587095


samples/android/app/src/main/AndroidManifest.xml
```diff
@@ -15,6 +15,7 @@
         android:icon="@mipmap/ic_launcher"
         android:label="@string/app_name"
         android:supportsRtl="true"
+        android:usesCleartextTraffic="true"
```
travis79: Is this just for tests or do we need to support cleartext traffic in normal operation?  If it's just for tests can we put this in a separate `AndroidManifest.xml` in the `androidTest` directory so as to not relax security on normal operation?

samples/ios/app/glean-sample-app/ViewController.swift
```diff
@@ -53,4 +55,14 @@ class ViewController: UIViewController {
         Custom.counter.add()
         Pings.shared.sample.send()
     }
+
+    @IBAction func enableToggled(_: Any) {
```
travis79: Even though this is an example, we probably should still set the initial value of the switch to the current `getUploadEnabled` in the `viewDidLoad()` function.

samples/android/app/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -45,6 +44,16 @@ open class MainActivity : AppCompatActivity() {
             )
         }
 
+        uploadSwitch.setOnCheckedChangeListener { _, isChecked ->
```
travis79: We should also set the initial state of the toggle to match `getUploadEnabled()`

docs/dev/core/internal/directory-structure.md
```diff
@@ -18,5 +18,7 @@ Within the `glean_data` directory are the following contents:
 
 - `pending_pings`: Pings are written here before they are picked up by the ping uploader to send to the submission endpoint.
 
+- `deletion_request`: The `deletion_request` ping is written here before it is picked up by the ping uploader. It is special in that it needs to get send once when the upload is disabled by the user.
```
Dexterp37: nit: can you expand a bit more on why this cannot be in the other directory?

glean-core/pings.yaml
```diff
@@ -54,3 +54,15 @@ events:
     - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
   notification_emails:
     - telemetry-client-dev@mozilla.com
+
+deletion_request:
+  description: >
+    This ping is submitted when a user opts out of
```
Dexterp37: Can you document on why this is needed? I think we need to clearly express our intent here, since this is a ping that will get sent when upload is disabled.

glean-core/pings.yaml
```diff
@@ -54,3 +54,15 @@ events:
...
+    sending technical and interaction data to Mozilla.
+  include_client_id: true
+  bugs:
+    - https://bugzilla.mozilla.org/1587095
+  data_reviews:
+    - https://bugzilla.mozilla.org/1587095#c3
+  notification_emails:
+    - telemetry-client-dev@mozilla.com
```
Dexterp37: nit: can you also add the glean team ML here and to the pings above?

glean-core/metrics.yaml
```diff
@@ -244,3 +244,18 @@ glean.error:
       - all_pings
     no_lint:
       - COMMON_PREFIX
+
+glean.deletion_request:
+  optout:
+    type: boolean
+    description:
+      Signals that the user opted out of sending technical and interaction data.
```
Dexterp37: Ah, we're not going for the "send if empty flag" for this. That's fine in the short term, provided that: (a) we file bout about adding that to `pings.yaml` to support this and other cases (b) document in this metric that this is really useless and it's only really needed to make glean send the ping out. Let's also reference the bug number for adding that feature, please.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
```diff
@@ -0,0 +1,147 @@
...
+    }
+
+    private fun waitForPingContent(
+        pingName: String,
+        maxAttempts: Int = 3
+    ): JSONObject?
+    {
+        val server = getPingServer()
+
+        var attempts = 0
```
Dexterp37: Wow, why do we need this? Are we failing because we are not receiving the pings?

glean-core/src/ping/mod.rs
```diff
@@ -240,8 +240,15 @@ impl PingMaker {
     ///
     /// The directory will be created inside the `data_path`.
     /// The `pings` directory (and its parents) is created if it does not exist.
-    fn get_pings_dir(&self, data_path: &Path) -> std::io::Result<PathBuf> {
-        let pings_dir = data_path.join("pending_pings");
+    fn get_pings_dir(&self, data_path: &Path, ping_type: Option<&str>) -> std::io::Result<PathBuf> {
+        // Use a special directory for deletion_request pings
+        let pings_dir = match ping_type {
+            Some(ping_type) if ping_type == "deletion_request" => {
```
Dexterp37: I just realized one thing: what if some library or application attempts to register a `deletion_request` ping themselves? Would Glean make it fail at build time? Would it race here? Not for fixing in this bug, but it might be worth filing a bug about this

glean-core/src/internal_pings.rs
```diff
@@ -0,0 +1,31 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use super::metrics::PingType;
+
+/// Glean-provided pings, all enabled by default.
+///
+/// These pings are defined in `glean-core/pings.yaml` and for now manually translated into Rust
```
Dexterp37: This will go away once we have the nice Rust API, right? If so, any chance you could add that as a comment here, mentioning the nice rust API bug number?

glean-core/metrics.yaml
Dexterp37: That would also be blocked on the nice Rust API, right? If so, any chance you could mention this as well?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt
```diff
@@ -0,0 +1,192 @@
...
+        internal val pingQueueLock = Any()
+
+        /**
+         * Build the constraints around which the worker can be run, such as whether network
+         * connectivity is required.
+         *
+         * @return [Constraints] object containing the required work constraints
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        fun buildConstraints(): Constraints = Constraints.Builder()
```
Dexterp37: Since this is likely to be shared across the two workers, what if we just used the function from the main upload worker class?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt
```diff
@@ -0,0 +1,192 @@
...
+ */
+class DeletionPingUploadWorker(context: Context, params: WorkerParameters) : Worker(context, params) {
+    companion object {
+        internal const val PING_WORKER_TAG = "mozac_service_glean_deletion_ping_upload_worker"
+
+        // Since ping file names are UUIDs, this matches UUIDs for filtering purposes
+        private const val FILE_PATTERN = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
+        private const val LOG_TAG = "glean/DeletionPingUploadWorker"
+        // NOTE: The `PINGS_DIR` must be kept in sync with the one in the Rust implementation.
+        internal const val PINGS_DIR = "deletion_request"
```
Dexterp37:
```suggestion
        internal const val DELETION_PING_DIR = "deletion_request"
```

Dexterp37: nit: to make sure we don't forget what this is for :) Not mandatory, really

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt
```diff
@@ -0,0 +1,192 @@
...
+
+        /**
+         * Function to deserialize and process all serialized ping files.  This function will ignore
+         * files that don't match the UUID regex and just delete them to prevent files from polluting
+         * the ping storage directory.
+         *
+         * @return Boolean representing the success of the upload task. This may be the value bubbled up
+         *         from the callback, or if there was an error reading the files.
+         */
+        fun process(): Boolean {
```
Dexterp37: Any chance we could just refactor out all this code or use the one from the main uploader?

samples/android/app/src/main/AndroidManifest.xml
badboy: I did not know that we can separate this out.
I'm not even sure why we need it, it worked before?
But maybe I updated the SDK used in my emulator and thus non-cleartext is now enforced.
Would be good if someone can run that test on their machine to see its effect.

Dexterp37: Yes, this is only required for `androidTest`, [see my comment here](https://github.com/mozilla-mobile/android-components/blob/5ec752f2587432e369b905f6de68cebabfd63ada/samples/glean/src/main/AndroidManifest.xml#L13-L15).

glean-core/metrics.yaml
badboy: to answer the first: yes, I think going for the flag is the better option. What do you think about me implementing that first and landing it, before we land the deletion ping here?

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
badboy: We should not, we're already waiting 20secs to receive something, let me simplify that

glean-core/src/ping/mod.rs
badboy: Same as with baseline and metrics: We don't allow it (that is: glean_parser will fail such a definition)

glean-core/src/internal_pings.rs
badboy: Depends. It could definitely be auto-generated, though it's core internal so having that generated on top of the nice API could be tricky

glean-core/metrics.yaml
Dexterp37: I have no strong opinion on this, I'd leave it to your judgment :-) If you want to play safe, given that nobody is going to rely on that data point, just land the current version without blocking on that flag, so that we have enough time to validate this once it hits Fenix.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt
badboy: Will refactor this to share more of the code

glean-core/src/ping/mod.rs
Dexterp37: Oh, right. Sorry for the noise :)

glean-core/src/internal_pings.rs
Dexterp37: Right! It would be so cool , though :-P

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt
Dexterp37: This can also happen as a follow-up, depending on your will to iterate on this PR :)

badboy: Follow-up it is! filing bug now

glean-core/pings.yaml
badboy: I'd like to wait on the phrasing on desktop, then just use that.

Dexterp37: Ok, but let's file a bug about this if it happens after merging this PR.

badboy:
Blocked by #528

glean-core/ios/Glean/Glean.swift
```diff
@@ -162,6 +162,13 @@ public class Glean {
 
                 if !enabled {
                     Dispatchers.shared.cancelBackgroundTasks()
+                    // If uploading is disabled, we need to send the deletion_request ping
+                    Dispatchers.shared.launchConcurrent {
```
Dexterp37: Can you file a bug for adding the 'deletion_request' to Python?

Dexterp37: Well, for the 'upload' part, since the creation itself happens in Rust.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1600259

PR #528, opened 2019-11-26
Add a new flag to pings: send_if_empty

badboy:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
```diff
@@ -17,14 +17,16 @@ import mozilla.telemetry.glean.rust.toByte
  */
 class PingType(
     internal val name: String,
-    includeClientId: Boolean
+    includeClientId: Boolean,
+    sendIfEmpty: Boolean = false
```
Dexterp37: Why are we providing a default value here? Isn't `glean_parser` taking care of adding it?

glean-core/ios/Glean/Metrics/Ping.swift
```diff
@@ -15,10 +15,10 @@ public class Ping {
     let includeClientId: Bool
 
     /// The public constructor used by automatically generated metrics.
-    public init(name: String, includeClientId: Bool) {
+    public init(name: String, includeClientId: Bool, sendIfEmpty: Bool = false) {
```
Dexterp37: As for the Kotlin code, shouldn't this be provided by the parsers? Why should we provide a default for this and not for the client id one?

glean-core/src/ping/mod.rs
```diff
@@ -196,7 +196,7 @@ impl PingMaker {
         let metrics_data = StorageManager.snapshot_as_json(glean.storage(), &ping.name, true);
         let events_data = glean.event_storage().snapshot_as_json(&ping.name, true);
 
-        if metrics_data.is_none() && events_data.is_none() {
+        if !ping.send_if_empty && metrics_data.is_none() && events_data.is_none() {
```
Dexterp37: Can we add a log line below stating that the ping is empty but we're sending it anyway due to the `send_if_empty` override? This would simplify debugging weird issues in the future.

glean-core/python/glean/metrics/ping.py
```diff
@@ -17,7 +17,7 @@ def __init__(self, name: str, include_client_id: bool):
         """
         self._name = name
         self._handle = _ffi.lib.glean_new_ping_type(
-            _ffi.ffi_encode_string(name), include_client_id
+            _ffi.ffi_encode_string(name), include_client_id, False
```
mdboom: Also should add this as a parameter and pass along to the FFI (as for the other language bindings).

glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
badboy: Considered it mostly for backwards-compatibility.
Even for files that are not regenerated by the parser they keep working.

glean-core/ios/Glean/Metrics/Ping.swift
badboy: As above

glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
Dexterp37: Thanks for explaining! Would you consider changing this so that any problem will surface? We don't expect files to be manually crafted except inside our core.

badboy: Sure, I can. guess I might need to land the parser changes and update it here first to make it all work though

Dexterp37: Sounds good to me!

PR #529, opened 2019-11-27
Use the zip release of the GitHub release tool

badboy:


travis79:
> Ok but.... ehm why? :)

Because, CI like Lockwise-iOS buddy build can't run `cargo` commands, so because Carthage can either checkout a project from github (or local) and build it from source, or use a pre-built binary framework, we needed to provide the binary in order to ease integration on iOS where build systems may not have all the tools to build Glean from source.

A-S does this for iOS binaries also.

badboy:
Should have expanded on this: The tool we are using to release the iOS binary framework comes in a zip file (the Linux build of it is tar.gz), so the job was failing. 
With this patch it shouldn't fail anymore

travis79:
> Should have expanded on this: The tool we are using to release the iOS binary framework comes in a zip file (the Linux build of it is tar.gz), so the job was failing.
> With this patch it shouldn't fail anymore

I was just about to ask about the `unzip` step and why it was needed.  Thanks for expanding on this.

Issue #10, opened 2019-05-08
Re-think current makefile


badboy:
From #8:
> On my machine this is `Sdk`. Will we run into issues on case-sensitive filesystems here?

Maybe we:

* Don't set the variable in the makefile (but fail if it is unset?)
* or Don't use the makefile at all anymore

PR #530, opened 2019-11-27
Add option to defer ping lifetime metric persistence

brizental:
Fixes [Bug 1596440](https://bugzilla.mozilla.org/show_bug.cgi?id=1596440)

Opening this as WIP so I can get validation that I am on the right way and also to discuss the name of the `defer_collection` configuration option -- keeping in mind that we might change this config option to a interval right after this gets PR accepted (see [comment #9](https://bugzilla.mozilla.org/show_bug.cgi?id=1596440#c9) and [comment #10](https://bugzilla.mozilla.org/show_bug.cgi?id=1596440#c10)).

I think the biggest part of this is done, that was adding the option and changing the way metrics are stored when the config option is set. I thought I had to change a little bit the `PingMaker`, but after looking at the code seems like my changes to `Database` were all that was necessary.

In my test I wasn't sure how to check that things were actually saved to the map and not rkv. I am not sure `assert!(db.ping_lifetime_data.is_some());` is enough... 

Finally, I hard-coded the value of `defer_collection` on the ffi, so I still need to change that.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
@@ -17,12 +17,13 @@ import mozilla.telemetry.glean.net.PingUploader
  * **CAUTION**: This must match _exactly_ the definition on the Rust side.
  *  If this side is changed, the Rust side need to be changed, too.
  */
-@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents", "delayPingLifetimeIO")
 internal class FfiConfiguration(
     dataDir: String,
     packageName: String,
     uploadEnabled: Boolean,
-    maxEvents: Int? = null
+    maxEvents: Int? = null,
+    delayPingLifetimeIO: Boolean
 ) : Structure() {
     /**
      * Expose all structure fields as actual fields,
@@ -37,6 +38,8 @@ internal class FfiConfiguration(
     public var uploadEnabled: Byte = uploadEnabled.toByte()
     @JvmField
     public var maxEvents: IntByReference = if (maxEvents == null) IntByReference() else IntByReference(maxEvents)
+    @JvmField
+    public var delayPingLifetimeIO: Byte = delayPingLifetimeIO.toByte()
 
     init {
         // Force UTF-8 string encoding when passing strings over the FFI
@@ -64,6 +67,7 @@ data class Configuration internal constructor(
     val channel: String? = null,
     val userAgent: String = DEFAULT_USER_AGENT,
     val maxEvents: Int? = null,
+    val delayPingLifetimeIO: Boolean = false,
     val logPings: Boolean = DEFAULT_LOG_PINGS,
     // NOTE: since only simple object or strings can be made `const val`s, if the
     // default values for the lines below are ever changed, they are required
@@ -96,6 +100,7 @@ data class Configuration internal constructor(
         serverEndpoint = serverEndpoint,
         userAgent = DEFAULT_USER_AGENT,
         maxEvents = maxEvents,
+        delayPingLifetimeIO = false,
         logPings = DEFAULT_LOG_PINGS,
         httpClient = httpClient,
         pingTag = null,
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -101,6 +101,7 @@ pub struct FfiConfiguration<'a> {
     package_name: FfiStr<'a>,
     upload_enabled: u8,
     max_events: Option<&'a i32>,
+    delay_ping_lifetime_io: u8,
 }
 
 /// Convert the FFI-compatible configuration object into the proper Rust configuration object.
@@ -112,12 +113,14 @@ impl TryFrom<&FfiConfiguration<'_>> for glean_core::Configuration {
         let application_id = cfg.package_name.to_string_fallible()?;
         let upload_enabled = cfg.upload_enabled != 0;
         let max_events = cfg.max_events.filter(|&&i| i >= 0).map(|m| *m as usize);
+        let delay_ping_lifetime_io = cfg.delay_ping_lifetime_io != 0;
 
         Ok(Self {
             upload_enabled,
             data_path,
             application_id,
             max_events,
+            delay_ping_lifetime_io,
         })
     }
 }
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -22,6 +22,12 @@ pub struct Database {
     // as the application lives: they don't need to be persisted
     // to disk using rkv. Store them in a map.
     app_lifetime_data: RwLock<BTreeMap<String, Metric>>,
+    // If the `delay_ping_lifetime_io` Glean config option is `true`,
+    // we will save metrics with 'ping' lifetime data in a map temporarily
+    // so as to persist them to disk using rkv in bulk on shutdown,
+    // or after a given interval, instead of everytime a new metric
+    // is created / updated.
+    ping_lifetime_data: Option<RwLock<BTreeMap<String, Metric>>>,
 }
 
 impl Database {
@@ -29,10 +35,15 @@ impl Database {
     ///
     /// This opens the underlying rkv store and creates
     /// the underlying directory structure.
-    pub fn new(data_path: &str) -> Result<Self> {
+    pub fn new(data_path: &str, delay_ping_lifetime_io: bool) -> Result<Self> {
         Ok(Self {
             rkv: Self::open_rkv(data_path)?,
             app_lifetime_data: RwLock::new(BTreeMap::new()),
+            ping_lifetime_data: if delay_ping_lifetime_io {
+                Some(RwLock::new(BTreeMap::new()))
+            } else {
+                None
+            },
         })
     }
 
@@ -117,6 +128,23 @@ impl Database {
             return;
         }
 
+        // Lifetime::Ping data is not persisted to disk if
+        // Glean has `delay_ping_lifetime_io` set to true
+        if lifetime == Lifetime::Ping {
+            if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+                let data = ping_lifetime_data
+                    .read()
+                    .expect("Can't read ping lifetime data");
+                for (key, value) in data.iter() {
+                    if key.starts_with(&iter_start) {
+                        let key = &key[len..];
+                        transaction_fn(key.as_bytes(), value);
+                    }
+                }
+                return;
+            }
+        }
+
         let store: SingleStore = unwrap_or!(
             self.rkv
                 .open_single(lifetime.as_str(), StoreOptions::create()),
@@ -168,6 +196,17 @@ impl Database {
                 .read()
                 .map(|data| data.contains_key(&key))
                 .unwrap_or(false);
+        }
+
+        // Lifetime::Ping data is not persisted to disk if
+        // Glean has `delay_ping_lifetime_io` set to true
+        if lifetime == Lifetime::Ping {
+            if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+                return ping_lifetime_data
+                    .read()
+                    .map(|data| data.contains_key(&key))
+                    .unwrap_or(false);
+            }
         }
 
         let store: SingleStore = unwrap_or!(
@@ -243,6 +282,18 @@ impl Database {
             return Ok(());
         }
 
+        // Lifetime::Ping data is not persisted to disk if
+        // Glean has `delay_ping_lifetime_io` set to true
+        if lifetime == Lifetime::Ping {
+            if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+                let mut data = ping_lifetime_data
+                    .write()
+                    .expect("Can't read ping lifetime data");
+                data.insert(final_key, metric.clone());
+                return Ok(());
+            }
+        }
+
         let encoded = bincode::serialize(&metric).expect("IMPOSSIBLE: Serializing metric failed");
         let value = rkv::Value::Blob(&encoded);
 
@@ -313,6 +364,27 @@ impl Database {
             return Ok(());
         }
 
+        // Lifetime::Ping data is not persisted to disk if
+        // Glean has `delay_ping_lifetime_io` set to true
+        if lifetime == Lifetime::Ping {
+            if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+                let mut data = ping_lifetime_data
+                    .write()
+                    .expect("Can't access ping lifetime data as writable");
+                let entry = data.entry(final_key);
+                match entry {
+                    Entry::Vacant(entry) => {
+                        entry.insert(transform(None));
+                    }
+                    Entry::Occupied(mut entry) => {
+                        let old_value = entry.get().clone();
+                        entry.insert(transform(Some(old_value)));
+                    }
+                }
+                return Ok(());
+            }
+        }
+
         let store_name = lifetime.as_str();
         let store = self.rkv.open_single(store_name, StoreOptions::create())?;
 
@@ -351,6 +423,16 @@ impl Database {
     ///
     /// * This function will **not** panic on database errors.
     pub fn clear_ping_lifetime_storage(&self, storage_name: &str) -> Result<()> {
+        // Lifetime::Ping might have data saved to `ping_lifetime_data`
+        // in case `delay_ping_lifetime_io` is set to true
+        if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+            ping_lifetime_data
+                .write()
+                .expect("Can't access ping lifetime data as writable")
+                .clear();
+            return Ok(());
+        }
+
         self.write_with_store(Lifetime::Ping, |mut writer, store| {
             let mut metrics = Vec::new();
             {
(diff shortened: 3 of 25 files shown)


glean-core/src/database/mod.rs
```diff
@@ -351,6 +423,14 @@ impl Database {
     ///
     /// * This function will **not** panic on database errors.
     pub fn clear_ping_lifetime_storage(&self, storage_name: &str) -> Result<()> {
+        let ping_lifetime_data = &self.ping_lifetime_data;
+        if let Some(ping_lifetime_data) = ping_lifetime_data {
+            ping_lifetime_data
+                .write()
+                .expect("Can't access ping lifetime data as writable")
+                .clear();
+        }
```
Dexterp37: Shouldn't we be returning here?

glean-core/src/database/mod.rs
```diff
@@ -443,6 +534,14 @@ impl Database {
             .write()
             .expect("Can't access app lifetime data as writable")
             .clear();
+
+        let ping_lifetime_data = &self.ping_lifetime_data;
```
Dexterp37: When we do `for lifetime in [Lifetime::User, Lifetime::Ping].iter()` on line 522, if the deferred write is on, we don't need to also clear the RKV store. Maybe not a big deal, but could save us some I/O.

glean-core/src/database/mod.rs
```diff
@@ -243,6 +284,17 @@ impl Database {
             return Ok(());
         }
 
+        if lifetime == Lifetime::Ping {
+            let ping_lifetime_data = &self.ping_lifetime_data;
```
Dexterp37: Both the reading and the writing parts for `Lifetime::Ping` and `Lifetime::Application` are kind of identical. Did you consider sharing the code in some utility functioN?

glean-core/src/database/mod.rs
```diff
@@ -707,4 +808,41 @@ mod test {
...
+            found_metrics += 1;
+            let metric_id = String::from_utf8_lossy(metric_name).into_owned();
+            assert_eq!(test_metric_id, metric_id);
+            match metric {
+                Metric::String(s) => assert_eq!(test_value, s),
+                _ => panic!("Unexpected data found"),
+            }
+        };
+
+        db.iter_store_from(Lifetime::Ping, test_storage, None, &mut snapshotter);
```
Dexterp37: One thing you should probably do is manually check if data was written to rkv. You might want to do something [like this](https://github.com/mozilla/glean/blob/15bd43b316e3829bd4d747923945f209ad9bb72c/glean-core/src/database/mod.rs#L214-L220).

glean-core/src/database/mod.rs
brizental: Yes, I agree. I am looking into doing something like that before marking this as "done" :)

glean-core/src/database/mod.rs
brizental: My thinking was that we might have something to clear on the map **and** in rkv. I don't implement this in this PR yet, but will do it right after... What do you think?

Dexterp37: If the implementation that makes use of this code will come in a separate PR, it will make more sense to have this code as part of that future PR. Otherwise it's hard to test and to make sense of it and, more importantly, if we need to backout the PR/features it gets a bit complicated as the PR itself is not "self-contained" :)

brizental: Makes sense, I will return for now then.

glean-core/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub struct Configuration {
     pub application_id: String,
     /// The maximum number of events to store before sending a ping containing events.
     pub max_events: Option<usize>,
+    /// Whether Glean should defer ping collection.
+    pub defer_collection: Option<bool>,
```
badboy: Why is this an option? What do the three states it could be in mean (`Some(true), Some(false), None`)?

badboy: Actually, do we need to store that value at all in the Glean struct?
The database already knows if it has the ping metrics store, no need to keep this flag around at all

glean-core/src/database/mod.rs
```diff
@@ -22,17 +22,28 @@ pub struct Database {
...
     ///
     /// This opens the underlying rkv store and creates
     /// the underlying directory structure.
-    pub fn new(data_path: &str) -> Result<Self> {
+    pub fn new(data_path: &str, defer_collection: bool) -> Result<Self> {
         Ok(Self {
             rkv: Self::open_rkv(data_path)?,
             app_lifetime_data: RwLock::new(BTreeMap::new()),
+            ping_lifetime_data: if defer_collection {
+                Some(RwLock::new(BTreeMap::new()))
```
badboy: If we defer collection, we might still have `Lifetime::Ping` data around on startup.
We would need to load this data from disk in order for metrics to be valid (otherwise we would later override them when they are persisted again)

glean-core/src/lib.rs
brizental: Hm, now that you say it, this doesn't really need to be an option.

> The database already knows if it has the ping metrics store, no need to keep this flag around at all

I get this part, can you explain it a little bit further?

badboy: Ah sorry. Misread the code, ignore that part, I thought this was the glean object.
It's the configuration object, so of course we need it

glean-core/src/database/mod.rs
brizental: This PR is only about [Bug 1596440](https://bugzilla.mozilla.org/show_bug.cgi?id=1596440) and @Dexterp37 advised me to do all the persisting when I work on [Bug 1596442](https://bugzilla.mozilla.org/show_bug.cgi?id=1596442), so, for now, there shouldn't ever be anything on disk.

I wonder if it would be easier to just work on both bugs in this PR. What do you think?

badboy: I'm fine with splitting it out then, that does indeed make sense and makes the review easier.

brizental:
@badboy @Dexterp37 this PR is ready for a final review :)

brizental:
@Dexterp37 @badboy this is ready for review.

I still need to see what's going on with CI and the new cbindgen version, but that shouldn't get much in the way. For context: I just downloaded cbindgen in version 0.10.0 and that broke some things and changed the generated header files (because CI was using 0.9.10). I added a commit with the update + header file changes to this PR, but that is still breaking the build.

iOS integration test is also failing in the CI, but that doesn't look like it is related to these chages.

.circleci/config.yml
```diff
@@ -142,7 +142,7 @@ jobs:
       - run:
           name: FFI header consistency check
           command: |
-            wget https://github.com/eqrion/cbindgen/releases/download/v0.9.1/cbindgen
+            wget https://github.com/eqrion/cbindgen/archive/v0.10.0.tar.gz -O cbindgen
```
badboy: hm, looks like cbindgen is currently missing a binary release of v0.10.0.
I reached out to emilio to fix that.
In the meantime you could downgrade your local install (`cargo install cbindgen --ver "=0.9.1"`) and regenerate the headers with that.

We can land the cbindgen upgrade in a separate PR then.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -17,12 +17,13 @@ import mozilla.telemetry.glean.net.PingUploader
...
  */
-@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents")
+@Structure.FieldOrder("dataDir", "packageName", "uploadEnabled", "maxEvents", "deferCollection")
 internal class FfiConfiguration(
     dataDir: String,
     packageName: String,
     uploadEnabled: Boolean,
-    maxEvents: Int? = null
+    maxEvents: Int? = null,
+    deferCollection: Boolean
```
Dexterp37: I think `deferCollection` is a bit unclear: we're not really deferring the collection, we're deferring writing to disk, but only for `lifetime: metrics`.

At the very least, this should be `deferPersistence`. Maybe even`delayPingLifetimeIO`? @badboy ?

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -37,6 +38,8 @@ internal class FfiConfiguration(
     public var uploadEnabled: Byte = uploadEnabled.toByte()
     @JvmField
     public var maxEvents: IntByReference = if (maxEvents == null) IntByReference() else IntByReference(maxEvents)
+    @JvmField
+    public var deferCollection: Byte = deferCollection.toByte()
```
Dexterp37: This is not being exposed nor documented in the `Configuration` object below. Is this intentional?

glean-core/src/database/mod.rs
```diff
@@ -313,6 +363,25 @@ impl Database {
             return Ok(());
         }
 
+        if lifetime == Lifetime::Ping {
```
Dexterp37: Please add a comment explaining why we're doing this.

glean-core/src/database/mod.rs
```diff
@@ -351,6 +420,14 @@ impl Database {
     ///
     /// * This function will **not** panic on database errors.
     pub fn clear_ping_lifetime_storage(&self, storage_name: &str) -> Result<()> {
+        let ping_lifetime_data = &self.ping_lifetime_data;
```
Dexterp37: Please add a comment explaining why we're doing this.

glean-core/src/database/mod.rs
```diff
@@ -413,6 +490,16 @@ impl Database {
             return Ok(());
         }
 
+        if lifetime == Lifetime::Ping {
```
Dexterp37: Please add a comment explaining why we're doing this.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
brizental: Yes, it's intentional because, as I understand it, this property will always be `false` for Kotlin and Swift, since this is something we are adding for desktop. Do you think it would be useful to make it configurable for mobile platforms?

Dexterp37: I think it's worth adding it here as well or, at the very least, document why it's not here.

docs/appendix/twig.md
```diff
@@ -12,3 +12,4 @@ They could be release notes, documentation, hopes, dreams, or whatever: so long
 * 2019-11-19: [GeckoView + Glean = Fenix performance metrics](https://www.a2p.it/wordpress/tech-stuff/mozilla/geckoview-glean-fenix-performance-metrics/)
 * 2019-11-21: [What is the Glean logo about?](https://dianaciufo.wordpress.com/2019/10/11/glean-graphic-identity-for-mozilla-firefox/)
 * 2019-11-22: [Glean in Private](https://chuttenblog.wordpress.com/2019/11/22/this-week-in-glean-glean-in-private/)
+* 2019-11-29: [Differences](https://fnordig.de/2019/11/29/this-week-in-glean/)
```
Dexterp37: This seems not relevant for this PR :) Rebase problem?

brizental: Yeap, wow, you were quick! I fixed that as saw and fixed it the second I sent it in, lol.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
brizental: I am happy with `delayPingLifetimeIO`, should I just change it to this and we can discuss this further on my next PR? Specially since this prop might change to an interval for the next one?

Dexterp37: Yes, let's change it here. Depending on how we go for it, we might not really need to offer ways to customise the timeout ;) But let's talk about that in the next PR :)

glean-core/src/database/mod.rs
brizental: I added code related to this, I am not sure it works 🤔

glean-core/src/database/mod.rs
brizental: I looked into this, and the code is similar but not exaclty the same specially because of the `if let`, I could figure out a way to this nicely. Let me know if you think of something :)

brizental: I was talking to @badboy and he suggested filing a bug to look into this after this PR is done. What do you think?

glean-core/src/database/mod.rs
brizental: This is going to require some refoactoring of this function and I don't know if it makes sense to do it, because if I do that it will be removed in my next PR.

glean-core/src/database/mod.rs
Dexterp37: Fine by me!

glean-core/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub struct Configuration {
     pub application_id: String,
     /// The maximum number of events to store before sending a ping containing events.
     pub max_events: Option<usize>,
+    /// Whether Glean should defer ping collection.
```
Dexterp37: Can you please expand on what this implies? Please note that this does not defer "ping" collection, but it defers persisting to disk metrics that have "lifetime: ping".

glean-core/src/database/mod.rs
```diff
@@ -22,17 +22,28 @@ pub struct Database {
     // as the application lives: they don't need to be persisted
     // to disk using rkv. Store them in a map.
     app_lifetime_data: RwLock<BTreeMap<String, Metric>>,
+    // If the `delay_ping_lifetime_io` Glean config option is `true`,
+    // we will save the 'ping' lifetime data in a map temporarily
```
Dexterp37:
```suggestion
    // we will save metrics with 'ping' lifetime data in a map temporarily
```

glean-core/src/lib.rs
brizental: Ah, sorry this one slipped my mind.

What do you think about "Whether Glean should delay persistence of metrics with 'ping' lifetime data."

Dexterp37: No problem :) Maybe "Whether Glean should delay persistence of data from metrics with `ping` lifetime"?

glean-core/ffi/src/lib.rs
```diff
@@ -101,6 +101,7 @@ pub struct FfiConfiguration<'a> {
     package_name: FfiStr<'a>,
     upload_enabled: u8,
     max_events: Option<&'a i32>,
+    delay_ping_lifetime_io: u8,
```
badboy: Do we actually need to expose this over FFI?
The platform layers don't expose it to users right now, just setting default values.

The only intended usage right now is Firefox and for that we plan to use the Rust API directly.
IMO it makes more sense to keep it out of the FFI for now and add it when necessary.
Wdyt, @Dexterp37?

badboy: Also @chutten 

Only minor thing: for the API preview I want to land tomorrow we will use the FFI. Can we live without the delay initially? The newer API i propose would not go through FFI.

Dexterp37: @badboy mh, I feel like exposing the option across all the platforms for consistency would be better/make it easier for understanding and documenting what's available where. However I have no strong opinion on this.

With respect to the FOGotype, no, we don't need to block on it. This is for project FOG, not for the fogotype :)

badboy: All right, then we keep it as is (and I'm personally gonna remove it from the FFI layer in 3 months if we never use it :P)

glean-core/src/database/mod.rs
```diff
@@ -707,4 +810,52 @@ mod test {
             );
         }
     }
+
+    #[test]
+    fn test_deferred_ping_collection() {
```
Dexterp37:
```suggestion
    fn test_deferred_ping_lifetime_collection() {
```

glean-core/src/database/mod.rs
Dexterp37: What makes you think this is not working?

glean-core/src/lib.rs
```diff
@@ -65,6 +65,8 @@ pub struct Configuration {
     pub application_id: String,
     /// The maximum number of events to store before sending a ping containing events.
     pub max_events: Option<usize>,
+    /// Whether Glean should delay persistence of data from metrics with ping lifetime
```
```suggestion
    /// Whether Glean should delay persistence of data from metrics with ping lifetime.
```

glean-core/tests/timing_distribution.rs
```diff
@@ -28,6 +28,7 @@ fn serializer_should_correctly_serialize_timing_distribution() {
         application_id: GLOBAL_APPLICATION_ID.into(),
         upload_enabled: true,
         max_events: None,
+        delay_ping_lifetime_io: false,
```
Dexterp37: Woha, configuration is repeated, always in the same way, across all the test files. I wonder if there's a way to share this and to create a `create_test_configuration` test-only function. That's for a follow-up PR, @brizental do you mind filing a bug for it?

brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1601209

glean-core/src/lib.rs
brizental: > What makes you think this is not working?

Just my lack of confidence in my understanding of this part of the code, heh.

glean-core/src/database/mod.rs
```diff
@@ -707,4 +810,52 @@ mod test {
...
+            match metric {
+                Metric::String(s) => assert_eq!(test_value, s),
+                _ => panic!("Unexpected data found"),
+            }
+        };
+
+        db.iter_store_from(Lifetime::Ping, test_storage, None, &mut snapshotter);
+        assert_eq!(1, found_metrics, "We only expect 1 Lifetime.Ping metric.");
+
+        let store: SingleStore = unwrap_or!(
```
Dexterp37: We discussed over Slack and agreed to slightly change this to make sure that something is being written to rkv at all :)

glean-core/src/lib.rs
brizental: Ok, to improve my confidence I made a little test where I record a metric with a `Lifetime::User` and assert that it `is_some()`. That worked and now I feel better about this test :) Thanks @Dexterp37 for suggesting that.

glean-core/src/database/mod.rs
brizental: Would the comment I added suffice here or do you mean for me to add the whole experiment I did to lowre my confusion?

I personally don't think the experiment adds to test and event makes it a little confusing (since it is code that is not related), but I want to know your opinion :)

Dexterp37: No, that's fine :)

PR #533, opened 2019-11-29
Count the number of times a Glean product was force-closed

Dexterp37:
This will require a new release and a PR in Android Components.


glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -39,6 +55,15 @@ internal class GleanLifecycleObserver : LifecycleEventObserver {
                 // because it belongs to the baseline ping and that ping is sent every
                 // time the app goes to background.
                 GleanBaseline.duration.start()
+
+                // Check if the "dirty bit" was set. If it was, increment the metric.
+                // Flip the dirty bit to 'true' after.
+                if (sharedPreferences.getBoolean("dirty", false)) {
```
travis79: Shouldn't this be checking if this is true before it increments the counter?  If the flag is set to true in ON_START and then it misses the ON_STOP, it should still be true when you check it here meaning we didn't have a "clean" shutdown.

Dexterp37: This line is doing what you suggest: `getBoolean` will attempt to read the stored value and, if there's no stored value, will return the default value (`false`)

glean-core/android/metrics.yaml
```diff
@@ -86,3 +86,21 @@ glean.validation:
...
     send_in_pings:
       - baseline
+
+  app_forceclosed_count:
+    type: counter
+    description:
+      This metric counts how many times a dirty bit was detected when starting
+      the Glean SDK, on Android.
+      This dirty bit is set to `true` when Android's lifecycle `ON_START` is hit
+      and flipped back to false on `ON_STOP`. When is force-closed, we might
```
badboy:
```suggestion
      and flipped back to false on `ON_STOP`. When an application is force-closed, it might
```

glean-core/android/metrics.yaml
```diff
@@ -86,3 +86,21 @@ glean.validation:
...
+      not get the `ON_STOP` and so the value would stay `true`. Next time Glean
```
badboy:
```suggestion
      not get the `ON_STOP` and so the value would stay `true`. The next time Glean
```

glean-core/android/metrics.yaml
```diff
@@ -86,3 +86,21 @@ glean.validation:
...
+      starts, it would increment this counter.
```
badboy:
```suggestion
      starts, it will increment this counter.
```

glean-core/android/metrics.yaml
```diff
@@ -86,3 +86,21 @@ glean.validation:
...
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1600340
+    data_reviews:
+      - TODO
+    notification_emails:
+      - glean-team@mozilla.com
+      - aplacitelli@mozilla.com
+    expires: 2019-12-31
```
badboy: So soon?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -4,17 +4,30 @@
...
  * functionality (where the actual work of sending pings is done).
  */
-internal class GleanLifecycleObserver : LifecycleEventObserver {
+internal class GleanLifecycleObserver(
+    private val applicationContext: Context
+) : LifecycleEventObserver {
+    // Using SharedPreferences to track the status here should be less of an I/O hit
+    // compared to plain file usage, given that most of its operations are off the main
+    // thread. This also gives us the nice guarantee that once the function is called
+    // the OS will keep SharedPreferences working, even though the process is being killed.
```
badboy: Not sure what this means. What does it mean to "keep SharedPreferences working"?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -4,17 +4,30 @@
...
+    internal val sharedPreferences: SharedPreferences by lazy {
+        applicationContext.getSharedPreferences("DirtyBitStatus", Context.MODE_PRIVATE)
```
badboy: No exact idea how shared preferences work, should this have any "namespacing" using the "Glean" word?

glean-core/android/metrics.yaml
```diff
@@ -86,3 +86,21 @@ glean.validation:
     expires: 2019-12-20
     send_in_pings:
       - baseline
+
+  app_forceclosed_count:
```
badboy: Is this missing a `send_in_pings` statement?

glean-core/android/metrics.yaml
Dexterp37: Yes, we're planning on shipping this to Nightly and see if that helps figuring out our ongoing investigations. We could always extend, if needed.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
Dexterp37: I mean that `"You don't need to worry about Android component lifecycles and their interaction with apply() writing to disk. The framework makes sure in-flight disk writes from apply() complete before switching states. "`, from [the docs](https://developer.android.com/reference/android/content/SharedPreferences.Editor.html#apply()).

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
Dexterp37: That's a good point. I did not want to use the class name, but having a Glean prefix sounds like a good idea :)

glean-core/android/metrics.yaml
Dexterp37: Yes, that's intentional. Glean sends this automatically in the `metrics` ping.

badboy: And here I am, always thinking that `send_in_pings: default` would be required.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
travis79: Okay, I don't think I was paying attention to the fact that `false` here was the default value for if the pref doesn't exist.  Carry on :)

Dexterp37:
ps: I love the way docs-linkcheck fails when dropping "TODO" in `data_reviews` fields.

PR #535, opened 2019-12-02
Update cbindgen to v0.10.0

brizental:


brizental:
Hm, I see there are failures in the iOS integration tests, but I can't reproduce them locally. I rebuild everything, so that is weird.

@badboy any ideia why that could happen?

badboy:
>     ✗ testViewControllerInteraction, Test crashed

That happens.... I restarted the test.

PR #538, opened 2019-12-03
Use up2date emulator

badboy:
Let's see how that goes.
I will rekick the integration test a couple of times if it succeeds.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -80,7 +80,7 @@ jobs:
       - run: make yamllint
   Check Swift formatting:
     macos:
-      xcode: "11.2.0"
+      xcode: "11.2.1"
     steps:
       - checkout
       - run:
@@ -202,7 +202,7 @@ jobs:
 
   iOS build and test:
     macos:
-      xcode: "11.2.0"
+      xcode: "11.2.1"
     steps:
       - run:
           name: Set Ruby Version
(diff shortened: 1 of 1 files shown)


.circleci/config.yml
```diff
@@ -202,7 +202,7 @@ jobs:
 
   iOS build and test:
     macos:
-      xcode: "11.2.0"
+      xcode: "11.2.1"
```
Dexterp37: Does this have implications for the generated code? Would this require apps to use the same version we specify here?

badboy: Nope, it's the same image: https://circle-macos-docs.s3.amazonaws.com/image-manifest/v1532/index.html
_Actual_ Xcode 11.2.0 was pulled, because it was next to unusable (contained code generator bugs, Apple itself discourages use of that).
CircleCI replaced 11.2.0 with 11.2.1 on its own.
I only updated that here to reflect better that we're using 11.2.1

badboy:
First run passed, trying again.

.circleci/config.yml
Dexterp37: Thanks for clarifying. Let's give this a couple of more shots on CI before merging.

badboy:
Second run passed, trying again.

badboy:
Third run passed, trying once more.

badboy:
Fourth run passed, landing!

PR #539, opened 2019-12-03
Python deletion request

badboy:


glean-core/python/glean/glean.py
```diff
@@ -213,6 +214,9 @@ def set_upload_enabled():
 
                 if original_enabled is False and cls.get_upload_enabled() is True:
                     cls._initialize_core_metrics()
+                else:
```
mdboom: Should the logic here be `if original_enabled is True and cls.get_upload_enabled() is False` (so that it only happens the first time on toggling, not if `False` is passed twice?

glean-core/python/glean/net/deletion_ping_upload_worker.py
```diff
@@ -0,0 +1,95 @@
...
+
+import logging
+from pathlib import Path
+import re
+
+
+log = logging.getLogger(__name__)
+
+
+class DeletionPingUploadWorker:
```
mdboom: It seems like we could remove a lot of code duplication here by just making this a subclass of `PingUploadWorker` and overriding `_PINGS_DIR`, and adding the `try/catch` for `FileNotFoundError` below.

glean-core/python/glean/_ffi.py
```diff
@@ -43,6 +43,7 @@ def _load_header(path: str) -> str:
 ffi = FFI()
 ffi.cdef(_load_header("glean.h"))
 lib = ffi.dlopen(str(Path(__file__).parent / get_shared_object_filename()))
+lib.glean_enable_logging()
```
mdboom: Makes sense, however, eventually we'll probably want to make this controllable through the Python logging APIs.  https://bugzilla.mozilla.org/show_bug.cgi?id=1600946

glean-core/python/glean/glean.py
badboy: Actually ... I think so, which means we need to change that on the Kotlin/Swift side as well

glean-core/python/glean/net/deletion_ping_upload_worker.py
badboy: Would you let this through if we tackle the deduplication in a follow-up, similar to https://bugzilla.mozilla.org/show_bug.cgi?id=1599427?

glean-core/python/glean/_ffi.py
badboy: 👍
Though we don't allow configuring this on other platforms either (it _is_ configurable by environment variables and by default shouldn't show logging I think).

I leave it here for now.

glean-core/python/glean/net/deletion_ping_upload_worker.py
mdboom: Sure.

PR #544, opened 2019-12-04
Refactor Android uploaders to share more code

badboy:
I moved shared functionality into free-standing functions

We could keep it in the companion object, but it felt a bit weird to re-use functions in another class across its companion object.
Also the one in the derived class just shadows the one from the superclass, so we need to use explicit calls.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -22,6 +22,103 @@ import java.io.FileNotFoundException
 import java.io.FileReader
 import java.io.IOException
 
+// Since ping file names are UUIDs, this matches UUIDs for filtering purposes
+private const val FILE_PATTERN = "[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}"
+private const val LOG_TAG = "glean/PingUploadWorker"
+
+/**
+ * Build the constraints around which the worker can be run, such as whether network
+ * connectivity is required.
+ *
+ * @return [Constraints] object containing the required work constraints
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+fun buildConstraints(): Constraints = Constraints.Builder()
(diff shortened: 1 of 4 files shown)


badboy:
Took it to a test run and also ran the UI tests (`./gradlew :glean-sample-app:connectedAndroidTest -PdisablePreDexg`).
Turns out: it was buggy.

Please re-review the latest 2 commits which 1) make it work, 2) make the tests pass

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -22,6 +22,105 @@ import java.io.FileNotFoundException
...
+private const val LOG_TAG = "glean/PingUploadWorker"
+
+/**
+ * Build the constraints around which the worker can be run, such as whether network
+ * connectivity is required.
+ *
+ * @return [Constraints] object containing the required work constraints
+ */
+@VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+fun buildConstraints(): Constraints = Constraints.Builder()
```
travis79: This needs marked as `internal`

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -22,6 +22,105 @@ import java.io.FileNotFoundException
...
+            } else {
+                // Delete files that don't match the UUID FILE_PATTERN regex
+                Log.d(LOG_TAG, "Pattern mismatch. Deleting ${file.name}")
+                file.delete()
+            }
+        }
+    }
+
+    return success
+}
```
travis79: I'm not _super_ excited about global stuff shared between multiple classes floating around the top of the PingUploadWorker.kt file.  Would this be better to have in a Util file?

badboy: It's not more or less global than it was before as static functions inside a companion object.

I don't like introducing "Util" files as a catch-all for non-class functions.

travis79: Okay, not a strong opinion on this, it was just a nit.

PR #545, opened 2019-12-04
On init and if upload is disabled, see if we need to resend a de…

badboy:
No new ping will be generated.
It will only sent a previously-failed ping, if any.

I would block that on #539, so that I can include the necessary Python changes as well


badboy:
There is! Let me do that.

badboy:
Looks like it currently already retries: https://github.com/mozilla/glean/blob/master/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/DeletionPingUploadWorker.kt#L188

As I understand it it's tried for a total runtime of 10 minutes (unless force-stopped I guess). Just makes it a bit harder to test.
I might just consider on Swift then, where we don't have this automatic retry.

PR #548, opened 2019-12-05
Bug 1601710 - Expose Glean-iOS testing functions for public use.

travis79:


travis79:
> I assume we will need to do this for all metric types as well?

Yes, I guess I was waiting to allow time for bike-shedding over making this public before going through all of the specific metric API test functions.

There is a way to allow the `@testable import` from a binary but it messes up the optimizations so it would still have the debug symbols included.  I have this vague idea that it might be possible to build a testing binary and a release binary...

PR #553, opened 2019-12-06
1592006: Implement events for Python bindings

mdboom:


glean-core/python/glean/_loader.py
```diff
@@ -87,7 +92,15 @@ def _get_metric_object(metric: glean_parser.metrics.Metric) -> Any:
 
     glean_metric.__doc__ = metric.description
 
-    return glean_metric
+    yield name, glean_metric
+
+    # Events also need to define an enumeration
+    if metric.type == "event":
```
Dexterp37: I wish we could just use glean_parser for these kind of things :(

glean-core/python/glean/util.py
```diff
@@ -22,3 +23,10 @@ def get_locale_tag() -> str:
     value = value.replace("_", "-")
 
     return value
+
+
+def time_ms() -> int:
```
Dexterp37: I bet there's no concept of `uint`, right?

glean-core/python/glean/_ffi.py
```diff
@@ -98,6 +98,13 @@ def ffi_encode_vec_string(strings: List[str]) -> Any:
     return result
 
 
+def ffi_encode_vec_int32(values: List[str]) -> Any:
```
badboy: `List[str]`? Shouldn't this be `List[int]`?

glean-core/python/glean/metrics/event.py
```diff
@@ -0,0 +1,199 @@
...
+                where additional richer context is needed. The maximum length
+                for values is 100.
+        """
+        if self._disabled:
+            return
+
+        timestamp = util.time_ms()
+
+        @Dispatcher.launch
+        def record():
```
badboy: I hope this just shadows the above function definition and thus won't cause any problems?

glean-core/python/glean/_ffi.py
mdboom: Good catch.

glean-core/python/glean/_loader.py
mdboom: Not sure I follow.  We *are* using glean_parser here.  This file is essentially what is required for any of the language generators (just generating objects, not code).  The parsing etc. still uses glean_parser underneath.

glean-core/python/glean/metrics/event.py
mdboom: It's a nested function, so not accessible from the outside.  We could really call it whatever we want -- it's never even used by name -- it's just a way to put a function on the dispatcher.  If Python had anonymous functions, I'd use that...

glean-core/python/glean/util.py
mdboom: That's correct: Python 3 only has a single `int` type.

glean-core/python/glean/_loader.py
Dexterp37: Yup, sorry, let me clarify: I wish this was as simple as just creating a `jinja` template in glean_parser :)

glean-core/python/glean/metrics/event.py
badboy: 👍

PR #559, opened 2019-12-09
Extend timeout in UI tests

badboy:
For some reason we see frequent timeout errors in these tests when run
on CI.
It's unclear if that is due to a too-short timeout or actual issues.

Let's increase it for now and see if it is still hit.


badboy:
Now the other test simply crashed -_-

PR #560, opened 2019-12-09
Add a summary of our third-party dependencies

badboy:
I will document the exact process in a followup

### DIFF
--- a/publish.gradle
+++ b/publish.gradle
@@ -103,6 +103,15 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
                         }
                     }
 
+                    def depLicenses = new XmlSlurper().parse(new File("${projectDir}/dependency-licenses.xml"))
+                    depLicenses.license.each { node ->
+                        license {
+                            name = node.name.text()
+                            url = node.url.text()
+                        }
+                    }
+
+
                     developers {
                         developer {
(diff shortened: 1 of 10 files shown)


mdboom:
IIUC, this extracts the Rust dependencies, but not Kotlin, Swift or Python dependencies.  Should we file bugs to tackle those?  (It's probably less critical, since there's no GPL-triggering static linking possible in Kotlin or Python, but it would still be nice to be complete...)

publish.gradle
```diff
@@ -103,6 +103,15 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
                         }
                     }
 
+                    def depLicenses = new XmlSlurper().parse(new File("${projectDir}/dependency-licenses.xml"))
```
Dexterp37: NEAT! So this gets added to our artefacts. Nice.

badboy:
Will file one.
The nice thing: cargo-about is extendeable and you can add "custom" licenses for other code!

publish.gradle
badboy: Yes, copied over from a-s :D

badboy:
Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1602447

PR #562, opened 2019-12-09
Document how to add Glean to an iOS project

badboy:


docs/user/adding-glean-to-your-project.md
```diff
@@ -99,16 +99,46 @@ There are [additional parameters](android-build-configuration-options.md) that c
...
 ```
 
+> **Important:** the `{latest-version}` placeholder should be replaced with the version number of the latest Glean SDK release.
+> You can find the version number on the [release page](https://github.com/mozilla/glean/releases/latest).
+
+Then check out and build the new dependency:
+
+```
+carthage bootstrap --platform iOS --color auto --cache-builds
+```
```
travis79: If they have just checked out fresh then `bootstrap` is perfect, if they are adding this to an existing application, then it might be more efficient using `update` instead:

```bash
carthage update glean --platform iOS
```
I find this is much faster, especially when updating the Glean version to a new release.

docs/user/adding-glean-to-your-project.md
```diff
@@ -99,16 +99,46 @@ There are [additional parameters](android-build-configuration-options.md) that c
...
+
+
 #### Integrating with the build system
 
-> **Note:** This is not yet documented, as the exact mechanism is not finalized. See [Bug 1589383](https://bugzilla.mozilla.org/show_bug.cgi?id=1589383) for more information.
+For integration with the build system you can follow the [Carthage Quick Start steps](https://github.com/Carthage/Carthage#quick-start).
+
+1. After building the dependency one drag the built `.framework` binaries from `Carthage/Build/iOS` into your application's Xcode project.
```
travis79: I actually did this by clicking the `+` icon in the General tab of the Xcode project settings and selecting "Add Other -> Files..." from the dialog that appeared.  This opened a file explorer window where I navigated to the `Carthage/Build/iOS` directory and selected `Glean.framework`

Dragging and dropping would have been nice to know, I think.  I should have suspected Apple would allow such nice things.

docs/user/adding-glean-to-your-project.md
```diff
@@ -99,16 +99,46 @@ There are [additional parameters](android-build-configuration-options.md) that c
...
+   /usr/local/bin/carthage copy-frameworks
+   ```
+
+3. Add the path to the Glean framework under "Input Files":
+
+   ```
+   $(SRCROOT)/Carthage/Build/iOS/Glean.framework
+   ```
+
+4. Add the paths to the copied framework to the "Output Files":
```
travis79: Note to self: Be sure and add a link about enabling the debug tools here!

docs/user/adding-glean-to-your-project.md
travis79: Technically we only support one build platform right now so the `--platform iOS` bit may be redundant.

docs/user/adding-glean-to-your-project.md
badboy: :D I just copied what Carthage documented.

PR #563, opened 2019-12-09
Python bindings: Increase test coverage

mdboom:


glean-core/python/glean/__init__.py
```diff
@@ -15,7 +15,7 @@
 __version__: str = "unknown"
 try:
     __version__ = str(get_distribution("glean-sdk").version)
-except DistributionNotFound:
+except DistributionNotFound:  # pragma: no cover
```
badboy: Increase coverage by .... ignoring stuff :D

Dexterp37: That's the way to do it!

PR #565, opened 2019-12-11
Add method to persist ping lifetime data on demand

brizental:
Fixes [Bug 1596442](https://bugzilla.mozilla.org/show_bug.cgi?id=1596442)

Continuation of #530 

I wasn't sure if I should add a method for this directly on the Glean object.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -5,6 +5,7 @@
 use std::collections::btree_map::Entry;
 use std::collections::BTreeMap;
 use std::fs;
+use std::str;
 use std::sync::RwLock;
 
 use rkv::{Rkv, SingleStore, StoreOptions};
@@ -24,9 +25,7 @@ pub struct Database {
     app_lifetime_data: RwLock<BTreeMap<String, Metric>>,
     // If the `delay_ping_lifetime_io` Glean config option is `true`,
     // we will save metrics with 'ping' lifetime data in a map temporarily
-    // so as to persist them to disk using rkv in bulk on shutdown,
-    // or after a given interval, instead of everytime a new metric
-    // is created / updated.
+    // so as to persist them to disk using rkv in bulk on demand.
     ping_lifetime_data: Option<RwLock<BTreeMap<String, Metric>>>,
 }
 
@@ -35,8 +34,11 @@ impl Database {
     ///
     /// This opens the underlying rkv store and creates
     /// the underlying directory structure.
+    ///
+    /// It also loads any Lifetime::Ping data that might be
+    /// persisted, in case `delay_ping_lifetime_io` is set.
     pub fn new(data_path: &str, delay_ping_lifetime_io: bool) -> Result<Self> {
-        Ok(Self {
+        let db = Self {
             rkv: Self::open_rkv(data_path)?,
             app_lifetime_data: RwLock::new(BTreeMap::new()),
             ping_lifetime_data: if delay_ping_lifetime_io {
@@ -44,7 +46,11 @@ impl Database {
             } else {
                 None
             },
-        })
+        };
+
+        db.load_ping_lifetime_data();
+
+        Ok(db)
     }
 
     /// Creates the storage directories and inits rkv.
@@ -75,6 +81,40 @@ impl Database {
         match metric_key {
             Some(k) => format!("{}#{}", storage_name, k),
             None => format!("{}#", storage_name),
+        }
+    }
+
+    /// Loads Lifetime::Ping data from rkv to memory,
+    /// if `delay_ping_lifetime_io` is set to true.
+    ///
+    /// Does nothing if it isn't or if there is not data to load.
+    fn load_ping_lifetime_data(&self) {
+        if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+            let mut data = ping_lifetime_data
+                .write()
+                .expect("Can't read ping lifetime data");
+
+            let store: SingleStore = unwrap_or!(
+                self.rkv
+                    .open_single(Lifetime::Ping.as_str(), StoreOptions::create()),
+                return
+            );
+
+            let reader = unwrap_or!(self.rkv.read(), return);
+            let mut iter = unwrap_or!(store.iter_start(&reader), return);
+
+            while let Some(Ok((metric_name, value))) = iter.next() {
+                let metric_name = match str::from_utf8(metric_name) {
+                    Ok(metric_name) => metric_name.to_string(),
+                    _ => continue,
+                };
+                let metric: Metric = match value.expect("Value missing in iteration") {
+                    rkv::Value::Blob(blob) => unwrap_or!(bincode::deserialize(blob), continue),
+                    _ => continue,
+                };
+
+                data.insert(metric_name, metric);
+            }
         }
     }
 
@@ -128,7 +168,7 @@ impl Database {
             return;
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immediately persisted to disk if
         // Glean has `delay_ping_lifetime_io` set to true
         if lifetime == Lifetime::Ping {
             if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
@@ -282,7 +322,7 @@ impl Database {
             return Ok(());
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immediately persisted to disk if
         // Glean has `delay_ping_lifetime_io` set to true
         if lifetime == Lifetime::Ping {
             if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
@@ -364,7 +404,7 @@ impl Database {
             return Ok(());
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immediately persisted to disk if
         // Glean has `delay_ping_lifetime_io` set to true
         if lifetime == Lifetime::Ping {
             if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
@@ -423,14 +463,13 @@ impl Database {
     ///
     /// * This function will **not** panic on database errors.
     pub fn clear_ping_lifetime_storage(&self, storage_name: &str) -> Result<()> {
-        // Lifetime::Ping might have data saved to `ping_lifetime_data`
+        // Lifetime::Ping data will be saved to `ping_lifetime_data`
         // in case `delay_ping_lifetime_io` is set to true
         if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
             ping_lifetime_data
                 .write()
                 .expect("Can't access ping lifetime data as writable")
                 .clear();
-            return Ok(());
         }
 
         self.write_with_store(Lifetime::Ping, |mut writer, store| {
@@ -495,20 +534,26 @@ impl Database {
             return Ok(());
         }
 
-        // Lifetime::Ping data is not persisted to disk if
-        // Glean has `delay_ping_lifetime_io` set to true
+        // Lifetime::Ping data will be saved to `ping_lifetime_data`
+        // in case `delay_ping_lifetime_io` is set to true
         if lifetime == Lifetime::Ping {
             if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
                 let mut data = ping_lifetime_data
                     .write()
                     .expect("Can't access app lifetime data as writable");
                 data.remove(&final_key);
-                return Ok(());
             }
         }
 
         self.write_with_store(lifetime, |mut writer, store| {
-            store.delete(&mut writer, final_key.clone())?;
+            if let Err(e) = store.delete(&mut writer, final_key.clone()) {
+                if self.ping_lifetime_data.is_some() {
+                    // If ping_lifetime_data exists, it might be
+                    // that data is in memory, but not yet in rkv.
+                    return Ok(());
+                }
+                return Err(e.into());
+            }
             writer.commit()?;
             Ok(())
         })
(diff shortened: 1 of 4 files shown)


glean-core/src/database/mod.rs
```diff
@@ -843,20 +880,49 @@ mod test {
...
-            .unwrap_or(None)
-            .is_none());
+        {
+            // Make sure data **wasn't** persisted with rkv.
+            let store: SingleStore = db.rkv
+                .open_single(Lifetime::Ping.as_str(), StoreOptions::create())
+                .unwrap();
+            let reader = db.rkv.read().unwrap();
+            assert!(store
+                .get(&reader, format!("{}#{}", test_storage, test_metric_id))
```
brizental: This is essentially what `get_storage_key` does, but that is not a public method.

glean-core/src/database/mod.rs
```diff
@@ -843,20 +880,49 @@ mod test {
...
+        db.persist_ping_lifetime_data().unwrap();
+
+        {
+            // Make sure that data **was** persisted with rkv.
+            let store: SingleStore = db.rkv
+                .open_single(Lifetime::Ping.as_str(), StoreOptions::create())
+                .unwrap();
+            let reader = db.rkv.read().unwrap();
+            assert!(store
+                .get(&reader, format!("{}#{}", test_storage, test_metric_id))
```
brizental: This is essentially what `get_storage_key` does, but that is not a public method.

glean-core/src/database/mod.rs
```diff
@@ -495,15 +491,16 @@ impl Database {
...
+        // in case `delay_ping_lifetime_io` is set to true
         if lifetime == Lifetime::Ping {
             if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
                 let mut data = ping_lifetime_data
                     .write()
                     .expect("Can't access app lifetime data as writable");
-                data.remove(&final_key);
-                return Ok(());
+                if data.remove(&final_key).is_some() {
+                    return Ok(());
```
mdboom: What if the data exists both in memory and on disk?  Shouldn't we clear it both places?  It looks like this will delete from memory, but leave it on disk.

glean-core/src/database/mod.rs
```diff
@@ -282,7 +279,7 @@ impl Database {
             return Ok(());
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immediatelly persisted to disk if
```
travis79:
```suggestion
        // Lifetime::Ping data is not immediately persisted to disk if
```

glean-core/src/database/mod.rs
```diff
@@ -364,7 +361,7 @@ impl Database {
             return Ok(());
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immeditally persisted to disk if
```
travis79:
```suggestion
        // Lifetime::Ping data is not immediately persisted to disk if
```

glean-core/src/database/mod.rs
```diff
@@ -537,13 +543,42 @@ impl Database {
...
+    /// Persist ping_lifetime_data to disk and clear it.
+    ///
+    /// Does nothing in case there is nothing to persist.
+    ///
+    /// ## Panics
+    ///
+    /// * This function will **not** panic on database errors.
+    pub fn persist_ping_lifetime_data(&self) -> Result<()> {
         if let Some(ping_lifetime_data) = &self.ping_lifetime_data {
+            {
```
travis79: Why do we need the extra scope/curly brackets here?

glean-core/src/database/mod.rs
brizental: Data *should* never be in both places. Once I persist it I clear `ping_lifetime_data`, so it should either be in one place or the other.

But computers don't work everytime, so maybe we should check if it is not on both places... What do you think?

glean-core/src/database/mod.rs
brizental: Because `ping_lifetime_data` holds a lock while reading the data. The brackets limit the scope of that lock, so that we can clear the data after reading it.

glean-core/src/database/mod.rs
mdboom: I see.  That makes sense, as long as we can't multi-thread around this (but I don't think we can, since we require a single handle to the Glean object for everything we do).  I think we can ignore my suggestion for now and readdress only if we discover it's an issue.

glean-core/src/database/mod.rs
```diff
@@ -843,20 +878,51 @@ mod test {
...
+                .open_single(Lifetime::Ping.as_str(), StoreOptions::create())
+                .unwrap();
+            let reader = db.rkv.read().unwrap();
+            assert!(store
+                .get(&reader, format!("{}#{}", test_storage, test_metric_id))
+                .unwrap_or(None)
+                .is_some());
+        }
+
+        // Verify that ping_lifetime_data was cleared and
```
Dexterp37: Mh, what are you testing here?

glean-core/src/database/mod.rs
```diff
@@ -537,13 +543,42 @@ impl Database {
...
+    }
 
+    /// Persist ping_lifetime_data to disk and clear it.
+    ///
+    /// Does nothing in case there is nothing to persist.
+    ///
+    /// ## Panics
+    ///
+    /// * This function will **not** panic on database errors.
+    pub fn persist_ping_lifetime_data(&self) -> Result<()> {
```
Dexterp37: I think this needs to be exposed somehow for external consumers. Where are you planning to expose it? Is that part of a different PR?

glean-core/src/database/mod.rs
brizental: I iterate over `Lifetime::Ping` stored data and make sure there is only one found metric, which would mean that we cleared the non-persisted data, since I just checked that data is indeed saved to rkv above.

That is not the most intuitive test, let me know if you have better ideas :)

glean-core/src/database/mod.rs
brizental: So, I wasn't sure just adding it as a public method to the Database module was enough.

I don't think it needs to be a different PR. Should I create a method in the Glean object to access this functionality?

glean-core/src/database/mod.rs
Dexterp37: What about this: we leave the existing test alone, then we can add another test that does the following:

- records a metric with `lifetime: ping`
- calls the persist function
- records a different metric with `lifetime: ping`
- checks that the first metric is in rkv and the second is in memory
- calls the persist function again
- checks that both metrics are in rkv and none is in memory

What do you think?

glean-core/src/database/mod.rs
Dexterp37: We should probably add it to both the Glean Rust object and then over FFI to the Glean object General APIs for each language we support and for which we exposed the option in #530

glean-core/src/database/mod.rs
travis79: Thanks for explaining, that makes sense!

glean-core/src/database/mod.rs
brizental: We can make that function public. I just wasn't sure if there was strong reasoning not to do that, hehe.

Dexterp37: Yeah, I mean, "is there a way to make it 'test-public'?" :D I guess we could just make it "package public"?

glean-core/src/database/mod.rs
brizental: Makes sense!

glean-core/src/database/mod.rs
brizental: I did this. I think it is clearer although it is a very long test :)

glean-core/src/database/mod.rs
brizental: Okay, as discussed AFK, I believe this should not be added to the **android** and **ios** bindings, as it is desktop specific. I will add it to the Rust object for now and maybe to the python bindings? What do you think?

I will also remove the options to defer ping lifetime data persistence from those platforms.

Does that sound good?

glean-core/src/database/mod.rs
```diff
@@ -843,20 +882,125 @@ mod test {
...
+        let test_metric_id2 = "telemetry_test.test_name2";
+        db.record_per_lifetime(
+            Lifetime::Ping,
+            test_storage,
+            test_metric_id2,
+            &Metric::String(test_value2.to_string()),
+        )
+        .unwrap();
+
+        {
```
Dexterp37: Can you comment on what we expect to find at this point? e.g. something along the lines of "We expect the first metric to be in rkv and the second to be recorded, but in memory, not in rkv, because.."

glean-core/src/lib.rs
```diff
@@ -543,6 +543,15 @@ impl Glean {
         metric.set_inactive(&self);
     }
 
+    /// Persist Lifetime::Ping data that might be in memory
+    /// in case `delay_ping_lifetime_io` is set or was set
+    /// at a previous time.
+    ///
+    /// If there is no data to persist, this function does nothing.
+    pub fn persist_ping_lifetime_data(&self) -> Result<()> {
```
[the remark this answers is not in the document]
brizental: Do you mean adding a section in the book about the Glean General API?

I see we now have language specific API documentation, generated from docstrings.

glean-core/src/database/mod.rs
brizental: Done :)

Also,
> Let's file a bug for adding that to the python bindings, no need to do it in this PR if it takes too much time.
Only now I saw that... I did it already.

glean-core/src/lib.rs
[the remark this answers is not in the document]
Dexterp37: > 
> 
> Do you mean adding a section in the book about the Glean General API?

Yes. We have clear sections for the specific metrics, but we have no page for the general API.

> I see we now have language specific API documentation, generated from docstrings.

Yes, we have API docs generated from code docstrings for [all languages](https://mozilla.github.io/glean/book/api/index.html), but that's for everything, not just for the general API.

glean-core/src/database/mod.rs
Dexterp37: Np ;) I guess @mdboom has to re-review then :D

glean-core/src/database/mod.rs
```diff
@@ -495,15 +491,20 @@ impl Database {
...
                 let mut data = ping_lifetime_data
                     .write()
                     .expect("Can't access app lifetime data as writable");
-                data.remove(&final_key);
-                return Ok(());
+                if data.remove(&final_key).is_some() {
+                    // If the metric is found in `ping_lifetime_data` there is
+                    // no need to check if it is also in rkv. Data should always
+                    // be either in one place or the other, since we clear
+                    // `ping_lifetime_data` upon persisting it to rkv.
```
badboy: I think this might lead to invalid data.

Consider this:

1. Ping-lifetime histogram data gets recorded, its stored in memory.
2. Ping-lifetime data is persisted, memory is cleared.
3. More ping-lifetime histogram data gets recorded, its stored in memory.
	* **NOTE**: This now starts from a fresh histogram
4. Ping-lifetime data is collected and send in a ping
    * **NOTE**: We're losing the first round of writes.

glean-core/src/database/mod.rs
```diff
@@ -537,13 +547,42 @@ impl Database {
             .write()
             .expect("Can't access app lifetime data as writable")
             .clear();
+    }
 
+    /// Persist ping_lifetime_data to disk and clear it.
```
badboy: See above why I think this is the wrong behavior.

glean-core/src/database/mod.rs
Dexterp37: Oh, great catch there.

brizental: Alright, right now how this works is: data is always **either** in memory or persisted. All the read / delete / update functions will check in both places and the create functions will always create in memory. Persistence will happen on demand and clear memory. I now understand how that can create problems.

The solution to these problems would be to change the way that works to something like: in memory data will mirror persisted data + any new changes and create / read / delete / update function will only apply to in memory data. Persisted data will then be updated on demand, but in memory will not be cleared.

> Is loading persisted data on startup coming in a follow up?

The way it is right now, I don't think this applies. With the new solution, yes that makes sense.

This is going to be a major change to the code on this PR. I think it might make sense to close this one and create a new one when I'm done. What  do you think?

badboy: I'm fine with re-using this PR, just force push if needed.
Though if it is easier for your own workflow starting fresh could work just as well, I don't mind either way.

The loading-on-startup part can be done in a follow up or with the other changes, I also don't mind that much.

glean-core/src/database/mod.rs
```diff
@@ -78,6 +84,40 @@ impl Database {
...
+                let metric_name = match str::from_utf8(metric_name) {
+                    Ok(metric_name) => metric_name.to_string(),
+                    _ => continue,
+                };
+                let metric: Metric = match value.expect("Value missing in iteration") {
+                    rkv::Value::Blob(blob) => unwrap_or!(bincode::deserialize(blob), continue),
+                    _ => continue,
+                };
+
+                data.insert(metric_name, metric.clone());
```
badboy: Do we need to clone here? I think `metric` is already an owned value, so we can avoid the clone.

glean-core/src/database/mod.rs
```diff
@@ -128,7 +168,7 @@ impl Database {
             return;
         }
 
-        // Lifetime::Ping data is not persisted to disk if
+        // Lifetime::Ping data is not immediatelly persisted to disk if
```
badboy:
```suggestion
        // Lifetime::Ping data is not immediately persisted to disk if
```

glean-core/src/database/mod.rs
```diff
@@ -495,20 +534,26 @@ impl Database {
...
-                return Ok(());
             }
         }
 
         self.write_with_store(lifetime, |mut writer, store| {
-            store.delete(&mut writer, final_key.clone())?;
+            if let Err(e) = store.delete(&mut writer, final_key.clone()) {
+                if self.ping_lifetime_data.is_some() {
+                    // If ping_lifetime_data exists, it might be
+                    // that data is in memory, but not yet in rkv.
```
badboy: Oh, that's a good catch! I haven't even thought about that.

glean-core/src/lib.rs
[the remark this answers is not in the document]
badboy: Yup, file a bug so we can document the general API properly (and then we can include this)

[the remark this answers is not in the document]
brizental: I can do that :)

mdboom:
I created [bug 1604191](https://bugzilla.mozilla.org/show_bug.cgi?id=1604191) to track adding proper async support to the Python bindings, at which time we can also add this deferred IO feature there.  I think that will be a requirement of being in devtools products, but is not a requirement for the current end-to-end testing use case.

brizental:
Thanks, I'll just fix linting and merge then :)

PR #566, opened 2019-12-11
1601262: Ping pattern analysis tool

mdboom:
I don't know if this belongs here, but I'm not sure where else it belongs.  Thoughts?

### DIFF
--- a/tools/analysis/ping-patterns/README.md
+++ b/tools/analysis/ping-patterns/README.md
@@ -1,0 +1,71 @@
+This directory contains a tool to analyse the patterns of baseline and metrics pings received on a per-client basis in Fenix.  These pattern analyses are then summarized in a single plot showing various patterns and issues found in the data.
+
+## Requirements
+
+A recent version of [matplotlib](https://matplotlib.org).
+
+## Collecting input data
+
+The input data is created using the following Redash query:
+
+```sql
+SELECT
+  client_info.client_id,
+  DATE(submission_timestamp) AS date,
+  'metrics' as ping_type,
+  client_info.app_display_version AS app_version,
+  client_info.telemetry_sdk_build AS telemetry_sdk_build,
+  client_info.android_sdk_version AS sdk,
+  ping_info.start_time AS start_time,
+  ping_info.end_time AS end_time,
+  0 AS duration,
+  ping_info.seq AS seq
+FROM
+  org_mozilla_fenix_nightly.metrics
+WHERE
+  DATE(submission_timestamp) > "2019-11-01"
+UNION ALL
+SELECT
+  client_info.client_id,
+  DATE(submission_timestamp) AS date,
+  'baseline' as ping_type,
+  client_info.app_display_version AS app_version,
+  client_info.telemetry_sdk_build AS telemetry_sdk_build,
+  client_info.android_sdk_version AS sdk,
+  ping_info.start_time AS start_time,
+  ping_info.end_time AS end_time,
+  metrics.timespan.glean_baseline_duration.value AS duration,
+  ping_info.seq AS seq
+FROM
+  org_mozilla_fenix_nightly.baseline
+WHERE
+  DATE(submission_timestamp) > "2019-11-01"
+```
+
+This query is also available [here](https://sql.telemetry.mozilla.org/queries/66682).
+
+Save the result of the query as a `.csv` file to use as input to this script.
+
+## Configuring the script
+
+Configuration is performed by editing the `config.py` script.  There are comments as to what the fields do there.
+
+## Running the script
+
+To run the script, pass the input dataset and the output directory on the commandline.
+
+```bash
+$ ./ping-patterns.py dataset.csv plots
+```
+
+## Output
+
+The output directory will contain one `.svg` file per client with a timeline of the baseline and metrics pings received from that client.
+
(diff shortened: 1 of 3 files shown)


tools/analysis/ping-patterns/README.md
```diff
@@ -0,0 +1,71 @@
+This directory contains a tool to analyse the patterns of baseline and metrics pings received on a per-client basis in Fenix.  These pattern analyses are then summarized in a single plot showing various patterns and issues found in the data.
```
Dexterp37: Is this really Fenix specific? Does it really make a difference if we're able to change the table name and use it, for example, to check Firefox Reality?

*Update* oh yes, I just figured out that it's due to the build id to date thing.

tools/analysis/ping-patterns/config.py
```diff
@@ -0,0 +1,26 @@
+"""
```
Dexterp37: This requires a license header.

tools/analysis/ping-patterns/ping-patterns.py
```diff
@@ -0,0 +1,499 @@
+#!/usr/bin/env python3
```
Dexterp37: This requires a license header.

tools/analysis/ping-patterns/README.md
```diff
@@ -0,0 +1,71 @@
...
+
+The output directory will contain one `.svg` file per client with a timeline of the baseline and metrics pings received from that client.
+
+Baseline pings are in blue.  The thick part of the line represents the active session as returned in the `baseline.duration` metric. 
+
+Metrics pings are in red.  Issues found with metrics pings are notated with a number to the right.  Hover over the number to display a tooltip with further information about the issue.
+
+Grey vertical lines are midnight local time.  Dashed grey vertical lines are 04:00 local time.
+
+Green vertical lines indicate the first ping coming from an interesting revision that contains a related fix.
```
Dexterp37: It would really be helpful to walk through an example: I found it hard at the beginning, but things became more clear when you walked me through it.

Dexterp37: @mdboom would you think this would be helpful? If so, I think it's ok doing this in a follow-up PR

Dexterp37:
> 
> 
> I don't know if this belongs here, but I'm not sure where else it belongs. Thoughts?

I don't like the idea of having Fenix specific things here, to be fair. We do have a team shared space: would it be fair to share it there instead?

tools/analysis/ping-patterns/README.md
mdboom: Yeah -- I think a worked example would be helpful.  Just need to put that together.

mdboom:
> > I don't know if this belongs here, but I'm not sure where else it belongs. Thoughts?
> 
> I don't like the idea of having Fenix specific things here, to be fair. We do have a team shared space: would it be fair to share it there instead?

What's the team shared space?  I don't really have a strong preference about where this lives as long as it's not "just on my computer" :)

Dexterp37:
> > I don't like the idea of having Fenix specific things here, to be fair. We do have a team shared space: would it be fair to share it there instead?
> 
> What's the team shared space? I don't really have a strong preference about where this lives as long as it's not "just on my computer" :)

Whoops, I just noticed you were not on it. I sent you an email.

badboy:
Counter argument to the shared team drive: Having it here makes it more discoverable for people outside of our team.

Another place for this might be docs.telemetry.mozilla.org (at least for the README.md content)

Dexterp37:
> 
> 
> Counter argument to the shared team drive: Having it here makes it more discoverable for people outside of our team.
> 
> Another place for this might be docs.telemetry.mozilla.org (at least for the README.md content)

For the record: a better place could be the Fenix repo. Since, in the current form, this would hardly be used outside of this team, I'd still recommend putting this in a shared team drive.

mdboom:
I'm not a huge fan of using Google Drive for code given the lack of version control etc.

I also don't think it belongs in Fenix -- it's just the version parsing that is Fenix specific, and there's a lot of ways this would be useful on other Glean-using products by modifying just that one function.

Maybe what we want is a new Github repo for analysis scripts like this?  Or maybe even just a gist (that we linked to from some central place) would be sufficient.  I agree these things are kind of "ephemeral" -- we're not making any guarantees we would maintain them as the rest of the system evolves, etc., which is why I'm not sure that *this* repo is the right one.

Dexterp37:
> I also don't think it belongs in Fenix -- it's just the version parsing that is Fenix specific, and there's a lot of ways this would be useful on other Glean-using products by modifying just that one function.

All right, let's merge this as it is. Can you please file a follow-up for abstracting away the version parsing stuff for making it useful for other products as well?

> Maybe what we want is a new Github repo for analysis scripts like this? Or maybe even just a gist (that we linked to from some central place) would be sufficient. I agree these things are kind of "ephemeral" -- we're not making any guarantees we would maintain them as the rest of the system evolves, etc., which is why I'm not sure that _this_ repo is the right one.

We've been on the gist road before: it suddenly became very complicated to manage.

Since I'm not able to suggest better options, let's stick to this and figure out next steps in the future, if needed.

badboy:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1607430

PR #568, opened 2019-12-11
Add glean registry files as `inputs` to the generation tasks

Dexterp37:
This ensures that, whenever a change is made to a registry file, docs and specific metrics APIs are regenerated accordingly.

This was tested by adding metrics to both the sample app and glean-core's registry files, then just building using the "Android Studio" hammer button. Doc changes were picked up and the APIs were correctly regenerated.

Without this change, sometimes, the code wouldn't be regenerated.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,9 +5,11 @@
 * Python bindings:
 
   * Support for events was added.
-  
-* The Glean Gradle Plugin correctly triggers docs and API updates when registry files
-  change, without requiring them to be deleted.
+
+* All platforms
+
+  * The Glean Gradle Plugin correctly triggers docs and API updates when registry files
+    change, without requiring them to be deleted.
 
 # v22.0.0 (2019-12-05)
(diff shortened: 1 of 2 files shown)


Dexterp37:
Implementation note: calling `inputs.file(s)` multiple times should be safe and it should not delete what was previously registered. When building, we get the correct `Glean SDK - generating docs for ...` message for all the expected files.

CHANGELOG.md
```diff
@@ -5,6 +5,9 @@
 * Python bindings:
 
   * Support for events was added.
+  
+* The Glean Gradle Plugin correctly triggers docs and API updates when registry files
```
mdboom: Maybe put this in an `Android` bullet point, since it only affects Android...?

Dexterp37:
> 
> 
> Looks good and makes sense. Now we just have to get the plugin distributed to our users :)

That's the "easy part", right? :-P

PR #569, opened 2019-12-12
Merge pull request #569 from travis79/Bug1591094-iOS-Timeout

travis79:


glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -153,15 +157,27 @@ class Dispatchers {
...
+            timeout.signal()
+        }
 
-        self.serialOperationQueue.addOperations(preInitOperations, waitUntilFinished: testingMode)
+        // Await the async task to complete up to the allowed time and log an error if it
+        // times out.
+        if timeout.wait(timeout: DispatchTime.now() + Constants.queueProcessingTimeout) == .timedOut {
+            logger.error("Timeout processing initial tasks queue")
+            // Turn off queuing to allow for normal background execution mode
+            queueInitialTasks.value = false
```
travis79: I had to duplicate this line here, otherwise the recording of the error on the next line would get lost when the queue was dumped.

glean-core/metrics.yaml
```diff
@@ -244,3 +244,16 @@ glean.error:
...
+    type: boolean
+    description:
+      Set to true if the tasks that are queued prior to Glean initialization
+      time out.
+    bugs:
+      - https://bugzilla.mozilla.org/1575848
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1575848#c5
+    notification_emails:
+      - telemetry-client-dev@mozilla.com
```
Dexterp37: Since you're touching this, any chance you could change it to the glean mailing list instead?

travis79: Yes, looks like all of them are pointing at this address, want me to change everything?

travis79: All addresses updated to `glean-team@mozilla.com`

glean-core/metrics.yaml
```diff
@@ -238,9 +238,22 @@ glean.error:
...
+
+  preinit_tasks_timeout:
+    type: boolean
+    description:
+      Set to true if the tasks that are queued prior to Glean initialization
+      time out.
+    bugs:
+      - https://bugzilla.mozilla.org/1575848
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1575848#c5
```
badboy: We kinda extend the population for this metric. This means it requires a new data review.

Dexterp37: Good call!

travis79: [The original review](https://bug1575848.bmoattachments.org/attachment.cgi?id=9104882) already states the population as:
```
7) What populations will you measure?
All products using Glean.
```

Having checked with @chutten via IRC, moving the metric only extends the mechanical population and not the approved population so we should be covered by the original review.

PR #571, opened 2019-12-12
Introduce the `InvalidOverflow` error

Dexterp37:
This is used by the timing distribution metric type whenever a passed sample is bigger than 10 minutes.

I wonder if this could be used in other metrics as well. I can't seem to find an immediate use for this outside of timing distributions.

### DIFF
--- a/docs/user/collected-metrics/metrics.md
+++ b/docs/user/collected-metrics/metrics.md
@@ -20,6 +20,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.error.invalid_label |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of times a metric was set with an invalid label. The labels are the `category.name` identifier of the metric. |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5)||never |
+| glean.error.invalid_overflow |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of times a metric was set a value that overflew. The labels are the `category.name` identifier of the metric. |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1591912#c3)||never |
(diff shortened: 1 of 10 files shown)


glean-core/src/metrics/timing_distribution.rs
```diff
@@ -226,10 +226,11 @@ impl TimingDistributionMetric {
     /// ## Notes
     ///
     /// Discards any negative value in `samples` and report an `ErrorType::InvalidValue`
-    /// for each of them.
+    /// for each of them. Reports an `ErrorType::InvalidOverflow` error for samples that
+    /// are longer than `MAX_SAMPLE_TIME`.
     pub fn accumulate_samples_signed(&mut self, glean: &Glean, samples: Vec<i64>) {
         let mut num_negative_samples = 0;
-        let mut num_too_log_samples = 0;
```
badboy: nice typo

docs/user/collected-metrics/metrics.md
```diff
@@ -21,6 +21,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.error.invalid_label |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of times a metric was set with an invalid label. The labels are the `category.name` identifier of the metric. |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5)||never |
+| glean.error.invalid_overflow |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of times a metric was set a value that overflew. The labels are the `category.name` identifier of the metric. |[1](TODO)||never |
```
Dexterp37: Note: this **requires** new data review. I'm currently creating the request and will update the bug accordingly.

glean-core/src/metrics/timing_distribution.rs
travis79: Maybe the number was too logarithmic?

Dexterp37: Hah :-D

PR #575, opened 2019-12-13
Don't send pings if upload is disabled

badboy:


glean-core/src/lib.rs
```diff
@@ -419,6 +419,10 @@ impl Glean {
     /// Returns true if a ping was assembled and queued, false otherwise.
     /// Returns an error if collecting or writing the ping to disk failed.
     pub fn send_ping(&self, ping: &PingType) -> Result<bool> {
+        if !self.is_upload_enabled() {
```
Dexterp37: Does it mean we can get rid of [this](https://github.com/mozilla/glean/blob/e6d59e1c21f7c288f6daa3b39c1e5b814cc759ca/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L532-L535) (and other platform specific checks)?

This would allow us to have "message level parity" across platforms and consolidate the behaviour in the Rust core.

badboy: Conceptually we should be able to get rid of them.
Code-wise it allows us to avoid the string copies when we know it's unnecessary

Dexterp37: > Code-wise it allows us to avoid the string copies when we know it's unnecessary

Given the frequency at which this happens, I'm not sure this is a performance hit. However, I have no strong opinion on this.

Any chance you could add a log message like the one we have in Kotlin/Swift here?

badboy: Will do!

badboy: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1603743

PR #581, opened 2019-12-17
Add the `invalid_overflow` error to Timing Distributions

Dexterp37:
This backports the changes from master to this branch, since we don't want to do a major version bump in A-C just before holidays.

### DIFF
--- a/docs/user/collected-metrics/metrics.md
+++ b/docs/user/collected-metrics/metrics.md
@@ -20,6 +20,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.error.invalid_label |[labeled_counter](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html) |Counts the number of times a metric was set with an invalid label. The labels are the `category.name` identifier of the metric. |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1499761#c5)||never |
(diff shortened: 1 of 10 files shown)


badboy:
Uhm, why 21.4.0 though?
We have some changes that would be good to have in fenix (specifically #545), meaning I'd rather have 22.1.0
(plus the fact that then 21.4.0 has feature changes that 22.0.0 does _not_ have seems very confusing to me)

Dexterp37:
> 
> 
> Uhm, why 21.4.0 though?
> We have some changes that would be good to have in fenix (specifically #545), meaning I'd rather have 22.1.0

Because Fenix, as A-C, is currently on [21.3.0](https://github.com/mozilla-mobile/android-components/blob/b823fece137983c3e94681eb83a84c5ea7626e17/buildSrc/src/main/java/Dependencies.kt#L32). 

Moving them to 22.0.0 is a bit risky, just before the holidays.

> (plus the fact that then 21.4.0 has feature changes that 22.0.0 does _not_ have seems very confusing to me)

I don't think we'll update A-C or Fenix to 22.0.0, we might just release another minor version once we come back.

badboy:
> Moving them to 22.0.0 is a bit risky, just before the holidays.

IMO 22.0.0 was a mistake and that should have been 21.4.0 already. I don't see a breaking change in there.

Dexterp37:
> 
> 
> > Moving them to 22.0.0 is a bit risky, just before the holidays.
> 
> IMO 22.0.0 was a mistake and that should have been 21.4.0 already. I don't see a breaking change in there.

All right, making this 22.1.0.

PR #582, opened 2019-12-17
Document debug pings and the HTTP header

badboy:


docs/dev/core/internal/debug-pings.md
```diff
@@ -0,0 +1,15 @@
...
+
+```
+X-Debug-ID: <tag>
+```
+
+`<tag>` is a alphanumeric string with a maximum length of 20 characters, used to identify pings in the Debug Ping Viewer.
+
+See [Debugging products using the Glean SDK](/user/debugging/index.md) for detailed information how to use this mechanism in applications.
+
+[debug-ping-viewer]: https://glean-debug-view-dev-237806.firebaseapp.com/
```
travis79: Should we mention this is currently only available to those with a Mozilla login?

badboy:
> Maybe we should link to this from the `tagPings` command [here](https://mozilla.github.io/glean/book/user/debugging/android.html) and in the same iOS information as well?

Hm, I'm unsure about that. It's for different audiences (folks using Glean and folks working on Glean itself)

mdboom:
> > Maybe we should link to this from the `tagPings` command [here](https://mozilla.github.io/glean/book/user/debugging/android.html) and in the same iOS information as well?
> 
> Hm, I'm unsure about that. It's for different audiences (folks using Glean and folks working on Glean itself)

Ok.  Fair enough.

PR #591, opened 2019-12-18
Stable serialization

badboy:


glean-core/src/event_database/mod.rs
```diff
@@ -21,7 +21,7 @@ use crate::Glean;
 use crate::Result;
 
 /// Represents the data for a single event.
-#[derive(Debug, Clone, Deserialize, Serialize)]
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
```
Dexterp37: What is `PartialEq` useful for? (trying to learn a bit more)

glean-core/src/metrics/experiment.rs
```diff
@@ -218,3 +217,60 @@ impl ExperimentMetric {
...
+                             97, 32, 107, 101, 121, 7, 0, 0, 0, 0, 0, 0, 0, 97,
+                             32, 118, 97, 108, 117, 101];
+
+
+        let mut data = HashMap::new();
+        data.insert("a key".to_string(), "a value".to_string());
+        let experiment_data = RecordedExperimentData { branch: "branch".into(), extra: Some(data), };
+
+        // We can't actually decode old experiment data.
+        // Luckily Glean never stored this in the database.
```
Dexterp37: For our future selves, I would add "at least before this commit".

glean-core/src/event_database/mod.rs
badboy: [PartialEq](https://doc.rust-lang.org/std/cmp/trait.PartialEq.html) give us `==` comparison, and [Eq](https://doc.rust-lang.org/std/cmp/trait.Eq.html) gives us that it's also reflexive (`a == a`).
This is required for the `assert_eq` test below.

PR #592, opened 2019-12-18
Allow to set client info data in glean-preview

badboy:
This would allow setting all required information for `client_info` in glean-preview.
This means `glean::initialize` in glean-preview now has a slightly different API compared to our Android and iOS implementations.
There are reasons for that:
* Our Rust implementation is different
	* Platform vs language
	* The Rust API fundamentally doesn't know everything about its environment, yet this data is required as early as possible.
* It exists right now only to support the FOGotype.
* It's useful as a discussion ground 

Notable: this does not require us to implement any specific metric API in glean-preview just yet, we defer to the underlying glean-core API.


glean-core/preview/src/core_metrics.rs
```diff
@@ -0,0 +1,113 @@
...
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use glean_core::{metrics::StringMetric, CommonMetricData, Lifetime};
+
+/// Metrics included in every ping as `client_info`.
+#[derive(Debug)]
+pub struct ClientInfoMetrics {
+    /// The build identifier generated by the CI system (e.g. "1234/A").
+    pub app_build: String,
```
Dexterp37: For other languages, this is coming as part of the configuration. We can try to be consistent with that as a beginning step, and then see if things need to be broken out differently.

glean-core/preview/src/core_metrics.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The user visible version string (e.g. "1.0.3").
+    pub app_display_version: String,
+    /// The channel the application is being distributed on.
+    pub app_channel: Option<String>,
```
Dexterp37: For other languages, this is coming as part of the configuration. We can try to be consistent with that as a beginning step, and then see if things need to be broken out differently.

glean-core/preview/src/core_metrics.rs
```diff
@@ -0,0 +1,113 @@
...
+    /// The name of the operating system.
+    pub os: String,
+    /// The user-visible version of the operating system (e.g. "1.2.3").
+    pub os_version: String,
+    /// The manufacturer of the device the application is running on.
+    pub device_manufacturer: String,
```
Dexterp37: Does this and the following even apply to Desktop?

glean-core/preview/src/core_metrics.rs
```diff
@@ -0,0 +1,113 @@
...
+#[derive(Debug)]
+pub struct ClientInfoMetrics {
+    /// The build identifier generated by the CI system (e.g. "1234/A").
+    pub app_build: String,
+    /// The user visible version string (e.g. "1.0.3").
+    pub app_display_version: String,
+    /// The channel the application is being distributed on.
+    pub app_channel: Option<String>,
+    /// The name of the operating system.
+    pub os: String,
```
Dexterp37: Can this be an internal implementation detail computed with `#[cfg(target_os = "windows")]`?

glean-core/preview/src/core_metrics.rs
mdboom: The information exists and you [can get it](https://github.com/mozilla/glean/blob/master/glean-core/python/glean/hardware.py).  Whether we want to is a separate question.

Dexterp37: > 
> 
> The information exists and you [can get it](https://github.com/mozilla/glean/blob/master/glean-core/python/glean/hardware.py). Whether we want to is a separate question.

Ah right! I'd be for stubbing out these values and add proper detection out of the prototype phase.

glean-core/preview/src/core_metrics.rs
chutten: What do you mean by "part of the configuration" in this case? As in part of the Configuration struct?

glean-core/preview/src/core_metrics.rs
chutten: Another question is whether this information is meaningful on Desktop (where, e.g., I could add HW components to wildly change how my manufactured model works compared to others).

glean-core/preview/src/core_metrics.rs
badboy: Oh right, I forgot that the channel might be passed in.

glean-core/preview/src/core_metrics.rs
badboy: Oh yes, the data of course does exist.
But what chutten said and also Firefox might have the right ways to detect this (or similar information) already, which is why I wouldn't want to have yet-another-way to read this out.

Unfortunately `device_manufacturer` _is_ a required field in the schema (to my knowledge), which is why I added it in the first place.

glean-core/preview/src/core_metrics.rs
badboy: We could, and if we wanted to support all the targets Rust supports we could use https://docs.rs/platforms/0.2.1/platforms/target/enum.OS.html to make it easy.

glean-core/preview/src/core_metrics.rs
Dexterp37: > 
> Unfortunately `device_manufacturer` _is_ a required field in the schema (to my knowledge), which is why I added it in the first place.

Since this is a prototype ping, we could just stub it out as we did for our other POCs, if @chutten agrees.

glean-core/preview/src/core_metrics.rs
Dexterp37: > 
> 
> We could, and if we wanted to support all the targets Rust supports we could use https://docs.rs/platforms/0.2.1/platforms/target/enum.OS.html to make it easy.

If we can use the RUST APIs on all the Rust targets, then this is the way to go, IMHO :)

badboy: If we wanted we could certainly move detection of that field into glean-core.

badboy: (putting this into glean-preview _for now_ would allow us to update m-c before the holidays and only cut a proper glean release when we're all back)

Dexterp37: Yes, it's fine to put this in glean-preview and we can then figure out if/when to move to glean-core. I have a slight preference toward stubbing out things vs providing new API surface (even if it needs to be scraped in the future), but I have no strong opinion on this. Your call!

glean-core/preview/src/core_metrics.rs
Dexterp37: Yes

badboy:
OS is detected, `hardware_*` is stubbed out ("unknown"), channel moved to the configuration object (same as other implementations), leaving only app_build and app_display_version to be passed in (and the C++ side should know these).

PR #594, opened 2019-12-18
Metrics recorded during init must not get into overdue `metrics` pings

Dexterp37:
This attempts to fix the problem of components/products recording metrics while or before Glean is still initializing and then flowing into overdue startup `metrics` pings that get generated.

Note that the current behaviour, without this fix, is flaky:

- We're always collecting & sending a `metrics` ping at startup if *any* component or product records a metric and a ping is overdue (not already sent on the current calendar day and we're after 4AM);
- Metrics collected during early startup will get into the generated overdue `metrics` ping, violating the principle of "measurement windows", letting data from a future measurement window fall into the ping related to a previous measurement window.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -519,7 +519,8 @@ open class GleanInternalAPI internal constructor () {
      * policies.
      *
      * If the ping currently contains no content, it will not be assembled and
-     * queued for sending.
+     * queued for sending, unless explicitly specified otherwise in the registry
+     * file.
      *
      * @param pingNames List of ping names to submit.
      * @return The async [Job] performing the work of assembling the ping
@@ -540,7 +541,8 @@ open class GleanInternalAPI internal constructor () {
      * policies.
      *
      * If the ping currently contains no content, it will not be assembled and
-     * queued for sending.
+     * queued for sending, unless explicitly specified otherwise in the registry
+     * file.
      *
      * @param pingNames List of ping names to submit.
      */
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -199,7 +199,7 @@ open class GleanInternalAPI internal constructor () {
             applicationContext,
             acMetadata.metricsPingLastSentDate
         )
-        metricsPingScheduler.schedule()
+        metricsPingScheduler.schedule(overduePingAsFirst = true)
 
         // Signal Dispatcher that init is complete
         @Suppress("EXPERIMENTAL_API_USAGE")
@@ -526,14 +526,33 @@ open class GleanInternalAPI internal constructor () {
      */
     @Suppress("EXPERIMENTAL_API_USAGE")
     internal fun submitPingsByName(pingNames: List<String>) = Dispatchers.API.launch {
+        submitPingsByNameSync(pingNames)
+    }
+
+    /**
+     * Collect and submit a list of pings for eventual upload by name, synchronously.
+     *
+     * Each ping will be looked up in the known instances of [PingType]. If the
+     * ping isn't known, an error is logged and the ping isn't queued for uploading.
+     *
+     * The ping content is assembled as soon as possible, but upload is not
+     * guaranteed to happen immediately, as that depends on the upload
+     * policies.
+     *
+     * If the ping currently contains no content, it will not be assembled and
+     * queued for sending.
+     *
+     * @param pingNames List of ping names to submit.
+     */
+    internal fun submitPingsByNameSync(pingNames: List<String>) {
         if (!isInitialized()) {
             Log.e(LOG_TAG, "Glean must be initialized before submitting pings.")
-            return@launch
+            return
         }
 
         if (!getUploadEnabled()) {
             Log.e(LOG_TAG, "Glean must be enabled before submitting pings.")
-            return@launch
+            return
         }
 
         val pingArray = StringArray(pingNames.toTypedArray(), "utf-8")
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
@@ -198,8 +198,12 @@ internal class MetricsPingScheduler(
     /**
      * Performs startup checks to decide when to schedule the next metrics ping
      * collection.
+     *
+     * @param overduePingAsFirst if this value is `true`, and Dispatchers tasks are
+     *        being enqueued due to startup, collect any overdue `metrics` at the
+     *        beginning of the queue.
      */
-    fun schedule() {
+    fun schedule(overduePingAsFirst: Boolean) {
         val now = getCalendarInstance()
         val lastSentDate = getLastCollectedDate()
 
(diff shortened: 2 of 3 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -28,7 +30,7 @@ internal object Dispatchers {
         private var queueInitialTasks = AtomicBoolean(true)
 
         // Use a [ConcurrentLinkedQueue] to take advantage of it's thread safety and no locking
-        internal val taskQueue: ConcurrentLinkedQueue<suspend CoroutineScope.() -> Unit> = ConcurrentLinkedQueue()
+        internal var taskQueue: TaskQueueType = LinkedList()
```
travis79: Are you just changing this to the LinkedList in order to be able to insert a task at the head?  I think that's a bad idea as the [documentation for LinkedList](https://developer.android.com/reference/java/util/LinkedList) states that:
> Note that this implementation is not synchronized.

I think that bypassing the queue like @mdboom suggests is a better solution that doesn't require us to use a non-threadsafe list.

travis79: There is also a [`ConcurrentLinkedDeque`](https://developer.android.com/reference/java/util/concurrent/ConcurrentLinkedDeque) which would allow for inserting at the head of the queue and still makes guarantees about concurrency.

Dexterp37: > 
> 
> There is also a [`ConcurrentLinkedDeque`](https://developer.android.com/reference/java/util/concurrent/ConcurrentLinkedDeque) which would allow for inserting at the head of the queue and still makes guarantees about concurrency.

Ah, I didn't see that. However, since we're not really using any concurrency feature already (as stated in the commit message), this is hardly of any uses. We're only paying the cost for determining the size of the queue each time we add a new element, as highlighted in their docs 

> `Beware that, unlike in most collections, the size method is NOT a constant-time operation. `

I'm happy with dropping this change, if the `executeTask` works with this as intented.

Dexterp37:
> 
> 
> Could we instead just use `executeTask`, as we already do for the `events` ping, which adds the task to the "real" Dispatcher, not the queue (and I think would have the same effect...)

Thanks for suggesting this, it made me go back to square 1 and think again/test a bit more. Turns out, both this approach and `executeTask` would still fail to order things properly, out of test mode, due to the nested use of the dispatchers. See [this POC](https://gist.github.com/Dexterp37/ba5fa31ad647d9fee89d5e921d698768).

Here's what's happening:

- We have a first dispatch [in the MPS](https://github.com/mozilla/glean/blob/7aa59365bc1019e777640134af2df94f686cd462/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt#L237) (this becomes "executeTask")
- It immediately calls [`Pings.metrics.submit()`](https://github.com/mozilla/glean/blob/7aa59365bc1019e777640134af2df94f686cd462/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt#L259)
- Which ends up calling [`Dispatchers.API.launch`](https://github.com/mozilla/glean/blob/7aa59365bc1019e777640134af2df94f686cd462/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L528) (**nested dispatcher evil!**)
- The second dispatched task gets executed *AT THE END* of the list, after the collected metrics previously enqueued tasks are executed.

I think, to really solve this, we might need to call `submitPingsByName` directly from the metrics ping scheduler, synchronously, avoiding the double dispatch and use that together with `executeTask`.

We might also probably want to find where double dispatch is happening and try to flag/prevent that (follow-up bug?)

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
travis79: No problem, I'll be glad when this metrics scheduling is all straightened out.

travis79:
> I think, to really solve this, we might need to call `submitPingsByName` directly from the metrics ping scheduler, synchronously, avoiding the double dispatch and use that together with `executeTask`.

I agree with this approach. I was just reconsidering this and I think that you are right, that this will probably be the best way.

mdboom:
> > I think, to really solve this, we might need to call `submitPingsByName` directly from the metrics ping scheduler, synchronously, avoiding the double dispatch and use that together with `executeTask`.
> 
> I agree with this approach. I was just reconsidering this and I think that you are right, that this will probably be the best way.

Yeah, +1 from me too.  Sorry I didn't think of it when I first suggested `executeTask`.

Dexterp37:
> 
> 
> > > I think, to really solve this, we might need to call `submitPingsByName` directly from the metrics ping scheduler, synchronously, avoiding the double dispatch and use that together with `executeTask`.
> > 
> > 
> > I agree with this approach. I was just reconsidering this and I think that you are right, that this will probably be the best way.
> 
> Yeah, +1 from me too. Sorry I didn't think of it when I first suggested `executeTask`.

That's fine! It didn't occur to me that we were having another problem with my implementation, until I tried to use `executeTask`. Interestingly, it didn't happen in `testMode` :-)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -526,14 +526,33 @@ open class GleanInternalAPI internal constructor () {
...
+     * Collect and submit a list of pings for eventual upload by name, synchronously.
+     *
+     * Each ping will be looked up in the known instances of [PingType]. If the
+     * ping isn't known, an error is logged and the ping isn't queued for uploading.
+     *
+     * The ping content is assembled as soon as possible, but upload is not
+     * guaranteed to happen immediately, as that depends on the upload
+     * policies.
+     *
+     * If the ping currently contains no content, it will not be assembled and
```
badboy: This is, as of now, only half correct (less relevant for the code of this PR though).
We can send empty pings now (`send_if_empty`).
I don't know how to properly phrase it though.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -228,15 +232,31 @@ internal class MetricsPingScheduler(
...
                 // context is that we want to make sure no other metric API adds data before
                 // the ping is collected. All the exposed metrics API dispatch calls to the
                 // engines through the `Dispatchers.API` context, so this ensures we are enqueued
                 // before any other recording API call.
-                @Suppress("EXPERIMENTAL_API_USAGE")
-                Dispatchers.API.launch {
-                    // This addresses (2).
-                    collectPingAndReschedule(now)
+                //
+                // * Do not change `Dispatchers.API.executeTask` to `Dispatchers.API.launch` as
```
badboy: Should we call out explicitly what executeTask does, namely it executes it right away (on the thread pool)?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -254,9 +274,27 @@ internal class MetricsPingScheduler(
...
-        Log.i(LOG_TAG, "Collecting the 'metrics' ping, now = ${now.time}")
-        Pings.metrics.submit()
+    internal fun collectPingAndReschedule(now: Calendar, startupPing: Boolean = false) {
+        Log.i(LOG_TAG, "Collecting the 'metrics' ping, now = ${now.time}, startup = $startupPing")
+        if (startupPing) {
+            // **IMPORTANT**
+            //
+            // During the Glean initialization, we require any metric recording to be
+            // batched up and replayed after any startup metrics ping is sent. To guarantee
+            // that, we dispatch this function from `Dispatchers.API.executaTask`. However,
```
badboy:
```suggestion
            // that, we dispatch this function from `Dispatchers.API.executeTask`. However,
```

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -254,9 +274,27 @@ internal class MetricsPingScheduler(
...
+            // Pings.metrics.submit() ends up calling `Dispatchers.API.launch` again which
+            // will delay the ping collection task after any pending metric recording is
+            // executed, breaking the 'metrics' ping promise of sending a startup 'metrics'
+            // ping only containing data from the previous session.
+            // To prevent that, we synchronously manually dispatch the 'metrics' ping, without
+            // going through our public API.
+            //
+            // * Do not change this line without checking what it implies for the above wall
+            // of text. *
```
badboy: Sounds like something a test should catch

Dexterp37: Yes, the provided test catches that. However I think it's still beneficial to call this out explicitly in the comment, so that we don't end up scratching our heads on "why is the test failing" months from now :)

PR #595, opened 2019-12-19
Fix docs for pings

badboy:


docs/user/pings/custom.md
```diff
@@ -6,6 +6,8 @@ Applications can define metrics that are sent in custom pings. Unlike the built-
 
 Custom pings must be defined in a `pings.yaml` file, which is in the same directory alongside your app's `metrics.yaml` file.
 
+Ping names are limited to lowercase ASCII characters and hyphens and a maximum of 30 characters.
```
travis79: Just a wording nit: 
```suggestion
Ping names are limited to lowercase letters and hyphens with a maximum of 30 characters.
```

docs/user/pings/custom.md
```diff
@@ -58,6 +60,26 @@ override fun onCreate() {
 
 </div>
 
+<div data-lang="Swift" class="tab">
+
+In Swift, this object must be registered with Glean from your startup code
+(such as in your application's `application` method or a function called from that method).
```
travis79: It probably wouldn't hurt to mention _which_ `application` method we recommend, as there are several.
```suggestion
(such as in your application's `UIApplicationDelegate` [`application(_:didFinishLaunchingWithOptions:)`](https://developer.apple.com/documentation/uikit/uiapplicationdelegate/1622921-application) method or a function called from that method).
```

docs/user/pings/custom.md
mdboom: Why remove "ASCII"?  It's probably important to state that you can't use é, for example.

travis79: ASCII `0X8A` == `é`

travis79: Again, just a nit.  I'm fine if `ASCII` stays there, I know what is meant by it and most everyone else would too.

badboy: Side stepped the whole issue by using the [same wording as in other places](https://docs.telemetry.mozilla.org/cookbooks/new_ping.html).

PR #597, opened 2019-12-19
1604819: Make parseISOTimeString faster

mdboom:


glean-core/android/src/main/java/mozilla/telemetry/glean/utils/DateUtils.kt
```diff
@@ -90,6 +100,17 @@ internal fun parseISOTimeString(date: String): Date? {
         date
     }
 
+    DATE_FORMAT_PATTERN_BY_LENGTH.get(correctedDate.length)?.let {
```
Dexterp37: Very smart!

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/DateUtils.kt
```diff
@@ -23,7 +23,17 @@ internal val DATE_FORMAT_PATTERNS = mapOf(
     TimeUnit.Day to "yyyy-MM-ddZ"
 )
 
+// A mapping from the length of the date string to the format that would parse
+// it.
```
badboy: Can we give an example here?
I guess the `Z` can match a literal Z or +0100 or something like that, right?

mdboom: The `Z` can only match `sHHMM` (where here `s` is a sign).  `X` matches the more general case.  https://developer.android.com/reference/java/text/SimpleDateFormat

Adding an example is a good idea.

PR #598, opened 2019-12-20
Let clippy fix our code

badboy:
With the release of Rust 1.40.0 a few new clippy lints were introduced.
For us they only apply to test code.
The following code automatically applied all of then:

  cargo +nightly fix -Z unstable-options --clippy

(This autofix feature requires Rust nightly, which is 2 versions ahead
of stable as always, but the code is still valid for our minimum supported Rust version)


mdboom:
The removal of unnecessary clones is really nice.

PR #599, opened 2019-12-20
Document queueing before Glean is initialized

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1600880

### DIFF
--- a/docs/user/general-api.md
+++ b/docs/user/general-api.md
@@ -1,7 +1,7 @@
 # The General API
 
 Glean has a minimal API available on its top-level `Glean` object.
-This API allows to enable and disable upload, register [custom pings][] and set [experiment data][experiments-api].
+This API allows to enable and disable upload, register [custom pings][custom-pings] and set [experiment data][experiments-api].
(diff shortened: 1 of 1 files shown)


docs/user/general-api.md
```diff
@@ -146,6 +146,13 @@ See the [custom pings documentation](pings/custom.md) about adding custom pings
 
 {{#include ../tab_footer.md}}
 
+## Behavior when uninitialized
+
+Metric recording that happens before Glean is initialized is queued and applied when Glean is initialized.
+To avoid unbounded memory growth the queue is bounded (currently to a maximum of 100 tasks). Further recordings are dropped.
+
+Ping collection and upload as well as the built-in pings are only available after initialization.
```
mdboom: I think this is a bit confusing: you *can* submit a ping prior to initialization, it will just be queued like everything else.

badboy: Ah, that is true. Though built-in pings are not handled, sending a followup.

PR #603, opened 2019-12-20
Set target architecture for glean-preview

badboy:


badboy:
>   testValidateBaselinePing, Asynchronous wait failed: Exceeded timeout of 5 seconds, with unfulfilled expectations: "Completed upload".

Gnah, interkitten again on that. Ignoring that on merge (let's fix THAT one once and for all in the new year)

PR #608, opened 2019-12-31
1602024: Improve the metrics ping scheduler

mdboom:
This uses a simple `Timer`, rather than the WorkManager for the MetricsPingScheduler.

As planned, we can also move the actual scheduling algorithm to Rust, but that is saved for future work.  This is just the simplest change possible to get us off of the WorkManager here.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
@@ -9,25 +9,16 @@ import android.content.SharedPreferences
 import androidx.annotation.VisibleForTesting
 import android.text.format.DateUtils
 import android.util.Log
-import androidx.lifecycle.Lifecycle
-import androidx.lifecycle.LifecycleEventObserver
-import androidx.lifecycle.LifecycleOwner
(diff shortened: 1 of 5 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -318,7 +318,7 @@ internal class MetricsPingScheduler(
  */
 internal class MetricsPingTimerTask(val scheduler: MetricsPingScheduler) : TimerTask() {
     companion object {
-        private const val LOG_TAG = "glean/MetricsPingTimerTask"
+        private const val LOG_TAG = "glean/MetricsPingTimer"
```
travis79: Why change this and not the class name?

mdboom: This is to workaround a weird limitation: the log tag can only be 23 characters :(  The WorkManager-related class that used to be here had the same issue.

mdboom: Ah -- but now that I'm more caffeinated, I think I see what you're getting at -- let's change the class name as well so they match.

PR #609, opened 2019-12-31
Remove app_forceclosed_count

mdboom:
CI is currently broken because the `app_forceclosed_count` metric has expired.  Since that data is no longer being collected, the test for it is failing.

I don't know if we plan to renew it, but if we just want to remove it, this should be enough to get CI going again.


badboy:
deferring to @Dexterp37 for this (and I accidentally removed him instead of me), don't think this is blocking landing anything before his return tomorrow

PR #614, opened 2020-01-03
1601625: Support Python 3.5

mdboom:


glean-core/python/glean/glean.py
```diff
@@ -43,34 +43,33 @@ class Glean:
     """
 
     # The handle to the underlying Rust object
-    _handle: int = 0
+    _handle = 0  # type: int
```
Dexterp37: Out of curiosity, are these type hints used by the linting tools or just useful for devs? (I'm fine with keeping them anyway)

mdboom: They are used by the static type checking tool `mypy`.  It helps catch a few things (but given that it's Python it's not nearly as useful as one would hope).  Python 3.6 added the special syntax for declaring the types of member variables, but in versions prior you have to use a special comment.  Either way, `mypy` can do its job, though.

PR #615, opened 2020-01-06
Release 23.0.0

mdboom:
A couple of complications beyond the usual here:

I merged the `v22.1.0` tag into master -- I don't think this was done earlier, and it fixes the changelog which previously didn't have an entry for that version.

Updated the `prepare_release.sh` script to update the version in `glean-preview`.

### DIFF
--- a/bin/prepare-release.sh
+++ b/bin/prepare-release.sh
@@ -90,6 +90,14 @@ run $SED -i.bak -E \
     "${WORKSPACE_ROOT}/${FILE}"
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
+### GLEAN-PREVIEW ###
+
+FILE=glean-core/preview/Cargo.toml
+run $SED -i.bak -E \
+    -e "/glean-core/!b;n;n;s/version = \"[0-9a-z.-]+\"/version = \"${NEW_VERSION}\"/" \
+    "${WORKSPACE_ROOT}/${FILE}"
+run rm "${WORKSPACE_ROOT}/${FILE}.bak"
+
 ### Update Cargo.lock
 
 cargo update -p glean-core -p glean-ffi
--- a/bin/prepare-release.sh
+++ b/bin/prepare-release.sh
@@ -75,6 +75,8 @@ fi
 
 ### GLEAN-CORE ###
 
+# Update the glean-core version
+
 FILE=glean-core/Cargo.toml
 run $SED -i.bak -E \
     -e "s/^version = \"[0-9a-z.-]+\"/version = \"${NEW_VERSION}\"/" \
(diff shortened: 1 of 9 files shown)


bin/prepare-release.sh
```diff
@@ -90,6 +90,14 @@ run $SED -i.bak -E \
     "${WORKSPACE_ROOT}/${FILE}"
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
+### GLEAN-PREVIEW ###
+
+FILE=glean-core/preview/Cargo.toml
+run $SED -i.bak -E \
+    -e "/glean-core/!b;n;n;s/version = \"[0-9a-z.-]+\"/version = \"${NEW_VERSION}\"/" \
```
badboy: glean-preview is currently at v0.0.4.
That makes it explicitly unstable (`< 1.0.0`, as per semver). I'd like to keep it that way.

badboy: Oh wait, this only changes the version it depends on, right? Then of course, go ahead!

glean-core/metrics.yaml
```diff
@@ -274,3 +274,6 @@ glean.error:
     notification_emails:
       - glean-team@mozilla.com
     expires: never
+      - all_pings
```
badboy: This line should not be there.

bin/prepare-release.sh
mdboom: Yes -- just to update the version of glean-core that glean-preview depends on.

glean-core/metrics.yaml
mdboom: Indeed.  Bad merge.

mdboom:
Let's hold off on merging this until #609 is resolved, which @Dexterp37 can comment on when he gets back.

Dexterp37:
> 
> 
> Let's hold off on merging this until #609 is resolved, which @Dexterp37 can comment on when he gets back.

Thanks for putting this together. I'm fine with cutting a release and I've approved #609

Dexterp37:
> I merged the `v22.1.0` tag into master -- I don't think this was done earlier, and it fixes the changelog which previously didn't have an entry for that version.

@mdboom , is there a missing step [in the docs](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html#new-point-releases)?

bin/prepare-release.sh
Dexterp37: Can we add a comment on what the sed is supposed to do?

mdboom: Sure thing.

mdboom:
> > I merged the `v22.1.0` tag into master -- I don't think this was done earlier, and it fixes the changelog which previously didn't have an entry for that version.
> 
> @mdboom , is there a missing step [in the docs](https://mozilla.github.io/glean/book/dev/cut-a-new-release.html#new-point-releases)?

As best I can tell, I think this release was tagged from a branch that wasn't on master.  Perhaps the missing step to document is that when you do this, you should merge that branch back into master (which is what I did to put this PR together).

PR #619, opened 2020-01-07
1602024: Improve the metrics ping scheduler

mdboom:


CHANGELOG.md
```diff
@@ -15,6 +15,9 @@
     migration and initialization.
   * Metrics with `lifetime: application` are now cleared when the application is started,
     after startup Glean SDK pings are generated.
+  * The metrics ping scheduler will now only send metrics pings while the
```
Dexterp37: Should this be in the "unreleased" section?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -101,25 +71,36 @@ internal class MetricsPingScheduler(
...
-        // is more suitable for this task, because of the inherent drifting. See
-        // https://developer.android.com/reference/androidx/work/PeriodicWorkRequest.html
-        // for more details.
-        val workRequest = OneTimeWorkRequestBuilder<MetricsPingWorker>()
-            .addTag(MetricsPingWorker.TAG)
-            .setInitialDelay(millisUntilNextDueTime, AndroidTimeUnit.MILLISECONDS)
-            .build()
+        // Cancel any existing scheduled work. Does not actually cancel a
+        // currently-running task.
+        timer?.cancel()
```
Dexterp37: nit: doesn't really change much but, since we're defining a `cancel()` function, would it make sense to use it here as well?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -205,6 +186,19 @@ internal class MetricsPingScheduler(
      */
     fun schedule(overduePingAsFirst: Boolean) {
```
Dexterp37: We're no longer calling `schedule()` from the lifecycle observer, I think we can get rid of the `overduePingAsFirst` and always assume that `schedule` is called from the Glean init path. We should also change the function docs accordingly.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -36,53 +27,32 @@ import java.util.concurrent.TimeUnit as AndroidTimeUnit
...
         const val DUE_HOUR_OF_THE_DAY = 4
-        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-        internal var isInForeground = false
-        internal var firstForeground = true
-
-        /**
-         * Function to cancel any pending metrics ping workers
-         */
-        internal fun cancel(context: Context) {
-            WorkManager.getInstance(context).cancelUniqueWork(MetricsPingWorker.TAG)
```
Dexterp37: I wonder if we should keep some code in the init to cancel any outstanding unique work: I'm not sure they get cleared automatically. What do you think? Or maybe we could land some metric to check if we find any existing work after the upgrade.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -350,74 +347,25 @@ internal class MetricsPingScheduler(
      */
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal fun getCalendarInstance(): Calendar = Calendar.getInstance()
-
-    /**
-     * Called when lifecycle events are triggered.
-     */
-    override fun onStateChanged(source: LifecycleOwner, event: Lifecycle.Event) {
```
Dexterp37: I love this **SO MUCH**.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
mdboom: Good idea.  This removes even more lines of code!

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
mdboom: From reading the API docs, you can cancel a scheduled timer, but you can't cancel work already in progress (i.e. once the timer has triggered, but before its task function has completed).  I've been careful in this PR to always cancel the timer before scheduling new work, so there will always only be one thing "pending" at any given time.  Is there something else you're suggesting?

Dexterp37: Sorry, I was unclear. I'm not concerned about cancelling the _timers_ we're now introducing. I'm concerned about cancelling any outstanding `WorkManager` job that was enqueued in previous versions of the Glean SDK.

I would not want `WorkManager` jobs to keep living forever. This shouldn't be the case, as they are not recurring jobs, but the `WorkManager` has the habit of surprising us every now and then :)

mdboom: Ah.  That's crazy town, but we know how crazy Android can be :)

I really hope this isn't possible (since it would have to run code from a previous revision of the app, which is hopefully completely gone), but I agree, to be extra sure, we should probably keep something like this in.

mdboom:
Makes sense.  Let's do reason code as a separate PR.  It already has a [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1557048)

PR #620, opened 2020-01-07
1605411 - Reset core client metrics when re-enabling upload

badboy:
We clear all metrics when upload is disabled (glean-core takes care of that),
so we need to reset the initial values only known to the platform when
re-enabling upload.

Now the client info and the channel are passed into the initialize
function of glean_preview.
Before this commit we didn't keep that data around, which means we also
couldn't restore it.
By introducing a small wrapper around the Glean instance we are able to
carry around that data and reuse it when necessary.

This now also adds tests for glean-preview with one big caveat:
Because glean_preview is a global-singleton, we need to run the tests
single-threaded to avoid different tests stomping over each other.
There's no explicit check in the tests, developers will notice problems
by tests just hard crashing on them.

Pass `--test-threads 1` to `cargo test --` or set an environment
variable: `RUST_TEST_THREADS=1`.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -42,7 +42,10 @@ commands:
           command: cargo test --all ---exclude glean-preview --verbose
       - run:
           name: Test glean-preview
-          command: cargo test -p glean-preview -- --test-threads=1
+          command:
+              # Because glean_preview is a global-singleton, we need to run the tests
+              # single-threaded to avoid different tests stomping over each other.
+              cargo test -p glean-preview -- --test-threads=1
 
   install-rustup:
     steps:
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -39,7 +39,10 @@ commands:
           rust-version: <<parameters.rust-version>>
       - run:
           name: Test
-          command: cargo test --all --verbose
+          command: cargo test --all ---exclude glean-preview --verbose
+      - run:
+          name: Test glean-preview
+          command: cargo test -p glean-preview -- --test-threads=1
 
   install-rustup:
     steps:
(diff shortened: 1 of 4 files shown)


.circleci/config.yml
```diff
@@ -39,7 +39,10 @@ commands:
           rust-version: <<parameters.rust-version>>
       - run:
           name: Test
-          command: cargo test --all --verbose
+          command: cargo test --all ---exclude glean-preview --verbose
+      - run:
+          name: Test glean-preview
+          command: cargo test -p glean-preview -- --test-threads=1
```
mdboom: Should we add a comment for why this has to be single-threaded?

glean-core/preview/src/lib.rs
```diff
@@ -75,46 +82,65 @@ where
...
-    let cfg = glean_core::Configuration {
+    let core_cfg = glean_core::Configuration {
         upload_enabled: cfg.upload_enabled,
-        data_path: cfg.data_path,
-        application_id: cfg.application_id,
+        data_path: cfg.data_path.clone(),
+        application_id: cfg.application_id.clone(),
         max_events: cfg.max_events,
         delay_ping_lifetime_io: cfg.delay_ping_lifetime_io,
     };
```
chutten: There's no adorably-short rustism for partially initing from another instance? Some sort of 
```rust
let core_cfg = glean_core::Configuration{data_path: cfg.data_path.clone(), ..cfg}; ?
```

Ah-ha, I knew I knew it from somewhere. [Struct update syntax](https://doc.rust-lang.org/book/ch05-01-defining-structs.html#creating-instances-from-other-instances-with-struct-update-syntax) should work here, right?

badboy: That only works if its of the same type (and cfg is a `glean_preview::Configuration`, not a `glean_core::Configuration`).

I was slightly annoyed by this as well, but given that it's all internal, it didn't bother me enough.

PR #621, opened 2020-01-07
Bug 1595771 - Break long pings up for logcat

travis79:


travis79:
> This is sufficiently complex that we should probably add a test. Particularly for breaking up messages of a non-multiple-of-4000.
> 
> I can't think of a way around this, but it's annoying that the output will no longer be copy-and-pastable. I find that really helpful to paste into an actual .json file to poke around at what's in there. (No action on that, just my grumbling...)

I agree, I was working on refactoring the breaking apart code into a function to make it easier to unit test so check back on the next commit.

Theoretically, you could reconstruct the file, it just won't be as easy since you will have to trim the LogCat headers out as you are stitching it back together.  Right now, you can't copy and paste it anyways since the JSON would be incomplete.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
```diff
@@ -36,7 +85,16 @@ class BaseUploader(d: PingUploader) : PingUploader by d {
             val json = JSONObject(data)
             val indented = json.toString(2)
 
-            Log.d(LOG_TAG, "Glean ping to URL: $path\n$indented")
+            if (indented.length + path.length <= MAX_LOG_PAYLOAD_SIZE_BYTES) {
```
Dexterp37: Do you really need this special case? Wouldn't the function still print only one chunk without this?

glean-core/android/src/test/java/mozilla/telemetry/glean/net/BaseUploaderTest.kt
```diff
@@ -95,4 +96,43 @@ class BaseUploaderTest {
...
+        for (chunk in chunks) {
+            val calculatedHeaderMsg =
+                "Glean ping to URL: $testPath [Part ${curChunk + 1} of ${chunks.size}]"
+            assertEquals(
+                "First line must contain correct header message",
+                calculatedHeaderMsg,
+                chunk.lines()[0]
+            )
+
+            if (chunk === chunks.last()) {
```
Dexterp37: Can we please also check the content of each chunk?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
travis79: It would, we would just get the "[Part 1 of 1]" message when it wasn't necessary.  I want to say that I did this more for a performance reason by short-circuiting to keep from having to create the list and do all the other (rather simple) calculations.

In the interest of simplifying things, I can remove this.

glean-core/android/src/test/java/mozilla/telemetry/glean/net/BaseUploaderTest.kt
travis79: Sure!

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
Dexterp37: You could also add a comment here stating the reason why we're "special casing" :) I'd be fine with that :)

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
```diff
@@ -22,6 +22,57 @@ import java.util.TimeZone
...
+        internal const val MAX_LOG_PAYLOAD_SIZE_BYTES = 4000
+
+        /**
+         * Function used to break apart large pings into an array of "chunks" that are compatible with
+         * Logcat's max payload size
+         */
+        @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
+        internal fun splitPingForLog(pingData: String, path: String): List<String> {
+            // Calculate the total number of chunks
+            var chunkCount = pingData.length / MAX_LOG_PAYLOAD_SIZE_BYTES
```
Dexterp37: Mh, wait a second: I randomly stumbled upon [`chunked`](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin.text/chunked.html), which seems to do exactly this. Maybe we can just use that out of the box?

PR #624, opened 2020-01-08
Remove data migration code

Dexterp37:
One of the most satisfying PR I've worked on.


Dexterp37:
> 
> 
> IIRC, there are also some date conversion functions in `DateUtils.kt` that are only needed by migration. If I'm right, maybe can remove those as well? (Since the Java date stuff is so problematic, if nothing else).

Oh, good catch indeed. I thought ktlint would warn for this, but it isn't. Android Studio tells me, though :-)

mdboom:
:fire:

PR #628, opened 2020-01-08
Bug 1591976 - Add the `uploadEnabled` parameter to `Glean.initialize`

travis79:


travis79:
> We should do this for Python as well. Up to you whether to do it here or create a new bug for that (and feel free to assign it to me).

😱 I thought I did!  I admit my Python skills are limited and rusty...

mdboom:
> > We should do this for Python as well. Up to you whether to do it here or create a new bug for that (and feel free to assign it to me).
> 
> scream I thought I did! I admit my Python skills are limited and rusty...

I think it was just missed in the docs (and then I missed the actually Python implementation). :+1: on the Python implementation here.

docs/user/general-api.md
```diff
@@ -174,7 +160,6 @@ When re-enabling, core Glean metrics will be recomputed at that time.
 <div data-lang="Swift" class="tab">
 
 `Glean.shared.setUploadEnabled()` should be called in response to the user enabling or disabling telemetry.
-This method should also be called at least once prior to calling `Glean.shared.initialize()`.
```
mdboom: Also remove this line from the Python section below.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -102,15 +102,19 @@ open class GleanInternalAPI internal constructor () {
...
     @JvmOverloads
     @Synchronized
     @MainThread
     fun initialize(
         applicationContext: Context,
-        configuration: Configuration = Configuration()
+        configuration: Configuration = Configuration(),
+        uploadEnabled: Boolean
     ) {
+        this.uploadEnabled = uploadEnabled
```
Dexterp37: I know the behaviour is basically the same, but can we please call `setUploadEnabled(uploadEnabled)` here instead?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -102,15 +102,19 @@ open class GleanInternalAPI internal constructor () {
...
      */
     @Suppress("ReturnCount", "LongMethod", "ComplexMethod")
     @JvmOverloads
     @Synchronized
     @MainThread
     fun initialize(
         applicationContext: Context,
-        configuration: Configuration = Configuration()
+        configuration: Configuration = Configuration(),
+        uploadEnabled: Boolean
```
Dexterp37: Since we're not making `uploadEnabled` optional, I think we need to move it before `configuration`: otherwise we make it mandatory to provide `Configuration()` even when not required by the embedder.

glean-core/ios/Glean/Glean.swift
```diff
@@ -60,7 +60,8 @@ public class Glean {
     ///
     /// - parameters:
     ///     * configuration: A Glean `Configuration` object with global settings.
-    public func initialize(configuration: Configuration = Configuration()) {
+    public func initialize(configuration: Configuration = Configuration(),
```
Dexterp37: Maybe the same reasoning for Android applies here as well?

samples/ios/app/glean-sample-app.xcodeproj/xcshareddata/xcschemes/glean-sample-app.xcscheme
```diff
@@ -0,0 +1,78 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<Scheme
```
Dexterp37: Mh, why this new file? Why was it missing before?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: It's not mandatory, you just have to use the label.  I don't have any preference to the order though and you are right, it's usually better to put parameters without defaults before those that have them (thinking of languages where that's a hard rule)

samples/ios/app/glean-sample-app.xcodeproj/xcshareddata/xcschemes/glean-sample-app.xcscheme
travis79: It apparently was..  I'm not sure where the previous one when but it's just a Xcode file that tells Xcodebuild which build configurations to use.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -102,13 +102,15 @@ open class GleanInternalAPI internal constructor () {
      * @param applicationContext [Context] to access application features, such
      * as shared preferences
      * @param configuration A Glean [Configuration] object with global settings.
+     * @param uploadEnabled A [Boolean] that determines the initial state of the uploader
```
Dexterp37: nit: since we moved it in the function signature, let's move it before `configuration` here as well

travis79:
Dang it, I forgot to commit that last change :/

PR #630, opened 2020-01-09
1607730 - Try fixing iOS integration test interkittens

badboy:
It's basically: Queue ping submission behind other metric operations.

I _think_ this might actually be the cause for both problems (the "crash" and the timeout).

Please see individual commit messages for detailed explanations.

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -225,10 +225,8 @@ public class Glean {
     /// If the ping currently contains no content, it will not be assembled and
     /// queued for sending.
     func submitPingsByName(pingNames: [String]) {
-        // This runs in the background to allow for processing of pings off of the
-        // main thread. Please note that the ping uploader will spawn other async
-        // operations if there are pings to upload.
-        Dispatchers.shared.launchConcurrent {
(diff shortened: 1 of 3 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -225,10 +225,8 @@ public class Glean {
     /// If the ping currently contains no content, it will not be assembled and
     /// queued for sending.
     func submitPingsByName(pingNames: [String]) {
-        // This runs in the background to allow for processing of pings off of the
-        // main thread. Please note that the ping uploader will spawn other async
-        // operations if there are pings to upload.
-        Dispatchers.shared.launchConcurrent {
+        // Queue submitting the ping behind all other metric operations to include them in the ping
+        Dispatchers.shared.launchAPI {
```
Dexterp37: Great catch. Whether or not this fixes the interkitten, I think we should keep it in, in line with [what we're doing for Android](https://github.com/mozilla/glean/blob/a7f366e7d5b13c51ce4b03a4a30f34a48e146cbc/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L513).

We do have a test to check this on the Android side, I think.

PR #631, opened 2020-01-09
Use anchors/tags to reduce duplication

mdboom:
This should have zero effect on what this does -- it's just a duplication-removal effort.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,5 +1,13 @@
 version: 2.1
 
+definitions:
+  - release_filters: &release-filters
+      branches:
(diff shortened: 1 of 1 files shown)


mdboom:
> Nice one!

Er... let me actually make it work first :)

.circleci/config.yml
```diff
@@ -1,5 +1,13 @@
 version: 2.1
 
+definitions:
+  - release_filters: &release-filters
```
badboy: I keep forgetting how powerful yaml is. It's mostly for the better that I don't know.

PR #633, opened 2020-01-09
Add locale metric to metrics ping

brizental:
Fixes [Bug 1601489](https://bugzilla.mozilla.org/show_bug.cgi?id=1601489)

This is how this will look in the metrics ping: 

```json
   {
      "client_info": {
        "android_sdk_version": "28",
        "app_build": "0",
        "app_display_version": "glean.version.name",
        "architecture": "armeabi-v7a",
        "client_id": "41dabd38-1b93-473b-8bc1-0e4284a65c53",
        "device_manufacturer": "unknown",
        "device_model": "robolectric",
        "first_run_date": "2020-01-09+01:00",
        "os": "Android",
        "os_version": "9",
        "telemetry_sdk_build": "23.0.1"
      },
      "metrics": {
        "string": {
          "glean.baseline.locale": "en-US"
        }
      },
      "ping_info": {
        "end_time": "2020-01-09T17:24+01:00",
        "ping_type": "metrics",
        "seq": 0,
        "start_time": "2020-01-09T17:24+01:00"
      }
    }
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,12 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v24.1.0...master)
 
+* General:
+  * Add `locale` to `client_info` section.
+  * **Deprecation Warning** Since `locale` is now in the `client_info` section, the one
+    in the baseline ping ([`glean.baseline.locale`](https://github.com/mozilla/glean/blob/c261205d6e84d2ab39c50003a8ffc3bd2b763768/glean-core/metrics.yaml#L28-L42))
+    is redundant and will be removed by the end of the quarter.
+
 # v24.1.0 (2020-01-16)
 
 [Full changelog](https://github.com/mozilla/glean/compare/v24.0.0...v24.1.0)
--- a/docs/user/pings/index.md
+++ b/docs/user/pings/index.md
@@ -52,6 +52,7 @@ Optional fields are marked accordingly.
 | `os_version` | String | The user-visible version of the operating system (e.g. "1.2.3") |
 | `android_sdk_version` | String | *Optional*. The Android specific SDK version of the software running on this hardware device (e.g. "23") |
 | `telemetry_sdk_build` | String | The version of the Glean SDK |
+| `locale` | String | The locale of the application (e.g. "es-ES") |
 
 All the metrics surviving application restarts (e.g. `client_id`, ...) are removed once the application using the Glean SDK is uninstalled.
 
--- a/glean-core/android/build.gradle
+++ b/glean-core/android/build.gradle
@@ -23,7 +23,7 @@ apply plugin: 'jacoco'
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
  */
-String GLEAN_PING_SCHEMA_GIT_HASH = "a56043b"
+String GLEAN_PING_SCHEMA_GIT_HASH = "f2a7ce4"
 String GLEAN_PING_SCHEMA_URL = "https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$GLEAN_PING_SCHEMA_GIT_HASH/schemas/glean/glean/glean.1.schema.json"
 
 // Set configuration for the glean_parser
--- a/glean-core/android/build.gradle
+++ b/glean-core/android/build.gradle
@@ -22,6 +22,10 @@ apply plugin: 'jacoco'
  * This defines the location of the JSON schema used to validate the pings
  * created during unit testing.
  * This uses a specific version of the schema identified by a git commit hash.
+ *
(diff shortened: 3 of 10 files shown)


brizental:
Hm, the tests failing here are tests such as `sending an empty ping doesn't queue work` and that is happening, because now Glean adds the locale to the metrics ping and even if the client doesn't add any metrics there, there will always be the locale there and so it never will be empty.

I could of course add an exception and say that if there is only locale in the metrics ping that means it is still empty, but that is not the prettiest solution. Second option would be to circle back and reconsider adding locale to the client_info.

Thoughts?

badboy:
The naming issue is indeed unfortunate. I defer a decision on that to Dexter.
Always triggering a metrics ping doesn't sound _too_ bad to me, I expect there to be mostly always other metric data in there (totally unproven though, I could be totally wrong and we're holding back sending a lot of empty metrics pings).

Re the test: that one can be fixed by using an adhoc-created custom ping, that's definitely gonna be empty then.

Dexterp37:
> The name "glean.baseline.locale" is a bit unfortunate. We could instead make a new metric "l10n.locale" or something, but then it would exist in different names in the baseline and metrics pings. Lots of suboptimal options here... Not sure of the best way forward.

Yeah :\ I wish we named it better back then :( We have a few options.

1. Rename the `glean.baseline.locale` to `l10n.locale` and call it a breaking change. (let's not!)
2. Add  `l10n.locale`  and mark `glean.baseline.locale` as deprecated in the changelog. Then remove this a the end of the quarter. We should send a PSA email to fx-data-dev announcing this.

I'd lean towards 2.

With respect to "always" sending the 'metrics' ping, this makes it no worse than the current state, already: GeckoView introduced a 'user' lifetime metric that will always make the 'metrics' ping available...

brizental:
After discussions in IRC, we came to the conclusion that the best option was to add this to the client_info. This is the case because then we don't always send the metrics ping. Always sending it turns out to change way more behaviour than expected.

All tests will fail here, because this changed need to be added to `mozilla-pipeline-schemas` first. I opened a Pr for that already: https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/482

I thought it would be best to leave the locale in the baseline ping body as well for now, even though it is redundant, so we can do as suggested be @Dexterp37 and remove that later.

Dexterp37:
> All tests will fail here, because this changes need to be added to `mozilla-pipeline-schemas` first. I opened a PR for that already: [mozilla-services/mozilla-pipeline-schemas#482](https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/482)

One trick here that can let you check your tests is to change [this line](https://github.com/mozilla/glean/blob/a7c16d21b46ed9feada0224dd484e47b4985ad62/glean-core/android/build.gradle#L27) to point to your mozilla-pipeline-schma branch ;-) 

> I thought it would be best to leave the locale in the baseline ping body as well for now, even though it is redundant, so we can do as suggested be @Dexterp37 and remove that later.

That's fine. Please do file a bug about removing that.

brizental:
> One trick here that can let you check your tests is to change this line to point to your mozilla-pipeline-schma branch ;-)

I've done this locally and all tests pass :) 

One more thing, just for illustration purposes, this is what the new client_info will look like. I have some manual testing locally too:

```json
"client_info": {
  "android_sdk_version": "25",
  "app_build": "1",
  "app_display_version": "1.0",
  "architecture": "x86_64",
  "client_id": "43359b2d-30f5-4bdf-a2f6-cbc956bcfa9a",
  "device_manufacturer": "unknown",
  "device_model": "Android SDK built for x86_64",
  "first_run_date": "2020-01-15+00:00",
  "locale": "en-US",
  "os": "Android",
  "os_version": "7.1.1",
  "telemetry_sdk_build": "24.0.0"
}
```

brizental:
Now blocke by: https://bugzilla.mozilla.org/show_bug.cgi?id=1609892

brizental:
Unblocked by 64ea5bf :) If all turns out green in CI I will just add the changes to the changelog and this should be ready for final review.

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v24.1.0...master)
 
+* General:
+  * Add `locale` to `client_info` section.
+  * **Deprecation Warning** Since `locale` is now in the `client_info` section, the one
+    in the baseline ping is redundant and will be removed by the end of the quarter.
```
Dexterp37: nit: please use the name of the metric, which is `glean.baseline.locale`. You can also add a [permalink to it](https://github.com/mozilla/glean/blob/c261205d6e84d2ab39c50003a8ffc3bd2b763768/glean-core/metrics.yaml#L28-L42)

PR #634, opened 2020-01-09
Automatically flush upload WorkManager jobs when using the `GleanTestLocalServer` rule

Dexterp37:
This PR changes the upload work units so that, when the `GleanTestLocalServer` rule is used in integration tests, no need for external hacks is required to force the `WorkManager` to run the jobs.

The underlying problem is that, when running in test mode, the `WorkManager` does not automatically run its jobs and there's no way to tell "just behave as in production". Moreover, we're forced to set the `WorkManager` in test mode, otherwise tests crash at startup.

**How to test**
The related PR in Android Components (mozilla-mobile/android-components#5549) was used to test this, by performing a [local substitution](https://mozilla.github.io/glean/book/dev/howtos/development-with-android-components.html) of the Glean SDK with this changes in A-C.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -86,6 +86,10 @@ open class GleanInternalAPI internal constructor () {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal var isMainProcess: Boolean? = null
 
+    // When sending pings to localhost, we're probably in instrumented tests. In that
+    // case pings are to be immediately submitted by the WorkManager,
+    internal var isSendingToLocalhost: Boolean = false
+
     /**
      * Initialize the Glean SDK.
      *
@@ -603,13 +607,16 @@ open class GleanInternalAPI internal constructor () {
      * TEST ONLY FUNCTION.
      * Sets the server endpoint to a local address for ingesting test pings.
      *
-     * The endpoint will be set as "http://localhost:<port>".
+     * The endpoint will be set as "http://localhost:<port>" and pings will be
+     * immediately sent by the WorkManager.
      *
      * @param port the local address to send pings to
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     internal fun testSetLocalEndpoint(port: Int) {
         Glean.enableTestingMode()
+
+        isSendingToLocalhost = true
 
         // We can't set the configuration unless we're initialized.
         assert(isInitialized())
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -88,7 +88,7 @@ open class GleanInternalAPI internal constructor () {
 
     // When sending pings to localhost, we're probably in instrumented tests. In that
     // case pings are to be immediately submitted by the WorkManager,
-    internal var isSendingToLocalhost: Boolean = false
+    internal var isSendingToTestEndpoint: Boolean = false
 
     /**
      * Initialize the Glean SDK.
@@ -616,7 +616,7 @@ open class GleanInternalAPI internal constructor () {
     internal fun testSetLocalEndpoint(port: Int) {
         Glean.enableTestingMode()
 
-        isSendingToLocalhost = true
+        isSendingToTestEndpoint = true
 
         // We can't set the configuration unless we're initialized.
         assert(isInitialized())
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -86,8 +86,8 @@ open class GleanInternalAPI internal constructor () {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal var isMainProcess: Boolean? = null
 
-    // When sending pings to localhost, we're probably in instrumented tests. In that
-    // case pings are to be immediately submitted by the WorkManager,
+    // When sending pings to a test endpoint, we're probably in instrumented tests. In that
+    // case pings are to be immediately submitted by the WorkManager.
     internal var isSendingToTestEndpoint: Boolean = false
 
     /**
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -151,7 +151,8 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                 buildWorkRequest<PingUploadWorker>(PING_WORKER_TAG)
             )
 
-            // TODO: only do th9is in test mode
+            // Only flush pings immediately if sending to a test endpoint,
+            // which means we're probably in instrumented tests.
             if (Glean.isSendingToTestEndpoint) {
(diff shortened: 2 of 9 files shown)


mdboom:
Hopefully a rebase on upstream/master will make the Python tests go green, but who knows...

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -87,6 +87,10 @@ open class GleanInternalAPI internal constructor () {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal var isMainProcess: Boolean? = null
 
+    // When sending pings to localhost, we're probably in instrumented tests. In that
+    // case pings are to be immediately submitted by the WorkManager,
+    internal var isSendingToLocalhost: Boolean = false
```
mdboom: I worry (a bit) about the name here, since if the user set the endpoint to localhost through the normal means (not `testSetLocalEndpoint`) it wouldn't be true.  Maybe `isSendingToTestEndpoint`?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -86,6 +86,10 @@ open class GleanInternalAPI internal constructor () {
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
     internal var isMainProcess: Boolean? = null
 
+    // When sending pings to localhost, we're probably in instrumented tests. In that
```
badboy: The comment here should be updated

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -86,6 +86,10 @@ open class GleanInternalAPI internal constructor () {
...
+    // case pings are to be immediately submitted by the WorkManager,
```
badboy: stray comma at the end should be a dot.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -147,7 +148,13 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
             WorkManager.getInstance(context).enqueueUniqueWork(
                 PING_WORKER_TAG,
                 ExistingWorkPolicy.KEEP,
-                buildWorkRequest<PingUploadWorker>(PING_WORKER_TAG))
+                buildWorkRequest<PingUploadWorker>(PING_WORKER_TAG)
+            )
+
+            // TODO: only do th9is in test mode
```
badboy: typo.
Will this todo be fixed here or in a follow-up? If so, it should get a bug filed.

glean-core/android/src/main/java/mozilla/telemetry/glean/testing/GleanTestLocalServer.kt
```diff
@@ -31,12 +35,23 @@ import org.junit.runner.Description
...
     private val localPort: Int
 ) : TestWatcher() {
     /**
      * Invoked when a test is about to start.
      */
     override fun starting(description: Description?) {
         Glean.testSetLocalEndpoint(localPort)
+
+        val config = Configuration.Builder()
+            // Use a single thread executor, we cannot make background upload
```
badboy: I'm not sure I understand how a single-thread executor means its not on the main thread?
Is this "single thread" separate from where it is started?

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/WorkManagerUtils.kt
```diff
@@ -0,0 +1,41 @@
...
+
+import android.content.Context
+import androidx.work.WorkInfo
+import androidx.work.WorkManager
+import androidx.work.testing.WorkManagerTestInitHelper
+import kotlinx.coroutines.runBlocking
+import kotlinx.coroutines.withTimeout
+
+/**
+ * TEST ONLY FUNCTION.
```
badboy: Does this need a `VisibleForTesting` annotation?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
Dexterp37: This TODO was already fixed, sorry, it was a stray comment. `isSendingToTestEndpoint` is already only set when using the Glean test rule.

glean-core/android/src/main/java/mozilla/telemetry/glean/testing/GleanTestLocalServer.kt
Dexterp37: Good point. I tried to rephrase the comment a bit to clarify. The problem is that the default test executor from the WorkManager runs things on the main thread, and we don't want that, as networking would crash the test. Instead we just use the simplest possible executor: a pool with 1 thread.

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/WorkManagerUtils.kt
Dexterp37: Unfortunately, no. While the function ends up triggered only during tests, it's not used directly only in tests. Instead, it's being used in the uploader classes, which are non-test classes.
One thing we could do, though, is make this `internal` only, as it's only meant to be used by our code.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/SharedTestUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+import java.util.concurrent.TimeUnit
+
+/**
+ * Waits for ping with the given name to be received
+ * in the test ping server.
+ *
+ * @param pingName the name of the ping to wait for
+ * @param maxAttempts the maximum number of attempts
+ */
+fun waitForPingContent(
```
brizental: I was doing something exactly like this to fix the android tests for #633 

I wonder if this should live in the test utils for glean core and not only samples.

Dexterp37: > I wonder if this should live in the test utils for glean core and not only samples.

Good call. @brizental would it be ok for you if I filed that as a follow-up bug?

brizental: Of coruse, no worries. I might fix that very quickly as a part of #633, if you don't mind :)

Dexterp37: I don't :)

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/SharedTestUtils.kt
```diff
@@ -0,0 +1,36 @@
...
+    pingName: String,
+    maxAttempts: Int = 3
+): JSONObject?
+{
+    val server = getPingServer()
+
+    var attempts = 0
+    do {
+        attempts += 1
+        val request = server.takeRequest(20L, TimeUnit.SECONDS)
```
travis79: 20 seconds seems like a long time for awaiting networking from a test.

Dexterp37: That's the standard time we've been using throughout the whole codebase: if the ping reaches the server in 100ms, it stops waiting. However, CI can be slow, and waits can be > 5 seconds.

PR #636, opened 2020-01-09
Reorganize and refactor the CircleCI config

mdboom:
This reorganizes the circleci config so that it's hopefully a little easier to manage.

Everything is organized into the following sections: Project, Rust, Android, iOS, Python, Docs

A few things have been refactored into commands that are reused.

Other than that, though, this is just a reorg/refactor and shouldn't change the behavior of the tasks in any way.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -86,9 +86,9 @@ commands:
     steps:
       - checkout
       - install-rustup
-      - install-android-targets
       - setup-rust-toolchain:
           rust-version: stable
+      - install-android-targets
       - run:
           name: Install missing Android SDK
           command: |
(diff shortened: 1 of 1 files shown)


mdboom:
Before merging, we should test the release tags by tagging a dummy release (which we can later delete).

.circleci/config.yml
```diff
@@ -85,7 +87,7 @@ commands:
       - checkout
       - install-rustup
       - install-android-targets
-      - test-setup:
+      - setup-rust-toolchain:
```
badboy: Guess we should probably first prepare the rust toolchain, then install the android targets
(simple fix: just swap these two tasks)

That was already wrong in the old config. It works because the stable toolchain is already installed anyway, but I'd prefer if this would be semantically correct here.

mdboom: Make sense!

PR #637, opened 2020-01-10
Remove failure as a direct dependency

badboy:
We can just implement the error trait ourselves and add some quick
Display formatting. It's not that much more code for us either.


badboy:
failure is mostly that simple (it does some of the conversion and nicety things under the hood, but for all we need it's really that simple.

PR #638, opened 2020-01-10
Update dependencies and remove unused

badboy:
In my quest to get glean into m-c, I'm trying to slim down our dependency tree.
This involves:

1) Updating what we depend on
2) Removing what we don't need (e.g. `color-backtrace`)

This is first and foremost pushed to get a CI run, I want to cross-check that against m-c and check that we're not accidentally cause crate duplication because of a version mismatch.


badboy:
We're good to go, only minor/patch updates, so m-c is just fine.

Dexterp37:
> This is first and foremost pushed to get a CI run, I want to cross-check that against m-c and check that we're not accidentally cause crate duplication because of a version mismatch.

Did you end up testing in m-c?

badboy:
Yup, did a local run of `cargo update -p gkrust-shared && mach vendor rust`. Only minimal changes.

PR #645, opened 2020-01-13
Switch to an auto-assign bot

badboy:
Everyone from the current Glean team will now be added as an reviewer
automatically, but only one per PR.
People can still re-assign of course.

Using https://github.com/kentaro-m/auto-assign/


badboy:
Sorry for the noise. Of course GitHub picks the config from master first and auto-assign will not overwrite that.

PR #650, opened 2020-01-15
Merge pull request #650 from mozilla/upgrade-ffi-support

badboy:


badboy:
oh true, nice that we have these checks :D

badboy:
> Huh, interesting diff for the header files. I do appreciate the standardized spacing.

cbindgen takes whatever docs the Rust part has.
I assume ffi-support changed the docs on the macro recently, which means we're now getting these in as well.

PR #653, opened 2020-01-15
1609149 - Remove `ping_type` from pings.

badboy:
Needs https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/485 to land first.

### DIFF
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
@@ -288,6 +288,9 @@ class MetricsPingSchedulerTest {
 
             // Fetch the ping from the server and decode its JSON body.
             val request = server.takeRequest(20L, AndroidTimeUnit.SECONDS)
+            val docType = request.path.split("/")[3]
+            assertEquals("The received ping must be a 'metrics' ping", "metrics", docType)
+
             val metricsJsonData = request.body.readUtf8()
             val metricsJson = JSONObject(metricsJsonData)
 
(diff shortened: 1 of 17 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -272,8 +272,6 @@ class MetricsPingSchedulerTest {
 
             // Validate the received data.
             checkPingSchema(metricsJson)
-            assertEquals("The received ping must be a 'metrics' ping",
```
Dexterp37: Can we retain the check and use the submission URL here and at the other locations?

badboy:
> The schema change was merged. Does it need to be deployed as well? Is it already live?

I'll check if it is live with the team. This only becomes relevant once this lands in a release used by applications though.

I'm adding a changelog entry, so we don't forget.

glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
badboy: adding it back

badboy:
I added back the checks where it makes sense.
In some parts we were already checking against the URL (in the helper functions), so we can avoid the additional check there.

badboy:
```
java.lang.AssertionError: Task queue is cleared expected:<0> but was:<3>
```

but only for `testReleaseUnitTest`.
Might be a race condition due to the recent changes? Need to look into that further first.

Dexterp37:
> Might be a race condition due to the recent changes? Need to look into that further first.

Maybe due to my async flush?

badboy:
Waiting for reason codes/baseline PRs to land and a release first.

PR #654, opened 2020-01-15
Don't block on dispatching tasks enqueued before init

Dexterp37:
This does not block the main thread when flushing enqueued tasks at initialization. This works and retains the current implemented behaviour, at least in Kotlin, since [this](https://github.com/mozilla/glean/blob/fa6aaac11d8d8eb1b883a085fdcc508dba275a6a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L257) already dispatches `setUploadEnabled` in the APIs threadpool.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
@@ -12,7 +12,6 @@ import kotlinx.coroutines.ObsoleteCoroutinesApi
 import kotlinx.coroutines.launch
 import kotlinx.coroutines.newSingleThreadContext
 import kotlinx.coroutines.runBlocking
-import kotlinx.coroutines.withTimeoutOrNull
 import mozilla.telemetry.glean.GleanMetrics.GleanError
 import java.util.concurrent.ConcurrentLinkedQueue
 import java.util.concurrent.atomic.AtomicBoolean
@@ -38,10 +37,6 @@ internal object Dispatchers {
             // setUploadEnabled(false) is not called so that we don't continue to queue tasks and
             // waste memory.
             const val MAX_QUEUE_SIZE = 100
-
-            // This is the number of milliseconds that are allowed for the initial tasks queue to
-            // process all of the queued tasks.
-            const val QUEUE_PROCESSING_TIMEOUT_MS = 5000L
         }
 
         // The number of items that were added to the queue after it filled up.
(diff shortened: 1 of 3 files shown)


Dexterp37:
@mdboom @travis79 @badboy what do you think?

This would shave off quite some time off the initialization if there's any enqueued task.

travis79:
> @mdboom @travis79 @badboy what do you think?
> 
> This would shave off quite some time off the initialization if there's any enqueued task.

I'm actually toying around with this on the iOS version as we speak to solve a problem with the metrics ping scheduler.

travis79:
This same thing needs done on the iOS side, feel free to file a bug and assign it to me if you want.

Dexterp37:
> 
> 
> We can also remove the `preInitTasksTimeout` metric in `metrics.yaml` here.

Yes, though I'm not sure I can do that in this PR since Swift still uses it. I'd leave it here, suggesting @travis79 or @badboy take the Swift part.

travis79:
> > We can also remove the `preInitTasksTimeout` metric in `metrics.yaml` here.
> 
> Yes, though I'm not sure I can do that in this PR since Swift still uses it. I'd leave it here, suggesting @travis79 or @badboy take the Swift part.

If you don't mind me adding a commit to this PR, I can do it right here!

Dexterp37:
> 
> 
> > > We can also remove the `preInitTasksTimeout` metric in `metrics.yaml` here.
> > 
> > 
> > Yes, though I'm not sure I can do that in this PR since Swift still uses it. I'd leave it here, suggesting @travis79 or @badboy take the Swift part.
> 
> If you don't mind me adding a commit to this PR, I can do it right here!

@travis79 any chance you could do that on a new PR?

travis79:
> > > > We can also remove the `preInitTasksTimeout` metric in `metrics.yaml` here.
> > > 
> > > 
> > > Yes, though I'm not sure I can do that in this PR since Swift still uses it. I'd leave it here, suggesting @travis79 or @badboy take the Swift part.
> > 
> > 
> > If you don't mind me adding a commit to this PR, I can do it right here!
> 
> @travis79 any chance you could do that on a new PR?

Not a problem at all.  Give me 10 minutes and I'll have one up.

badboy:
@Dexterp37 are the TODOs from the initial description resolved?

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -142,29 +133,13 @@ internal object Dispatchers {
...
-                    Log.e(LOG_TAG, "Timeout processing initial tasks queue")
-                    GleanError.preinitTasksTimeout.set(true)
-                    queueInitialTasks.set(false)
-                    taskQueue.clear()
-                } else {
-                    Log.i(LOG_TAG, "Initial tasks queue completed successfully")
+                // This must happen after `queueInitialTasks.set(false)` is run, or it
+                // would be added to a full task queue and be silently dropped.
+                if (overflowCount > 0) {
+                    GleanError.preinitTasksOverflow.addSync(MAX_QUEUE_SIZE + overflowCount)
```
Dexterp37: Note: since the tasks are now dispatched asynchronously, this is now a sync call within the "flush" job.

Dexterp37:
> 
> 
> @Dexterp37 are the TODOs from the initial description resolved?

Yes, some resolved, some do not apply anymore given #656 - that's why I removed them.

Dexterp37:
> 
> 
> This looks good to me! I am not sure whether the test was necessary since it's just testing the features built into coroutines, but it makes for a safety net if we fall into async hell due to some change in the future.

There's one twist here: the test makes sure that we never mistakenly switch back to making it on the main thread :) It's not testing that coroutines work as expected, but rather that our expectations of flushing being performed OTMT stands true

PR #655, opened 2020-01-15
Bug 1586738 - Add Glean-iOS MetricsPingScheduler

travis79:
This adds the MetricsPingScheduler to the iOS bindings.

### DIFF
--- a/glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
+++ b/glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
@@ -1,0 +1,213 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+/// MetricsPingScheduler facilitates scheduling the periodic assembling of metrics pings,
+/// at a given time, trying its best to handle the following cases:
+///
+/// - ping is overdue (due time already passed) for the current calendar day;
+/// - ping is soon to be sent in the current calendar day;
+/// - ping was already sent, and must be scheduled for the next calendar day.
+class MetricsPingScheduler {
+    // This struct is used for organizational purposes to keep the class constants in a single place
+    struct Constants {
+        static let logTag = "glean/MetricsPingSched"
+        static let lastMetricsPingSentDateTime = "last_metrics_ping_iso_datetime"
+        static let dueHourOfTheDay = 4
+        static let lastVersionOfAppUsed = "last_version_of_app_used"
+    }
+
+    private let logger = Logger(tag: Constants.logTag)
+
+    var timer: Timer?
+
+    /// Schedules the metrics ping collection at the due time.
+    ///
+    /// - parameters:
+    ///     * now:  A `Date` representing the current date/time
+    ///     * sendTheNextCalendarDay:  Determines whether to schedule collection for the next calendar day
+    ///                            or to attempt to schedule it for the current calendar day. If the latter and
+    ///                            we're overdue for the expected collection time, the task is scheduled for
+    ///                            immediate execution.
+    func schedulePingCollection(_ now: Date, sendTheNextCalendarDay: Bool) {
+        var fireDate = Calendar.current.date(
+            bySettingHour: Constants.dueHourOfTheDay,
+            minute: 0,
+            second: 0,
+            of: now
+        )!
+
+        // Invalidiate the timer if it's already running
+        timer?.invalidate()
+
+        // Setup timer and schedule
+        if sendTheNextCalendarDay {
+            let tomorrow = Calendar.current.date(
+                byAdding: .day,
+                value: 1,
+                to: fireDate,
+                wrappingComponents: true
+            )!
+            fireDate = tomorrow
+        }
+
+        logger.debug("Scheduling the 'metrics' ping for \(fireDate)")
+        timer = Timer(
+            fireAt: fireDate,
+            interval: 0.0,
+            target: self,
+            selector: #selector(collectPingAndReschedule),
+            userInfo: nil,
+            repeats: false
+        )
+    }
+
+    /// Determines if the application is a differnet version from the last time it was run. This is used to prevent
+    /// mixing data from multiple versions of the application in the same ping.
+    ///
+    /// - returns: `true` if the version is different, `false` if the version is the same.
+    func isDifferentVersion() -> Bool {
+        // Determine if the version has changed since the last time we ran.
+        let currentVersion = AppInfo.displayVersion
+        if let lastVersion = UserDefaults.standard.string(forKey: Constants.lastVersionOfAppUsed) {
+            if currentVersion != lastVersion {
+                UserDefaults.standard.set(currentVersion, forKey: Constants.lastVersionOfAppUsed)
+                return true
+            }
+        }
+        return false
+    }
+
+    /// Check if the provided time is after the ping due time.
+    ///
+    /// - parameters:
+    ///     * now: A `Date` representing the current time
+    ///     * dueHourOfTheDay: An `Int` representing the due hour of the day, in the [0...23] range
+    ///
+    /// - returns: `true` if `now` is past the due hour of the day.
+    func isAfterDueTime(_ now: Date, dueHourOfTheDay: Int = Constants.dueHourOfTheDay) -> Bool {
+        return now > Calendar.current.date(
+            bySettingHour: dueHourOfTheDay,
+            minute: 0,
+            second: 0,
+            of: now
+        )!
+    }
+
+    /// Performs startup checks to decide when to schedule the next metrics ping collection.
+    func schedule() {
+        let now = Date()
+
+        // If the version of the app is different from the last time we ran the app,
+        // schedule the metrics ping for immediate collection. We only need to perform
+        // this check at startup (when overduePingAsFirst is true).
+        if isDifferentVersion() {
+            Dispatchers.shared.serialOperationQueue.addOperation {
+                self.collectPingAndReschedule(now)
+            }
+            return
+        }
+
+        let lastSentDate = getLastCollectedDate()
+        logger.debug("The 'metrics' ping was last sent on \(String(describing: lastSentDate))")
+
+        // We expect to cover 3 cases here:
+        //
+        // (1) - the ping was already collected the current calendar day; only schedule
+        //       one for collecting the next calendar day at the due time;
+        // (2) - the ping was NOT collected the current calendar day, and we're later than
+        //       the due time; collect the ping immediately;
+        // (3) - the ping was NOT collected the current calendar day, but we still have
+        //       some time to the due time; schedule for sending the current calendar day.
+
+        let alreadySentToday = lastSentDate != nil && Calendar.current.isDateInToday(lastSentDate!)
+        if alreadySentToday {
+            // The metrics ping was already sent today. Schedule it for the next
+            // calendar day. This addresses (1).
+            logger.info("The 'metrics' ping was already sent today, \(now).")
+            schedulePingCollection(now, sendTheNextCalendarDay: true)
+        } else if isAfterDueTime(now) {
+            logger.info("The 'metrics' ping is scheduled for immediate collection, \(now)")
+            // **IMPORTANT**
+            //
+            // The reason why we're collecting the "metrics" ping in the `Dispatchers`
+            // queue is that we want to make sure no other metric API adds data before
+            // the ping is collected. All the exposed metrics API dispatch calls to the
+            // engines through the `Dispatchers.API` context, so this ensures we are enqueued
+            // before any other recording API call.
+            //
+            // * Do not change `Dispatchers.shared.serialOperationQueue.addOperation` to
+            // `Dispatchers.shared.launchAPI` as this would break startup overdue ping
+            //  collection. * `addOperation` schedules the task for immediate execution on the
+            // `Dispatchers` serial execution queue, before any other enqueued task. For more
+            // context, see bug 1604861 and the implementation of
+            // `collectPingAndReschedule`.
+            Dispatchers.shared.serialOperationQueue.addOperation {
+                self.collectPingAndReschedule(now, startupPing: true)
+            }
+        } else {
+            // This covers (3).
+            logger.info("The 'metrics' collection is scheduled for today, \(now)")
+            schedulePingCollection(now, sendTheNextCalendarDay: false)
+        }
+    }
+
+    /// Triggers the collection of the "metrics" ping and schedules the next collection.
+    ///
+    /// - parameters:
+    ///     * now: A `Date` representing the current time
+    @objc func collectPingAndReschedule(_ now: Date, startupPing: Bool = false) {
+        logger.info("Collecting the 'metrics' ping, now = \(now), startup = \(startupPing)")
+        if startupPing {
+            // **IMPORTANT**
+            //
+            // During the Glean initialization, we require any metric recording to be
+            // batched up and replayed after any startup metrics ping is sent. To guarantee
+            // that, we dispatch this function from `Dispatchers.API.executeTask`. However,
+            // Pings.metrics.submit() ends up calling `Dispatchers.API.launch` again which
+            // will delay the ping collection task after any pending metric recording is
+            // executed, breaking the 'metrics' ping promise of sending a startup 'metrics'
+            // ping only containing data from the previous session.
+            // To prevent that, we synchronously manually dispatch the 'metrics' ping, without
+            // going through our public API.
+            //
+            // * Do not change this line without checking what it implies for the above wall
+            // of text. *
+            Glean.shared.submitPingsByNameSync(pingNames: ["metrics"])
+        } else {
+            GleanMetrics.Pings.shared.metrics.submit()
+        }
+        // Update the collection date: we don't really care if we have data or not, let's
+        // always update the sent date.
+        updateSentDate(now)
+        // Reschedule the collection.
+        schedulePingCollection(now, sendTheNextCalendarDay: true)
+    }
+
+    /// Get the date the metrics ping was last collected.
+    ///
+    /// - returns: An Optional `Date` representing the date the metrics ping was last collected, or null if no metrics
+    ///            ping was previously collected.
+    func getLastCollectedDate() -> Date? {
+        var lastCollectedDate: Date?
+
+        if let loadedDate = UserDefaults.standard.string(forKey: Constants.lastMetricsPingSentDateTime) {
+            lastCollectedDate = Date.fromISO8601String(dateString: loadedDate, precision: .millisecond)
+        } else {
+            logger.error("MetricsPingScheduler last stored ping time was not valid")
+        }
+
+        return lastCollectedDate
+    }
+
+    /// Update the persisted date when the metrics ping is sent.
+    ///
+    /// - parameters:
+    ///     * date: The `Date` to store.
+    func updateSentDate(_ date: Date = Date()) {
+        UserDefaults.standard.set(
+            date.toISO8601String(precision: .millisecond),
+            forKey: Constants.lastMetricsPingSentDateTime
+        )
+    }
+}
--- a/glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
+++ b/glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
@@ -51,15 +51,17 @@ class MetricsPingScheduler {
             fireDate = tomorrow
         }
 
-        logger.debug("Scheduling the 'metrics' ping for \(fireDate)")
-        timer = Timer(
-            fireAt: fireDate,
-            interval: 0.0,
-            target: self,
-            selector: #selector(collectPingAndReschedule),
-            userInfo: nil,
-            repeats: false
+        logger.debug(
+            "Scheduling the 'metrics' ping for \(fireDate), in \(fireDate - now) seconds."
         )
+
+        // Set the timer to fire at the `fireDate`
+        timer = Timer.scheduledTimer(withTimeInterval: fireDate - now, repeats: false) { _ in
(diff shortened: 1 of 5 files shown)


glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,213 @@
...
+            }
+            return
+        }
+
+        let lastSentDate = getLastCollectedDate()
+        logger.debug("The 'metrics' ping was last sent on \(String(describing: lastSentDate))")
+
+        // We expect to cover 3 cases here:
+        //
+        // (1) - the ping was already collected the current calendar day; only schedule
```
badboy:
```suggestion
        // (1) - the ping was already collected on the current calendar day; only schedule
```

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,213 @@
...
+            //
+            // The reason why we're collecting the "metrics" ping in the `Dispatchers`
+            // queue is that we want to make sure no other metric API adds data before
+            // the ping is collected. All the exposed metrics API dispatch calls to the
+            // engines through the `Dispatchers.API` context, so this ensures we are enqueued
+            // before any other recording API call.
+            //
+            // * Do not change `Dispatchers.shared.serialOperationQueue.addOperation` to
+            // `Dispatchers.shared.launchAPI` as this would break startup overdue ping
+            //  collection. * `addOperation` schedules the task for immediate execution on the
```
badboy:
```suggestion
            //  collection. 
            // * `addOperation` schedules the task for immediate execution on the
```

missing newline here.

badboy: oh wait, was the `*` meant as a highlight of this paragraph? I assumed this should become a list.
Looks weird inline a bit, but I don't care as much.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,213 @@
...
+        // Update the collection date: we don't really care if we have data or not, let's
+        // always update the sent date.
+        updateSentDate(now)
+        // Reschedule the collection.
+        schedulePingCollection(now, sendTheNextCalendarDay: true)
+    }
+
+    /// Get the date the metrics ping was last collected.
+    ///
+    /// - returns: An Optional `Date` representing the date the metrics ping was last collected, or null if no metrics
```
badboy:
```suggestion
    /// - returns: A `Date` representing the date the metrics ping was last collected, or nil if no metrics
```

IMo the optional is implied by the "or nil"

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,331 @@
...
+        fakeDate = Calendar.current.date(from: fakeNow)!
+        XCTAssertFalse(
+            mps.isAfterDueTime(fakeDate, dueHourOfTheDay: 4),
+            "isAfterDueTime must report false before the due time on the same calendar day"
+        )
+        // The same hour
+        fakeNow.year = 2015
+        fakeNow.month = 6
+        fakeNow.day = 11
+        fakeNow.hour = 4
```
badboy: I see below we set minute to 1 later on.
Can we explicitly set it to 0 here to make this clear?

travis79:
> I need to give this another look, especially the tests.
> For a first pass I found a few nitpicky things, but overall this looks like some very solid work already.

Please do check it carefully, the last two tests in the file were especially troublesome due to async hell.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,213 @@
...
+                byAdding: .day,
+                value: 1,
+                to: fireDate,
+                wrappingComponents: true
+            )!
+            fireDate = tomorrow
+        }
+
+        logger.debug("Scheduling the 'metrics' ping for \(fireDate)")
+        timer = Timer(
```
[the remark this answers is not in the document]
travis79: The tolerance is only a feature when you use the repeating Timer, I think.  Let me re-read the docs and I'll add a tolerance factor if it will be used.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,213 @@
...
+    ///                            or to attempt to schedule it for the current calendar day. If the latter and
+    ///                            we're overdue for the expected collection time, the task is scheduled for
+    ///                            immediate execution.
+    func schedulePingCollection(_ now: Date, sendTheNextCalendarDay: Bool) {
+        var fireDate = Calendar.current.date(
+            bySettingHour: Constants.dueHourOfTheDay,
+            minute: 0,
+            second: 0,
+            of: now
+        )!
```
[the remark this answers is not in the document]
travis79: I think that's what the Swift language strives for!

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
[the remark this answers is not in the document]
badboy: Yeah, we might not want to add it right away. Right now we make a promise about 4am, not 4am + random tolerance. Let's keep it that way until we learned how it performs.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
travis79: List is fine with me.  I typically only use the asterisk for second level lists.  This comment was cargo-culted from the Android version so adjusting the formatting for line length may have made it look a bit weird.

glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -36,20 +47,19 @@ extension Date {
...
-        ]
-
         let dateFormatter = DateFormatter()
         dateFormatter.dateFormat = dateFormatPatterns[precision]
         return dateFormatter.date(from: dateString)
     }
+
+    /// Convenience function to convert a Date to an ISO8601 string
+    ///
+    /// - returns: An ISO8601 `String?` representing the current value of the `Date` object
```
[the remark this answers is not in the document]
travis79: You are right, the Android version was returning an optional value here so another case of it bleeding through in my implementation :\

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+
+        // Set our expectation that will be fulfilled by the stub above
+        expectation = expectation(description: "Metrics Ping Received")
+
+        // Initialize Glean the SECOND time: it will send the expected string metric (stored from
+        // the previous run) but must not send the canary string, which would be sent the next time
+        // the "metrics" ping is collected after this one.
+        // Glean.shared.initialize(uploadEnabled: true)
+        Glean.shared.initialize(uploadEnabled: true)
+        wait(for: [expectation!], timeout: TimeInterval(5.0))
```
badboy: In other tests we use `waitForExpectations` and provide an error message:

https://github.com/mozilla/glean/blob/master/samples/ios/app/glean-sample-appUITests/DeletionRequestPingTest.swift#L48-L50

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+        Glean.shared.testDestroyGleanHandle()
+    }
+
+    func testGleanPreservesLifetimeApplicationMetrics() {
+        // Reset Glean and do not start it right away
+        clearPingDirectory()
+        Glean.shared.testDestroyGleanHandle()
+        Dispatchers.shared.setTaskQueuing(enabled: true)
+
+        // Set the last time the "metrics" ping was set to now. This is required for us to not
```
badboy:
```suggestion
        // Set the last time the "metrics" ping was sent to now. This is required for us to not
```

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+        // Create a metric and set its value. We expect this to be sent in the first ping
+        // that gets generated the SECOND time we start Glean.
+        let testMetric = StringMetricType(
+            category: "telemetry",
+            name: "test_applifetime_metric",
+            sendInPings: ["metrics"],
+            lifetime: Lifetime.application,
+            disabled: false
+        )
+        let expectedValue = "I-will-survive!"
```
badboy: I WILL SURVIVE!

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+
+        // Set our expectation that will be fulfilled by the stub above
+        expectation = expectation(description: "Metrics Ping Received")
+
+        // Initialize Glean the SECOND time: it will send the expected string metric (stored from
+        // the previous run) but must not send the canary string, which would be sent the next time
+        // the "metrics" ping is collected after this one.
+        // Glean.shared.initialize(uploadEnabled: true)
+        Glean.shared.resetGlean(clearStores: false)
+        wait(for: [expectation!], timeout: TimeInterval(5.0))
```
badboy: As above.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+
+            // Ensure there is only the expected metric
+            XCTAssertEqual(1, strings?.count, "Must contain only the expected metric")
+
+            // Check the received metric's value against the expected value
+            let receivedValue = strings?["telemetry.test_applifetime_metric"] as? String
+            XCTAssertEqual(expectedValue, receivedValue, "Values must match")
+
+            DispatchQueue.main.async {
+                // let the response get processed before we mark the expectation fulfilled
```
badboy: I wonder if we should do this for our other tests as well 🤔 right now we fulfill, then return an HTTP status.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,214 @@
...
+            logger.info("The 'metrics' collection is scheduled for today, \(now)")
+            schedulePingCollection(now, sendTheNextCalendarDay: false)
+        }
+    }
+
+    /// Triggers the collection of the "metrics" ping and schedules the next collection.
+    ///
+    /// - parameters:
+    ///     * now: A `Date` representing the current time
+    @objc func collectPingAndReschedule(_ now: Date, startupPing: Bool = false) {
```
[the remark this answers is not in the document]
travis79: In Swift, selectors are most commonly found when working with the target/action paradigm that you find in classes such as Timer. The selector tells it who to notify when the timer fires (the target) and what selector should be called (the action). When the target is notified by calling the action function, the timer that fired is passed in so you can see the source of the trigger.  You are right, I'm not sure why it didn't see that the selector didn't match the prototype :/.  And I should be using the `scheduledTimer()` function, also.  Let me rework this to be correct, _and_ add a test to prove it :)

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,214 @@
...
+                byAdding: .day,
+                value: 1,
+                to: fireDate,
+                wrappingComponents: true
+            )!
+            fireDate = tomorrow
+        }
+
+        logger.debug("Scheduling the 'metrics' ping for \(fireDate)")
+        timer = Timer(
```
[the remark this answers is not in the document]
travis79: You are correct, see the other comment about the `selector` for more info.  I need to rework this a bit and add a test.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,214 @@
...
+        )!
+    }
+
+    /// Performs startup checks to decide when to schedule the next metrics ping collection.
+    func schedule() {
+        let now = Date()
+
+        // If the version of the app is different from the last time we ran the app,
+        // schedule the metrics ping for immediate collection. We only need to perform
+        // this check at startup (when overduePingAsFirst is true).
```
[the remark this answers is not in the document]
travis79: > Do we have any way to modify the clock used in the test to e.g. simulate the time hopping a bit better?
> That way we could also test timezones changes more reliably.

We can change timezones but setting the date/time, even for tests seems to be possible only by scripting it from CLI.  [This SO post](https://stackoverflow.com/questions/12628859/is-there-any-way-to-change-the-date-time-as-seen-on-the-ios-simulator-other-tha) was the best I found from a quick search.  Go Apple!

travis79:
> Do we have any way to modify the clock used in the test to e.g. simulate the time hopping a bit better?
> That way we could also test timezones changes more reliably.

We can change timezones but setting the date/time, even for tests seems to be possible only by scripting it from CLI.  [This SO post](https://stackoverflow.com/questions/12628859/is-there-any-way-to-change-the-date-time-as-seen-on-the-ios-simulator-other-tha) was the best I found from a quick search.  Go Apple!

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+        }
+    }
+
+    private func clearPingDirectory() {
+        // Get the ping directory
+        let pingDir = HttpPingUploader(configuration: Configuration())
+            .getOrCreatePingDirectory()
+
+        // Clear the directory to ensure we start fresh
+        // Verify all the files were removed, including the bad ones
```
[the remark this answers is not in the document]
travis79: After I got the underlying issue corrected, it turns out we don't need to do this.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -0,0 +1,334 @@
...
+        fakeNow.month = 6
+        fakeNow.day = 11
+        fakeNow.hour = 3
+        fakeNow.minute = 0
+        fakeDate = Calendar.current.date(from: fakeNow)!
+        XCTAssertFalse(
+            mps.isAfterDueTime(fakeDate, dueHourOfTheDay: 4),
+            "isAfterDueTime must report false before the due time on the same calendar day"
+        )
+        // The same hour
```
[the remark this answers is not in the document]
travis79: The timer could fire a little off of the calculated fire time, I don't know what the degree of accuracy is for it. Fortunately, the timer function isn't going through the `schedule()` function like startup does so it's not going to check what time it is before it fires, it simply collects the ping (if there's anything to collect) and reschedules a timer for the following 4 AM interval.

I questioned why we don't send unless it's > 4, and not >= 4 when I copied the logic from the Android implementation.  I'm happy to file a bug to follow up and discuss if you think we need to handle this differently.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
travis79: 😆

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
travis79: Happy to file a bug if you think it's worth it.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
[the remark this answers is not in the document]
travis79: Test is added, but for whatever reason the timer fires _after_ the wait.  😞 I'll look into it more on Tuesday when I get back.

Dexterp37:
Given @mdboom experience with the Android implementation, I'd recommend he takes at least one high level look at this.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
badboy: Yup, definitel worth filing. It's an easy change.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,216 @@
...
+    ///
+    /// - parameters:
+    ///     * now:  A `Date` representing the current date/time
+    ///     * sendTheNextCalendarDay:  Determines whether to schedule collection for the next calendar day
+    ///                            or to attempt to schedule it for the current calendar day. If the latter and
+    ///                            we're overdue for the expected collection time, the task is scheduled for
+    ///                            immediate execution.
+    func schedulePingCollection(_ now: Date, sendTheNextCalendarDay: Bool) {
+        var fireDate = Calendar.current.date(
+            bySettingHour: Constants.dueHourOfTheDay,
```
mdboom: Weird API.  Was confused by `bySetting` on hour, but not the others, but seems to do the same thing for each of them...

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,216 @@
...
+    /// Performs startup checks to decide when to schedule the next metrics ping collection.
+    func schedule() {
+        let now = Date()
+
+        // If the version of the app is different from the last time we ran the app,
+        // schedule the metrics ping for immediate collection. We only need to perform
+        // this check at startup (when overduePingAsFirst is true).
+        if isDifferentVersion() {
+            Dispatchers.shared.serialOperationQueue.addOperation {
+                self.collectPingAndReschedule(now)
```
mdboom: This should pass `startupPing = true`.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,216 @@
...
+    /// - returns: `true` if the version is different, `false` if the version is the same.
+    func isDifferentVersion() -> Bool {
+        // Determine if the version has changed since the last time we ran.
+        let currentVersion = AppInfo.displayVersion
+        if let lastVersion = UserDefaults.standard.string(forKey: Constants.lastVersionOfAppUsed) {
+            if currentVersion != lastVersion {
+                UserDefaults.standard.set(currentVersion, forKey: Constants.lastVersionOfAppUsed)
+                return true
+            }
+        }
```
mdboom: We should probably sent an update ping if we can't determine the earlier version, to match the logic in Android.  This only sends one if it *can* load the old version *and* it doesn't match.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -0,0 +1,216 @@
...
+            )!
+            fireDate = tomorrow
+        }
+
+        logger.debug(
+            "Scheduling the 'metrics' ping for \(fireDate), in \(fireDate - now) seconds."
+        )
+
+        // Set the timer to fire at the `fireDate`
+        timer = Timer.scheduledTimer(withTimeInterval: fireDate - now, repeats: false) { _ in
```
mdboom: This API just takes standard `Calendar.Date` instances.  This Swift stuff is amazing :)

travis79: The function actually takes a `TimeInterval` which is a time and a time unit. I just added an extension to the `Date` class that allows me to get a `TimeInterval` when I subtract two dates :).

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
travis79: [Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1610516)

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
badboy: Swift (and its predecessor ObjC) have this thing where they combine the first argument "hour" with _what_ it should do, and following arguments don't have that.

It does read weird if you don't know it, but once you do it kinda reads like a sentence (I am _not_ the biggest fan of this convention)

travis79: Swift often strives to read like a sentence, so you get some interesting parameter/function names sometimes.  In this case, you use the `Calendar.current` to generate a date with some adjusted elements.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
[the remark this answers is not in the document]
travis79: FYI: You shouldn't use `DispatchSemaphore` in tests, `XCTExpectation` works much better as it allows background stuff to happen when you `waitForExpectations`

glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -36,20 +47,25 @@ extension Date {
...
+    /// - returns: An ISO8601 `String` representing the current value of the `Date` object
+    func toISO8601String(precision: TimeUnit) -> String {
+        let dateFormatter = DateFormatter()
+        dateFormatter.dateFormat = dateFormatPatterns[precision]
+        return dateFormatter.string(from: self)
+    }
+
+    /// Overloads the operator so that subtraction between two dates results in a TimeInterval representing
+    /// the difference between them
+    static func - (lhs: Date, rhs: Date) -> TimeInterval {
```
badboy: `static func -` ... it _does_ look weird.

PR #656, opened 2020-01-16
Bug 1609052 - Don't block on dispatching tasks

travis79:
This is the iOS counterpart to #654


glean-core/ios/GleanTests/DispatchersTest.swift
```diff
@@ -30,8 +30,9 @@ class DispatchersTest: XCTestCase {
             "Tasks have not executed while in queue"
         )
 
-        // Now trigger the queue to fire the tasks
-        Dispatchers.shared.flushQueuedInitialTasks()
+        // Now trigger the queue to fire the tasks, we must use the `waitUntilFinished`
+        // parameter to prevent async issues.
+        Dispatchers.shared.flushQueuedInitialTasks(waitUntilFinished: true)
```
Dexterp37: Does `waitUntilFinished` basically [block the caller thread](https://developer.apple.com/documentation/foundation/operationqueue/1408358-addoperations)?

If so, that's something we **do not** want to do. Please see the final version of #654

travis79: This is JUST for the test, otherwise the foreground task executes the test assertion before the tasks have executed.

travis79: Basically, I'm using an API feature to await the task rather than introducing some more async'ery to await the task on my own.  The actual application does not await the tasks to complete, it queues them on the background thread and moves on (see the default value for the parameter)

mdboom: Could we assert we are in testing mode if `waitUntilFinished` is `true`?  Barring that, maybe add a docstring comment that it should only be set to `true` in tests...?

travis79: That's a GREAT idea!  Let me add that to the PR.

travis79: I was a little less worried since this is an internal API, but it's good to document it and put up a safeguard.

Dexterp37:
Calling @badboy for a second set of eyes given his Swift expertise.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -156,28 +156,22 @@ class Dispatchers {
...
-            // Add all of the queued operations to the `operationQueue` which will cause them to be
-            // executed serially in the order they were collected.
-            self.serialOperationQueue.addOperations(self.preInitOperations, waitUntilFinished: true)
-            timeout.signal()
+    ///
+    /// - parameters:
+    ///     * waitUntilFinished: **USE FOR TESTING ONLY**  Boolean to force awaiting the tasks to complete
+    ///                          in order to make it easier for tests to deal with async issues.
+    func flushQueuedInitialTasks(waitUntilFinished: Bool = false) {
+        // Assert we are in testing mode if the waitUntilFinished flag is true
```
badboy: This comment is essentially the same as the code and thus could go.

badboy:
Oh, I guess the `preinitTasksTimeout` metric is obsolete now and could be removed as part of this.

travis79:
> Oh, I guess the `preinitTasksTimeout` metric is obsolete now and could be removed as part of this.

There's another bug/PR around that's doing that already (I think).  @Dexterp37?

badboy:
Please add a changelog entry (see hell, ... I mean #666)

Dexterp37:
> 
> 
> > Oh, I guess the `preinitTasksTimeout` metric is obsolete now and could be removed as part of this.
> 
> There's another bug/PR around that's doing that already (I think). @Dexterp37?

No, there's not. I left that in from my PR otherwise we'd break Swift :)

travis79:
> > > Oh, I guess the `preinitTasksTimeout` metric is obsolete now and could be removed as part of this.
> > 
> > 
> > There's another bug/PR around that's doing that already (I think). @Dexterp37?
> 
> No, there's not. I left that in from my PR otherwise we'd break Swift :)

Got it, I've removed it as part of the latest commit.  I've also added recording to the `preinitTasksOverflow`

travis79:
> Please add a changelog entry (see hell, ... I mean #666)

CHANGELOG.md updated

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -156,32 +160,34 @@ class Dispatchers {
...
+            self.preInitOperations,
+            waitUntilFinished: waitUntilFinished
+        )
 
         // Turn off queuing to allow for normal background execution mode
         queueInitialTasks.value = false
 
+        // This must happen after `queueInitialTasks.set(false)` is run, or it
+        // would be added to a full task queue and be silently dropped.
+        if overflowCount > 0 {
```
Dexterp37: If this is added, you should probably also add a test for it and a changelog entry. I'd recommend leaving this out of this PR, and take bug https://bugzilla.mozilla.org/show_bug.cgi?id=1609734 after this :)

travis79: Okay, works for me

PR #657, opened 2020-01-16
1609482: Add instrumentation about overflowing the pre-init task queue

mdboom:


glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -166,6 +175,12 @@ internal object Dispatchers {
         private fun addTaskToQueue(block: suspend CoroutineScope.() -> Unit) {
             if (taskQueue.size >= MAX_QUEUE_SIZE) {
                 Log.e(LOG_TAG, "Exceeded maximum queue size, discarding task")
+
+                // This value ends up in the `preinit_tasks_overflow` metric, but we
+                // can't record directly there, because that would only add
+                // the recording to an already-overflowing task queue and would be
+                // silently dropped.
+                overflowCount += 1
```
travis79: I'm guessing accumulating and then adding them all at once is to give the total queue size and for performance?

mdboom: Yes.  We want to know now just "did it overflow" but "by how much".  And calling the metric record function here doesn't work because the queue is already overflowing, and that metric record would likewise just get dropped, hence the member variable and recording it later...

PR #663, opened 2020-01-20
Report `invalid_state` when stopping non started TimingDistribution

Dexterp37:
This will be reported when stop_and_accumulate() is called without a prior call to start()

### DIFF
--- a/glean-core/tests/timing_distribution.rs
+++ b/glean-core/tests/timing_distribution.rs
@@ -314,3 +314,33 @@ fn large_nanoseconds_values() {
     // Check that we got the right sum and number of samples.
     assert_eq!(val.sum() as u64, time);
 }
+
+#[test]
+fn stopping_non_existing_id_records_an_error() {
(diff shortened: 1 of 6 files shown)


glean-core/tests/timing_distribution.rs
```diff
@@ -314,3 +314,33 @@ fn large_nanoseconds_values() {
...
     assert_eq!(val.sum() as u64, time);
 }
+
+#[test]
+fn stopping_non_existing_id_records_an_error() {
+    let (glean, _t) = new_glean(None);
+
+    let mut metric = TimingDistributionMetric::new(
+        CommonMetricData {
+            name: "non_existing_id".into(),
```
brizental: I don't understand why the naming choice of `non_existing_id` here.

Dexterp37: Because when you stop a timing distribution, it looks for the id of the timer passed in. For this metric we're using a non existing id to trigger an error recording :)

PR #664, opened 2020-01-20
Drop the Glean handle and move state into glean-core

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1602751

This is finally addressing the changes to move Glean's state into the glean-core crate, dropping all use of a "glean handle" through the FFI.

This will ultimatively allow others to use Glean through FFI without relying on our direct platform wrappers (think application-service's use across dynamic library boundaries).

This is a huge change and touches all parts of the code base, so I'd like review from a bigger set of eyes.
I tried structuring the commits to make sense on their own.
The "Remove the glean handle from *" parts are mostly automatic simple code changes, the other commits contain the actual logic changes.

This leaves the user-facing API untouched (Android/iOS consumers will keep working).

### DIFF
--- a/glean-core/ffi/src/boolean.rs
+++ b/glean-core/ffi/src/boolean.rs
@@ -4,7 +4,7 @@
 
 use ffi_support::FfiStr;
 
-use crate::{define_metric, handlemap_ext::HandleMapExtension, GLEAN};
+use crate::{define_metric, handlemap_ext::HandleMapExtension, with_glean_value};
 
 define_metric!(BooleanMetric => BOOLEAN_METRICS {
     new           -> glean_new_boolean_metric(),
@@ -12,8 +12,8 @@ define_metric!(BooleanMetric => BOOLEAN_METRICS {
 });
 
 #[no_mangle]
-pub extern "C" fn glean_boolean_set(glean_handle: u64, metric_id: u64, value: u8) {
-    GLEAN.call_infallible(glean_handle, |glean| {
+pub extern "C" fn glean_boolean_set(metric_id: u64, value: u8) {
+    with_glean_value(|glean| {
         BOOLEAN_METRICS.call_infallible(metric_id, |metric| {
             metric.set(glean, value != 0);
         })
@@ -21,12 +21,8 @@ pub extern "C" fn glean_boolean_set(glean_handle: u64, metric_id: u64, value: u8
 }
 
 #[no_mangle]
-pub extern "C" fn glean_boolean_test_has_value(
-    glean_handle: u64,
-    metric_id: u64,
-    storage_name: FfiStr,
-) -> u8 {
-    GLEAN.call_infallible(glean_handle, |glean| {
+pub extern "C" fn glean_boolean_test_has_value(metric_id: u64, storage_name: FfiStr) -> u8 {
+    with_glean_value(|glean| {
         BOOLEAN_METRICS.call_infallible(metric_id, |metric| {
             metric
                 .test_get_value(glean, storage_name.as_str())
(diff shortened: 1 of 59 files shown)


glean-core/src/lib.rs
```diff
@@ -55,6 +57,26 @@ lazy_static! {
         Uuid::parse_str("c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0").unwrap();
 }
 
+static GLEAN: OnceCell<Mutex<Glean>> = OnceCell::new();
```
Dexterp37: nit: can you add a comment here explaining why we need a global Glean object?

glean-core/src/database/mod.rs
```diff
@@ -16,15 +16,29 @@ use crate::Glean;
 use crate::Lifetime;
 use crate::Result;
 
-#[derive(Debug)]
 pub struct Database {
     rkv: Rkv,
+    user_store: SingleStore,
```
Dexterp37: nit: can we add a comment here mentioning what you said in the commit message?

glean-core/src/database/mod.rs
```diff
@@ -219,13 +243,11 @@ impl Database {
             }
         }
 
-        let store: SingleStore = unwrap_or!(
```
Dexterp37: This looks like a potential perf hit, which we got rid of.

glean-core/python/glean/glean.py
```diff
@@ -41,8 +41,8 @@ class Glean:
     >>> Glean.initialize(application_id="my-app", application_version="0.0.0", upload_enabled=True)
     """
 
-    # The handle to the underlying Rust object
-    _handle = 0  # type: int
+    # Whether Glean was initialized
```
Dexterp37: @mdboom could you please take a look at least at the Python changes?

badboy:
> @badboy did you test this with our instrumented tests as well?

I did forget that we don't run these for Android just yet on CI. I get to it.

badboy:
> One thing that is not completely clear to me is how `Keep the stores open all the time` ([00d45ed](https://github.com/mozilla/glean/commit/00d45ed3bd6a22e068ae4e1939a15d03bcfcb967)) directly relates to "Drop the Glean handle and move state into glean-core". Would you mind explaining that a bit further?

You are right in that this is unrelated. I will pull this out of this PR and submit it as its own (and include more of a reason "why" I did it)

glean-core/ffi/src/boolean.rs
```diff
@@ -4,29 +4,25 @@
 
 use ffi_support::FfiStr;
 
-use crate::{define_metric, handlemap_ext::HandleMapExtension, GLEAN};
+use crate::{define_metric, handlemap_ext::HandleMapExtension, with_glean_value};
```
mdboom: nit: This function name didn't immediately make sense to me.  Maybe `with_glean_singleton`?

glean-core/ffi/src/counter.rs
```diff
@@ -15,29 +15,21 @@ define_metric!(CounterMetric => COUNTER_METRICS {
...
-    storage_name: FfiStr,
-) -> u8 {
-    GLEAN.call_infallible(glean_handle, |glean| {
-        COUNTER_METRICS.call_infallible(metric_id, |metric| {
-            metric
-                .test_get_value(glean, storage_name.as_str())
-                .is_some()
-        })
+pub extern "C" fn glean_counter_test_has_value(metric_id: u64, storage_name: FfiStr) -> u8 {
+    let glean = glean_core::global_glean().lock().unwrap();
```
mdboom: Why the use of this, rather than `with_glean_value` as in `boolean.rs`?

glean-core/ffi/src/lib.rs
```diff
@@ -35,8 +35,48 @@ use from_raw::*;
...
+    let mut error = ffi_support::ExternError::success();
+    let res = ffi_support::abort_on_panic::call_with_result(&mut error, || {
+        let mut glean = glean_core::global_glean().lock().unwrap();
+        callback(&mut glean)
+    });
+    handlemap_ext::log_if_error(error);
+    res
+}
+
+pub(crate) fn with_glean_value<R, F>(callback: F) -> R::Value
```
mdboom: Maybe add some docs about when to use `with_glean` vs. `with_glean_value`?

glean-core/ffi/src/counter.rs
badboy: Oh, thanks for noticing, seems like a left-over.

glean-core/ffi/src/lib.rs
badboy: Docs coming.

badboy:
```
BUILD SUCCESSFUL in 19s
103 actionable tasks: 4 executed, 99 up-to-date
```

Something is happening when I run `./gradlew connectedAndroidTest`, so tests pass :)

PR #665, opened 2020-01-21
Keep the stores open all the time

badboy:
A "store" is a handle to the database.
Glean has 3 different ones and can keep them open throughout its lifetime.

Not frequently re-opening database handles also might be a performance
improvement.

Extracted from #664 because it's unrelated. Not directly related to a filed bug.

### DIFF
--- a/glean-core/benchmark/README.md
+++ b/glean-core/benchmark/README.md
@@ -1,0 +1,38 @@
+# Glean Benchmarks
+
+The `Glean SDK` is a modern approach for a Telemetry library and is part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+
+## Benchmarks
+
+This crates provides simple benchmarks for Glean, based on the [criterion benchmark framework](https://bheisler.github.io/criterion.rs/book/criterion_rs.html).
+The library itself does not contain additional code.
+
+### Available benchmarks
+
+* [`benches/bench_basic.rs`](benches/bench_basic.rs) - Setting metrics and submitting a custom ping
+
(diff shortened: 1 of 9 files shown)


badboy:
We didn't know at submission time, it was mostly a hunch from Dexter.
I did ran a very simplistic benchmark (setting 2 metrics and submitting a ping) and get reliable 5% improvements compared to what's in current master.

```
glean set and submit    time:   [94.437 us 94.502 us 94.577 us]
                        change: [-5.3865% -5.1006% -4.8247%] (p = 0.00 < 0.05)
                        Performance has improved.
```

That is on my Linux machine (on my Mac it's unreliable and significantly slower, milliseconds instead of microseconds)

I'll work on adding the benchmark as part of this as well.

badboy:
So the numbers are unreliable even on my Linux workstation. The benchmark is also _very_ simplistic and nowhere near a real use case.
Nonetheless I included the benchmark now as an extra crate so we can re-run/extend it if needed.

badboy:
holding back on this to investigate what this means for I/O operations.

badboy:
All good, I/O operations unchanged for the startup case.

/cc @travis79, please give it a final review

glean-core/benchmark/README.md
```diff
@@ -0,0 +1,38 @@
...
+cargo bench -p benchmark
+```
+
+This is also available as a `make` task:
+
+```
+make bench-rust
+```
+
+Any additional run compare results to the preceding run.
```
travis79: `compare` -> `compares`

PR #668, opened 2020-01-21
1608585: Fix labeled regex

mdboom:


glean-core/tests/labeled.rs
```diff
@@ -247,6 +247,7 @@ fn dynamic_labels_regex_mimsatch() {
         "1.not_fine",
         "this.$isnotfine",
         "-.not_fine",
+        "this.is_not_fine.2",
```
Dexterp37: Why is this not fine? Is this different from [this](https://github.com/mozilla/glean/pull/668/files#diff-15c101f06a477efbf7196d7571cfe05cR27)?

glean-core/src/metrics/labeled.rs
```diff
@@ -33,7 +33,9 @@ lazy_static! {
     /// * `1.not_fine`
     /// * `this.$isnotfine`
     /// * `-.not_fine`
-    static ref LABEL_REGEX: Regex = Regex::new("^[a-z_][a-z0-9_-]{0,29}(\\.[a-z0-9_-]{0,29})*$").unwrap();
+    static ref LABEL_REGEX: Regex = Regex::new(
+        "^[a-z_][a-z0-9_-]{0,29}(\\.[a-z_][a-z0-9_-]{0,29})*$"
```
Dexterp37: We also need to change the docs [here](https://mozilla.github.io/glean/book/user/metrics/labeled_counters.html#limits), if required.

glean-core/tests/labeled.rs
```diff
@@ -291,7 +292,6 @@ fn dynamic_labels_regex_allowed() {
         "this_is_fine_too",
         "this.is_still_fine",
         "thisisfine",
-        "this.is_fine.2",
```
Dexterp37: This is a bit scary: can we check the new regex against the data from Fenix from the last month? This might affect search data, which uses dynamic labels.

glean-core/tests/labeled.rs
mdboom: BQ may put these labels into a structure, and the names of each element can't begin with a number.  Fortunately, no one is doing this yet.

glean-core/tests/labeled.rs
mdboom: That's a good idea.  I'll run an analysis.

glean-core/tests/labeled.rs
Dexterp37: Ouch :| So would `this.is_not_fine.2` be fine but not `2this.is_not_fine` ?

mdboom: Neither is fine.

glean-core/tests/labeled.rs
mdboom: With [this query](https://sql.telemetry.mozilla.org/queries/67873/), I could get the names of all of the search engine identifiers, and they all parse against the new regex.

mdboom: They all pass.

Dexterp37: Nice!

glean-core/src/metrics/labeled.rs
Dexterp37: @mdboom do the docs require updates (see the previous comment)? If so, we should also update [this page](https://mozilla.github.io/glean/book/user/metrics/labeled_strings.html).

CHANGELOG.md
```diff
@@ -34,6 +34,8 @@
   * The metrics ping will now include `reason` codes that indicate why it was
     submitted.
   * The version of `glean_parser` has been upgraded to 1.17.3
+  * Labels in labeled metrics now have a correct, and slightly stricter, regular expression.
```
Dexterp37: I think this should now be moved above in the changelog

PR #669, opened 2020-01-21
Bug 1609734 - Swift: Adds use of overflow metric when pre-init queue overflows

travis79:


travis79:
@brizental wins the reviewer roulette!  I can walk you through any of this you may have questions about, if that makes it more comfortable for you to review.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -176,6 +185,13 @@ class Dispatchers {
         // Turn off queuing to allow for normal background execution mode
         queueInitialTasks.value = false
 
+        // This must happen after `queueInitialTasks.set(false)` is run, or it
+        // would be added to a full task queue and be silently dropped.
+        if preInitTaskCount > Constants.maxQueueSize {
+            GleanMetrics.GleanError.preinitTasksOverflow.add(preInitTaskCount)
+        }
+        preInitTaskCount = 0
```
brizental: Just out of curiosity, why do you need zero it out here again?

brizental:
> @brizental wins the reviewer roulette! I can walk you through any of this you may have questions about, if that makes it more comfortable for you to review.

Thanks @travis79 ! I only had one question, the rest was pretty straight forward.

glean-core/ios/Glean/Dispatchers.swift
travis79: Mostly for tests which don't necessarily destroy the copy of the `MetricsPingScheduler` held by Glean.  I'll add a comment to help clarify why it was done.

PR #672, opened 2020-01-23
Make Glean init in Kotlin fully async

Dexterp37:
I tested this both using Fenix and the Glean SDK sample app: this seem to work as expected. Accumulated data is correctly accounted for and any pre-init dispatched piece of information is sent.

This PR additionally fixes some weird bugs we had in our testing code that made our tests hang forever under certain circumstances. Se the indivudual commit messages.

**TODO**

- [x] Take this for a manual QA, make sure it doesn't break
- [x] Think a bit more what changing `flushQueuedInitialTasks` implies
- [x] Test this in Fenix

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -15,6 +15,8 @@ import androidx.annotation.VisibleForTesting
 import androidx.lifecycle.ProcessLifecycleOwner
 import com.sun.jna.StringArray
 import kotlinx.coroutines.Job
+import kotlinx.coroutines.MainScope
+import kotlinx.coroutines.launch
 import mozilla.telemetry.glean.config.Configuration
 import mozilla.telemetry.glean.config.FfiConfiguration
 import mozilla.telemetry.glean.utils.getLocaleTag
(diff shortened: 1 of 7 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -135,82 +137,78 @@ open class GleanInternalAPI internal constructor () {
...
-        // At this point, all metrics and events can be recorded.
-        // This should only be called from the main thread. This is enforced by
-        // the @MainThread decorator and the `assertOnUiThread` call.
-        ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+            // At this point, all metrics and events can be recorded.
+            // This should only be called from the main thread. This is enforced by
+            // the @MainThread decorator and the `assertOnUiThread` call.
+            MainScope().launch {
+                ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
+            }
```
mdboom: Do we need to do this here, or could we do it outside of the coroutine (near the top where things like `applicationContext` are set?)

Dexterp37: I'm not sure yet, I don't think there's anything preventing us from moving it there, before the glean init even starts.

The only two reasons for not doing that would be:

- we don't want to register this if the glean init fails;
- we really want to test this in Fenix before making a call!

mdboom: Fair enough on both points.  I think I was just coming at it from the angle of -- going off the main thread only to come back to it is more complicated.  But, as you say, we don't really know until we try it in the context of Fenix.

Dexterp37: I'll make sure to try that first once I get rid of the failing test and have a proper Fenix build for testing!

Dexterp37: I tested that, but I think we should retain the current behaviour even just to make sure that we don't mistakenly register observers if init fails. Thoughts?

mdboom: That's a good point about not registering if we exit early for some other error during init.  :+1: on this, then!

PR #673, opened 2020-01-23
Create upload module with "get ping request" functionality

brizental:
Part of the resolution for [Bug 1605077](https://bugzilla.mozilla.org/show_bug.cgi?id=1605077)

What the changes here try to accomplish:

- Setup initial structure for the upload module.
- Create one of the main functions this module will have. Specifically the one that returns a new `Request` for one of the pending pings.

I thought best to send this incomplete and have different PRs for:

- The function to process a ping upload response.
- Adding interface for the new functionality to the Glean object.

I will also create a different bug for adding logic to deal with the deletion request ping. I am aware this code doesn't handle that at all.

For more details on what this new module will try to accomplish, please refer to:
- [Implementation proposal for moving the ping uploading logic to glean-core](https://docs.google.com/document/d/1YEhrudNl5aHBnht3FCjT66Bj58ckDw3jO-_V56mnqHs/edit?usp=sharing)

PS.: I know this is a lot of code to review. I hope the file structure will help in reading it :)

### DIFF
--- a/glean-core/src/upload/mod.rs
+++ b/glean-core/src/upload/mod.rs
@@ -1,0 +1,192 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Manages the pending pings queue and directory.
+//!
+//! * Keeps track of pending pings, loading any unsent ping from disk on startup;
+//! * Exposes `get_next_ping` API for the platform layer to request next ping in line;
+//! * Exposes `process_ping_upload_response` API to check the HTTP response from the ping upload
+//!   and either delete the corresponding ping from disk or re-enqueue it for sending.
+
+// !IMPORTANT!
+// Remove the next line when this module's functionality is in the Glean object.
+// This is here just to not have lint error for now.
+#![allow(dead_code)]
+
+use std::collections::VecDeque;
+use std::path::PathBuf;
+use std::str::FromStr;
+use std::sync::atomic::{AtomicBool, Ordering};
+use std::sync::{Arc, RwLock};
+use std::thread;
+
+use log;
+use serde_json::Value as JsonValue;
+
+use request::PingRequest;
+use util::process_pings_dir;
+
+mod request;
+mod util;
+
+/// Manages the pending pings queue and directory.
+#[derive(Debug)]
+pub struct PingUploadManager {
+    /// A FIFO queue storing a `PingRequest` for each pending ping.
+    queue: Arc<RwLock<VecDeque<PingRequest>>>,
+    /// A flag, signaling if we are done processing the pending pings directories.
+    processed_pending_pings: Arc<AtomicBool>,
+}
+
+impl PingUploadManager {
+    /// Create a new PingUploadManager.
+    ///
+    /// This will spawn a new thread and processes the pending pings folder
+    /// filling up the queue with whatever pings are in there
+    /// ordered by oldest to newest file modified.
+    pub fn new(data_path: &str) -> Self {
+        let queue = Arc::new(RwLock::new(VecDeque::new()));
+        let processed_pending_pings = Arc::new(AtomicBool::new(false));
+
+        let data_path = PathBuf::from_str(&data_path).expect("data_path must be a valid path.");
+        let local_queue = queue.clone();
+        let local_flag = processed_pending_pings.clone();
+        let _ = thread::Builder::new()
+            .name("glean.upload_manager.process_pings_directory".to_string())
+            .spawn(move || match process_pings_dir(&data_path) {
+                Ok(requests) => {
+                    let mut local_queue = local_queue
+                        .write()
+                        .expect("Can't write to pending pings queue.");
+                    local_queue.extend(requests.into_iter());
+                    local_flag.store(true, Ordering::SeqCst);
+                }
+                Err(e) => log::info!("Error processing pending pings directories! {}", e),
+            });
+
+        Self {
+            queue,
+            processed_pending_pings,
+        }
+    }
+
+    fn has_processed_pending_pings(&self) -> bool {
+        self.processed_pending_pings.load(Ordering::SeqCst)
+    }
+
+    /// Creates a `PingRequest` and adds it to the queue.
+    pub fn enqueue_ping(&self, uuid: &str, url: &str, body: JsonValue) {
+        let mut queue = self
+            .queue
+            .write()
+            .expect("Can't write to pending pings queue.");
+        let request = PingRequest::new(uuid, url, body);
+        queue.push_back(request);
+    }
+
+    /// Clear pending pings queue.
+    pub fn clear_ping_queue(&self) {
+        let mut queue = self
+            .queue
+            .write()
+            .expect("Can't write to pending pings queue.");
+        queue.clear();
+    }
+
+    /// Get the next `PingRequest` in queue.
+    pub fn get_next_ping(&self) -> Option<PingRequest> {
+        let mut queue = self
+            .queue
+            .write()
+            .expect("Can't write to pending pings queue.");
+        queue.pop_front()
+    }
+}
+
+#[cfg(test)]
+mod test {
+    use crate::metrics::PingType;
+    use serde_json::Value as JsonValue;
+
+    let uuid = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
+    let url = "http://example.com";
+    let body = json!("{}")
+
+    #[test]
+    fn test_doesnt_error_when_there_are_no_pending_pings() {
+        let data_dir = tempfile::tempdir().unwrap();
+        let tmpname = data_dir.path().display().to_string();
+
+        // Create a new upload_manager
+        let upload_manager = PingUploadManager::new(&tempname);
+
+        // Try and get the next request
+        let request = upload_manager.get_next_ping();
+
+        // Verify request was not returned
+        assert!(request.is_none());
+    }
+
+    #[test]
+    fn test_returns_ping_request_when_there_is_one() {
+        let data_dir = tempfile::tempdir().unwrap();
+        let tmpname = data_dir.path().display().to_string();
+
+        // Create a new upload_manager
+        let upload_manager = PingUploadManager::new(&tempname);
+
+        // Enqueue a ping
+        upload_manager.enqueue_ping(uuid, url, body);
+
+        // Try and get the next request
+        let request = upload_manager.get_next_ping();
+
+        // Verify request was returned
+        assert!(request.is_some());
+    }
+
+    #[test]
+    fn test_returns_as_many_ping_request_as_there_are() {
+        let data_dir = tempfile::tempdir().unwrap();
+        let tmpname = data_dir.path().display().to_string();
+
+        // Create a new upload_manager
+        let upload_manager = PingUploadManager::new(&tempname);
+
+        // Enqueue a ping multiple times
+        let n = 10;
+        for _ in 0..n {
+            upload_manager.enqueue_ping(uuid, url, body);
+        }
+
+        // Verify a request is returned for each submitted ping
+        for _ in 0..n {
+            assert!(upload_manager.get_next_ping().is_some());
+        }
+
+        // Verify that after all requests are returned, none are left
+        assert!(upload_manager.get_next_ping().is_none());
+    }
+
+    #[test]
+    fn clearing_the_queue_works_correctly() {
+        let data_dir = tempfile::tempdir().unwrap();
+        let tmpname = data_dir.path().display().to_string();
+
+        // Create a new upload_manager
+        let upload_manager = PingUploadManager::new(&tempname);
+
+        // Submit the ping multiple times
+        let n = 10;
+        for _ in 0..n {
+            upload_manager.enqueue_ping(uuid, url, body);
+        }
+
+        // Clear the queue
+        glean.upload_manager.clear_ping_queue();
+
+        // Verify there really isn't any ping in the queue.
+        assert!(glean.upload_manager.get_next_ping().is_none());
+    }
+}
--- a/glean-core/src/upload/request.rs
+++ b/glean-core/src/upload/request.rs
@@ -1,0 +1,56 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Ping request representation.
+
+use std::collections::HashMap;
+
+use chrono::prelude::{DateTime, Utc};
+use serde_json::Value as JsonValue;
+
+/// Represents a request to upload a ping.
+#[derive(Debug)]
+pub struct PingRequest {
+    /// The Job ID to identify this request,
+    /// this is the same as the ping UUID.
+    pub uuid: String,
+    /// The path for the server to upload the ping to.
+    pub url: String,
+    /// The body of the request.
+    pub body: JsonValue,
+    /// A map with all the headers to be sent with the request.
+    pub headers: HashMap<String, String>,
+}
+
+impl PingRequest {
+    /// Creates a new PingRequest.
+    ///
+    /// Automatically creates the default request headers.
+    /// Clients may add more headers such as `userAgent` to this list.
+    pub fn new(uuid: &str, url: &str, body: JsonValue) -> Self {
+        Self {
+            uuid: uuid.into(),
+            url: url.into(),
+            body,
+            headers: Self::create_request_headers(),
+        }
+    }
+
+    /// Creates the default request headers.
+    fn create_request_headers() -> HashMap<String, String> {
+        let mut headers = HashMap::new();
+        let date: DateTime<Utc> = Utc::now();
+        headers.insert("Date".to_string(), date.to_string());
+        headers.insert("X-Client-Type".to_string(), "Glean".to_string());
+        headers.insert(
+            "Content-Type".to_string(),
+            "application/json; charset=utf-8".to_string(),
+        );
+        headers.insert(
+            "X-Client-Version".to_string(),
+            env!("CARGO_PKG_VERSION").to_string(),
+        );
+        headers
+    }
+}
--- a/glean-core/src/upload/util.rs
+++ b/glean-core/src/upload/util.rs
@@ -1,0 +1,281 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Pings directory processing utilities.
+
+use std::cmp::Ordering;
+use std::fs::{self, File};
+use std::io::{BufRead, BufReader};
+use std::path::{Path, PathBuf};
+
+use log;
+use serde_json::Value as JsonValue;
+use uuid::Uuid;
+
+use super::PingRequest;
+use crate::Result;
+
+/// Attempt to delete a file.
+///
+/// This won't panic if not able to delete, just log.
+fn delete_file(path: &Path) {
+    match fs::remove_file(path) {
+        Err(e) => log::error!("Error deleting file {}. {}", path.display(), e),
+        _ => log::info!("Deleted file {}", path.display()),
+    };
+}
+
+/// Get the file name from a path as a &str.
+///
+/// This won't panic if not able to get file name, just log.
+fn get_file_name_as_str(path: &Path) -> Option<&str> {
+    match path.file_name() {
+        None => {
+            log::warn!("Error getting file name from path: {}", path.display());
+            None
+        }
+        Some(file_name) => {
+            let file_name = file_name.to_str();
+            if file_name.is_none() {
+                log::warn!("File name is not valid unicode: {}", path.display());
+            }
+            file_name
+        }
+    }
+}
+
+/// Get the pending pings directory path.
+fn get_pings_dir(data_path: &Path) -> PathBuf {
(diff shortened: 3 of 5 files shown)


glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+            if file_name.is_none() {
+                log::warn!("File name is not valid unicode: {}", path.display());
+            }
+            file_name
+        }
+    }
+}
+
+/// Creates requests for ping upload and process responses from them.
+pub struct PingUploadHelper {
```
Dexterp37: Bikeshedding: `UploadRequestManager`? `UploadManager`? `UploadScheduler`?

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    /// to prevent files from polluting the pings directory.
+    ///
+    /// Any directories will also be deleted. There should only be
+    /// files inside the `pending_pings` directory.
+    fn process_pings_dir(&self) -> Result<()> {
+        for entry in self.pending_pings_dir.read_dir()? {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+
+                if path.is_dir() {
```
Dexterp37: Is this a new behaviour?

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+    fn get_next_ping(&self) -> Option<(Uuid, PathBuf)> {
+        let mut attempts = 0;
+        loop {
+            if let Some(uuid) = self.queue.process() {
+                // Create the ping file path from the UUID
+                let path = self.pending_pings_dir.join(uuid.to_string());
+                return Some((uuid, path));
+            }
+
+            if let Err(e) = self.process_pings_dir() {
```
Dexterp37: What about this:

- only to the `process_pings_dir` when this module is initialized/built
- do that off the main thread
- every time a new ping is created (outside of this module), in addition to storing it to disk, also add it to the queue.

This way we would not require to process the pings dir when the queue is empty and we'd guarantee that the full processing only happens once, at startup, off the main thread.

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+
+                if let Some(file_name) = get_file_name_as_str(&path) {
+                    if !PING_FILE_NAME_PATTERN.is_match(file_name) {
+                        log::warn!("Pattern mismatch {}", path.display());
+                        delete_file(&path);
+                        continue;
+                    }
+
+                    // Safe unwrap, already verified if valid UUID
+                    self.queue.enqueue(Uuid::parse_str(file_name).unwrap());
```
Dexterp37: If we have pending pings, at startup, we would probably want to send them in a specific order (oldest first!), like we do on desktop. This would make the behaviour consistent once we jump to project FOG.

Note that this doesn't need to happen as part of this PR, but please file a bug and make sure the current design allows for that.

glean-core/src/upload/queue.rs
```diff
@@ -0,0 +1,138 @@
...
+            queue: RwLock::new(VecDeque::new()),
+            processing: RwLock::new(HashSet::new()),
+        }
+    }
+
+    /// Enqueue a ping
+    ///
+    /// Pings that are in the processing set will be ignored
+    pub fn enqueue(&self, uuid: Uuid) {
+        let processing = self
```
Dexterp37: When would we happen to enqueue a ping that's already being processed?

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+            queue: PingsQueue::new(),
+            pending_pings_dir: data_path.join("pending_pings"),
+        }
+    }
+
+    /// Creates a `PingRequest` for the first possible ping in the queue.
+    ///
+    /// If one of the files is malformed or generates errors, it will
+    /// keep trying while there are still pending pings.
+    pub fn create_next_ping_request(&self) -> Option<PingRequest> {
```
Dexterp37: Is this the function that needs to be called from the outside to get any available work?

If so, please note that, [by design](https://docs.google.com/document/d/17D5K--z8GC-vU8t1sKwziTwexQTtDzYB99xDzhUIjiI/edit#heading=h.ap1pjdulp5vm), this should return whether or not work is available or if the caller needs to wait. Are you planning on doing that in a separate bug?

glean-core/src/upload/mod.rs
brizental: Hm, just checked. Yes it is, for some reason I wrote it down as something this function should do. If you think it doesn't make sense, I can remove it.

Dexterp37: Let's keep it simple for now and just reproduce what we currently have with respect to this. We can always add later.

glean-core/src/upload/mod.rs
brizental: Yes, I agree that would be ideal. I thought of something similar, but didn't want to digress too much from the current way things work for this PR. I tried to "copy" as much as I could from the Android implementation. 

Would you agree to make this change in a future PR, specifically the one where I add interface for the upload module to the Glean object? I think that would make sense, since this change would require that the upload helper is initialized upon initialization of the Glean object.

glean-core/src/upload/mod.rs
brizental: I am not sure this is what you mean, but the queue always will return oldest first.

I could maybe loop the folder in a way that the oldest files are enqueued first, thus processed first, if that is what you mean.

glean-core/src/upload/mod.rs
Dexterp37: Note that the current Android implementation already runs off the main thread, since it runs on a thread owned by the WorkManager. If we were to enable this in the current form, we would see a performance regression due to disk I/O.

glean-core/src/upload/mod.rs
Dexterp37: > I could maybe loop the folder in a way that the oldest files are enqueued first, thus processed first, if that is what you mean.

Yes, that's what I mean, sorry if I was unclear :-)

glean-core/src/upload/queue.rs
brizental: In case we process the directory while some request is still being processed.

Dexterp37: That would not happen if we were to scan only once at startup, right?

glean-core/src/upload/mod.rs
brizental: Yes, I am. I thought about this, but for the stage the module is in right now, I can't see a situation where we would tell the caller to wait. Please correct me if I am wrong.

Upon implementing, I actually realized that I am not sure at which time we would tell the caller to wait other than for throttling purposes, so I thought we could do this when looking into that.

Dexterp37: That's fine, I wanted to make sure about that :)

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,357 @@
...
+
+use crate::Result;
+use queue::PingsQueue;
+use request::PingRequest;
+
+mod queue;
+mod request;
+
+lazy_static! {
+    /// Ping file names are UUIDs, this matches UUIDs for filtering purposes
```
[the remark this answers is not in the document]
brizental: This nit is probably going to pop up in many of my comments, apologies.

* off to add `.` to all my docstrings.

glean-core/src/upload/mod.rs
brizental: I like `UploadManager`. I'll wait and see if more people have opinions on this before changing it :)

glean-core/src/upload/queue.rs
brizental: Right.

glean-core/src/upload/mod.rs
brizental: I just remembered where I got this from, https://github.com/mozilla/glean/blob/master/glean-core/tests/common/mod.rs#L115-L118

glean-core/src/upload/mod.rs
brizental: Why do you think so?

Dexterp37: Because naming is hard :-)

What is this API doing in the first place? We're calling it to understand if there's anything to upload or if we need to wait. I'm not saying `get_ping_to_upload` is a good name, but I'm not sure `create_next_ping_request` is as well!

The `next` within the function name reminds me of some Win32 APIs that were like "GetFirstSomething" / "GetNextSomething": the next implies, for me, that there was a "first" :) In addition to that, the fact that it creates a "request" object is an implementation detail: the real objective of calling this function is, really, just getting the indications on what to do (upload or wait)

brizental: > Because naming is hard :-)

LOL

---

This is very interesting to me, we understand the name so differently! I agree with your argument about the request part being an implementation detail and could be left out of the function name. But the `next` to me in necessary because as I understand it, it implies that there is more than one and we should keep coming back until there is none.

It's like "Next, please!" when you are calling the next person in a line.

My suggestion is to rename it to `get_next_ping_to_upload`, merging our two suggestions.

Dexterp37: > My suggestion is to rename it to `get_next_ping_to_upload`, merging our two suggestions.

Sounds good to me :-) Let's see what @badboy thinks :)

glean-core/src/upload/mod.rs
brizental: Talked to @badboy about this and we think the best approach would be to do what you said and spawn a new thread on initialization to process de directory, but also also process the files and save Request objects to the queue so that we only do disk I/O on this step.

brizental: I did this, it works but it resulted in a lot of changes in the structure of the code.

glean-core/src/upload/util.rs
```diff
@@ -0,0 +1,262 @@
...
+    if let (Some(Ok(url)), Some(Ok(body))) = (lines.next(), lines.next()) {
+        if let Ok(parsed_body) = serde_json::from_str::<JsonValue>(&body) {
+            return Some(PingRequest::new(uuid, &url, parsed_body));
+        } else {
+            log::warn!("Can't parse ping contents as JSON.");
+        }
+    } else {
+        log::warn!("Ping file is not formatted as expected.");
+    }
+    delete_file(path);
```
badboy: I don't think we actually want to delete the file until we know for sure it has been sent.
This is the only persistence we have for those.

glean-core/src/upload/request.rs
```diff
@@ -0,0 +1,56 @@
...
+
+use std::collections::HashMap;
+
+use chrono::prelude::{DateTime, Utc};
+use serde_json::Value as JsonValue;
+
+/// Represents a request to upload a ping.
+#[derive(Debug)]
+pub struct PingRequest {
+    /// The Job ID to identify this request later.
```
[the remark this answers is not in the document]
brizental: Yes.

glean-core/src/upload/request.rs
```diff
@@ -0,0 +1,56 @@
...
+    pub uuid: String,
+    /// The path for the server to upload the ping to.
+    pub url: String,
+    /// The body of the request, should be valid JSON.
```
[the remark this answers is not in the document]
brizental: Ah right, that is a comment from when this was a String, earlier in the development process :) Will be removed, thanks for noticing.

glean-core/src/upload/util.rs
```diff
@@ -0,0 +1,262 @@
...
+use log;
+use regex::Regex;
+use serde_json::Value as JsonValue;
+
+use super::PingRequest;
+use crate::Result;
+
+lazy_static! {
+    /// Ping file names are UUIDs, this matches UUIDs for filtering purposes.
+    static ref PING_FILE_NAME_PATTERN: Regex = Regex::new("[0-9a-fA-F]{8}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{4}-[0-9a-fA-F]{12}").unwrap();
```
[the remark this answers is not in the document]
brizental: Ah, didn't even think of that while copy pasting the regex part from Kotlin! Much prefer not using the regex :)

glean-core/src/upload/util.rs
brizental: I delete only if the file is not formatted properly or if ping contents are not valid JSON. I also delete on the `process_pings_dir` function in case the file name is not a valid uuid.

I will send a different PR after this one to add the functionality where the upload module also receives the response from the upload and decides if should delete the file or try again.

glean-core/src/upload/util.rs
```diff
@@ -0,0 +1,262 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Pings directory processing utilities.
+
+use std::cmp::Ordering;
+use std::fs::{metadata, remove_file, File};
```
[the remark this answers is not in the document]
brizental: I agree.

glean-core/src/upload/util.rs
```diff
@@ -0,0 +1,262 @@
...
+    } else {
+        log::warn!("Ping file is not formatted as expected.");
+    }
+    delete_file(path);
+    None
+}
+
+/// Process the pings directory and return a vector
+/// of `PingRequest`s corresponding to each valid ping
+/// file in the directory. This vector will be ordered by
```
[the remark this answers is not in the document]
brizental: I was honestly randomly breaking the lines wherever I thought it looked ok. I appreciate having some structure to the line breaking :) Thanks!

glean-core/src/upload/util.rs
badboy: Oooh, sorry. Indeed. All good then.

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+use util::process_pings_dir;
+
+mod request;
+mod util;
+
+/// Manages the pending pings queue and directory.
+#[derive(Debug)]
+pub struct PingUploadHelper {
+    /// A FIFO queue storing a `PingRequest` for each pending_ping.
+    queue: Arc<RwLock<VecDeque<PingRequest>>>,
```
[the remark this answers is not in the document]
brizental: So that we can have a reference to the queue in two threads at the same time.

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+                let mut local_queue = local_queue
+                    .write()
+                    .expect("Can't write to pending pings queue.");
+                for request in requests {
+                    local_queue.push_back(request);
+                }
+            }
+            Err(e) => log::info!("Error processing pending pings directories! {}", e),
+        })
+        .join()
```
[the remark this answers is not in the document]
brizental: Alright, so this was a actually a question that I had here and you just answered it, thanks :) Now I understand that if I just remove the `join` altogether we will never know if the thread was successfull or not, what would be the better approach here?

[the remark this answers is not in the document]
badboy: Do we need to know about the thread's state?
When it finishes it works it should put stuff into the shared queue.
If someone asks and that queue is empty than we don't hand out anything. Next time the uploader asks and there is something we hand it out.

I think just letting it run is fine.

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        assert!(request.is_some());
+    }
+
+    #[test]
+    fn test_returns_as_many_ping_request_as_there_are() {
+        let (mut glean, _) = new_glean();
+        // Register a ping for testing
+        let ping_type = PingType::new("test", true, true);
+        glean.register_ping_type(&ping_type);
+        // Submit the ping multiple times
```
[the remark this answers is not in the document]
brizental: That generates lint errors :/

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,212 @@
...
+        }
+
+        // Start glean again and verify that the submitted pings get enqueued
+        {
+            let (glean, _) = new_glean(Some(tempdir));
+
+            // Getting this thread to sleep for just 100ms is enough for the
+            // spawned thread to lock the queue, which guarantees that,
+            // when we try `get_next_ping`, we will only get it after we are done enqueueing.
+            thread::sleep(Duration::from_millis(100));
```
brizental: This is not ideal and fails intermittently on CI, as you may notice. I discussed it with @badboy and we figured that:
1. Testing multi-threaded stuff is hard;
2. What could be done is:
    2.1. Have a flag that will say whether processing of the pending_pings dir is done;
    2.2. When trying to `get_next_ping` either: a) tell the client to wait in case not done processing; b) return a `PingRequest` if there is one; c) return `None` if there is none.

I have started working on that, and it seems like it will be quite a big change. Just for the record :)

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,160 @@
...
+        .expect("Could not process pending pings.");
+
+        Self { queue }
+    }
+
+    /// Creates a `PingRequest` and adds it to the queue.
+    pub fn enqueue_ping(&self, uuid: &str, url: &str, body: JsonValue) {
+        let mut queue = self
+            .queue
+            .write()
```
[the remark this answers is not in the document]
badboy: At that point do we _really_ care about the exact order?
They are already delayed if they are read on startup, they will get processed eventually.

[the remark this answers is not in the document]
Dexterp37: > 
> 
> At that point do we _really_ care about the exact order?

That's a good point. One thing we could do is leave it as it is and document this caveat. We can get to that in the future, if needed.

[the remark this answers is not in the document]
brizental: > To understand how to structure this a bit better, I'd also recommend taking a look at how Firefox Desktop handles this :-) See, for example, this code (and follow along there)

From what I see in this code, desktop works with a flag signaling if processing of pending pings on disk is done. That is what I have started to do with 0f7dc5b and what is described in https://github.com/mozilla/glean/pull/673#discussion_r375815726.

For the sake of not making this PR any longer than it already is, I think it would be best to finish it now and have follow ups for:

- Using the flag to return different statuses inclusing the "come back later" status from `get_next_ping`;
- Adding support for deletion request pings;
- Adding functionatility to process upload responses;
- Adding all of it to the Glean object;
- And then finally adding it to the FFI.

As suggested by @badboy on Riot, we can put all this behind a feature flag for now, if we think necessary. It probably is necessary because even though this code doesn't interefere with the way ping uploading is done on the platform side, it does add tasks to init which is not ideal if it's not going to be used anywhere.

Let me know your thoughts :)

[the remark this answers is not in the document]
Dexterp37: Works for me!

glean-core/src/lib.rs
```diff
@@ -458,10 +463,15 @@ impl Glean {
...
 
+                // Enqueue the ping for sending
+                self.upload_manager
+                    .enqueue_ping(&doc_id, &url_path, content);
+
                 log::info!(
                     "The ping '{}' was submitted and will be sent as soon as possible",
                     ping.name
                 );
+
```
Dexterp37: nit: stray newline?

glean-core/src/lib.rs
```diff
@@ -140,6 +142,7 @@ pub struct Configuration {
 #[derive(Debug)]
 pub struct Glean {
     upload_enabled: bool,
+    upload_manager: PingUploadManager,
```
Dexterp37: Since this is a partial implementation, do we really want to enable it here?

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+//!
+//! * Keeps track of pending pings, loading any unsent ping from disk on startup;
+//! * Exposes `get_next_ping` API for the platform layer to request next ping in line;
+//! * Exposes `process_ping_upload_response` API to check the HTTP response from the ping upload
+//!   and either delete the corresponding ping from disk or re-enqueue it for sending.
+
+// !IMPORTANT!
+// Remove the next line when this module's functionality is in the Glean object.
+// This is here just to not have lint error for now.
+#![allow(dead_code)]
```
Dexterp37: Isn't this already in the Glean object?

glean-core/src/upload/mod.rs
```diff
@@ -0,0 +1,225 @@
...
+
+        // Try and get the next request
+        let request = glean.upload_manager.get_next_ping();
+
+        // Verify request was returned
+        assert!(request.is_some());
+    }
+
+    // #[test]
+    // fn test_pings_not_sent_pings_are_loaded_on_init() {
```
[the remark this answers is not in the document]
brizental: I decided to delete this test. I think there is really no reason to leave it there, since it doesn't work and will be different when I actually make it work. Is that alright?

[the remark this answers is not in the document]
Dexterp37: It is!

glean-core/src/lib.rs
brizental: Removed it in 67ccda3. I tried putting it behind a feature flag, but honestly deleting it is nicer and simpler for this situation.

glean-core/src/upload/mod.rs
brizental: It was, but since it is not anymore I'll leave this :)

glean-core/src/upload/mod.rs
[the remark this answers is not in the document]
brizental: Perfect, I'll file bugs for all the things listed there.

brizental:
@Dexterp37 Thanks :) Is it alright to squash all this into one commit with a good description?

Dexterp37:
> 
> 
> @Dexterp37 Thanks :) Is it alright to squash all this into one commit with a good description?

Fine by me!

PR #674, opened 2020-01-23
1611177: Document that uploadEnabled=false clears metrics

mdboom:


docs/user/general-api.md
```diff
@@ -50,12 +50,13 @@ class SampleApplication : Application() {
         Glean.registerPings(Pings)
 
         // Initialize the Glean library.
-        Glean.initialize(applicationContext, uploadEnabled = true)
+        Glean.initialize(applicationContext, uploadEnabled = settings().isTelemetryEnabled)
```
Dexterp37: I'd be tempted to leave this line as it was or explain that `settings().isTelemetryEnabled` is something which is not part of the SDK but whatever contains the user preference/intent for having telemetry enabled.

docs/user/general-api.md
```diff
@@ -50,12 +50,13 @@ class SampleApplication : Application() {
...
     }
 }
 ```
 
-Once initialized, if collection is enabled, the Glean SDK will automatically start collecting [baseline metrics](pings/metrics.md) and sending its [pings](pings/index.md).
+Once initialized, if `uploadEnabled` is true, the Glean SDK will automatically start collecting [baseline metrics](pings/metrics.md) and sending its [pings](pings/index.md).
```
Dexterp37:
```suggestion
Once initialized, if `uploadEnabled` is true, the Glean SDK will automatically start collecting [baseline metrics](pings/metrics.md) and sending its [pings](pings/index.md), according to their respective schedule.  
```

docs/user/general-api.md
mdboom: I think I prefer to add a comment about it -- the purpose of this change is to make it clear that this value should come from a user preference, and not be a constant (and also to match the Swift example below, but that's secondary)...

Dexterp37: Comment sounds good (maybe also in the swift example?)

PR #676, opened 2020-01-24
Merge pull request #676 from mozilla/shellcheck

badboy:
The now-popular shellcheck tool is a static analysis tool for shell scripts.
It can detect common mistakes or potential bugs in shell scripts.

As the sdk_generator.sh is intended to be used by Glean users, we should
ensure it works across all systems and configurations.
This commit ran shellcheck against it and fixed all warnings and errors:

* Use shell arrays instead of appending to strings
* Properly quote paths to not fail when they contain spaces (not too uncommon on macOS after all)
* Rely on bash for expansion and math

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -83,6 +83,9 @@ swiftlint: ## Run swiftlint to lint Swift code
 yamllint: ## Run yamllint to lint YAML files
 	yamllint glean-core .circleci
 
+shellcheck: ## Run shellcheck against important shell scripts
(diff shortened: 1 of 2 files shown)


badboy:
> LGTM, your bash-fu is impressive!

it's down there on the same level as gradle-knowledge (and here I mostly followed what shellcheck told me in their wiki, e.g. https://github.com/koalaman/shellcheck/wiki/SC2128)

PR #677, opened 2020-01-24
Document changing labels [docs only]

Dexterp37:
This unfortunately requires a GeckoView note :(

### DIFF
--- a/docs/user/metrics/labeled_counters.md
+++ b/docs/user/metrics/labeled_counters.md
@@ -18,7 +18,7 @@ stability:
     ...
 ```
 
-> **Note:** removing or changing labels, including their order in the registry file, is permitted. Special care must be taken when touching GeckoView metrics exfiltrated through the Glean SDK, as the index of the labels is used to exfiltrate Gecko data through the Glean SDK.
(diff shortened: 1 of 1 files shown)


docs/user/metrics/labeled_counters.md
```diff
@@ -18,6 +18,8 @@ stability:
     ...
 ```
 
+> **Note:** removing or changing labels, including their order in the registy file, is permitted. Special care must be taken when touching GeckoView metrics exfiltrated through the Glean SDK, as the index of the labels is used to exfiltrate Gecko data through the Glean SDK.
```
badboy:
```suggestion
> **Note:** removing or changing labels, including their order in the registry file, is permitted. Special care must be taken when touching GeckoView metrics exfiltrated through the Glean SDK, as the index of the labels is used to exfiltrate Gecko data through the Glean SDK.
```

docs/user/metrics/labeled_counters.md
```diff
@@ -18,6 +18,8 @@ stability:
     ...
 ```
 
+> **Note:** removing or changing labels, including their order in the registry file, is permitted. Special care must be taken when touching GeckoView metrics exfiltrated through the Glean SDK, as the index of the labels is used to exfiltrate Gecko data through the Glean SDK.
```
mdboom: Do we want to add something like: 

"Avoid reusing labels that were removed in the past"?

Also, maybe add: "It is best practice to add documentation about removed labels to the `description` field so that analysts will know of their existence and meaning in historical data."  (Or maybe some future version of the probe dictionary will handle that for us automatically -- by shown known, yet decommissioned, labels).

Dexterp37: Yes! Makes sense, adding it

PR #679, opened 2020-01-27
Use simpler conversion between local and fixed offset

badboy:
For reasons chrono's API can be more cumbersome than I would like, but
converting between the system's local time and the same
system-independent fixed offset is doable in fewer lines.

Solution from
https://github.com/chronotope/chrono/issues/104#issuecomment-255655945


Dexterp37:
Is this related to bug 1611770?

badboy:
> Is this related to bug 1611770?

Indirectly. Because of that bug I looked at our code and found this ... a bit too complex piece of code.

It doesn't fix what's causing the crash.

PR #680, opened 2020-01-30
Add "open on github" link to docs

badboy:
After a comment from @mdboom I opted for quickly writing a plugin to add a "open on github" link to each page.

It looks like this (see the bottom):

image

### DIFF
--- a/docs/tabs.css
+++ b/docs/tabs.css
@@ -39,3 +39,10 @@
     border-top: none;
     padding: 6px 12px;
 }
+
+footer {
+  font-size: 0.8em;
+  text-align: center;
+  border-top: 1px solid black;
+  padding: 5px 0;
+}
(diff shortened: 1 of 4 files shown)


docs/tabs.css
```diff
@@ -39,3 +39,10 @@
     border-top: none;
     padding: 6px 12px;
 }
+
+footer {
+  font-size: 0.8em;
+  text-align: center;
+  border-top: 1px solid black;
+  padding: 5px 0;
+}
```
mdboom: This is probably ok for now, but strictly speaking this isn't for the language tabs feature, so maybe it goes in its own css file?

badboy: That is true, but we have so few CSS rules, I'd rather keep them in a single file. I opted for renaming that file instead.

PR #682, opened 2020-02-03
Disambiguate device model

tdsmith:
"Device model" can mean different things on Android; clarify that this is `Build.MODEL` and not e.g. `Build.DEVICE`.


mdboom:
We should also update [here](https://github.com/mozilla/glean/blob/master/glean-core/metrics.yaml#L96).

tdsmith:
Done!

PR #683, opened 2020-02-03
1612940: Don't include null extras in experiments

mdboom:
I couldn't find a magic serde incantation to have a different behavior on JSON vs. bincode, so I came up with this.  But maybe there's something I missed...

### DIFF
--- a/glean-core/src/storage/mod.rs
+++ b/glean-core/src/storage/mod.rs
@@ -232,8 +232,6 @@ mod test {
         assert!(empty_snapshot.is_none());
     }
 
-    // Experiment's API tests: the next test comes from glean-ac's
-    // ExperimentsStorageEngineTest.kt.
     #[test]
     fn test_experiments_json_serialization_empty() {
         let t = tempfile::tempdir().unwrap();
@@ -246,7 +244,6 @@ mod test {
         let snapshot = StorageManager
             .snapshot_experiments_as_json(glean.storage(), "glean_internal_info")
             .unwrap();
-        print!("{:?}", snapshot);
         assert_eq!(
             json!({"some-experiment": {"branch": "test-branch"}}),
             snapshot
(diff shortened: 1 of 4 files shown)


Dexterp37:
> 
> 
> > I couldn't find a magic serde incantation to have a different behavior on JSON vs. bincode, so I came up with this. But maybe there's something I missed...
> 
> Hm, if I recall correctly serde does have some magic for this. Wouldn't [`#[serde(skip_serializing_if = "Option::is_none")]`](https://serde.rs/field-attrs.html) work is this case?

Yeah, but this would cause a problem if we ever needed to deserialize with bincode (it happened already for some other metric type)

glean-core/src/storage/mod.rs
```diff
@@ -231,4 +231,31 @@ mod test {
             StorageManager.snapshot_experiments_as_json(glean.storage(), "glean_internal_info");
         assert!(empty_snapshot.is_none());
     }
+
+    // Experiment's API tests: the next test comes from glean-ac's
```
badboy: I guess with glean-ac gone for good, it doesn't make sense to keep these comments around.

glean-core/src/storage/mod.rs
```diff
@@ -231,4 +231,31 @@ mod test {
...
+        let name = t.path().display().to_string();
+        let glean = Glean::with_options(&name, "org.mozilla.glean", true).unwrap();
+
+        let metric = ExperimentMetric::new(&glean, "some-experiment".to_string());
+
+        metric.set_active(&glean, "test-branch".to_string(), None);
+        let snapshot = StorageManager
+            .snapshot_experiments_as_json(glean.storage(), "glean_internal_info")
+            .unwrap();
+        print!("{:?}", snapshot);
```
badboy: Stray debug print

PR #649, opened 2020-01-14
1557048: Add reason codes to the metrics ping

mdboom:
Requires https://github.com/mozilla/glean_parser/pull/157 and https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/484 to be merged first.

This was a real iceberg.

~Also, no Swift or Python support yet, but thought it worth getting feedback on this much first.~

PR #687, opened 2020-02-05
Send a `baseline` at startup if the previous session ended abruptly

Dexterp37:
This attempts to address the problem of 'force-close' interfering with the `baseline` ping. It implements the following logic:

**In addition to the current behaviour:**

- when the Glean SDK is initialized, it checks for a dirty flag:
  - if the flag is set, it generates a baseline ping with reason `dirty_startup` (a new standard field being added to all pings, see #649 );
  - if the flag is not set, then set it to true;
- when going to background, clear the dirty flag after computing the foreground duration and submitting the baseline ping.

This will guarantee that whenever the product using the Glean SDK is force-closed or crashes before it has the chance to send a baseline ping with a specific reason, signaling that there was a dirty flag set.

_Important_: there will be no foreground duration reported with the baseline ping with reason “dirty-startup”.

**TODO**

- [x] Add test coverage in Kotlin
- [x] Add a changelog entry

### DIFF
--- a/docs/user/pings/baseline.md
+++ b/docs/user/pings/baseline.md
@@ -3,13 +3,13 @@
 ## Description
 
 This ping is intended to provide metrics that are managed by the library itself, and not explicitly set by the application or included in the application's `metrics.yaml` file.
-If the application crashes no `baseline` ping is sent, no additional ping is generated with the data from before the crash.
 
 > **Note:** As the `baseline` ping was specifically designed for mobile operating systems, it is not sent when using the Glean Python bindings.
 
 ## Scheduling
 
-The `baseline` ping is automatically sent when the application is moved to the [background](index.md#defining-background-state).
+The `baseline` ping is automatically submitted with a `reason: background` when the application is moved to the [background](index.md#defining-background-state).
+The `baseline` ping will be submitted at startup with a `reason: dirty_startup`, if the previous session was not cleanly closed. This only happens from the second start.
 
 ## Contents
 
@@ -17,7 +17,7 @@ The baseline ping includes the following fields:
 
 | Field name | Type | Description |
 |---|---|---|
-| `duration` | Timespan | The duration, in seconds, of the last foreground session. [^1]  |
+| `duration` | Timespan | The duration, in seconds, of the last foreground session. Only available if `reason: background`. [^1]  |
 | `locale` | String | The locale of the application. |
 
 [^1]: A quick note about the `duration` metric: it measures specifically the time that the user spent in the last foreground session.
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -66,7 +66,7 @@ open class GleanInternalAPI internal constructor () {
     private lateinit var applicationContext: Context
 
     // Note: we set `applicationContext` early during startup so this should be fine.
-    private val gleanLifecycleObserver by lazy { GleanLifecycleObserver(applicationContext) }
+    private val gleanLifecycleObserver by lazy { GleanLifecycleObserver() }
 
     private lateinit var gleanDataDir: File
 
@@ -200,6 +200,21 @@ open class GleanInternalAPI internal constructor () {
         // Signal Dispatcher that init is complete
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.flushQueuedInitialTasks()
+
+        // Check if the "dirty flag" is set. That means the product was probably
+        // force-closed. If that's the case, submit a 'baseline' ping with the
+        // reason "dirty_startup". We only do that from the second run.
+        if (!isFirstRun) {
+            @Suppress("EXPERIMENTAL_API_USAGE")
+            Dispatchers.API.launch {
+                if (LibGleanFFI.INSTANCE.glean_is_dirty_flag_set().toBoolean()) {
+                    submitPingByNameSync("baseline", "dirty_startup")
+                    // Note: while in theory we should set the "dirty flag" to true
+                    // here, in practice it's not needed: if it hits this branch, it
+                    // means the value was `true` and nothing needs to be done.
+                }
+            }
+        }
 
         // At this point, all metrics and events can be recorded.
         // This should only be called from the main thread. This is enforced by
@@ -476,7 +491,7 @@ open class GleanInternalAPI internal constructor () {
      * Handle the background event and send the appropriate pings.
      */
     internal fun handleBackgroundEvent() {
-        Pings.baseline.submit()
+        Pings.baseline.submit(Pings.baselineReasonCodes.background)
         Pings.events.submit()
     }
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
@@ -4,7 +4,6 @@
 
 package mozilla.telemetry.glean.scheduler
 
-import android.content.Context
 import androidx.lifecycle.Lifecycle
 import androidx.lifecycle.LifecycleEventObserver
 import androidx.lifecycle.LifecycleOwner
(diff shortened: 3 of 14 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -468,14 +483,15 @@ open class GleanInternalAPI internal constructor () {
...
-    internal fun testCollect(ping: PingType): String? {
+    internal fun testCollect(ping: PingTypeBase): String? {
         return LibGleanFFI.INSTANCE.glean_ping_collect(ping.handle)?.getAndConsumeRustString()
     }
 
     /**
      * Handle the background event and send the appropriate pings.
      */
     internal fun handleBackgroundEvent() {
+        // TODO: fixme - we should sent "reason: background" for the baseline ping.
```
Dexterp37: @mdboom this is where it gets tricky if we don't get the API right

docs/user/pings/baseline.md
```diff
@@ -3,21 +3,21 @@
...
 This ping is intended to provide metrics that are managed by the library itself, and not explicitly set by the application or included in the application's `metrics.yaml` file.
-If the application crashes no `baseline` ping is sent, no additional ping is generated with the data from before the crash.
 
 > **Note:** As the `baseline` ping was specifically designed for mobile operating systems, it is not sent when using the Glean Python bindings.
 
 ## Scheduling
 
-The `baseline` ping is automatically sent when the application is moved to the [background](index.md#defining-background-state).
+The `baseline` ping is automatically submitted with a `reason: background` when the application is moved to the [background](index.md#defining-background-state).
+The `baseline` ping will be submitted at startup with a `reason: dirty_startup`, if the previous session was not cleanly closed. This only happens from the second start.
```
badboy: Just as a clarification: `reasons` will end up as normal text fields in the table? Therefore whether to use underscores or dashes is just a stylistic question?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
```diff
@@ -4,20 +4,20 @@
...
 import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
+import mozilla.telemetry.glean.rust.LibGleanFFI
+import mozilla.telemetry.glean.rust.toByte
 
 /**
  * Connects process lifecycle events from Android to Glean's handleEvent
  * functionality (where the actual work of sending pings is done).
  */
-internal class GleanLifecycleObserver(
-    private val applicationContext: Context
```
badboy: What was the reason we took in the context?

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -244,6 +246,44 @@ class GleanTest {
...
+        ), false)
+
+        try {
+            // Trigger worker task to upload the pings in the background
+            triggerWorkManager(context)
+
+            val request = server.takeRequest(20L, TimeUnit.SECONDS)
+
+            val baselineJson = JSONObject(request.body.readUtf8())
+            assertEquals("baseline", baselineJson.getJSONObject("ping_info")["ping_type"])
```
badboy: This is going away (#653). Over there we're checking the URL for the right type now.
(That patch is currently stuck, I should get on to it).

If you already adopt the other way to check this then we won't run into merge problems one way or another.

glean-core/ffi/src/lib.rs
```diff
@@ -303,6 +303,16 @@ pub extern "C" fn glean_clear_application_lifetime_metrics() {
...
 }
 
+#[no_mangle]
+pub extern "C" fn glean_set_dirty_flag(flag: u8) {
+    with_glean_value_mut(|glean| glean.set_dirty_flag(flag != 0));
+}
+
+#[no_mangle]
+pub extern "C" fn glean_is_dirty_flag_set() -> u8 {
+    with_glean(|glean| Ok(glean.is_dirty_flag_set()))
```
badboy:
```suggestion
    with_glean_value(|glean| glean.is_dirty_flag_set())
```

would be the alternative way to write this (we also do this for `get_upload_enabled`). It's effectively the same code, so your judgement.

glean-core/pings.yaml
```diff
@@ -19,10 +19,20 @@ baseline:
...
   bugs:
     - https://bugzilla.mozilla.org/1512938
+    - https://bugzilla.mozilla.org/1599877
   data_reviews:
     - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
   notification_emails:
     - glean-team@mozilla.com
+  reasons:
+    dirty_startup: |
+      The ping was submitted at startup, because he application process was
```
badboy:
```suggestion
      The ping was submitted at startup, because the application process was
```

glean-core/src/lib.rs
```diff
@@ -595,6 +598,51 @@ impl Glean {
...
+    /// - on mobile: set to `false` when going to background or shutting down,
+    ///   set to `true` at startup and when going to foreground.
+    /// - on non-mobile platforms: set to `true` at startup and `false` at
+    ///   shutdown.
+    ///
+    /// At startup, before setting its new value, if the "dirty flag" value is
+    /// `true`, then Glean knows it did not exit cleanly and can implement
+    /// coping mechanisms (e.g. sending a `baseline` ping).
+    pub fn set_dirty_flag(&self, new_value: bool) {
+        // Sequence numbers are stored as a counter under a name that includes the storage name
```
badboy: Sequence numbers?

glean-core/src/lib.rs
```diff
@@ -595,6 +598,51 @@ impl Glean {
...
+        self.get_dirty_bit_metric().set(self, new_value);
+    }
+
+    /// ** This is not meant to be used directly.**
+    ///
+    /// Check the stored value of the "dirty flag".
+    pub fn is_dirty_flag_set(&self) -> bool {
+        let dirty_bit_metric = self.get_dirty_bit_metric();
+        match StorageManager.snapshot_metric(
```
badboy: If we have more usage of this we should refactor that into a `get_value` function on the boolean metrics (no need to do this here now)

docs/user/pings/baseline.md
Dexterp37: Yes, exactly.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/GleanLifecycleObserver.kt
Dexterp37: We used to need the `Context` because we were using `SharedPreferences`, [in this PR](https://github.com/mozilla/glean/commit/1338a1dbbd3021e17b35bc848859c4daff6e0803#diff-66b8e9961a42e4203fd9a68980025f35R21), to save the state of the dirty bit. We are now doing this in Rust :-)

glean-core/src/lib.rs
Dexterp37: Yes, this makes sense, Let's see how much this will get used from now on.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -201,6 +201,21 @@ open class GleanInternalAPI internal constructor () {
...
         Dispatchers.API.flushQueuedInitialTasks()
 
+        // Check if the "dirty flag" is set. That means the product was probably
+        // force-closed. If that's the case, submit a 'baseline' ping with the
+        // reason "dirty_startup". We only do that from the second run.
+        if (!isFirstRun) {
+            @Suppress("EXPERIMENTAL_API_USAGE")
+            Dispatchers.API.launch {
+                if (LibGleanFFI.INSTANCE.glean_is_dirty_flag_set().toBoolean()) {
+                    submitPingByNameSync("baseline", "dirty_startup")
```
mdboom: Can we do `Pings.baseline.submit(Pings.baselineReasonCodes.dirtyStartup)` instead to make use of the enum?

mdboom: Oh, wait, we can't because of the sync requirement.  Carry on.

glean-core/src/lib.rs
```diff
@@ -576,6 +579,50 @@ impl Glean {
...
+            self.storage(),
+            INTERNAL_STORAGE,
+            &dirty_bit_metric.meta().identifier(self),
+        ) {
+            Some(Metric::Boolean(b)) => b,
```
mdboom: Could use `unwrap_or(false)` instead of this `match` clause here...

Dexterp37: By changing to that, we get a `expected bool, found enum `metrics::Metric`` - leaving this as it is, we can always fixup later!

PR #689, opened 2020-02-06
Run Kotlin tests only once in debug configuration

badboy:
The default test target runs both testDebugUnitTest and testReleaseUnitTest, effectively running everything twice.
This seems like a waste of resources (and time) and just running it once seems fine.

Now we could decide to run testReleaseUnitTest on CI, but testDebugUnitTest for developer machines.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -323,7 +323,7 @@ jobs:
       - skip-if-doc-only
       - run:
           name: Android tests
-          command: ./gradlew --no-daemon test
+          command: ./gradlew --no-daemon :glean:testDebugUnitTest
           environment:
             GRADLE_OPTS: -Xmx2048m
             TARGET_CFLAGS: -DNDEBUG
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -322,21 +322,22 @@ jobs:
       - android-setup
       - skip-if-doc-only
       - run:
+          name: Restrict to Linux builds only
+          command: |
+              echo "rust.targets=linux-x86-64" > local.properties
+      - run:
           name: Android tests
           command: ./gradlew --no-daemon :glean:testDebugUnitTest
           environment:
             GRADLE_OPTS: -Xmx2048m
             TARGET_CFLAGS: -DNDEBUG
-      - store_artifacts:
-          path: glean-core/android/build/rustJniLibs/android
-          destination: libs
       - run:
           name: Save test results
           command: |
-            mkdir -p ~/test-results/junit/
-            mkdir -p ~/test-results/tests/
-            cp -a glean-core/android/build/reports/tests ~/test-results/
-            find glean-core/android/build -type f -regex ".*/build/test-results/.*xml" -exec cp {} ~/test-results/junit/ \;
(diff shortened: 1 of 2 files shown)


badboy:
This is a first push to see if that's an actual improvement (resource & timing) on CI.

badboy:
Weird Python failure:

```
ImportError while loading conftest '/home/circleci/project/glean-core/python/tests/conftest.py'.
glean-core/python/tests/conftest.py:14: in <module>
    from glean import testing
glean-core/python/.venv/lib/python3.5/site-packages/glean_sdk-24.1.0-py3.5-linux-x86_64.egg/glean/__init__.py:12: in <module>
    from ._loader import load_metrics, load_pings
glean-core/python/.venv/lib/python3.5/site-packages/glean_sdk-24.1.0-py3.5-linux-x86_64.egg/glean/_loader.py:20: in <module>
    from glean_parser.parser import parse_objects  # type: ignore
glean-core/python/.venv/lib/python3.5/site-packages/glean_parser-1.15.5-py3.5.egg/glean_parser/parser.py:19: in <module>
    from .metrics import Metric
glean-core/python/.venv/lib/python3.5/site-packages/glean_parser-1.15.5-py3.5.egg/glean_parser/metrics.py:14: in <module>
    from . import util
glean-core/python/.venv/lib/python3.5/site-packages/glean_parser-1.15.5-py3.5.egg/glean_parser/util.py:19: in <module>
    import jinja2
/home/circleci/project/glean-core/python/.venv/lib/python3.5/site-packages/Jinja2-3.0.0a1-py3.5.egg/jinja2/__init__.py:8: in <module>
    ???
E     File "/home/circleci/project/glean-core/python/.venv/lib/python3.5/site-packages/Jinja2-3.0.0a1-py3.5.egg/jinja2/bccache.py", line 207
E       dirname = f"_jinja2-cache-{os.getuid()}"
E                                              ^
E   SyntaxError: invalid syntax
make: *** [Makefile:63: test-python] Error 4
```

cc @mdboom

badboy:
First run clocked in at 6min 37s, which is in the range it was before.
I added another small tweak to avoid building stuff we don't need.

badboy:
Restricting to only linux builds brings it down to 3mins.

I think running the Kotlin tests twice is insignificant in comparison.

badboy:
Cancelling the iOS builds in the meantime, rerunning them when we made a decision.

mdboom:
> Weird Python failure:

I saw this yesterday.  Jinja2 made a release that is no longer compatible with Python 3.5.

The fix is over in `glean_parser` [here](https://github.com/mozilla/glean_parser/commit/4d5bd7119819e88a13845ea17e3ec98a60b74781#diff-2eeaed663bd0d25b7e608891384b7298).  We can update `glean_parser` now, or otherwise it will happen as part of #653.

mdboom:
I'm *usually* a fan of testing both for things like C++ or Rust where this might generate significantly different behavior.  I'm not aware of a single bug in our Kotlin code that was debug- or release-only, though, so I think this is probably a good idea.  And doing debug only is the right choice because (I presume) that gives us better backtraces.

PR #696, opened 2020-02-11
Have `get_next_ping` signal when ping directory is being processed

brizental:
Resolves [Bug 1614281](https://bugzilla.mozilla.org/show_bug.cgi?id=1614281)

While trying to use `get_next_ping` from the upload module, the requester will now receive three possible return values:
* `Processing` - which means the pings directory is currently being processed and the requester should come back later;
* `Empty` - which means the pings queue is empty and requester can stop asking for more pings;
* `Next(PingRequest)` - if no processing is going on and there are pings in the queue, requester will get the oldest ping in it.

This also includes minor fixes to tests in the upload module.

### DIFF
--- a/glean-core/src/upload/mod.rs
+++ b/glean-core/src/upload/mod.rs
@@ -5,7 +5,7 @@
 //! Manages the pending pings queue and directory.
 //!
 //! * Keeps track of pending pings, loading any unsent ping from disk on startup;
-//! * Exposes `get_next_ping` API for the platform layer to request next ping in line;
+//! * Exposes `get_upload_task` API for the platform layer to request next upload task;
 //! * Exposes `process_ping_upload_response` API to check the HTTP response from the ping upload
 //!   and either delete the corresponding ping from disk or re-enqueue it for sending.
 
@@ -30,40 +30,67 @@ use util::process_pings_dir;
 mod request;
 mod util;
 
+/// When asking for the next ping request to upload,
+/// the requester may receive one out of three possible tasks.
+#[derive(PartialEq, Debug)]
+pub enum PingUploadTask {
+    /// A PingRequest popped from the front of the queue.
+    /// See [`PingRequest`](struct.PingRequest.html) for more information.
+    Upload(PingRequest),
+    /// A flag signaling that the pending pings directories are not done being processed,
+    /// thus the requester should wait and come back later.
+    Wait,
+    /// A flag signaling that the pending pings queue is empty and requester is done.
+    Done,
+}
+
 /// Manages the pending pings queue and directory.
 #[derive(Debug)]
 pub struct PingUploadManager {
     /// A FIFO queue storing a `PingRequest` for each pending ping.
     queue: Arc<RwLock<VecDeque<PingRequest>>>,
-    /// A flag, signaling if we are done processing the pending pings directories.
+    /// A flag signaling if we are done processing the pending pings directories.
+    ///
+    /// This does not indicate that processing of the directory was successful,
+    /// only that it did happen.
     processed_pending_pings: Arc<AtomicBool>,
 }
 
 impl PingUploadManager {
     /// Create a new PingUploadManager.
     ///
-    /// This will spawn a new thread and processes the pending pings folder
-    /// filling up the queue with whatever pings are in there
-    /// ordered by oldest to newest file modified.
+    /// Spawns a new thread and processes the pending pings directory,
+    /// filling up the queue with whatever pings are in there.
+    ///
+    /// # Arguments
+    ///
+    /// * `data_path` - Path to the pending pings directory.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if unable to spawn a new thread.
     pub fn new(data_path: &str) -> Self {
         let queue = Arc::new(RwLock::new(VecDeque::new()));
         let processed_pending_pings = Arc::new(AtomicBool::new(false));
 
-        let data_path = PathBuf::from_str(&data_path).expect("data_path must be a valid path.");
+        let data_path = PathBuf::from_str(data_path).expect("data_path must be a valid path.");
         let local_queue = queue.clone();
         let local_flag = processed_pending_pings.clone();
         let _ = thread::Builder::new()
-            .name("glean.upload_manager.process_pings_directory".to_string())
-            .spawn(move || match process_pings_dir(&data_path) {
-                Ok(requests) => {
-                    let mut local_queue = local_queue
-                        .write()
-                        .expect("Can't write to pending pings queue.");
-                    local_queue.extend(requests.into_iter());
-                    local_flag.store(true, Ordering::SeqCst);
+            .name("glean.upload_manager.process_pings_dir".to_string())
+            .spawn(move || {
+                match process_pings_dir(&data_path) {
+                    Ok(requests) => {
+                        let mut local_queue = local_queue
+                            .write()
+                            .expect("Can't write to pending pings queue.");
+                        local_queue.extend(requests.into_iter());
+                    }
+                    Err(e) => log::info!("Error processing pending pings directories! {}", e),
                 }
-                Err(e) => log::info!("Error processing pending pings directories! {}", e),
-            });
+                local_flag.store(true, Ordering::SeqCst);
+            })
+            .expect("Unable to spawn thread to process pings directories.");
 
         Self {
             queue,
@@ -71,7 +98,7 @@ impl PingUploadManager {
         }
     }
 
-    fn has_processed_pending_pings(&self) -> bool {
+    fn has_processed_pings_dir(&self) -> bool {
         self.processed_pending_pings.load(Ordering::SeqCst)
     }
 
@@ -85,7 +112,7 @@ impl PingUploadManager {
         queue.push_back(request);
     }
 
-    /// Clear pending pings queue.
+    /// Clears the pending pings queue.
     pub fn clear_ping_queue(&self) {
         let mut queue = self
             .queue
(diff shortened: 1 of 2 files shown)


glean-core/src/upload/mod.rs
```diff
@@ -30,49 +30,82 @@ use util::process_pings_dir;
 mod request;
 mod util;
 
+/// The state of the upload manager queue at the time of asking.
+#[derive(PartialEq, Debug)]
+pub enum PingsQueueStatus {
```
brizental: Not very sure about naming choices here, regarding the enum and its options. Would appreciate some better suggestions if anyone has them :)

glean-core/src/upload/mod.rs
```diff
@@ -30,49 +30,82 @@ use util::process_pings_dir;
...
+    /// * `data_path` - Path to the folder where the pending pings directories are.
+    ///
+    /// # Return value
+    ///
+    /// `thread::JoinHandle`, the spawned thread handle (for tests synchronization purposes).
+    ///
+    /// # Panics
+    ///
+    /// Will panic if unable to spawn a new thread.
+    pub fn process_pending_pings(&self, data_path: &str) -> thread::JoinHandle<()> {
```
brizental: Most of the changes I made to this function, even the change where I remove it from `::new` into it's own function, were so that I could test this in an easier manner. These changes don't however change how it used to work, apart from:
1)  Now it panicks in case not able to spawn a new thread, I just thought that made sense;
2) It will change the flag even when not sucessful processing the directory. I also just thought that made sense.

glean-core/src/upload/mod.rs
```diff
@@ -94,99 +127,179 @@ impl PingUploadManager {
...
+        let data_path = dir.path().to_str().unwrap();
+
+        // Attempt at processing the directory
+        let handle = upload_manager.process_pending_pings(&data_path);
+
+        // With the processing ongoing, flag should be `true`
+        assert_eq!(upload_manager.is_processing_pending_pings(), true);
+        // ...and `get_next_ping` should return `Processing`
+        assert_eq!(upload_manager.get_next_ping(), PingsQueueStatus::Processing);
+
```
brizental: I am not sure the two above assertions are not going to create intermittents, because I don't have such fine grained control over order of execution on the spawned thread. @badboy what do you think?

glean-core/src/upload/mod.rs
badboy: What about describing what the receiver needs to do:
It's an `UploadJob` / `UploadTask`:

* `Upload(PingRequest)` or `Send` - as your `Next`
* `Wait(seconds)` -  Wait the amount of seconds before asking again (the receiver doesn't care why it needs to wait. This will then also apply once we add throttling.
* `None`/`Nothing`/`Done` - No more job/task, end your work.

What do you think?

glean-core/src/upload/mod.rs
badboy: Seems like good choices.

glean-core/src/upload/mod.rs
```diff
@@ -30,49 +30,82 @@ use util::process_pings_dir;
...
+    ///
+    /// # Return value
+    ///
+    /// `thread::JoinHandle`, the spawned thread handle (for tests synchronization purposes).
+    ///
+    /// # Panics
+    ///
+    /// Will panic if unable to spawn a new thread.
+    pub fn process_pending_pings(&self, data_path: &str) -> thread::JoinHandle<()> {
+        self.processing_pending_pings.store(true, Ordering::SeqCst);
```
badboy: IMO we should not toggle back and forth.

On start we're in a "not processed" state, then we process it, and then processing is done.
Given that we require this to be called early we should not go back and forth in the states.

glean-core/src/upload/mod.rs
badboy: True. I think the flag is merely an implementation detail.

If we go with what I said before that we don't toggle back and forth between the states what we would want to test here is:

* When starting, we have not processed the directory yet and don't get a ping out
* We start processing and wait for it to finish.
* After the async processing is done, we get a ping out

glean-core/src/upload/mod.rs
badboy: Paging in @Dexterp37 for some bikeshedding.

glean-core/src/upload/mod.rs
brizental: Yes, the flag before was either `processed` ou not-`processed`. I changed it to be, `processing` or not-`processing`. That allowed me to write a test to verify if we got all three states correctly.

I hand't considered just not adding a specific test for that, as you suggest :)

glean-core/src/upload/mod.rs
brizental: I'll wait for @Dexterp37 's opinions before settling this, but I will comment on the `Wait(seconds)`. I do agree that will make sense for throttling, but I didn't add it here this time because I didn't find it necessary yet.

Do you think I should add it now? And if so, how many seconds should I default to?

glean-core/src/upload/mod.rs
brizental: Yes, I agree with you. And that would mean just removing this test altogether, because what you are describing is what the test just before this one does.

glean-core/src/upload/mod.rs
Dexterp37: > 
> 
> What about describing what the receiver needs to do:
> It's an `UploadJob` / `UploadTask`:
> 
>     * `Upload(PingRequest)` or `Send` - as your `Next`

+1 on `Upload`
 
>     * `Wait(seconds)` -  Wait the amount of seconds before asking again (the receiver doesn't care why it needs to wait. This will then also apply once we add throttling.

+1 on this!

>     * `None`/`Nothing`/`Done` - No more job/task, end your work.
> 
> 
> What do you think?

I wonder if we even need this third one. @brizental when will the product use the APIs? Do you expect the APIs to be called right after startup (only once, then governed by the `Wait(seconds)` for the rest of the time)? Or do you expect the product to poll at specified times?

glean-core/src/upload/mod.rs
Dexterp37: Yes, both are good ideas! Please comment in the code on *why*, though. Especially why you're changing the flag even though processing has failed.

glean-core/src/upload/mod.rs
```diff
@@ -94,99 +127,179 @@ impl PingUploadManager {
         queue.clear();
     }
 
-    /// Get the next `PingRequest` in queue.
-    pub fn get_next_ping(&self) -> Option<PingRequest> {
+    /// Gets the next `PingRequest` in queue.
```
Dexterp37: nit: can you add a docs for the return type? :)

glean-core/src/upload/mod.rs
badboy: We do need a third one. The idea was for the upload job on the platform side to ask for jobs and upload them and only stop working when instructed.

Consider what happens to pending pings at startup, say 3:
* On Glean.init we start an upload worker, that one asks the API for something to upload
* It gets 3 jobs, one at a time, then finishes

brizental: > I wonder if we even need this third one. @brizental when will the product use the APIs? Do you expect the APIs to be called right after startup (only once, then governed by the Wait(seconds) for the rest of the time)? Or do you expect the product to poll at specified times?

This is an interesting question. I imagined the upload worker would be called at exaclty the same times as it is called right now (on init, whenever pings are submitted). I guess we do have the ability to change it up here and just give the platform a timeframe as to "when it should come back", so that the UploadWorker would be started on init and just told to come back every given interval.

This raises questions that I don't know how to answer such as: when **should** it come back?

Now, taking a step back, if we stick to what I initially was certain of: that the upload worker will ask for "next pings" on init and whenever it submits new pings we do need the `None` option. Because the platform needs to know when to stop asking for pings.

If we change it up to some other flow, it seems reasonable to me that we still keep the `None` option, for the same reason. But I am not so sure about it.

badboy: I don't think we need to change where the upload worker is started.
Just right now we don't explicitly start one to look for pending pings from previous runs. This just happens as a side effect of other pings (e.g. startup metrics pings or similar).

I'm fine with leaving out `Wait` for now until we actually implement backoff.

Dexterp37: > Now, taking a step back, if we stick to what I initially was certain of: that the upload worker will ask for "next pings" on init and whenever it submits new pings, we do need the `None` option. Because the platform needs to know when to stop asking for pings.

Perfect, the above makes sense to me :-) Let's retain the `None` option and leave `Wait` for when the backoff is implemented.

glean-core/src/upload/mod.rs
```diff
@@ -30,48 +30,77 @@ use util::process_pings_dir;
...
     processed_pending_pings: Arc<AtomicBool>,
 }
 
 impl PingUploadManager {
     /// Create a new PingUploadManager.
     ///
-    /// This will spawn a new thread and processes the pending pings folder
-    /// filling up the queue with whatever pings are in there
-    /// ordered by oldest to newest file modified.
+    /// Spawns a new thread and processes the pending pings directory,
```
brizental: I am sorry for the back and forth, but I didn't think it made sense anymore to not do this on init once I changed the flag as per @badboy 's review. I moved it back to the init, which ultimately I think is better, because the caller doesn't need to init and also call the `process_pings_dir` function.

The only bad thing about this is that it required some _hacks_ to get the tests to work everytime, but I don't think they are too bad. Let me know your thoughts.

glean-core/src/upload/mod.rs
```diff
@@ -94,99 +123,172 @@ impl PingUploadManager {
...
-
-        // Try and get the next request
-        let request = upload_manager.get_next_ping();
+        let dir = tempfile::tempdir().unwrap();
+        let tmpname = dir.path().display().to_string();
+        let upload_manager = PingUploadManager::new(&tmpname);
+
+        // HACK!
+        // Wait for processing of pending pings directory to finish.
+        while !upload_manager.has_processed_pings_dir() {
```
Dexterp37: Is there any way, in Rust, to just wait on [this thread](https://github.com/mozilla/glean/pull/696/files#diff-80512c19c26eacea30d18cd504ce6439R81) to be finished?

glean-core/src/upload/mod.rs
```diff
@@ -94,99 +123,172 @@ impl PingUploadManager {
...
+
+    use super::*;
     use crate::metrics::PingType;
-    use serde_json::Value as JsonValue;
+    use crate::tests::new_glean;
 
-    let uuid = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
-    let url = "http://example.com";
-    let body = json!("{}")
+    const UUID: &str = "c0ffeec0-ffee-c0ff-eec0-ffeec0ffeec0";
```
Dexterp37: nit: thinking of this a bit, can we change its value? This has other meanings within the glean repo, so let's use a different fake UUID.

glean-core/src/upload/mod.rs
brizental: I looked, but didn't find it :( Maybe @badboy knows?

badboy: If we store the handle, then we can access it and `.join()` on it (that's what the other test did before refactoring)

brizental: Yes, but that would required that the `process_pings_dir` live outside of the init function and got called separately.

Since we start the `processed_pings_dir` flag as `false` and change it to `true` once the processing is done, it is kind of required that it get's called on init or immediatelly after, so I thought it didn't really make sense for it to be called outside of init.

brizental: One possibility would be to return a tuple from this `::new` function like `(PingUploadManager, JoinHandle)` and then we culd join on the test, but I didn't go with that because it looked bad.

Dexterp37: I'm definitely missing something here, but can't we save the [handle from the builder](https://github.com/mozilla/glean/pull/696/files#diff-80512c19c26eacea30d18cd504ce6439R81) in the `PingUploadTask`, giving it just enough visibility to make useful in tests? @badboy ?

badboy: We would store it in `PingUploadManager`

brizental: Do you think I should make that change? I realised I can add the attribute to the `PingUploadManager` struct and then put it under a `#[cfg(test)]`, which would make it only part of this struct for testing. 

One last argument in favour of the current hack, is that is mimics exactly how we plan on calling this function in the future. I could even change it to:
```rust
        while upload_manager.get_next_ping() == PingUploadTask::Wait {
            thread::sleep(Duration::from_millis(10));
        }
```
To make that clearer.

badboy: Looking at it again, `join` takes the handle by value, so we need to store `Option<JoinHandle>` to make it work.
We only `join` in a test, but I kinda dislike sprinkling `#[cfg(test)]` around.

I think for now the `while !upload_manager.has_processed_pings_dir` is a hack, but an okay-ish one. We can come back to it later of course.

Dexterp37: > ```rust
>         while upload_manager.get_next_ping() == PingUploadTask::Wait {
>             thread::sleep(Duration::from_millis(10));
>         }
> ```
> 
> To make that clearer.

Using `Wait` makes sense here. I'm fine with this change.

brizental: > Looking at it again, join takes the handle by value, so we need to store Option<JoinHandle> to make it work.

Yes, I was testing that out and just did exactly this :)

glean-core/src/upload/mod.rs
```diff
@@ -30,48 +30,77 @@ use util::process_pings_dir;
 mod request;
 mod util;
 
+/// When asking for the next ping request to upload,
+/// the requester may receive one out of three possible tasks.
+///
+/// This enum is a representation of these choices.
```
badboy: this comment is superfluous

glean-core/src/upload/mod.rs
```diff
@@ -94,98 +123,171 @@ impl PingUploadManager {
...
-        upload_manager.clear_ping_queue();
+        // Create a new upload_manager
+        let data_path = dir.path().to_str().unwrap();
+        let upload_manager = PingUploadManager::new(data_path);
+
+        // Wait for processing of pending pings directory to finish.
+        let mut upload_task = upload_manager.get_next_ping();
+        while upload_task == PingUploadTask::Wait {
+            thread::sleep(Duration::from_millis(10));
+            upload_task = upload_manager.get_next_ping();
```
brizental: This really makes me rethink the name of the `get_next_ping` function... `get_upload_task` seems more suitable now.

glean-core/src/upload/mod.rs
```diff
@@ -30,48 +30,77 @@ use util::process_pings_dir;
...
 /// Manages the pending pings queue and directory.
 #[derive(Debug)]
 pub struct PingUploadManager {
     /// A FIFO queue storing a `PingRequest` for each pending ping.
     queue: Arc<RwLock<VecDeque<PingRequest>>>,
-    /// A flag, signaling if we are done processing the pending pings directories.
+    /// A flag signaling if we are done processing the pending pings directories.
+    ///
+    /// This flags value does not mean that the processing of the directory was successful,
+    /// only that it happened or not yet.
```
badboy: I would rephrase that slightly:

```
    /// This does not indicate that processing of the directory was successful, 
	/// only that it did happen.
```

glean-core/src/upload/mod.rs
```diff
@@ -30,48 +30,77 @@ use util::process_pings_dir;
...
     ///
-    /// This will spawn a new thread and processes the pending pings folder
-    /// filling up the queue with whatever pings are in there
-    /// ordered by oldest to newest file modified.
+    /// Spawns a new thread and processes the pending pings directory,
+    /// filling up the queue with whatever pings are in there.
+    ///
+    /// # Arguments
+    ///
+    /// * `data_path` - Path to the folder where the pending pings directories are.
```
badboy:
```suggestion
    /// * `data_path` - Path to the pending pings directory.
```

glean-core/src/upload/mod.rs
```diff
@@ -94,98 +123,171 @@ impl PingUploadManager {
...
+        // Wait for processing of pending pings directory to finish.
+        while upload_manager.get_next_ping() == PingUploadTask::Wait {
+            thread::sleep(Duration::from_millis(10));
+        }
+
         // Enqueue a ping multiple times
         let n = 10;
         for _ in 0..n {
-            upload_manager.enqueue_ping(UUID, URL, json!({}));
+            upload_manager.enqueue_ping(UUID, URL, json!("{}"));
```
badboy: `json!({})` and `json!("{}")` are different things.

The first one results in an empty json object (literally just `{}`).
the second one results in the string "{}".

We want the first one: an empty object.

glean-core/src/upload/mod.rs
badboy: I do agree.

glean-core/src/upload/mod.rs
```diff
@@ -94,98 +123,171 @@ impl PingUploadManager {
...
+        // Verify there really isn't any ping in the queue
+        assert_eq!(upload_manager.get_next_ping(), PingUploadTask::Done);
+    }
+
+    #[test]
+    fn test_fills_up_queue_successfully_from_disk() {
+        let (mut glean, dir) = new_glean(None);
+
+        // Register a ping for testing
+        let ping_type = PingType::new("test", true, true, vec![]);
```
badboy:
```suggestion
        let ping_type = PingType::new("test", true, /* send_if_empty */ true, vec![]);
```

sometimes it would help having named arguments.
IMO spelling it out this way (or by adding it to the comment before this line) makes it clearer why we can submit the ping multiple times.

glean-core/src/upload/mod.rs
badboy: Didn't you change that? Was that change somehow lost again?

brizental: Hm, I made some mistake on rebasing this... Thanks for noticing.

glean-core/src/upload/mod.rs
brizental: Also an error on rebasing, but I actually didn't know this.

PR #703, opened 2020-02-12
Bug 1614785 - Add clearing of app lifetime metrics after sending metrics ping

travis79:
This is the iOS equivalent to [Bug 1604862](https://bugzilla.mozilla.org/show_bug.cgi?id=1604862)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,10 @@
   * The baseline ping will now include `reason` codes that indicate why it was
     submitted. If an unclean shutdown is detected (e.g. due to force-close), this
     ping will be sent at startup with `reason: dirty_startup`.
+  * Per [Bug 1614785](https://bugzilla.mozilla.org/show_bug.cgi?id=1614785), the
+    clearing of application lifetime metrics now occurs after the metrics ping is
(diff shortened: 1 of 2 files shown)


travis79:
> Let's add a CHANGELOG entry and a test (maybe there's already a Kotlin one that could just be translated...)

Sure thing on the CHANGELOG entry.

The test is already there (see [`testGleanPreservesLifetimeApplicationMetrics()`](https://github.com/mozilla/glean/blob/41f583603fd8cb7bc5b2318f4fa77d629dd4a61c/glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift#L229).  I had apparently added the test when working on the iOS `MetricsPingScheduler` but missed the changes to `Glean.initialize()`.  After looking at the Kotlin PR for this, I saw that was where that test was added.  Oddly enough, it passed without the changes included in this PR, perhaps a side effect of platform differences?

travis79:
I'm waiting on the iOS tests to pass.  Once they do, I'll push my CHANGELOG update commit but I don't plan on retriggering the iOS tests (so long as they pass the first time).

travis79:
Okay, looks like the tests all passed. Pushing my CHANGELOG.md commit now.

PR #706, opened 2020-02-12
1613174: Run glinter as part of loading Python metrics

mdboom:


glean-core/python/glean/_loader.py
```diff
@@ -135,6 +137,9 @@ def load_metrics(
 
     filepath = [Path(x) for x in filepath]
 
+    if glean_parser.lint.glinter(filepath, config, file=sys.stderr):
+        raise ValueError("glinter checks failed")
```
Dexterp37: This currently breaks runtime, right? On other platforms we only print _warnings_, we don't actually break the build, unless users enable "warning as errors".

If we were to follow the same pattern, we'd only print to stderr without raising an exception: what do you think?

mdboom: Ah, good point.  That would be consistent with our other platforms.  I'll make this just print warnings for now and we can consider flipping the switch to warnings-as-errors in lock step for everyone later.

PR #707, opened 2020-02-12
Bug 1614948 - Make `Glean.initialize` run off the main thread in iOS

travis79:


glean-core/ios/Glean/Glean.swift
```diff
@@ -73,90 +73,84 @@ public class Glean {
             return
         }
 
-        self.registerPings(Pings.shared)
-
         self.configuration = configuration
+        setUploadEnabled(uploadEnabled)
+
+        // Execute startup off the main thread
+        Dispatchers.shared.launchConcurrent {
```
Dexterp37: I'm a bit confused by the different queues on iOS, but I guess that's due to my lack of knowledge of this platform. On the Kotlin side, the whole init runs in the `Dispatchers.API` context (as the first thing after the currently executed operation, if any) preventing any other `Dispatchers.API.launch` call from being executed. Is this the case here? Can things executed with `Dispatchers.shared.launchConcurrent` run at the same time of `Dispatchers.shared.serialOperationQueue`?

Dexterp37:
Adding @badboy for his swift expertise.

glean-core/ios/Glean/Glean.swift
```diff
@@ -73,90 +73,84 @@ public class Glean {
...
+            if !isFirstRun {
                 glean_clear_application_lifetime_metrics()
                 self.initializeCoreMetrics()
             }
-        }
 
-        // Signal Dispatcher that init is complete
-        Dispatchers.shared.flushQueuedInitialTasks()
+            // Signal Dispatcher that init is complete
+            Dispatchers.shared.flushQueuedInitialTasks()
```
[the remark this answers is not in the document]
travis79: Not a problem, there is a `waitUntilAllTasksAreFinished()` function that accomplishes this nicely.  It probably would have saved me some trouble with a test that broke due to making this async...

glean-core/ios/Glean/Glean.swift
```diff
@@ -73,90 +73,84 @@ public class Glean {
...
                     self.submitPingByNameSync(
                         pingName: "baseline",
                         reason: "dirty_startup"
                     )
                 }
             }
-        }
 
-        self.observer = GleanLifecycleObserver()
+            self.observer = GleanLifecycleObserver()
```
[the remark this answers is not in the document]
travis79: Not on iOS, fortunately :)

glean-core/ios/Glean/Glean.swift
travis79: The concurrent queue is just that, it allows more than one task to run on it concurrently.  I'm using that particular queue here because if we added this to the 'serial' queue, any serial tasks that needed to execute during the init would be blocked until the init returned, which would not allow us to run and await the flushing of the queued tasks (or any other `launchAPI` block)

travis79:
> In addition to my comments below, did you take this for a test run?

Yes, it seemed to shave some milliseconds off the launch of the sample app, I was planning on testing this in Lockwise and doing a more in depth profile since Lockwise does more 'real' work on startup.

Dexterp37:
> 
> 
> > In addition to my comments below, did you take this for a test run?
> 
> Yes, it seemed to shave some milliseconds off the launch of the sample app, I was planning on testing this in Lockwise and doing a more in depth profile since Lockwise does more 'real' work on startup.

Great! Did you use the debug view to generate some pings and make sure they look ok? That's my main concern :)

travis79:
> > > In addition to my comments below, did you take this for a test run?
> > 
> > 
> > Yes, it seemed to shave some milliseconds off the launch of the sample app, I was planning on testing this in Lockwise and doing a more in depth profile since Lockwise does more 'real' work on startup.
> 
> Great! Did you use the debug view to generate some pings and make sure they look ok? That's my main concern :)

The pings appeared to be okay, even without awaiting the `flushInitialTaskQueue`.  I did have some problems with one test (the one about preserving application lifetime metrics) that just required me to think about when it was appropriate to check if the metric was cleared after startup.

travis79:
After @badboy has a look, I'll retrigger the iOS tests one final time

glean-core/ios/Glean/Glean.swift
travis79: After re-reading your question, allow me to expand on this a bit:  during the init, and up until we flush the queue, any `launchAPI` tasks would be just added to the queue.  Since the flushing of the queue happens almost at the end of `initialize`, it could be possible for tasks that run via `launchAPI` on the serial queue while `initialize` is sending the "dirty_startup" baseline ping, registering the lifecycle observers, or sending the deletion request ping.  If you see any problem with this, please let me know as I will have to change how `flushQueuedInitialTasks` works because, as I mentioned in the previous comment, executing `initialize` on the serial queue would prevent executing any of the queued tasks until after `initialize` returned, and we wouldn't be able to await them.  In this case, we could just execute them synchronously, I suppose...

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -164,22 +164,17 @@ class Dispatchers {
         concurrentOperationsQueue.cancelAllOperations()
     }
 
-    /// Stop queuing tasks and process any tasks in the queue.
+    ///  Stop queuing tasks and process any tasks in the queue.
```
badboy: Double space at the beginning. We do single-space usually.

glean-core/ios/Glean/Glean.swift
```diff
@@ -498,5 +492,6 @@ public class Glean {
         // Init Glean.
         testDestroyGleanHandle()
         initialize(uploadEnabled: true, configuration: configuration)
+        Dispatchers.shared.concurrentOperationsQueue.waitUntilAllOperationsAreFinished()
```
badboy: I understand this is so that it effectively becomes a synchronous init when testing.
Do we do something similar on the Kotlin side as well?

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -166,13 +164,11 @@ class MetricsPingScheduler {
...
             // context, see bug 1604861 and the implementation of
             // `collectPingAndReschedule`.
-            Dispatchers.shared.serialOperationQueue.addOperation {
-                self.collectPingAndReschedule(
-                    now,
-                    startupPing: true,
-                    reason: .overdue
-                )
-            }
+            self.collectPingAndReschedule(
```
badboy: This negates the whole block of comments above it.
Can you clarify why that is fine? Possibly adding that as a comment.

I suspect it is because now it's happening in the async startup, but there's no guarantuee other than us calling it there, or is there?

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
```diff
@@ -201,13 +197,10 @@ class MetricsPingScheduler {
...
             // batched up and replayed after any startup metrics ping is sent. To guarantee
-            // that, we dispatch this function from `Dispatchers.API.executeTask`. However,
-            // Pings.metrics.submit() ends up calling `Dispatchers.API.launch` again which
-            // will delay the ping collection task after any pending metric recording is
-            // executed, breaking the 'metrics' ping promise of sending a startup 'metrics'
-            // ping only containing data from the previous session.
-            // To prevent that, we synchronously manually dispatch the 'metrics' ping, without
-            // going through our public API.
+            // that we synchronously and manually dispatch the 'metrics' ping without
+            // going through our public API.  Since this part of the `Glean.initialize`
```
badboy: I'm not super happy to tie this so closely to where it  is called in initialize.
I'd rather phrase this in terms of `startupPing`, because a call with `startupPing = true` should only ever happen at most once per application run.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -259,18 +259,11 @@ class MetricsPingSchedulerTests: XCTestCase {
...
         stub(condition: isHost(host)) { data in
-            let body = (data as NSURLRequest).ohhttpStubs_HTTPBody()
-            let json = try! JSONSerialization.jsonObject(with: body!, options: []) as? [String: Any]
-            XCTAssert(json != nil)
-            let metrics = json?["metrics"] as? [String: Any]
-            let strings = metrics?["string"] as? [String: Any]
-
-            // Ensure there is only the expected metric
-            XCTAssertEqual(1, strings?.count, "Must contain only the expected metric")
+            let docType = (data as NSURLRequest).url!.path
```
badboy: In other tests we are a bit more explicit, checking the right fragment of the path:

```swift
request.url?.path.split(separator: "/")[2]
```

glean-core/ios/Glean/Glean.swift
badboy: Worth noting: The only other call to `launchConcurrent` left is inside the uploader to upload already written data in a thread.
The only way to get there is by sending a ping, which itself is gated by a `launchAPI` (for all external users at least). Thus there's no other `launchCorrent` happening before this task here flushes the queue.

badboy: I therefore think using this queue for the init and the other queue for the normal tasks is the correct usage.

glean-core/ios/Glean/Glean.swift
travis79: I had originally put this here to try and solve an issue with a particular test that broke with the async changes.  I've now determined that this wait is no longer necessary so I'll remove it.

glean-core/ios/Glean/Scheduler/MetricsPingScheduler.swift
travis79: This is because this part of `schedule` no longer needs to be called off the main thread because `initialize` is called off the main thread, and since this is being called synchronously (in the background thread) before we flush the queued tasks, it's guaranteed to be called before any metric API adds data when the queue is flushed. 

I'll update the comment to reflect this.

glean-core/ios/Glean/Glean.swift
Dexterp37: @travis79 can you add a comment about what @badboy said?

travis79: Yes, sorry but this is already merged.  I'll open a new PR and add some usage notes about the use of `launchConcurrent` vs `launchAPI`

travis79: > @travis79 can you add a comment about what @badboy said?

[Bug filed to take care of the documentation](https://bugzilla.mozilla.org/show_bug.cgi?id=1615603)

PR #711, opened 2020-02-13
Process ping upload response in upload module

brizental:
Resolves [Bug 1614283](https://bugzilla.mozilla.org/show_bug.cgi?id=1614283)

**I didn't create a whole new file, I just renamed `utils.rs` to `directory.rs`** ~I added some comments to the file to make it easier to see what was changed, apart from the rename. Git shows it as if I had written it from scratch and deleted to other one, wich is not the case.~ I changed the commits so that you can see exaclty what was changed in the file, by looking at 2ac4708.

Added the processing functionality, copied over from https://github.com/mozilla/glean/blob/master/glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt#L69.

Also, because I was looking at the android code in depth again, I realized some of the logging in there was not mirrored in this new implementation. I believe we would like to have at least all the logs we have there, so I copied over what was missing.


glean-core/src/upload/directory.rs
```diff
@@ -0,0 +1,342 @@
...
+                log::warn!("File name is not valid unicode: {}", path.display());
+            }
+            file_name
+        }
+    }
+}
+
+/// Manages the pending pings directories.
+#[derive(Debug)]
+pub struct PingDirectoryManager {
```
brizental: I moved the functinality of this file into a struct, mainly so that it would retain the data_path without me having to pass it to every function.

glean-core/src/upload/directory.rs
```diff
@@ -0,0 +1,342 @@
...
+    /// Attempts to delete a ping file.
+    ///
+    /// ## Arguments
+    ///
+    /// * `uuid` - The UUID of the ping file to be processed
+    ///
+    /// ## Panics
+    ///
+    /// Won't panic if unable to delete the file.
+    pub fn delete_file(&self, uuid: &str) {
```
brizental: Having the `data_path` saved, allowed me to do things such as only pass the `uuid` to the `delete_file` and `process_file` functions.

Apart from that, I only changed a little bit the structure of the docstrings here to match the structure of it in the other files and adapted the tests to this new structure.

glean-core/src/upload/mod.rs
```diff
@@ -140,6 +145,73 @@ impl PingUploadManager {
...
+    /// * **Any other error**
+    ///   For any other error, a warning is logged and the ping is re-enqueued.
+    ///   _Known other errors:_
+    ///   * 500 - internal error
+    ///
+    /// # Arguments
+    ///
+    /// `uuid` - The UUID of the ping in question.
+    /// `status` - The HTTP status of the response.
+    pub fn process_ping_upload_response(&self, uuid: &str, status: u16) {
```
brizental: Should I spawn a new thread for all the IO that happens here?

Dexterp37: Do you mean for the `directory_manager.delete_file(uuid)` parts? Yes, that would be ideal.

glean-core/src/upload/mod.rs
```diff
@@ -293,4 +360,78 @@ mod test {
         // Verify that after all requests are returned, none are left
         assert_eq!(upload_manager.get_upload_task(), PingUploadTask::Done);
     }
+
+    #[test]
+    fn test_processes_correctly_different_upload_responses() {
```
Dexterp37: nit: can you please break this test into smaller unit tests? see [this Kotlin code](https://github.com/mozilla/glean/blob/master/glean-core/android/src/test/java/mozilla/telemetry/glean/net/HttpURLConnectionUploaderTest.kt) for some related test coverage. You might also think to convert these tests to Rust.

glean-core/src/upload/directory.rs
```diff
@@ -0,0 +1,342 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! Pings directory processing utilities.
+
+use std::cmp::Ordering;
```
Dexterp37: Was this moved with `git mv`?

brizental: Yes, unfortunatelly I didn't realise if I did that and then made changes to file and added those changes, git disconsidered the mv. Anyways, I am fixing the commits and will push the new ones in a bit.

glean-core/src/upload/mod.rs
brizental: Yes. About converting the tests to Rust, do you see any specific test case that I haven't covered in my tests? I did based myself off of these and can't see any that made sense and I didn't copy over.

glean-core/src/upload/mod.rs
```diff
@@ -140,6 +145,100 @@ impl PingUploadManager {
...
+    ///
+    /// # Return value
+    ///
+    /// Returns the JoinHandle for the spawned thread for test syncronization purposes.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if unable to spawn a new thread.
+    pub fn process_ping_upload_response(&self, uuid: &str, status: u16) -> thread::JoinHandle<()> {
+        let mut task = "";
```
brizental: What would be a better way to implement this? With this approach I get a lint error about unused_assignments.

badboy:
(side note: here's a link to skip the file move in the review: https://github.com/mozilla/glean/pull/711/files/6ca876bc0c3b1e0883b455e9d8da7dfc42186fb0..542bb9a6c65dfb9cb3e7db5d129186770276b8a7)

glean-core/src/upload/directory.rs
```diff
@@ -45,87 +38,154 @@ fn get_file_name_as_str(path: &Path) -> Option<&str> {
...
+    /// Get the pending pings directory path.
+    pub fn get_dir(&self) -> PathBuf {
+        self.data_path.join("pending_pings")
+    }
+
+    /// Attempts to delete a ping file.
+    ///
+    /// ## Arguments
+    ///
+    /// * `uuid` - The UUID of the ping file to be processed
```
badboy:
```suggestion
    /// * `uuid` - The UUID of the ping file to be deleted.
```

glean-core/src/upload/directory.rs
```diff
@@ -45,87 +38,154 @@ fn get_file_name_as_str(path: &Path) -> Option<&str> {
...
+        let path = self.get_dir().join(uuid);
+        match fs::remove_file(&path) {
+            Err(e) => log::error!("Error deleting file {}. {}", path.display(), e),
+            _ => log::info!("Files was deleted {}", path.display()),
+        };
+    }
+
+    /// Reads a ping file and returns a `PingRequest` from it.
+    ///
+    /// If the file is not properly formatted, it will be deleted.
```
badboy:
```suggestion
    /// If the file is not properly formatted, it will be deleted and `None` will be returned.
```

glean-core/src/upload/directory.rs
```diff
@@ -45,87 +38,154 @@ fn get_file_name_as_str(path: &Path) -> Option<&str> {
...
+    /// ## Arguments
+    ///
+    /// * `uuid` - The UUID of the ping file to be processed
+    ///
+    /// ## Panics
+    ///
+    /// Will panic if unable to read the file.
+    pub fn process_file(&self, uuid: &str) -> Option<PingRequest> {
+        let path = self.get_dir().join(uuid);
+        let file = File::open(path).expect("Should be able to read ping file.");
```
badboy: And if not?
It might be missing/inaccessible/something else. Let's handle that gracefully and return None.

```suggestion
        let file = match File::open(path).expect("Should be able to read ping file.") {
        	Ok(file) => file,
        	Err(_) => return None,
        };
```

glean-core/src/upload/directory.rs
```diff
@@ -45,87 +38,154 @@ fn get_file_name_as_str(path: &Path) -> Option<&str> {
...
+    /// Attempts to delete a ping file.
+    ///
+    /// ## Arguments
+    ///
+    /// * `uuid` - The UUID of the ping file to be processed
+    ///
+    /// ## Panics
+    ///
+    /// Won't panic if unable to delete the file.
+    pub fn delete_file(&self, uuid: &str) {
```
badboy: Does this need to be public?

glean-core/src/upload/directory.rs
```diff
@@ -140,14 +200,19 @@ mod test {
 
     #[test]
     fn test_doesnt_panic_if_no_pending_pings_directory() {
+        // Create a new directory_manager
         let dir = tempfile::tempdir().unwrap();
-        assert!(process_pings_dir(&dir.path()).is_err());
+        let tmpname = dir.path().display().to_string();
+        let directory_manager = PingDirectoryManager::new(&tmpname);
```
badboy: He, if the manager takes a `PathBuf` we don't even need to go through the display implementation.

glean-core/src/upload/directory.rs
```diff
@@ -140,14 +200,19 @@ mod test {
 
     #[test]
     fn test_doesnt_panic_if_no_pending_pings_directory() {
+        // Create a new directory_manager
```
badboy: That's clear from the code below.

glean-core/src/upload/mod.rs
```diff
@@ -49,6 +47,8 @@ pub enum PingUploadTask {
 pub struct PingUploadManager {
     /// A FIFO queue storing a `PingRequest` for each pending ping.
     queue: Arc<RwLock<VecDeque<PingRequest>>>,
+    /// A manager for the pending pings directories.
+    directory_manager: Arc<RwLock<PingDirectoryManager>>,
```
badboy: Do we really need it behind a lock and shared?
It's essentially a one-off object solely for processing.
To me it seems like it's better to create it when needed in the thread and keep another instance around in the upload manager

glean-core/src/upload/mod.rs
badboy: Don't use a string for what is essentially an enum to begin with.

glean-core/src/upload/mod.rs
```diff
@@ -140,6 +145,99 @@ impl PingUploadManager {
...
+                );
+                task = "re-enqueue";
+            }
+        };
+
+        // Spawn a new thread to perform the task defined above.
+        let local_queue = self.queue.clone();
+        let local_manager = self.directory_manager.clone();
+        let uuid = uuid.to_owned();
+        thread::Builder::new()
```
badboy: Hm, if we continue with threading on the Rust side we eventually should keep a thread pool around (creating short-lived threads is not free).

One question: do we _need_ the thread here for I/O? Shouldn't this be off-thread already on the platform side?

glean-core/src/upload/mod.rs
```diff
@@ -140,6 +145,99 @@ impl PingUploadManager {
...
+    ///   * 500 - internal error
+    ///
+    /// # Arguments
+    ///
+    /// `uuid` - The UUID of the ping in question.
+    /// `status` - The HTTP status of the response.
+    ///
+    /// # Return value
+    ///
+    /// Returns the JoinHandle for the spawned thread for test syncronization purposes.
```
badboy:
```suggestion
    /// Returns a handle for the spawned I/O thread. Callers can use this to synchronize on the I/O processing.
```

I'm not a fan of making a return value solely documented for a test purpose.
What do you think about the alternate phrasing?

Is a (non-test) caller supposed to do anything with the thread?

glean-core/src/upload/mod.rs
brizental: Yes, it should be called off the main thread. That is mostly the reason I asked if I should spawn a thread for it in the first place. It is not immediatelly obvious to me I should, since it is already off the main thread on the platform side.

@Dexterp37 advised me to spawn a thread for disk IO whenever possible. Could ou maybe elaborate more on why, Alessio?

glean-core/src/upload/mod.rs
brizental: > To me it seems like it's better to create it when needed in the thread and keep another instance around in the upload manager.

Hm, I hdan't considered that. I'll give it a try.

glean-core/src/upload/directory.rs
brizental: Yes, it is used in the `process_ping_upload_response` function.

glean-core/src/upload/mod.rs
brizental: It _can_ do something with the thread if it wants to. But for the uses I forsee for it, I don't think it will. I do prefer your phrasing though.

glean-core/src/upload/mod.rs
Dexterp37: @badboy @brizental is this the case on all platforms? I'm sure that's the case for Android, but I'm not sure that's the case for Python/iOS.

If that's a problem for this PR, I'm fine with taking this to a separate one and use a thread-pool instead (which is a good suggestion regardless!)

badboy: The upload manager runs on the separate concurrent queue on iOS.
Python is largely unsolved right now (there's an open bug how we want to handle it).

I'd be ok to simplify it here to not use the thread, file a bug and we can investigate/discuss our stance on doing threading more wide-spread in Rust

Dexterp37: All right, sounds good to me :) Let's at the very least add a comment here, explaining why we're ok for this being on the caller thread, putting in words the assumption of not being called on the main thread.

Let's also mention the known state of the other platforms.

glean-core/src/upload/directory.rs
```diff
@@ -45,87 +38,154 @@ fn get_file_name_as_str(path: &Path) -> Option<&str> {
...
+    /// Creates a new directory manager.
+    ///
+    /// # Arguments
+    ///
+    /// * `data_path` - Path to the pending pings directory.
+    ///
+    /// # Panics
+    ///
+    /// Will panic if unable to parse the data_path.
+    pub fn new(data_path: &str) -> Self {
```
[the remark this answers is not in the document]
brizental: The problem here is that for me to be able to use this `data_path` value in the spawned thread it has to add `+ Send + 'static`. Do you know a better way to work around this? By adding the `+ Send + 'static` the easiest way is to pass a string slice.

I could also keep the type like you say, but to be able to move that to a spawned thread I would have to turn it into a PathBuf and then into a String.

I am sure I must be missing something here, hehe. Let me know your thoughts, @badboy :)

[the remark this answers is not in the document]
brizental: I was able to get it working with something like this:
```rust
pub fn new<P: Into<PathBuf>>(data_path: P) -> Self {
        let queue = Arc::new(RwLock::new(VecDeque::new()));
        let directory_manager = PingDirectoryManager::new(data_path);
        let processed_pending_pings = Arc::new(AtomicBool::new(false));

        let local_queue = queue.clone();
        let local_flag = processed_pending_pings.clone();

        /* Important part below */
        let local_data_path = directory_manager.data_path.to_str().unwrap().to_owned();

        let _ = thread::Builder::new()
        ...etc
```

But like I said, I must be missing something.

glean-core/src/upload/mod.rs
brizental: Ok, I filed a bug (https://bugzilla.mozilla.org/show_bug.cgi?id=1615597) and will add comments :)

glean-core/src/upload/directory.rs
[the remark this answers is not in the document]
badboy: Turn the data path into a `PathBuf` as early as possible:
```rust
let data_path = data_path.into()
```

Make `PingDirectoryManager` clonable and move that into the thread.

```rust
let dir_manager = directory_manager.clone()
```

requires

```
#[derive(Clone)]
struct PingDirectoryManager
```

(and `struct PingDirectoryManager` then takes the same generic `P: PathBuf` on `new`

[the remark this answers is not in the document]
brizental: Thank you, that works beautifully :)

PR #712, opened 2020-02-14
Upgrade to glean_parser v1.18.2

mdboom:


docs/user/collected-metrics/metrics.md
```diff
@@ -35,14 +50,38 @@ The following metrics are added to the ping:
...
 This ping contains no metrics.
 
 ## metrics
+
 This is a built-in ping that is assembled out of the box by the Glean SDK.
+
 See the Glean SDK documentation for the [`metrics` ping](https://mozilla.github.io/glean/book/user/pings/metrics.html).
+
+Reasons this ping may be sent:
+    - `overdue`: The last ping wasn't submitted on the current calendar day, but it's after
```
Dexterp37: Oh, I think there's still a bug here: if you see the rendered version, you'll notice that we still need a newline before this otherwise `overdue` will get printed on the same line as "Reasons this ping may be sent"

mdboom: Good catch.  I'll have to go back to glean_parser and make a new release ... blablabla...

PR #716, opened 2020-02-18
Clarify the usage of 'lifetime: application' [doc only]

Dexterp37:
This might help with future cases similar to the one for the GFX metrics.

### DIFF
--- a/docs/user/metric-parameters.md
+++ b/docs/user/metric-parameters.md
@@ -27,7 +27,7 @@
 
 - `lifetime`: Defines the lifetime of the metric. Different lifetimes affect when the metrics value is reset.
   - `ping` (default): The metric is reset each time it is sent in a ping.
-  - `application`: The metric is related to an application run, and is reset when the application restarts.
+  - `application`: The metric is related to an application run, and is reset when the application restarts. Care should be taken when using this lifetime: if the metric is sent with the [`metrics` ping](pings/metrics.md), set its value as early as possible and make sure that the value is set on every application run, otherwise it may be missing from some pings.
   - `user`: The metric is part of the user's profile.
     
 - `send_in_pings`: Defines which pings the metric should be sent on. If not specified, the metric is sent on the "default ping", which is the `events` ping for events and the `metrics` ping for everything else. Most metrics don't need to specify this unless they are sent on [custom pings](pings/custom.md).


docs/user/metric-parameters.md
```diff
@@ -27,7 +27,7 @@
 
 - `lifetime`: Defines the lifetime of the metric. Different lifetimes affect when the metrics value is reset.
   - `ping` (default): The metric is reset each time it is sent in a ping.
-  - `application`: The metric is related to an application run, and is reset when the application restarts.
+  - `application`: The metric is related to an application run, and is reset when the application restarts. Care should be taken when using this lifetime: if the metric is sent with the [`metrics` ping](pings/metrics.md), set its value as early as possible to minimise the possibility of it being missing from the ping.
```
travis79: nit: `minimise` would probably be correct for British English, but I believe the more common spelling would be `minimize`

```suggestion
  - `application`: The metric is related to an application run, and is reset when the application restarts. Care should be taken when using this lifetime: if the metric is sent with the [`metrics` ping](pings/metrics.md), set its value as early as possible to minimize the possibility of it being missing from the ping.
```

chutten: In absence of a style guide, I think you can use whichever form you'd like. Doesn't even have to be English : D

(( fwiw, Canadian English goes with `ize` in most cases as well, so I probably have a bias here. ))

travis79: I'm also biased, which is why I relegated this to the level of "nit" :)

chutten:
> Is it not the case that the issue with GV metrics is that GV isn't always initialized? i.e. only if the user views web content? So maybe the recommendation is "only use for things that are sent every time during application startup?"

Yes and no. GV's problem appears to be a mismatch between "application" and "ping" lifecycles. If ever the ping outlives the application (even in our new extended vision of "application" lifecycles wrt the at-startup "metrics" ping pretending it's from the previous app session), then attempting to pair "ping"-lifetime metrics to "application"-lifetime metrics will always have a problem where the latter will sometimes be null.

Looking at it a different way: if we recommend that "application"-lifetime metrics be only ever set during app startup, they become indistinguishable from "user" lifetime metrics: present in every ping, reflective of whatever the most recent value was that was set.

Dexterp37:
> 
> 
> > Is it not the case that the issue with GV metrics is that GV isn't always initialized? i.e. only if the user views web content? So maybe the recommendation is "only use for things that are sent every time during application startup?"
> 
> Yes and no.

I think "yes": the problem we had would have been present even on custom pings. There were some product workflows that did not initialize GeckoView _at all_. When that happened, while application lifetime metrics were cleared at startup, they were not set again.

> GV's problem appears to be a mismatch between "application" and "ping" lifecycles. If ever the ping outlives the application (even in our new extended vision of "application" lifecycles wrt the at-startup "metrics" ping pretending it's from the previous app session), then attempting to pair "ping"-lifetime metrics to "application"-lifetime metrics will always have a problem where the latter will sometimes be null.

Pairing lifetimes is a separate issue, IMHO: for this specific case ping generated at startup will have application-lifetime metrics from the previous session. There will be "holes" if a new metric ping is generated *but* GV doesn't start.

> Looking at it a different way: if we recommend that "application"-lifetime metrics be only ever set during app startup, they become indistinguishable from "user" lifetime metrics: present in every ping, reflective of whatever the most recent value was that was set.

I don't think I agree with the "they become equal" part, but you do have a point about the recommendation. I think this whole problem really boils down to "If you expect some data to be present, make sure you set it".

chutten:
> I don't think I agree with the "they become equal" part, but you do have a point about the recommendation. I think this whole problem really boils down to "If you expect some data to be present, make sure you set it".

WHAAAAAAAT? :expressionless:

Dexterp37:
I took another stab at this, trying to make it clearer and less controversial :)

PR #717, opened 2020-02-18
Support deletion request pings in upload module

brizental:
Resolves [Bug 1614280](https://bugzilla.mozilla.org/show_bug.cgi?id=1614280)

After a good discussion in the bug I decided to take a rather simplistic approach to implementing this.

This is my understanding:
1. I don't need to worry about stale ping files when a user has disabled upload, that is taken care of elsewhere (https://github.com/mozilla/glean/blob/master/glean-core/src/lib.rs#L332).
2. Even if a client has upload enabled, if there is a deletion-request ping file we should send that.

My solution was to have the DirectoryManager always check the deletion_request directory, regardless of upload enabled or not.

### DIFF
--- a/glean-core/src/upload/directory.rs
+++ b/glean-core/src/upload/directory.rs
@@ -14,7 +14,9 @@ use serde_json::Value as JsonValue;
 use uuid::Uuid;
 
 use super::PingRequest;
-use crate::Result;
+
+pub const PENDING_PINGS_DIRECTORY: &str = "pending_pings";
+pub const DELETION_REQUEST_PINGS_DIRECTORY: &str = "deletion_request";
 
 /// Get the file name from a path as a &str.
 ///
@@ -56,9 +58,12 @@ impl PingDirectoryManager {
         }
     }
 
-    /// Get the pending pings directory path.
-    pub fn get_dir(&self) -> PathBuf {
-        self.data_path.join("pending_pings")
+    /// Get the paths to the directories that may contain ping files.
+    fn get_dirs(&self) -> [PathBuf; 2] {
+        [
+            self.data_path.join(PENDING_PINGS_DIRECTORY),
+            self.data_path.join(DELETION_REQUEST_PINGS_DIRECTORY),
+        ]
     }
 
     /// Attempts to delete a ping file.
@@ -71,7 +76,13 @@ impl PingDirectoryManager {
     ///
     /// Won't panic if unable to delete the file.
     pub fn delete_file(&self, uuid: &str) {
-        let path = self.get_dir().join(uuid);
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            None => {
+                log::error!("Cannot find ping file to delete {}", uuid);
+                return;
+            }
+        };
         match fs::remove_file(&path) {
             Err(e) => log::error!("Error deleting file {}. {}", path.display(), e),
             _ => log::info!("Files was deleted {}", path.display()),
@@ -86,13 +97,22 @@ impl PingDirectoryManager {
     ///
     /// * `uuid` - The UUID of the ping file to be processed
     pub fn process_file(&self, uuid: &str) -> Option<PingRequest> {
-        let path = self.get_dir().join(uuid);
-        let file = match File::open(path) {
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            None => {
+                log::error!("Cannot find ping file to process {}", uuid);
+                return None;
+            }
+        };
+        let file = match File::open(&path) {
             Ok(file) => file,
-            Err(_) => return None,
+            Err(e) => {
+                log::error!("Error reading ping file {}. {}", path.display(), e);
+                return None;
+            }
         };
 
-        log::info!("Processing ping: {}", uuid);
+        log::info!("Processing ping at: {}", path.display());
 
         // The way the ping file is structured,
         // first line should always have the url
@@ -128,20 +148,17 @@ impl PingDirectoryManager {
     ///
     /// # Return value
     ///
-    /// `Result<Vec<PingRequest>>` - see [`PingRequest`](struct.PingRequest.html) for more information.
-    ///                              This will only be `Err` in case it is unable to read the pings directory.
-    pub fn process_dir(&self) -> Result<Vec<PingRequest>> {
-        let pings_dir = self.get_dir();
-
-        log::info!("Processing persisted pings at {}", pings_dir.display());
+    /// `Vec<PingRequest>` - see [`PingRequest`](struct.PingRequest.html) for more information.
+    pub fn process_dir(&self) -> Vec<PingRequest> {
+        log::info!("Processing persisted pings.");
 
         // Walk the pings directory and process each file in it,
         // deleting invalid ones and ignoring unreadable ones.
         // Create a vector of tuples: (modified_date, PingRequest)
         // using the contents and metadata of all valid files.
-        let mut pending_pings: Vec<_> = pings_dir
-            .read_dir()?
-            .filter_map(|entry| entry.ok())
+        let mut pending_pings: Vec<_> = self
+            .get_ping_entries()
+            .into_iter()
             .filter_map(|entry| {
                 let path = entry.path();
                 if let Some(file_name) = get_file_name_as_str(&path) {
@@ -175,10 +192,35 @@ impl PingDirectoryManager {
         });
 
         // Return the vector leaving only the `PingRequest`s in it
-        Ok(pending_pings
+        pending_pings
             .into_iter()
             .map(|(_, request)| request)
-            .collect())
+            .collect()
+    }
+
+    /// Get all the ping entries in all ping directories.
+    fn get_ping_entries(&self) -> Vec<fs::DirEntry> {
+        let mut result = Vec::new();
+        for dir in &self.get_dirs() {
+            if let Ok(entries) = dir.read_dir() {
+                result.extend(entries.filter_map(|entry| entry.ok()))
+            };
+        }
+        result
+    }
+
+    /// Get the path for a ping file based on its uuid.
+    ///
+    /// Will look for files in each ping directory until something is found.
+    /// If nothing is found, returns `None`.
+    fn get_file_path(&self, uuid: &str) -> Option<PathBuf> {
+        for dir in &self.get_dirs() {
+            let path = dir.join(uuid);
+            if path.exists() {
+                return Some(path);
+            }
+        }
+        None
     }
 }
 
@@ -198,7 +240,7 @@ mod test {
         let directory_manager = PingDirectoryManager::new(dir.path());
 
         // Verify that processing the directory didn't panic
-        assert!(directory_manager.process_dir().is_err());
+        assert_eq!(directory_manager.process_dir().len(), 0);
     }
 
     #[test]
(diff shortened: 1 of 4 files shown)


glean-core/src/upload/directory.rs
```diff
@@ -175,10 +180,35 @@ impl PingDirectoryManager {
...
+            };
+        }
+        result
+    }
+
+    /// Get the path for a ping file based on its uuid.
+    ///
+    /// Will look for files in each ping directory until something is found.
+    /// If nothing is found, returns `None`.
+    fn get_file_path(&self, uuid: &str) -> Option<PathBuf> {
```
brizental: I did this so that we didn't need to always know which kind of ping it is we are searching for. 

Since most times we will be looking for ordinary pings, it doesn't really change much, we never even go into the deletion_request directory (I was careful about ordering the dirs slice to have the pending_pings folder path first). 

If we are indeed looking for a deletion-request ping file, then it will scan the pending_pings directory before going into the deletion_request directory, which is not ideal but shouldn't be a big problem, since the pending_pings directory will be (or should be) empty by then.

glean-core/src/upload/directory.rs
```diff
@@ -71,7 +73,13 @@ impl PingDirectoryManager {
     ///
     /// Won't panic if unable to delete the file.
     pub fn delete_file(&self, uuid: &str) {
-        let path = self.get_dir().join(uuid);
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            _ => {
+                log::info!("Cannot find ping file to delete {}", uuid);
```
Dexterp37: nit: is this an error? should we use `log::error!` instead?

glean-core/src/upload/directory.rs
```diff
@@ -56,9 +55,12 @@ impl PingDirectoryManager {
         }
     }
 
-    /// Get the pending pings directory path.
-    pub fn get_dir(&self) -> PathBuf {
-        self.data_path.join("pending_pings")
+    /// Get the paths to the directories that may contain ping files.
+    fn get_dirs(&self) -> [PathBuf; 2] {
```
Dexterp37: TIL about fixed size array declaration in Rust :)

glean-core/src/upload/directory.rs
```diff
@@ -86,13 +94,16 @@ impl PingDirectoryManager {
     ///
     /// * `uuid` - The UUID of the ping file to be processed
     pub fn process_file(&self, uuid: &str) -> Option<PingRequest> {
-        let path = self.get_dir().join(uuid);
-        let file = match File::open(path) {
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            _ => return None,
```
Dexterp37: Do we expect this to happen? If yes, can you document why? If not, can you at least log an error?

glean-core/src/upload/directory.rs
Dexterp37: > Since most times we will be looking for ordinary pings, it doesn't really change much, we never even go into the deletion_request directory (I was careful about ordering the dirs slice to have the pending_pings directory path first).

At the very least, we should document this requirement where you are declaring the dirs slice.

> If we are indeed looking for a deletion-request ping file, then it will scan the pending_pings directory before going into the deletion_request directory, which is not ideal but shouldn't be a big problem, since the pending_pings directory will be (or should be) empty by then.

Does this really matter? Why is order important in this case?

I think we should probably do a better job at _not_ sending data if upload is off. Removing non-'deletion' pings can still happen while the product asks Glean for things to upload, meaning we could be sending pings even though upload was off, because of racing. Thoughts?

glean-core/src/upload/mod.rs
```diff
@@ -377,7 +369,7 @@ mod test {
         }
 
         // Get the pending ping directory path
-        let pending_pings_dir = upload_manager.directory_manager.get_dir();
+        let pending_pings_dir = dir.path().join("pending_pings");
```
Dexterp37: Let's declare the directory name as a constant in `directory.rs` and use it here.

glean-core/src/upload/mod.rs
```diff
@@ -417,7 +409,7 @@ mod test {
...
```
Dexterp37: Let's declare the directory name as a constant in `directory.rs` and use it here.

glean-core/src/upload/directory.rs
```diff
@@ -56,9 +55,12 @@ impl PingDirectoryManager {
         }
     }
 
-    /// Get the pending pings directory path.
-    pub fn get_dir(&self) -> PathBuf {
-        self.data_path.join("pending_pings")
+    /// Get the paths to the directories that may contain ping files.
+    fn get_dirs(&self) -> [PathBuf; 2] {
+        [
+            self.data_path.join("pending_pings"),
```
Dexterp37: Let's add a comment about order being important here, since it will break tests otherwise.

glean-core/src/upload/directory.rs
brizental: > Why is order important in this case?

It is important based on the assumption that most pings we send are not deletion-request pings. Order is important to make sure we scan pending_pings directory before falling back to deletion_request directory, thus ensuring we don't perform the usually pointless step of reading the deletion_request directory.

Dexterp37: > 
> 
> > Why is order important in this case?
> 
> It is important based on the assumption that most pings we send are not deletion-request pings. Order is important to make sure we scan pending_pings directory before falling back to deletion_request directory, thus ensuring we don't perform the usually pointless step of reading the deletion_request directory.

So it is important because of performance? Not because we break functionality?

brizental: > I think we should probably do a better job at not sending data if upload is off. Removing non-'deletion' pings can still happen while the product asks Glean for things to upload, meaning we could be sending pings even though upload was off, because of racing. Thoughts?

This comment has made me go back and look again at the steps in `set_upload_enabled`. It goes like this:
1. Submit the deletion-request ping;
2. Change upload enabled flag;
3. Clear pending_pings_directory;
4. Clear stored metrics;
5. etc;

I guess if it is possible to have a race condition there and get a ping to send before finishing clearing all metrics on the rust side, we have that problem already.

In android at least we lock the the PingUploader, so that we don't iterate over pings while deleting them.

I guess I could in some way lock the pings queue from the moment we start clearing all the stores and queues until we finish it, so that we don't need to rely on the platform side doing that, but that should happen when I add the upload manager functionality to the Glean object, which will happen in my next PR.

glean-core/src/upload/directory.rs
brizental: Yes, it is an error.

glean-core/src/upload/directory.rs
brizental: No we dont expect it. I will log an error.

glean-core/src/upload/directory.rs
brizental: > So it is important because of performance? Not because we break functionality?

Yes, we skip some steps this way. So you could say, performance :)

Dexterp37: Ok, setting aside the race from the previous comment, if performance is the only remaining issue, I think it's a minor one: we expect this to be off the main thread. While this still means I/O, it doesn't necessarily mean frequent.

glean-core/src/upload/directory.rs
```diff
@@ -14,7 +14,9 @@ use serde_json::Value as JsonValue;
 use uuid::Uuid;
 
 use super::PingRequest;
-use crate::Result;
+
+pub const PENDING_PINGS_DIRECTORY: &str = "pending_pings";
+pub const DELETION_REQUEST_PINGS_DIRECTORY: &str = "deletion_request";
```
brizental: These values are also used in the `ping` module, where they are hardcoded. Would it make sense to have these constants be a shared on the crate level? So that any module can access them?

glean-core/src/upload/directory.rs
brizental: Is this still relevant after https://github.com/mozilla/glean/pull/717#discussion_r381227058?

glean-core/src/upload/directory.rs
badboy: Yes.

glean-core/src/upload/directory.rs
```diff
@@ -56,9 +58,12 @@ impl PingDirectoryManager {
...
     }
 
-    /// Get the pending pings directory path.
-    pub fn get_dir(&self) -> PathBuf {
-        self.data_path.join("pending_pings")
+    /// Get the paths to the directories that may contain ping files.
+    fn get_dirs(&self) -> [PathBuf; 2] {
+        [
+            self.data_path.join(PENDING_PINGS_DIRECTORY),
+            self.data_path.join(DELETION_REQUEST_PINGS_DIRECTORY),
```
badboy: We might eventually avoid the constant reallocation here and just store the final paths. But for now we can keep it like this.

glean-core/src/upload/directory.rs
```diff
@@ -71,7 +76,13 @@ impl PingDirectoryManager {
     ///
     /// Won't panic if unable to delete the file.
     pub fn delete_file(&self, uuid: &str) {
-        let path = self.get_dir().join(uuid);
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            _ => {
```
badboy: Personal preference: I would always list out the `None` here instead of a catch-all underscore.

glean-core/src/upload/directory.rs
brizental: This is a very small change and kind of a leftover from development. This function used to have logic in it, now that it soesn't anymore you are right that I could just store the final paths. If you think it is best to change it later alright, but  I can also just do it right now.

Dexterp37: Let's do it now please, this would allow us to move less unrelated code in follow-up prs.

glean-core/src/upload/directory.rs
```diff
@@ -86,13 +95,22 @@ impl PingDirectoryManager {
     ///
     /// * `uuid` - The UUID of the ping file to be processed
     pub fn process_file(&self, uuid: &str) -> Option<PingRequest> {
-        let path = self.get_dir().join(uuid);
-        let file = match File::open(path) {
+        let path = match self.get_file_path(uuid) {
+            Some(path) => path,
+            None => {
```
Dexterp37: @badboy if you have a strong opinion on this, any chance we could codify it in some style guide and/or linter?

glean-core/src/upload/directory.rs
Dexterp37: > I guess I could in some way lock the pings queue from the moment we start clearing all the stores and queues until we finish it, so that we don't need to rely on the platform side doing that, but that should happen when I add the upload manager functionality to the Glean object, which will happen in my next PR.

Let's add this as a TODO here (in a comment) and mention when it will happen (bug number).

brizental: Hm,  I am not sure that it makes sense to add a comment like this here. This function is considerably distant from where that code will be.

Since implementing this is what I will do immediatelly after closing this PR, do you agree to just add a comment to that bug, so that we know that is one important thing to do on it? (I did that already, btw. https://bugzilla.mozilla.org/show_bug.cgi?id=1614284#c1)

PR #718, opened 2020-02-18
Bug 1615603 - Better document the iOS background queues and their usage [doc only]

travis79:


glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -65,6 +75,11 @@ class Dispatchers {
     /// recording of metrics and things that need to execute in order. Since this executes the tasks on
     /// a non-concurrent (serial) queue, the tasks are executed in the order that they are launched.
     ///
+    /// **Important Note:** This queue should only be used for performing API functions, and not for
```
badboy: what are "API functions"? 
IMO that's too vague, this whole thing is an API (`Dispatchers.API`) making it more confusing.

glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -7,7 +7,12 @@ import Foundation
 /// This class manages background execution of Glean tasks.
 ///
 /// This class makes use of the higher level `Operation` and `OperationQueue` API's in order to allow
-/// for observable background operation with the capabilities to pause, cancel, and resume tasks
+/// for observable background operation with the capabilities to pause, cancel, and resume tasks.
+///
+/// There are two main queues created and used as part of `Dispatchers`.  The `serialOperationQueue`
+/// is a serially executed queue meant for API tasks and is serviced through `launchAPI`, the
```
badboy: maybe we should mention it like "serially, single-threaded queue" ?

glean-core/ios/Glean/Dispatchers.swift
badboy: What about something along these lines:

> **Note**: Tasks that should be processed in order and finish before successive tasks are run should be launched using `launchAPI`. That includes all metric recording functions.

And then on the `launchConcurrent` one something like

> **Note**: Tasks that can be processed asynchronously and can block for a long time, but which should not block other tasks, such as recording data, should be launched using `launchConcurrent`. That for example includes an upload task.

(ok, that first sentence is too long now, maybe you can come up with a better version?)

travis79: These are good suggestions, let me see what I can do with this...

PR #719, opened 2020-02-19
Validate a simple generated custom ping against the ping schema

badboy:
Ping schema is taken from https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/f2a7ce4/schemas/glean/glean/glean.1.schema.json
and vendored.
It was changed to include an "$originUrl" as a way to track which one we imported.

---

Through #717 we noticed we only validate pings against the schema in Kotlin and Python test suites.

This is for several reasons:

* Validation is done by `glean_parser`, a Python project. A Python environment is only around for Kotlin and Python test suites
* Pings generated by _only_ `glean_core` are not trivially valid, due to `client_info` data misssing. We can set it, but it's not done by default.

But we only require _jsonschema validation_, and that's independent of other work from `glean_parser`, so one thing we can do is use a Rust project for the Rust part.
Additionally we do have a project that acts as kind of a "platform layer": `glean_preview`. That allows us to set `client_info` data as part of the normal flow.

Notes:
* I vendored the schema from above URL (and also added the exact URL as `$sourceURL` in the schema itself. I wanted to avoid requiring an internet connection and also a way to download stuff
* This uses [valico](https://github.com/rustless/valico), the first crate that supports jsonschema validation and was stable. I have not looked at their code yet (I did check the license)
  * for example they say they support jsonschema draf-4, but we use draft-6 (?!), however the test works.
* Currently contains `doc only` in the commit to avoid runnning the tests. This should be a discussion if we want to do it like this, if so I can remove that and run CI tests.

This is what it would look like if it fails to validate:

(long output hidden behind a detail view)
<details>

```
[Running cargo test -p glean-preview --test schema]
   Compiling glean-preview v0.0.5 (/Users/jrediger/mozilla/src/glean.rs/glean-core/preview)
    Finished test [unoptimized + debuginfo] target(s) in 2.25s
     Running target/debug/deps/schema-486af64b29843148

running 1 test
test validate_against_schema ... FAILED

failures:

---- validate_against_schema stdout ----
thread 'validate_against_schema' panicked at 'Data: Object(
    {
        "client_info": Object(
            {
                "telemetry_sdk_build": String(
                    "25.0.0",
                ),
            },
        ),
        "ping_info": Object(
            {
                "end_time": String(
                    "2020-02-19T15:59+01:00",
                ),
                "seq": Number(
                    0,
                ),
                "start_time": String(
                    "2020-02-19T15:59+01:00",
                ),
            },
        ),
    },
), Validation: ValidationState {
    errors: [
        Required {
            path: "/client_info/app_build",
        },
        Required {
            path: "/client_info/app_display_version",
        },
        Required {
            path: "/client_info/architecture",
        },
        Required {
            path: "/client_info/device_manufacturer",
        },
        Required {
            path: "/client_info/device_model",
        },
        Required {
            path: "/client_info/first_run_date",
        },
        Required {
            path: "/client_info/os",
        },
        Required {
            path: "/client_info/os_version",
        },
    ],
    missing: [],
}', glean-core/preview/tests/schema.rs:81:5
note: run with `RUST_BACKTRACE=1` environment variable to display a backtrace.
```

</details>

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -228,7 +228,7 @@ jobs:
     resource_class: "medium+"
     steps:
       - test-rust:
-          rust-version: "1.36.0"
+          rust-version: "1.41.0"
 
   Rust FFI header check:
     docker:
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -324,9 +324,11 @@ version = "0.0.5"
 dependencies = [
  "env_logger 0.7.1 (registry+https://github.com/rust-lang/crates.io-index)",
  "glean-core 25.1.0",
+ "jsonschema-valid 0.3.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "lazy_static 1.4.0 (registry+https://github.com/rust-lang/crates.io-index)",
  "log 0.4.8 (registry+https://github.com/rust-lang/crates.io-index)",
  "once_cell 1.2.0 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde_json 1.0.44 (registry+https://github.com/rust-lang/crates.io-index)",
  "tempfile 3.1.0 (registry+https://github.com/rust-lang/crates.io-index)",
 ]
 
@@ -357,6 +359,15 @@ dependencies = [
 ]
 
 [[package]]
+name = "iri-string"
+version = "0.3.0"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+dependencies = [
+ "nom 5.1.1 (registry+https://github.com/rust-lang/crates.io-index)",
+ "serde 1.0.104 (registry+https://github.com/rust-lang/crates.io-index)",
+]
+
+[[package]]
 name = "iso8601"
 version = "0.3.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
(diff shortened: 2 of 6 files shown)


badboy:
> I think using a Draft 4 schema checker is a bad idea. Generally the additions after that are _more_ specific, so passing on Draft 4 doesn't really indicate passing on Draft 6. The use of different drafts has bitten us before, so I think we need to use Draft 6 here. 

I initially hoped the crate would just fail to use a schema from a newer draft.

> It's not personal, but are there reasons not to use `jsonschema-valid`? I know the developer and I'm sure he'd be amenable to any changes required :)

My choise wasn't personal either. I'm happy to switch this over, I didn't take a deeper look at your crate and made the wrong assumption it's not ready to use.
I'd be even happier to use something we* maintain (*you), makes our task easier if we ever need changes,

mdboom:
> I'd be even happier to use something we* maintain (*you), makes our task easier if we ever need changes,

Yeah -- it's basically feature complete and passes the standard JSON schema test suite (which I put a lot more faith in than any tests I would write).  The main known issue I have with it is that it uses a somewhat unusual-for-Rust callback pattern to report errors back.  Iterators would be more appropriate, but I've been struggling with the borrow checker in a branch on that stuff for a while.  If/when generators are available, that will make all of that much easier.

badboy:
>  The main known issue I have with it is that it uses a somewhat unusual-for-Rust callback pattern to report errors back. Iterators would be more appropriate, but I've been struggling with the borrow checker in a branch on that stuff for a while. If/when generators are available, that will make all of that much easier.

See, that's a thing we can tackle as we* (*you) control the crate. I'm happy to switch over then.

badboy:
Switched to @mdboom's `jsonschema-valid` crate (that was simple), but it's blocked on some updates over at mdboom/jsonschema-valid#5

PR #720, opened 2020-02-19
1610521: Document the timing and memory distribution function

mdboom:


mdboom:
image

.dictionary
```diff
@@ -97,6 +98,7 @@ ios
 janerik
 ktlint
 lang
+lfloor
```
mdboom: The aspell filter for markdown doesn't correctly ignore math chunks.  It's ok to just throw them in the dictionary for now, but it might get annoying if we try to make our documentation look like an academic paper.

docs/user/metrics/memory_distribution.md
```diff
@@ -3,6 +3,10 @@
 Memory distributions are used to accumulate and store memory sizes.
 
 Memory distributions are recorded in a histogram where the buckets have an exponential distribution, specifically with 16 buckets for every power of 2.
+That is, the function from a value to a bucket index is:
```
badboy:
```suggestion
That is, the function from a value \\( x \\) to a bucket index is:
```

PR #721, opened 2020-02-19
Document adding a new metric type in Python

mdboom:


docs/dev/core/new-metric-type/python.md
```diff
@@ -0,0 +1,135 @@
...
+        """
+        if ping_name is None:
+            ping_name = self._send_in_pings[0]
+
+        return _ffi.lib.glean_counter_test_get_num_recorded_errors(
+            self._handle, error_type.value, _ffi.ffi_encode_string(ping_name),
+        )
+```
+
+The new metric type also needs to be imported from `glean-core/python/glean/metrics/__init__.py` and added to the `_TYPE_MAPPING` in `glean-core/python/glean/_loader.py`.
```
travis79: nit: Would showing examples of the import and type mapping entries be useful?

PR #723, opened 2020-02-19
Spellcheck all the things

mdboom:
This expands the spellchecker to run on all of the markdown files checked into git (except DEPENDENCIES.md).

This fixes a minor bug with the spellcheck script so it will always spellcheck the CHANGELOG.

The issue is that when the docs are built, `./CHANGELOG.md` is copied to `docs/appendix/changelog.md`.  The spellcheck script only looked at `.md` in the `docs` directory, so the changelog would be spellchecked only if the developer had previously built the docs.  Notably, it meant the `CHANGELOG.md` was not spellchecked on CI.  This should at least keep things consistent, and not force users to spellcheck things in the CHANGELOG that were previously missed by CI.


tools/analysis/ping-patterns/README.md
```diff
@@ -1,8 +1,8 @@
-This directory contains a tool to analyse the patterns of baseline and metrics pings received on a per-client basis in Fenix.  These pattern analyses are then summarized in a single plot showing various patterns and issues found in the data.
+This directory contains a tool to analyze the patterns of baseline and metrics pings received on a per-client basis in Fenix.  These pattern analyses are then summarized in a single plot showing various patterns and issues found in the data.
```
mdboom: :canada: -> :us:

PR #729, opened 2020-02-20
Add interface for the upload module to Glean object

brizental:
Resolves [Bug 1614284](https://bugzilla.mozilla.org/show_bug.cgi?id=1614284).

I figured glean-ffi already does the job of locking the Glean object until completion of most functions, because it usually uses [`with_glean`](https://github.com/mozilla/glean/blob/master/glean-core/ffi/src/lib.rs#L45-L57) and [`with_glean_mut`](https://github.com/mozilla/glean/blob/master/glean-core/ffi/src/lib.rs#L66-L78), which locks Glean.

Nonetheless, I decided to lock the queue until done clearing all the stores, because what if we are not interfacing with glean through the FFI?

@badboy Is there a good way to still keep this behind a feature flag?

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -38,7 +38,6 @@ pub mod ping;
 pub mod storage;
 mod util;
 
-#[cfg(feature = "upload")]
 mod upload;
 
 pub use crate::common_metric_data::{CommonMetricData, Lifetime};
@@ -51,6 +50,7 @@ use crate::internal_pings::InternalPings;
 use crate::metrics::{Metric, MetricType, PingType};
 use crate::ping::PingMaker;
 use crate::storage::StorageManager;
+use crate::upload::{PingUploadManager, PingUploadTask};
 use crate::util::{local_now_with_offset, sanitize_application_id};
 
 const GLEAN_SCHEMA_VERSION: u32 = 1;
@@ -160,6 +160,7 @@ pub struct Glean {
     start_time: DateTime<FixedOffset>,
     max_events: usize,
     is_first_run: bool,
+    upload_manager: PingUploadManager,
 }
 
 impl Glean {
@@ -183,12 +184,13 @@ impl Glean {
             event_data_store,
             core_metrics: CoreMetrics::new(),
             internal_pings: InternalPings::new(),
-            data_path: PathBuf::from(cfg.data_path),
+            data_path: PathBuf::from(&cfg.data_path),
             application_id,
             ping_registry: HashMap::new(),
             start_time: local_now_with_offset(),
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
+            upload_manager: PingUploadManager::new(&cfg.data_path),
         };
         glean.on_change_upload_enabled(cfg.upload_enabled);
         Ok(glean)
@@ -323,6 +325,10 @@ impl Glean {
 
     /// Clear any pending metrics when telemetry is disabled.
     fn clear_metrics(&mut self) {
+        // Clear the pending pings queue and acquire the lock
+        // so that it can't be accessed until this function is done.
+        let _queue_lock = self.upload_manager.clear_ping_queue();
+
         // There is only one metric that we want to survive after clearing all
         // metrics: first_run_date. Here, we store its value so we can restore
         // it after clearing the metrics.
@@ -402,6 +408,15 @@ impl Glean {
         self.max_events
     }
 
+    /// Gets the next `PingUploadTask`.
+    ///
+    /// # Return value
+    ///
+    /// `PingUploadTask` - see [`PingUploadTask`](enum.PingUploadTask.html) for more information.
+    pub fn get_upload_task(&self) -> PingUploadTask {
+        self.upload_manager.get_upload_task()
+    }
+
     /// Take a snapshot for the given store and optionally clear it.
     ///
     /// ## Arguments
@@ -471,6 +486,9 @@ impl Glean {
                     log::warn!("IO error while writing ping to file: {}", e);
                     return Err(e.into());
                 }
+
+                self.upload_manager
+                    .enqueue_ping(&doc_id, &url_path, content);
 
                 log::info!(
                     "The ping '{}' was submitted and will be sent as soon as possible",
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -36,6 +36,7 @@ mod internal_pings;
 pub mod metrics;
 pub mod ping;
 pub mod storage;
+#[cfg(feature = "upload")]
 pub mod upload;
 mod util;
 
@@ -49,6 +50,7 @@ use crate::internal_pings::InternalPings;
 use crate::metrics::{Metric, MetricType, PingType};
 use crate::ping::PingMaker;
 use crate::storage::StorageManager;
+#[cfg(feature = "upload")]
 use crate::upload::{PingUploadManager, PingUploadTask};
 use crate::util::{local_now_with_offset, sanitize_application_id};
 
@@ -159,6 +161,7 @@ pub struct Glean {
     start_time: DateTime<FixedOffset>,
     max_events: usize,
     is_first_run: bool,
+    #[cfg(feature = "upload")]
     upload_manager: PingUploadManager,
 }
 
@@ -183,6 +186,7 @@ impl Glean {
             event_data_store,
             core_metrics: CoreMetrics::new(),
             internal_pings: InternalPings::new(),
+            #[cfg(feature = "upload")]
             upload_manager: PingUploadManager::new(&cfg.data_path),
             data_path: PathBuf::from(cfg.data_path),
             application_id,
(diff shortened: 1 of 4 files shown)


glean-core/src/upload/mod.rs
```diff
@@ -303,12 +313,45 @@ mod test {
         }
 
         // Clear the queue
-        upload_manager.clear_ping_queue();
+        let _ = upload_manager.clear_ping_queue();
```
brizental: TIL: `let _ = <something>` drops the value at the end of the statement :)

glean-core/src/upload/mod.rs
```diff
@@ -217,7 +227,7 @@ mod test {
     use crate::{tests::new_glean, PENDING_PINGS_DIRECTORY};
 
     const UUID: &str = "40e31919-684f-43b0-a5aa-e15c2d56a674"; // Just a random UUID.
-    const URL: &str = "http://example.com";
+    const URL: &str = "/submit/app_id/ping_name/schema_version/doc_id";
```
Dexterp37: nit: this is not really an URL anymore, that's more of a path :)

glean-core/src/lib.rs
```diff
@@ -183,12 +184,13 @@ impl Glean {
...
             core_metrics: CoreMetrics::new(),
             internal_pings: InternalPings::new(),
-            data_path: PathBuf::from(cfg.data_path),
+            data_path: PathBuf::from(&cfg.data_path),
             application_id,
             ping_registry: HashMap::new(),
             start_time: local_now_with_offset(),
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
+            upload_manager: PingUploadManager::new(&cfg.data_path),
```
```suggestion
            #[cfg(feature = "upload")]
            upload_manager: PingUploadManager::new(&cfg.data_path),
```

Would this work?

glean-core/src/lib.rs
```diff
@@ -38,7 +38,6 @@ pub mod ping;
 pub mod storage;
 mod util;
 
-#[cfg(feature = "upload")]
```
Dexterp37: If you kept this and tried to guard the field below?

glean-core/src/lib.rs
brizental: It would work, the problem is I'd have to spread these around all over the code.

glean-core/src/upload/mod.rs
brizental: Yes... I agree, i thought about changing this, but didn't. I will also change it in PingRequest object.

glean-core/src/lib.rs
brizental: If I keep this, whener I try to use the upload stuff below would produce an error, and each of the erroring line would require this annotation.

glean-core/src/lib.rs
Dexterp37: Not sure this is a problem. But maybe @badboy has another way to still do this while not enabling this on master

glean-core/src/lib.rs
Dexterp37: Yes, I'd be fine with this unless there's a better option. We don't really want this enabled on master untill we test the full thing.

glean-core/src/lib.rs
brizental: I did it ec7aae6 :)

glean-core/src/lib.rs
```diff
@@ -36,10 +36,9 @@ mod internal_pings;
 pub mod metrics;
 pub mod ping;
 pub mod storage;
-mod util;
-
 #[cfg(feature = "upload")]
-mod upload;
+pub mod upload;
```
badboy: What's the reason to make this `pub`?
I don't think we need to, as we expose for example `get_upload_task` as a method on `Glean`.

glean-core/src/lib.rs
```diff
@@ -402,6 +412,20 @@ impl Glean {
         self.max_events
     }
 
+    /// Gets the next task for an uploader. Which can be either:
+    ///
+    /// Wait - which means the requester should ask again later;
```
badboy: I think it would be better to refer to the `PingUploadTask` documentation.
That way we avoid duplicating this doc and we don't need to remember to update this when we change `PingUploadTask`

glean-core/src/lib.rs
```diff
@@ -402,6 +412,20 @@ impl Glean {
...
+    /// Upload(PingRequest) - which means there is a ping to upload. This wraps the actual request object;
+    /// Done - which means there are no more pings queued right now.
+    ///
+    /// # Return value
+    ///
+    /// `PingUploadTask` - an enum representing the three possible tasks.
```
badboy: Same here, if we just say `the possible tasks` we can extend the enum when we need to

glean-core/src/upload/mod.rs
```diff
@@ -98,22 +98,32 @@ impl PingUploadManager {
...
             .expect("Can't write to pending pings queue.");
-        let request = PingRequest::new(uuid, url, body);
+        let request = PingRequest::new(uuid, path, body);
         queue.push_back(request);
     }
 
-    /// Clears the pending pings queue.
-    pub fn clear_ping_queue(&self) {
+    /// Clears the pending pings queue, leaves the deletion-request pings.
+    pub fn clear_ping_queue(&self) -> RwLockWriteGuard<'_, VecDeque<PingRequest>> {
```
badboy: Why do we need to return the lock guard?

glean-core/src/upload/mod.rs
```diff
@@ -98,22 +98,32 @@ impl PingUploadManager {
...
         let mut queue = self
             .queue
             .write()
             .expect("Can't write to pending pings queue.");
+
+        let mut deletion_requests = Vec::new();
+        for ping in queue.iter() {
```
badboy: This whole loop can be simplified to:

```rust
queue.retain(|ping| ping.is_deletion_request());
```

See [`retain`](https://doc.rust-lang.org/std/collections/struct.VecDeque.html#method.retain)

glean-core/src/lib.rs
brizental: Hm, I am not sure really. If I don't make this public tho, android tests fail.

glean-core/src/lib.rs
brizental: Yes, I agree. If I do that docs-linkcheck also fails, when looping through the docs for glean-preview.

So, yesterday I was trying to fix the docs-linkcheck bug by making the upload module public. It did not fix it, but it did fix de Android tests I have 0 ideas why. Now that this is behing a feature flag, it should not brak anything. Still the problem should show up again later.

glean-core/src/upload/mod.rs
brizental: So that we can acquire it when clearing all the stores. Here: https://github.com/mozilla/glean/pull/729/files#diff-baecdb62c1f9893931c0b7b7b9a4a947R334

glean-core/src/lib.rs
badboy: huh? it shouldn't make _tests_ fail.

glean-core/src/lib.rs
badboy: for now we can refer to it without a link.

glean-core/src/lib.rs
brizental: Agree.

brizental: I will remove this, it won't break tests now because of the feature flag. I can understand what is going on when I start working on the android bindings. What do you think?

badboy: Sure! 
Once we build the platform layer for this we need to think how we are going to deal with this feature flag anyway.

glean-core/src/upload/mod.rs
badboy: Oh I see.

brizental:
Hey folks, I am sorry, but I just realized I had forgotten to also add interface to the `process_ping_upload_response` function. You can see that here: c528f67

glean-core/src/upload/request.rs
```diff
@@ -28,15 +28,24 @@ impl PingRequest {
...
         }
     }
 
+    pub fn is_deletion_request(&self) -> bool {
+        // The path format should be `/submit/<app_id>/<ping_name>/<schema_version/<doc_id>`
+        self.path
+            .split('/')
+            .nth(3)
+            .expect("URL path does not follow the expected pattern.")
+            == "deletion-request"
```
badboy: We shouldn't panic on invalid paths, instead treat them as not-a-deletion-ping.
(We don't validate the URL on load, do we?)

```suggestion
            .map(|url| url == "deletion-request)
            .unwrap_or(false)
```

That should work

glean-core/src/lib.rs
badboy: Did you intend to still change this?

At the very least this should be a proper markdown list:

```
/// * `Wait` - ...
/// * `Upload(PingRequest)` - ...
```

brizental: Hm, I am sorry for resolving this one and the other review comment I understood I didn't need to change anything. I now see I understood incorrectly.

glean-core/src/upload/request.rs
brizental: > We don't validate the URL on load, do we?

No.

PR #734, opened 2020-02-21
Ensure we only generate a deletion-request ping when toggling from on to off

badboy:
The old code already did what we wanted, because when `upload_enabled` is false the ping submission will exit early.
This just makes it more explicit, easier to read and understand without that context and we avoid a useless log line.

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -276,14 +276,14 @@ impl Glean {
     pub fn set_upload_enabled(&mut self, flag: bool) -> bool {
         log::info!("Upload enabled: {:?}", flag);
 
-        // When upload is disabled, submit a deletion-request ping
-        if !flag {
-            if let Err(err) = self.internal_pings.deletion_request.submit(self, None) {
-                log::error!("Failed to send deletion-request ping on optout: {}", err);
-            }
-        }
-
         if self.upload_enabled != flag {
+            // When upload is disabled, submit a deletion-request ping
+            if !flag {
+                if let Err(err) = self.internal_pings.deletion_request.submit(self, None) {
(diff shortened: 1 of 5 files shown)


glean-core/src/lib.rs
```diff
@@ -276,14 +276,14 @@ impl Glean {
...
 
-        // When upload is disabled, submit a deletion-request ping
-        if !flag {
-            if let Err(err) = self.internal_pings.deletion_request.submit(self, None) {
-                log::error!("Failed to send deletion-request ping on optout: {}", err);
+        if self.upload_enabled != flag {
+            // When upload is disabled, submit a deletion-request ping
+            if !flag {
+                if let Err(err) = self.internal_pings.deletion_request.submit(self, None) {
+                    log::error!("Failed to send deletion-request ping on optout: {}", err);
```
```suggestion
                    log::error!("Failed to submit deletion-request ping on optout: {}", err);
```

glean-core/tests/ping.rs
```diff
@@ -64,6 +64,20 @@ fn disabling_upload_clears_pending_pings() {
...
+    let (mut glean, _) = new_glean(None);
+
+    // Disabling upload generates a deletion ping
+    glean.set_upload_enabled(false);
+    assert_eq!(1, get_deletion_pings(glean.get_data_path()).unwrap().len());
+
+    // Re-setting it to `false` should not generate an additional ping.
+    // As we didn't clear the pending ping, that's the only one that sticks around.
+    glean.set_upload_enabled(false);
+    assert_eq!(1, get_deletion_pings(glean.get_data_path()).unwrap().len());
```
Dexterp37: Can you also add, since we're paranoid, that flipping to true doesn't trigger one?

PR #737, opened 2020-02-25
Python: Implement timespan metric type

mdboom:


glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimespanMetricTypeTest.kt
```diff
@@ -251,6 +251,7 @@ class TimespanMetricTypeTest {
         metric.setRawNanos(timespanNanos)
         metric.stop()
 
-        assertNotEquals(timespanNanos, metric.testGetValue())
+        // If that was recorded, we would expect it to record 1 second
+        assertNotEquals(1, metric.testGetValue())
```
Dexterp37: Mh, should this be aserting on "testHasValue" instead?

mdboom: It should have *a* value due to the start/stop, not the value explicitly set, which is 1000000000 nanoseconds (1 sec).

PR #738, opened 2020-02-26
Expose upload module functionality through the FFI

brizental:
Resolves [Bug 1614286](https://bugzilla.mozilla.org/show_bug.cgi?id=1614286)

### DIFF
--- a/glean-core/ffi/examples/glean_app.c
+++ b/glean-core/ffi/examples/glean_app.c
@@ -31,9 +31,23 @@ int main(void)
 
   glean_counter_add(metric, 2);
 
-  char *payload = glean_ping_collect(store1, NULL);
-  printf("Payload:\n%s\n", payload);
-  glean_str_free(payload);
+  glean_submit_ping_by_name("store1", NULL);
+
+  // task == 0 (wait)
+  // task == 1 (upload)
+  // task == 2 (done)
+  FfiPingUploadTask task = glean_get_upload_task();
+  while (task.tag == 1) {
+    printf("tag: %d\n", task.tag);
+    if (task.tag == 1) {
+      printf("uuid: %s\n", task.upload.uuid);
+      printf("path: %s\n", task.upload.path);
+      printf("body: %s\n", task.upload.body);
+    }
+    glean_process_ping_upload_response(task.upload.uuid, 200);
+    task = glean_get_upload_task();
+  }
+  printf("tag: %d\n", task.tag);
 
   glean_destroy_counter_metric(metric);
 
--- a/glean-core/ffi/examples/glean_app.c
+++ b/glean-core/ffi/examples/glean_app.c
@@ -33,18 +33,23 @@ int main(void)
 
   glean_submit_ping_by_name("store1", NULL);
 
-  // task == 0 (wait)
-  // task == 1 (upload)
-  // task == 2 (done)
+  // Since we disabled upload and submitted a ping above,
+  // we expect to have at least two pending pings: a deletion-request and a store1.
+  //
+  // The upload task.tag will either be:
+  // 0 - "wait", glean is still parsing the pending_pings directory;
+  // 1 - "upload", there is a new ping to upload and the task will also include the request data;
+  // 2 - "done", there are no more pings to upload.
+  //
+  //  NOTE: If, there are other ping files inside tmp/glean_data directory
+  // they will also be consumed here by `glean_process_ping_upload_response`.
   FfiPingUploadTask task = glean_get_upload_task();
   while (task.tag == 1) {
     printf("tag: %d\n", task.tag);
-    if (task.tag == 1) {
-      printf("uuid: %s\n", task.upload.uuid);
-      printf("path: %s\n", task.upload.path);
-      printf("body: %s\n", task.upload.body);
-    }
-    glean_process_ping_upload_response(task.upload.uuid, 200);
+    printf("path: %s\n", task.upload.path);
+    printf("body: %s\n", task.upload.body);
+
+    glean_process_ping_upload_response(task, 200);
     task = glean_get_upload_task();
   }
   printf("tag: %d\n", task.tag);
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -70,6 +70,29 @@ typedef const int32_t *RawIntArray;
 typedef const char *const *RawStringArray;
 
 /**
+ * A FFI-compatible representation for the PingUploadTask
+ */
+enum FfiPingUploadTask_Tag {
+  Wait,
+  Upload,
+  Done,
+};
+typedef uint8_t FfiPingUploadTask_Tag;
+
+typedef struct {
+  FfiPingUploadTask_Tag tag;
+  const char *uuid;
+  const char *path;
+  const char *body;
+  const char *headers;
+} Upload_Body;
+
+typedef union {
+  FfiPingUploadTask_Tag tag;
+  Upload_Body upload;
+} FfiPingUploadTask;
+
+/**
  * Configuration over FFI.
  *
  * **CAUTION**: This must match _exactly_ the definition on the Kotlin side.
@@ -194,6 +217,8 @@ uint8_t glean_event_test_has_value(uint64_t metric_id, FfiStr storage_name);
 char *glean_experiment_test_get_data(FfiStr experiment_id);
 
 uint8_t glean_experiment_test_is_active(FfiStr experiment_id);
+
+FfiPingUploadTask glean_get_upload_task(void);
 
 /**
  * # Safety
@@ -389,6 +414,8 @@ uint64_t glean_new_uuid_metric(FfiStr category,
 uint8_t glean_on_ready_to_submit_pings(void);
 
 char *glean_ping_collect(uint64_t ping_type_handle, FfiStr reason);
+
+void glean_process_ping_upload_response(FfiStr uuid, uint16_t status);
 
 void glean_quantity_set(uint64_t metric_id, int64_t value);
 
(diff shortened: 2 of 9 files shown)


glean-core/ffi/examples/glean_app.c
```diff
@@ -31,9 +31,23 @@ int main(void)
...
   glean_counter_add(metric, 2);
 
-  char *payload = glean_ping_collect(store1, NULL);
-  printf("Payload:\n%s\n", payload);
-  glean_str_free(payload);
+  glean_submit_ping_by_name("store1", NULL);
+
+  // task == 0 (wait)
+  // task == 1 (upload)
+  // task == 2 (done)
```
Dexterp37: Can you add a proper comment about what this is doing?

glean-core/ffi/examples/glean_app.c
```diff
@@ -31,9 +31,23 @@ int main(void)
...
+  FfiPingUploadTask task = glean_get_upload_task();
+  while (task.tag == 1) {
+    printf("tag: %d\n", task.tag);
+    if (task.tag == 1) {
+      printf("uuid: %s\n", task.upload.uuid);
```
Dexterp37: nit: let's collapse the uuid and path in a single line, would you? :-)

glean-core/ffi/examples/glean_app.c
```diff
@@ -31,9 +31,23 @@ int main(void)
...
-  glean_str_free(payload);
+  glean_submit_ping_by_name("store1", NULL);
+
+  // task == 0 (wait)
+  // task == 1 (upload)
+  // task == 2 (done)
+  FfiPingUploadTask task = glean_get_upload_task();
+  while (task.tag == 1) {
+    printf("tag: %d\n", task.tag);
+    if (task.tag == 1) {
```
Dexterp37: Isn't this already evaluated by the while?

glean-core/ffi/glean.h
```diff
@@ -69,6 +69,29 @@ typedef const int32_t *RawIntArray;
...
+enum FfiPingUploadTask_Tag {
+  Wait,
+  Upload,
+  Done,
+};
+typedef uint8_t FfiPingUploadTask_Tag;
+
+typedef struct {
+  FfiPingUploadTask_Tag tag;
+  const char *uuid;
```
Dexterp37: Should these (and the ones below) be of type `FfiStr`?

glean-core/ffi/glean.h
```diff
@@ -69,6 +69,29 @@ typedef const int32_t *RawIntArray;
...
+  const char *path;
+  const char *body;
+  const char *headers;
```
Dexterp37: Should this rather be a string array, given the plural?

glean-core/ffi/src/lib.rs
```diff
@@ -324,4 +326,67 @@ pub extern "C" fn glean_is_first_run() -> u8 {
     with_glean_value(|glean| glean.is_first_run())
 }
 
+/// A FFI-compatible representation for the PingUploadTask
+#[repr(u8)]
+pub enum FfiPingUploadTask {
+    Wait,
+    Upload {
+        uuid: *const c_char,
```
Dexterp37: Should all of these be `FfiStr`?

glean-core/src/upload/request.rs
```diff
@@ -37,6 +37,27 @@ impl PingRequest {
...
+        self.path.to_owned()
+    }
+
+    /// Get the body as a JSON formatted string.
+    pub fn body(&self) -> String {
+        serde_json::to_string(&self.body).unwrap()
+    }
+
+    /// Get the headers as a JSON formatted string.
+    pub fn headers(&self) -> String {
```
Dexterp37: Why does this need to be a json formatted string?

glean-core/ffi/glean.h
brizental: I decided to go with JSON encoded string here because I thought it made more sense. This is what this string will look like in this case (I added line breaks to make it easier to visualize, it won't really have them): 
```json
{
    "X-Client-Type":"Glean",
    "Content-Type":"application/json; charset=utf-8","X-Client-Version":"25.0.0",
    "Date":"2020-02-26 09:54:40.648697 UTC"
}
```

glean-core/ffi/examples/glean_app.c
brizental: Yes, oops.

glean-core/ffi/examples/glean_app.c
brizental: Hm, I think that will make the output harder to visualize. The uuid is already in the path, in this case I can just remove it altogether and leave path and body only.

glean-core/ffi/src/lib.rs
brizental: Honestly, I didn't because I started having multiple problems with lifetimes when I tried. I didn't go deep enough to understand if I *should* use it here or not. Because I have to return the `FfiPingUploadTask`, the only way I found to do that with the `FfiStr` was to add a 'static lifetime to it, which the documentation strictly says never to do.

The other functions that return strings, also return them as `*mut c_char`, so I decided to also do that for this one. 

I will try and understand if I should or not use `FfiStr` here and try to come back with an answer, but if you or @badboy already know the answer that would be cool too :)

glean-core/src/upload/request.rs
brizental: Like I said in the other comment, I thought it made sense to serialize this as JSON. It will make it easier for platforms to deserialize it later.

glean-core/ffi/src/lib.rs
badboy: `FfiStr` is used to accept strings (and then convert them and help with lifetimes).
On the way out of the FFI we need to use `*const c_char` (actually maybe `*mut c_char`?).

badboy: Oh and one thing I realize: we need to ensure the platform side frees these strings again (by calling `glean_str_free` or maybe even better by passing back the `FfiPingUploadTask` somehow, so the Rust side can reconstruct the `CString` and free them.

glean-core/ffi/examples/glean_app.c
Dexterp37: Yeah, path and body sounds like enough.

glean-core/ffi/glean.h
Dexterp37: I'm not sure this is the best option here. We don't require it to be JSON for sending headers, and Android would require to parse it/break it into a key value map even if it's JSON (see [here](https://searchfox.org/glean/rev/7f4d8b01a47750b2cb633b69f724128c2396b628/glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt#117) and [here](https://searchfox.org/mozilla-mobile/source/android-components/components/service/glean/src/main/java/mozilla/components/service/glean/net/ConceptFetchHttpUploader.kt#67)).

Since we don't really require JSON, let's report back two arrays: one for the keys and ones for the values. If it makes it more convenient to just keep it JSON to avoid the nasty two arrays passing, make sure to encapsulate and abstract this away to the other FFI side. Please also rename this to something more meaningful such as `headersAsJSON` or `jsonHeaders`. Let's also add a big comment about why we're doing this and not going the key-value map route.

badboy: I think encoding it as JSON is preferably to the 2-array solution.
That will also make it easier to free it properly (there's already `glean_str_free`).
I do agree we can do the map-to-JSON serialization in glean-ffi instead of doing that in glean-core.

brizental: > I do agree we can do the map-to-JSON serialization in glean-ffi instead of doing that in glean-core.

The only thing that made me put it in glean-core at first was because to parse it as json on the ffi we would have to add serde_json as a dependency there. Probably not a big deal, but let me know if you think I should go ahead with it.

glean-core/ffi/src/lib.rs
brizental: > (actually maybe *mut c_char?)

I noticed that is what the other functions return, why is that? FWIW `*const c_char` seems to be working for now.

badboy: Because it is a mutable pointer after all, as we're giving up ownership (effectively).

glean-core/ffi/src/upload.rs
```diff
@@ -0,0 +1,90 @@
...
+    pub fn uuid_as_str(&self) -> Result<&str> {
+        if let FfiPingUploadTask::Upload { uuid, .. } = self {
+            assert!(!uuid.is_null());
+            unsafe {
+                CStr::from_ptr(*uuid)
+                    .to_str()
+                    .map_err(|_| glean_core::Error::utf8_error())
+            }
+        } else {
+            Ok("")
```
Dexterp37: Is it common practice in Rust to return empty results? Shouldn't be return value be something like `Result<(), &str>`? Or what are you trying to do here?

badboy: Indeed, this seems like it should be an error (if there's no `uuid` we can't return it)

badboy: Which brings up the question if that should not just be in the one place where it is used directly, instead of making it a function (within `glean_process_ping_upload_response`)

brizental: I actually was not sure which error to return here, forgot to comment it. None of the errors in our error module seemed to fit. Should I add a new one?

brizental: Sure, not making it a separate function eliminates that problem. I'll do that.

glean-core/ffi/glean.h
badboy: Not a big deal at all, it's a dependency anyway and as it's all compiled together there's no duplication.

PR #740, opened 2020-02-26
Add script to update schema and run checks on CI

badboy:
Previously we ran schema checks in Python (`test_glean.py`) and Kotlin (several).
These relied on fetching the schema from the mozilla-pipeline-schemas repository.
For reproducability it uses a fixed git hash across both tests.

Now we also bundle a schema into the repository and run a single test
against it from glean-preview.

This is a small imrprovement, but it still means we can miss updating it
(or miss updating all places in the repository to the same version).
The previous commit already introduced a script to automate that.

This CI task now runs an explicit check, failing if the upstream version
changed.

This does mean that unrelated PRs can now fail due to schema changes,
that may not affect them.

In a future world we could have a nightly task that runs this check and
opens a PR with the changed files.

### DIFF
--- a/bin/update-schema.sh
+++ b/bin/update-schema.sh
@@ -1,0 +1,92 @@
+#!/bin/bash
+
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Update the bundled schema version tests run against.
+#
+# The schema is pulled from mozilla-pipeline-schemas at the following URL:
+#
+#     https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/$HASH/schemas/glean/glean/glean.1.schema.json
+#
+# References in the code base are updated.
+#
+# Usage: update-schema.sh <commit hash>
+#
+# Environment:
+#
+# DRY_RUN - Do not modify files or run destructive commands when set.
+# VERB    - Log commands that are run when set.
+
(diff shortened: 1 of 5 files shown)


badboy:
The new CI job is this here: https://circleci.com/gh/mozilla/glean/48116
It's in the list here on GitHub, but not yet marked required.

bin/update-schema.sh
```diff
@@ -0,0 +1,92 @@
...
+  echo "Vendoring schema from ${FULL_URL}"
+  run curl --silent --fail --show-error --location --retry 5 --retry-delay 10 "$FULL_URL" --output "$SCHEMA_PATH"
+
+  # Update the android/build.gradle version
+
+  FILE=glean-core/android/build.gradle
+  run $SED -i.bak -E \
+      -e "s/^String GLEAN_PING_SCHEMA_GIT_HASH = \"[0-9a-z.-]+\"/String GLEAN_PING_SCHEMA_GIT_HASH = \"${COMMIT_HASH}\"/" \
+      "${WORKSPACE_ROOT}/${FILE}"
+  run rm "${WORKSPACE_ROOT}/${FILE}.bak"
```
mdboom: Either here or as a follow-on -- it might be nice to use a vendored file for this too (rather than a git hash).  Also for the Python checking below.  That would save the network request / caching etc.

badboy: True, but I don't think the glean_parser supports that currently. I'll open a bug for that.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1618263

PR #743, opened 2020-02-27
Mini optimizations

badboy:
I read through the code a bit again and these were 2 little things I noticed.


badboy:
> Awesome! I always tend to run away from `'static` lifetimes, but I think it is on me to actually understand when they are ok or not. Thanks for doing this :)

The good thing about lifetimes is: you can't get it wrong, the compiler will stop you.

PR #746, opened 2020-03-02
Add missed word

badboy:
[doc only]


badboy:
Merging without r+ as its just a spellcheck dictionary fix to unbreak master.

PR #747, opened 2020-03-02
1618263 - Use vendored schema in all tests

badboy:


glean-core/python/tests/conftest.py
```diff
@@ -4,25 +4,21 @@
...
-    "https://raw.githubusercontent.com/mozilla-services/"
-    "mozilla-pipeline-schemas/{}/schemas/glean/glean/"
-    "glean.1.schema.json"
-).format(GLEAN_PING_SCHEMA_GIT_HASH)
+# This defines the location of the JSON schema used to validate the pings
+# created during unit testing. This uses the vendored schema.
+#
+# Use `bin/update-schema.sh latest` to update it to the latest upstream version.`
+this_dir = os.path.dirname(__file__)
+GLEAN_PING_SCHEMA_PATH = os.path.join(this_dir, "../../../glean.1.schema.json")
```
mdboom: Let's use [pathlib](https://docs.python.org/3/library/pathlib.html) to be consistent with the rest of the Glean Python bindings:

At top:

```
from pathlib import Path
```

```suggestion
this_dir = Path(__file__.parent)
GLEAN_PING_SCHEMA_PATH = path.parent.parent.parent / "glean.1.schema.json"
```

badboy: `parent.parent.parent` - lol

mdboom: Where is the `greatgrandparent` property?

badboy: don't make me implement that!

PR #748, opened 2020-03-02
[doc only] Fix docs about using plugin, and use a-c linkage everywhere

mdboom:


docs/user/adding-glean-to-your-project.md
```diff
@@ -38,17 +38,17 @@ Add this to your Gradle configuration:
 implementation "org.mozilla.components:service-glean:{latest-version}"
 ```
 
-> **Important:** the `{latest-version}` placeholder in the above link should be replaced with the version number of the Glean SDK used by the project.
+> **Important:** the `{latest-version}` placeholder in the above link should be replaced with the version of android components used by the project.
```
badboy:
```suggestion
> **Important:** the `{latest-version}` placeholder in the above link should be replaced with the version of Android Components used by the project.
```

PR #750, opened 2020-03-02
Documentation about choosing metric types

mdboom:
The bones of this are done, but there are parts that I might need some help explaining.  Comments below.

### DIFF
--- a/docs/user/adding-new-metrics.md
+++ b/docs/user/adding-new-metrics.md
@@ -69,7 +69,7 @@ If the metric is still needed after its expiration date, it should go back for [
 
 The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available:
 
-- `ping` (default): The metric is cleared each time it is sent in the ping.
+- `ping` (default): The metric is cleared each time it is submitted in the ping.
   This is the most common case, and should be used for metrics that are highly dynamic, such as things computed in response to the user's interaction with the application.
   
 - `application`: The metric is related to an application run, and is cleared every time the application restarts.
@@ -101,7 +101,7 @@ In this diagram, the ping "lifetimes" are represented as rectangles, but the mom
 
 - **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
 
-- **F. User, set on change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
+- **F. User, set on init and change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
 
 Note that for all of the metric configurations, the toggle of the preference off and on during Ping 4 is completely missed.  If you need to create a ping containing one, and only one, value for this metric, consider using a [custom ping](pings/custom.html) to create a ping whose lifetime matches the lifetime of the value.
 
--- a/docs/user/adding-new-metrics.md
+++ b/docs/user/adding-new-metrics.md
@@ -113,7 +113,7 @@ If the timing at which the metric is sent in the ping needs to closely match the
 
 #### Reuse names from other applications
 
-There's a lot of value using the same name for analogous metrics collected across different products. For example, BigQuery makes it simple to join columns with the same name across multiple tables. Therefore, we encourage you to investigate if a similar metric is already being collected by another product and use the same naming whenever feasible.
+There's a lot of value using the same name for analogous metrics collected across different products. For example, BigQuery makes it simple to join columns with the same name across multiple tables. Therefore, we encourage you to investigate if a similar metric is already being collected by another product. If it is, there may be an opportunity for code reuse across these products, and if all the projects are using Glean, it's easy for libraries to send their own metrics. If sharing the code doesn't make sense, at a minimum we recommend using the same metric name for similar actions and concepts whenever possible.
 
 #### Make names unique within an application
 
--- a/docs/user/adding-new-metrics.md
+++ b/docs/user/adding-new-metrics.md
@@ -72,7 +72,7 @@ The `lifetime` parameter of a metric defines when its value will be cleared. The
 - `ping` (default): The metric is cleared each time it is submitted in the ping.
   This is the most common case, and should be used for metrics that are highly dynamic, such as things computed in response to the user's interaction with the application.
   
-- `application`: The metric is related to an application run, and is cleared every time the application restarts.
+- `application`: The metric is related to an application run, and is cleared after the application restarts and any Glean-owned ping, due at startup, is submitted.
   This should be used for things that are constant during the run of an application, such as the operating system version.
   In practice, these metrics are generally set during application startup.
   A common mistake---using the `ping` lifetime for these type of metrics---means that they will only be included in the first ping sent during a particular run of the application.
@@ -99,9 +99,9 @@ In this diagram, the ping "lifetimes" are represented as rectangles, but the mom
 
 - **D. Application, set on init and change**: The default value is included in Ping 1, and the changed value is included in Pings 2 and 3. Even though the application startup causes it to be cleared, it is set again, and all subsequent pings also have the value.
 
-- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
+- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset (e.g. on Android, when the product is uninstalled), it is included in all subsequent pings.
 
-- **F. User, set on init and change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
+- **F. User, set on init and change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.  This would be true even if the "turbo mode" preference were never changed again.
 
 Note that for all of the metric configurations, the toggle of the preference off and on during Ping 4 is completely missed.  If you need to create a ping containing one, and only one, value for this metric, consider using a [custom ping](pings/custom.html) to create a ping whose lifetime matches the lifetime of the value.
 
--- a/docs/user/adding-new-metrics.md
+++ b/docs/user/adding-new-metrics.md
@@ -1,19 +1,144 @@
 # Adding new metrics
 
-All metrics that your project collects must be defined in a `metrics.yaml` file.
-This file should be at the root of the application or library module.
-The format of that file is documented [here](https://mozilla.github.io/glean_parser/metrics-yaml.html).
-
 When adding a new metric, the workflow is:
 
-* Decide on which [metric type](metrics/index.md) you want to use.
+* Consider the question you are trying to answer with this data, and choose the [metric type](metrics/index.md) and parameters to use.
 * Add a new entry to [`metrics.yaml`](#The-metricsyaml-file).
 * Add code to your project to record into the metric by calling the Glean SDK.
 
 > **Important**: Any new data collection requires documentation and data-review.
 This is also required for any new metric automatically collected by the Glean SDK.
 
-## The `metrics.yaml` file
+## Choosing a metric type
+
+The following is a set of questions to ask about the data being collected to help better determine which metric type to use.
+
+### Is it a single measurement?
+
+If the value is true or false, use a [boolean metric](metrics/boolean.html).
+
+If the value is a string, use a [string metric](metrics/string.html). For example, to record the name of the default search engine. Beware: string metrics are exceedingly general, and you are probably best served by selecting the most specific metric for the job, since you'll get better error checking and richer analysis tools for free. For example, avoid storing a number in a string metric --- you probably want a [counter metric](metrics/counter.html) instead.
+
+If you need to store multiple string values in a metric, use a [string list metric](metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
+
+<!-- TODO: If you do know the list of values you might store and you just want to record their presence, consider using a [labeled boolean](metrics/labeled_boolean.html) instead. -->
+
+<!-- If you have a related set of metrics that you want to record strings for, and you don't know the things the strings relate to at build time, use a [labeled string metric](metrics/labeled_strings.html). -->
+
+For all of the simple metric types in this section that measure single values, it is especially important to consider how the lifetime of the value relates to the ping it is being sent in. Since these metrics don't perform any aggregation on the client side, when a ping containing the metric is submitted, it will contain only the "last known" value for the metric, potentially resulting in **data loss**.  There is further discussion of [metric lifetimes](#When-should-Glean-automatically-reset-the-measurement) below.
+
+### Are you counting things?
+
+If you want to know how many times something happened, use a [counter metric](metrics/counter.html).  If you are counting a group of related things, or you don't know what all of the things to count are at build time, use a [labeled counter metric](metrics/labeled_counters.html).
+
+If you need to know when the things being counted happened relative to other things, consider using an [event](metrics/event.html).
+
+### Are you measuring time?
+
+If you need to record an absolute time, use a [datetime metric](metrics/datetime.html). Datetimes are recorded in the user's local time, according to their device's real time clock, along with a timezone offset from UTC. Datetime metrics allow specifying the resolution they are collected at, and to stay [lean][lean-data], they should only be collected at the minimum resolution required to answer your question.
+
+If you need to record how long something takes you have a few options. 
+
+If you need to measure the total time spent doing a particular task, look to the [timespan metric](metrics/timespan.html). Timespan metrics allow specifying the resolution they are collected at, and to stay [lean][lean-data], they should only be collected at the minimum resolution required to answer your question.
+Note that this metric should only be used to measure time on a single thread. If multiple overlapping timespans are measured for the same metric, an invalid state error is recorded. 
+
+If you need to measure the relative occurrences of many timings, use a [timing distribution](metrics/timing_distribution.html). It builds a histogram of timing measurements, and is safe to record multiple concurrent timespans on different threads.
+
+If you need to know the time between multiple distinct actions that aren't a simple "begin" and "end" pair, consider using an [event](metrics/event.html).
+
+### Do you need to know the order of events relative to other events?
+
+If you need to know the order of actions relative to other actions, such as, the user performed tasks A, B, and then C, and this is meaningfully different from the user performing tasks A, C and then B, (in other words, the order is meaningful beyond just the *fact* that a set of tasks were performed), use an [event metric](metrics/event.html).
+
+Importantly, events are the most expensive metric type to record, transmit, store and analyze, so they should be used sparingly, and only when none of the other metric types are sufficient for answering your question.
+
+### For how long do you need to collect this data?
+
+Think carefully about how long the metric will be needed, and set the `expires` parameter to disable the metric at the earliest possible time. 
+This is an important component of Mozilla's [lean data practices][lean-data].
+
+When the metric passes its expiration date (determined at build time), it will automatically stop collecting data.
+
+When a metric's expiration is within in 14 days, emails will be sent from `telemetry-alerts@mozilla.com` to the `notification_emails` addresses associated with the metric.
+At that time, the metric should be removed, which involves removing it from the `metrics.yaml` file and removing uses of it in the source code.
+Removing a metric does not affect the availability of data already collected by the pipeline.
+
+If the metric is still needed after its expiration date, it should go back for [another round of data review](https://wiki.mozilla.org/Firefox/Data_Collection) to have its expiration date extended.
+
+### When should Glean automatically clear the measurement?
+
+The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available:
+
+- `ping` (default): The metric is cleared each time it is sent in the ping.
+  This is the most common case, and should be used for metrics that are highly dynamic, such as things computed in response to the user's interaction with the application.
+  
+- `application`: The metric is related to an application run, and is cleared every time the application restarts.
+  This should be used for things that are constant during the run of an application, such as the operating system version.
+  In practice, these metrics are generally set during application startup.
+  A common mistake---using the `ping` lifetime for these type of metrics---means that they will only be included in the first ping sent during a particular run of the application.
+  
+- `user`: The metric is part of the user's profile.
+  This should be used for things that change only when the user's profile is created.
+  It is rare to use this lifetime outside of some metrics that are built in to Glean, such as `client_id`.
+
+While lifetimes are important to understand for all metric types, they are particularly important for the metric types that record single values and don't aggregate on the client (`boolean`, `string`, `labeled_string`, `string_list`, `datetime` and `uuid`), since these metrics will send the "last known" value and missing the earlier values could be a form of unintended data loss.
+
+#### A lifetime example
+
+Let's work through an example to see how these lifetimes play out in practice. Let's suppose we have a user preference, "turbo mode", which defaults to `false`, but the user can turn it to `true` at any time.  We want to know when this flag is `true` so we can measure its affect on other metrics in the same ping.  In the following diagram, we look at a time period that sends 4 pings across two separate runs of the application. We assume here, that like Glean's built-in [metrics ping](pings/metrics.html), the developer writing the metric isn't in control of when the ping is submitted. 
+
+In this diagram, the ping "lifetimes" are represented as rectangles, but the moment the ping is "submitted" is represented by its right edge. The user changes the "turbo mode" setting from `false` to `true` in the first run, and then toggles it again twice in the second run. 
+  
+![Metric lifetime timeline](metric-lifetime-timeline.svg)
+
+- **A. Ping lifetime, set on change**: The value isn't included in Ping 1, because Glean doesn't know about it yet.  It is included in the first ping after being recorded (Ping 2), which causes it to be cleared.
+
+- **B. Ping lifetime, set on init and change**: The default value is included in Ping 1, and the changed value is included in Ping 2, which causes it to be cleared.  It therefore misses Ping 3, but when the application is started, it is recorded again and it is included in Ping 4.  However, this causes it to be cleared again and it is not in Ping 5.
+
+- **C. Application lifetime, set on change**: The value isn't included in Ping 1, because Glean doesn't know about it yet. After the value is changed, it is included in Pings 2 and 3, but then due to application restart it is cleared, so it is not included until the value is manually toggled again.
+
+- **D. Application, set on init and change**: The default value is included in Ping 1, and the changed value is included in Pings 2 and 3. Even though the application startup causes it to be cleared, it is set again, and all subsequent pings also have the value.
+
+- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
+
+- **F. User, set on change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
+
+Note that for all of the metric configurations, the toggle of the preference off and on during Ping 4 is completely missed.  If you need to create a ping containing one, and only one, value for this metric, consider using a [custom ping](pings/custom.html) to create a ping whose lifetime matches the lifetime of the value.
+
+#### What if none of these lifetimes are appropriate?
+
+If the timing at which the metric is sent in the ping needs to closely match the timing of the metrics value, the best option is to use a [custom ping](pings/custom.html) to manually control when pings are sent.
+
+### What should this new metric be called?
+
+#### Reuse names from other applications
+
+There's a lot of value using the same name for analogous metrics collected across different products. For example, BigQuery makes it simple to join columns with the same name across multiple tables. Therefore, we encourage you to investigate if a similar metric is already being collected by another product and use the same naming whenever feasible.
+
+#### Make names unique within an application
+
+Metric identifiers (the combination of a metric's category and name) must be unique across all metrics that are sent by a single application.
+This includes not only the metrics defined in the app's `metrics.yaml`, but the `metrics.yaml` of any Glean-using library that the application uses, including Glean itself.
+Therefore, care should be taken to name things specifically enough so as to avoid namespace collisions.
+In practice, this generally involves thinking carefully about the `category` of the metric, more than the `name`.
+
+> **Note:** Duplicate metric identifiers are not currently detected at build time. See [bug 1578383](https://bugzilla.mozilla.org/show_bug.cgi?id=1578383) for progress on that.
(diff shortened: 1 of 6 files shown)


docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
-    ...
+If you need to record a closely-related group of strings, or you don't know the things the strings are related to at build time, use a [labeled string metric](metrics/labled_strings.html).
 
-category2.subcategory:  # Categories can contain subcategories
-  metric:
-    ...
+For all of these simple value types, you should consider how the value changes in the application and how that relates to the ping it is being sent in. When the ping is submitted, it will contain the "last known" value for the metric, and then the value is reset according to its [lifetime](#What-is-the-lifetime-of-the-data).
 
-```
+If you need to know the current state of a value, such as a user preference, to provide context for other metrics, be aware that the lifetime of the value may not be the same as the lifetime of the ping it is in. You can address this either with a custom ping that matches the lifetime of the value, or by using a user lifetime metric so the value will persist across pings.
```
mdboom: Aside from just being hard to explain:

I think for metrics that want to record the status of a preference (like the environment stuff on desktop) so that it can be related to other metrics, I think the best recommendation in our toolbox is a user lifetime metric, but I know others have *opinions* about that.

docs/user/adding-new-metrics.md
```diff
@@ -66,33 +73,70 @@ The `lifetime` parameter of a metric defines when it will be reset. There are th
 - `user`: The metric is part of the user's profile.
   This should be used for things that change only when the user's profile is created.
   It is rare to use this lifetime outside of some metrics that are built in to Glean, such as `client_id`.
```
mdboom: The lifetime language here hasn't changed, but probably needs to be improved as it relates to the other discussion of lifetimes in the "simple value" section (which is where getting the lifetimes matters most, IMHO).

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+If you need to measure the relative occurrences of many timings, use a [timing distribution](metric/timing_distribution.html). It builds a histogram of timing measurements, and is safe to use in a multi-threaded context.
 
-“There are only two hard things in Computer Science: cache invalidation and naming things” -- attributed to Phil Karlton.
+If you need to know the time between distinct events that aren't a simple "begin" and "end" pair, consider using an [event](metric/event.html).
+
+### Do you need to know the order of events relative to other events?
+
+If you need to know the order of events relative to other events, such as, the user did A, then B, then C, use an [event metric](metric/event.html).
+
+Events are the most expensive metric type to record, transmit, store and analyse, so they should be used sparingly, and only when one of the other metric types aren't sufficient for answering your question.
```
mdboom: Also perhaps a higher potential for accidental PII, but maybe that's hard to back up with facts.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+
+If you need to record how long something takes you have a couple of options:
+
+If you need to measure the total time spent doing a particular task, look to the [timespan metric](metric/timespan.html). Note that it can only be used on a single thread.
 
-## Recommendations for defining new metrics
+If you need to measure the relative occurrences of many timings, use a [timing distribution](metric/timing_distribution.html). It builds a histogram of timing measurements, and is safe to use from multiple threads.
 
-“There are only two hard things in Computer Science: cache invalidation and naming things” -- attributed to Phil Karlton.
+If you need to know the time between distinct events that aren't a simple "begin" and "end" pair, consider using an [event](metric/event.html).
```
Dexterp37: At first I was... "eugh" with this sentence. But it's really a legitimate use-case I didn't think about! Good catch.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+
+### Do you need to know the order of events relative to other events?
+
+If you need to know the order of events relative to other events, such as, the user did A, then B, then C, use an [event metric](metric/event.html).
```
Dexterp37:
```suggestion
If you need to know the order of actions relative to other actions, such as, the user did A, then B, then C, use an [event metric](metric/event.html).
```

What about this to reduce the repetition of "event"?

docs/user/adding-new-metrics.md
Dexterp37: Good point.

Another thing that might be relevant (but that I really struggle to phrase properly) is that events push the computational efforts to the "analysis" step (e.g. think of some of our users using events to compute, at analysis time, the "foreground duration"). Other metric types make sure the effort is at the SDK level, handled by the SDK.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+
+Events are the most expensive metric type to record, transmit, store and analyse, so they should be used sparingly, and only when one of the other metric types aren't sufficient for answering your question.
+
+### For how long do you need to collect this data?
+
+Think carefully about how long the metric will be needed, and set the `expires` parameter to expire the metric at the soonest time possible.
```
Dexterp37: Is it worth linking to our [lean data](https://www.mozilla.org/en-US/about/policy/lean-data/) website here, for a more in depth read? Specifically https://www.mozilla.org/en-US/about/policy/lean-data/stay-lean/ (point 4) might be relevant to this particular sentence.

docs/user/adding-new-metrics.md
Dexterp37: Yes, good idea. What improvements did you have in mind? Any relevant example we might just embed in the docs?

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
 > **Important**: Any new data collection requires documentation and data-review.
 This is also required for any new metric automatically collected by the Glean SDK.
 
-## The `metrics.yaml` file
+## Choosing a metric type
 
-The [`metrics.yaml` file](https://mozilla.github.io/glean_parser/metrics-yaml.html) defines the metrics your application or library will send.
-They are organized into categories.
-The overall organization is:
+### Is it a simple (scalar) measurement?
```
Dexterp37: I think @chutten and DS should really have a look at this too!

docs/user/adding-new-metrics.md
Dexterp37: > I think the best recommendation in our toolbox is a user lifetime metric, but I know others have _opinions_ about that.

:-P 

I personally think that the status of a preference (exactly like the environment on desktop) is `application` lifetime: the _specific_ environment on Desktop is built at startup (ok, in a staged way) and lives throughout the life of Firefox (or as long as it is changed).

A relevant example, for desktop environment, is search defaults: it's set [later at startup](https://searchfox.org/mozilla-central/rev/91f6c02fcf4c16f78fdc4417f61f192688294066/toolkit/components/telemetry/app/TelemetryEnvironment.jsm#1418) and, in the rare circumstance of pings being generated before that, it will not be present in pings. Still, that's not a "user lifetime": it lives as long as the application lives.

I think the core of the issue here is understanding what it's being measured and when/how it should be available: that's indeed an hard task :)

docs/user/adding-new-metrics.md
travis79:
```suggestion
Think carefully about how long the metric will be needed, and set the `expires` parameter to expire the metric at the earliest possible time.
```

travis79: nit: 'soonest' just sounded weird here for some reason.

docs/user/adding-new-metrics.md
travis79: Maybe expanding on the gfx metric example of the `compositor` would be a good way to illustrate the dangers of using application lifetime metrics with the metrics ping, and that it requires some extra care? That seemed like a good real-world case that shows where application lifetime metrics can be weird.

docs/user/adding-new-metrics.md
travis79: As I said above, I think any 'real-life' examples we could provide that illustrate our intentions with lifetimes would be useful.

docs/user/metrics/index.md
```diff
@@ -26,9 +28,9 @@ There are different metrics to choose from, depending on what you want to achiev
 
 * [Events](event.md): Records events e.g. individual occurrences of user actions, say every time a view was open and from where.
 
-* [Custom Distribution](custom_distribution.md): Used to record the distribution of a value that needs fine-grained control of how the histogram buckets are computed.
+* [Custom Distribution](custom_distribution.md): Used to record the distribution of a value that needs fine-grained control of how the histogram buckets are computed.  **Custom distributions are only available for values that come from Gecko.**
```
badboy:
```suggestion
* [Custom Distribution](custom_distribution.md): Used to record the distribution of a value that needs fine-grained control of how the histogram buckets are computed.  **Custom distributions are only available for values that come from Gecko on Android.**
```

We soon will have Glean in Gecko, but still not support those.

docs/user/metrics/index.md
```diff
@@ -26,9 +28,9 @@ There are different metrics to choose from, depending on what you want to achiev
...
 
-* [Quantity](quantity.md): Used to record a single non-negative integer value. For example, the width of the display in pixels.
+* [Quantity](quantity.md): Used to record a single non-negative integer value. For example, the width of the display in pixels.  **Quantities are only available for values that come from Gecko.**
```
badboy:
```suggestion
* [Quantity](quantity.md): Used to record a single non-negative integer value. For example, the width of the display in pixels.  **Quantities are only available for values that come from Gecko on Android.**
```

docs/user/adding-new-metrics.md
chutten: We can leave it as a warning that the lifetime might not match and let the rest of the documentation suggest solutions appropriate to the reader. Or we could be a bit more precise about "If it is important that every ping contain some value for this metric, it should have a 'ping' lifetime and be sent on a custom ping you control."

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
-The `metrics.yaml` file is used to generate code in the target language (e.g. Kotlin, Swift, ...) that becomes the public API to access your application's metrics.
+If you want to know how many times something happened, use a [counter metric](metrics/counter.html).  If you are counting a group of related things, or you don't know what all of the things to count are at build time, use a [labeled counter metric](metrics/labeled_counter.html).
+
+If you need to know when the things being counted happened relative to other things, use an [event](metrics/event.html).
+
+### Are you measuring time?
+
+If you need to record an absolute time, use a [datetime metric](metrics/datetime.html). Datetimes are recorded in the user's local time, along with a timezone offset from UTC.
+
+If you need to record how long something takes you have a couple of options:
```
chutten: This is the first instance of this structure in the guide. Everything else has been a paragraph of related metric types with increasing complexity, and this is more of a list. I suggest we pick a structure for the whole guide and use it throughout.

Also "a couple" to me means "two". There are three options presented : )

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+
+If you need to measure the total time spent doing a particular task, look to the [timespan metric](metric/timespan.html). Note that it can only be used on a single thread.
```
chutten: Might be worth a thumbnail of what happens if you misuse it.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
 
-## Recommendations for defining new metrics
+If you need to measure the relative occurrences of many timings, use a [timing distribution](metric/timing_distribution.html). It builds a histogram of timing measurements, and is safe to use from multiple threads.
```
chutten: Does this mean "start" and "stopandaccumulate" can be on different threads? Is a distribution with a single sample a suggested solution for someone who needs a cross-thread timespan?

docs/user/adding-new-metrics.md
chutten: Any "Eugh" is a candidate for a new metric type to me. Let's say a "timing list" where each step in the multi-part process is represented as a count of nanos from the first `start()`. ...but if what we're hoping is to hear from people who want this metric type, we can't provide a workaround for them in this guide.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
 
-“There are only two hard things in Computer Science: cache invalidation and naming things” -- attributed to Phil Karlton.
+If you need to know the time between distinct events that aren't a simple "begin" and "end" pair, consider using an [event](metric/event.html).
+
+### Do you need to know the order of events relative to other events?
+
+If you need to know the order of actions relative to other actions, such as, the user did A, then B, then C, use an [event metric](metric/event.html).
```
chutten: I'd add something about how "And this is meaningfully different than if the user did A then C then B." The order has to be meaningful, otherwise you could just count A and B and C.

docs/user/adding-new-metrics.md
chutten: It can be backed up with facts: events are structured strings. They contain more bytes per record than any other metric type besides labeled strings, and contain even more bytes overall because a labeled string can only store one string per label: events can record different values each time you hit record.

Put another way, if you were looking for a way to exfiltrate a copy of wikipedia from Fenix to the Data Platform via the Glean SDK, events would be the way you'd choose to do it.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
+
+Events are the most expensive metric type to record, transmit, store and analyse, so they should be used sparingly, and only when one of the other metric types aren't sufficient for answering your question.
+
+### For how long do you need to collect this data?
+
+Think carefully about how long the metric will be needed, and set the `expires` parameter to expire the metric at the earliest possible time.
+When a metric is about to expire, emails will be sent to the `notification_emails` addresses associated with the metric.
```
chutten: I think if we specify from what email address and what service these emails will come we will build trust in them.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
-“There are only two hard things in Computer Science: cache invalidation and naming things” -- attributed to Phil Karlton.
+If you need to know the time between distinct events that aren't a simple "begin" and "end" pair, consider using an [event](metric/event.html).
+
+### Do you need to know the order of events relative to other events?
+
+If you need to know the order of actions relative to other actions, such as, the user did A, then B, then C, use an [event metric](metric/event.html).
+
+Events are the most expensive metric type to record, transmit, store and analyse, so they should be used sparingly, and only when one of the other metric types aren't sufficient for answering your question.
+
+### For how long do you need to collect this data?
```
chutten: One last segment in this guide should probably exhort the reader to peruse the options in the documentation. This guide will not be complete as niche datatypes do not make sense to have the reader be guided to (like the UUID metric type, for example). And then consider reaching out to us on #glean:mozilla.org, or to request a new Metric Type if all else fails.

docs/user/adding-new-metrics.md
```diff
@@ -6,54 +6,61 @@ The format of that file is documented [here](https://mozilla.github.io/glean_par
...
-    bugs:
-      - 123456789
-    data_reviews:
-      - http://example.com/path/to/data-review
-    expires: 2019-06-01  # <-- Update to a date in the future
+If the value is a string, use a [string metric](metrics/string.html).  If you want to store multiple strings in a group, use a [string list metric](metrics/string_list.html).
 
-  double_click:
-    ...
+If you need to record a closely-related group of strings, or you don't know the things the strings are related to at build time, use a [labeled string metric](metrics/labled_strings.html).
```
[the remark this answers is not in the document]
mdboom: Honestly, it's hard to find motivating examples for this one.  Even the real-world use cases that do exist probably shouldn't have been using this.

@Dexterp37 : Do you remember the motivation for this one?

docs/user/adding-new-metrics.md
mdboom: How about:

```
This metric can safely measure multiple concurrent timespans on different threads.
```

(Cross-thread timespans aren't really the intended use case, though since we use integer ids to identify the timespans, we can't stop you).  But I think this level of detail belongs in the reference of the timing distribution metric itself.

docs/user/adding-new-metrics.md
mdboom: I think if you want to answer, how long was spent between performing action A and action B, events are exactly the thing that was designed for.  I think the proposed timing list metric might be redundant to all of that.  Let me reword this to be a little more on-the-nose about the events use case.

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
+    True      :3, 12s  
+
+  section Application<br/>(init and change)
+    False     :0, 3s
+    True      :3, 12s
+    True      :15, 10s
+
+  section User<br/>(change)
+    True      :3, 22s
+```
```
mdboom: image

docs/user/adding-new-metrics.md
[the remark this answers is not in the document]
Dexterp37: Mh, no, I don't really remember. If in doubt, maybe we can disable that and re-enable if needed. I'm not sure that's currently used. The only example I can find is the [one in our docs](https://mozilla.github.io/glean/book/user/metrics/labeled_strings.html#configuration)

docs/user/adding-new-metrics.md
Dexterp37: Can we break out this lifetime example part to a separate PR?

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
   It is rare to use this lifetime outside of some metrics that are built in to Glean, such as `client_id`.
 
-### Naming things
+While lifetimes are important to understand for all metric types, they are particularly important for the metric types that record single values and don't aggregate on the client (`boolean`, `string`, `labeled_string`, `string_list`, `datetime` and `uuid`), since these metrics will send the "last known" value and missing the earlier values could be a form of unintended data loss.
+
+#### A lifetime example
+
+Let's work through an example to see how these lifetimes play out in practice. Let's suppose we have a user preference, "turbo mode", which defaults to `false`, but the user can turn it to `true` at any time.  We want to know when this flag is `true` so we can measure its affect on other metrics in the same ping.  In the following diagram, we look at a time period that sends 4 pings across two separate runs of the application. We assume here, that like Glean's built-in [metrics ping](pings/metrics.html), the developer writing the metric isn't in control of when the ping is submitted.
```
Dexterp37: Do we want to know _when_ it gets to true or what's its status when a ping is generated? I think the latter, right?

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
+In this diagram, the ping "lifetimes" are represented as rectangles, but the moment the ping is "submitted" is represented by its right edge. The user changes the "turbo mode" setting from `false` to `true` at 3 seconds.
+  
+```mermaid
+gantt
+  axisFormat %S
+  title Metric lifetimes
+  dateFormat  s
+
+  section Metrics<br/>pings
+    Ping 1 :active, p1, 0, 2s
```
Dexterp37: Would it make sense to note the value reported by these metrics ping?

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
+
+- **Application (change)**: The lifetime is set to `application` and the value is recorded into the metric with it changes. The value isn't included in Ping 1, because Glean doesn't know about it yet. After the value is changed, it is included in Pings 2 and 3, but then due to application restart it is cleared, so it is not included in any subsequent pings.
+
+- **Application (init and change)**: The lifetime is set to `application` and the value is recorded at application initialization and when it changes. The default value is included in Ping 1, and the changed value is included in Pings 2 and 3. Even though the application startup causes it to be cleared, it is set again, and all subsequent pings also have the value.
+
+- **User (change)**: The lifetime is set to `user` and the value is recorded when it changes. Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
```
Dexterp37: It's worth noting that this fails to capture if the value was flipped between two pings (i.e. flag is on, then off, then again on ). We should probably highlight the disadvantages of this option as well, otherwise we risk this seen as the holy grail of the solutions

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
+    True      :3, 12s
+    True      :15, 10s
+
+  section User<br/>(change)
+    True      :3, 22s
+```
+
+- **Ping (change)**: The lifetime is set to `ping` and the value is recorded into the metric when it changes. The value isn't included in Ping 1, because Glean doesn't know about it yet.  It is included in the first ping after being recorded (Ping 2), which causes it to be cleared and it is not included in any subsequent pings.
+
+- **Ping (init and change)**: The lifetime is set to `ping` and the value is recorded at application initialization and when it changes. The default value is included in Ping 1, and the changed value is included in Ping 2, which causes it to be cleared.  It therefore misses Ping 3, but when the application is started, it is recorded again and it is included in Ping 4.  However, this causes it to be cleared again and it is not in Ping 5.
```
Dexterp37: Is "Ping" here meaning a custom ping? Or the 'metrics' ping? I think we should disambiguate as it gets confusing: 'ping' is a lifetime and a ping name

docs/user/adding-new-metrics.md
mdboom: Yes, the latter.  That's what I meant by: "We want to know when this flag is `true` so we can measure its affect on other metrics in the same ping."  Any suggestions to make that more clear?

docs/user/adding-new-metrics.md
mdboom: Yes, but I'm not sure how to fit that in the visualization -- the value is different for each of the metric examples.  (I'm feeling a bit underwhelmed by mermaid for this -- I might need to fire up Inkscape and do it by hand, but I was trying to avoid that...)

docs/user/adding-new-metrics.md
mdboom: Good point -- missing flipping twice during a ping lifetime is a disadvantage of *every* approach, though.  It's definitely worth noting that.

docs/user/adding-new-metrics.md
mdboom: It's ping lifetime.  Will disambiguate.

docs/user/adding-new-metrics.md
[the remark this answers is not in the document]
mdboom: There is this usage in sync telemetry in the wild: https://github.com/mozilla-mobile/android-components/blob/f046b61335ad4f2f026e65536421540d31f33b97/components/support/sync-telemetry/metrics.yaml#L116

It's fine -- I don't think it's a problem we should ask to fix, but the use case doesn't seem very compelling, somehow.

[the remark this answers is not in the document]
Dexterp37: Oh , right! They are using this metric as there's no better way, right now, to report errors for them :( I agree with you: that's not a very compelling use-case.

docs/user/adding-new-metrics.md
Dexterp37: Sorry, I think I don't :\ Let's leave it as it is.

docs/user/adding-new-metrics.md
Dexterp37: Oh, right :)

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,216 @@
...
+If you need to know the order of actions relative to other actions, such as, the user performed tasks A, B, and then C, and this is meaningfully different from the user performing tasks A, C and then B, (in other words, the order is meaningful beyond just the *fact* that a set of tasks were performed), use an [event metric](metrics/event.html).
+
+Importantly, events are the most expensive metric type to record, transmit, store and analyze, so they should be used sparingly, and only when none of the other metric types are sufficient for answering your question.
+
+### For how long do you need to collect this data?
+
+Think carefully about how long the metric will be needed, and set the `expires` parameter to disable the metric at the earliest possible time. 
+This is an important component of Mozilla's [lean data practices][lean-data].
+
+When a metric is to expire in 14 days, emails will be sent from `telemetry-alerts@mozilla.com` to the `notification_emails` addresses associated with the metric.
```
chutten:
```suggestion
When a metric's expiry is within 14 days, emails will be sent from `telemetry-alerts@mozilla.com` to the `notification_emails` addresses associated with the metric.
```

docs/user/adding-new-metrics.md
chutten: The solution there is to cut a ping when the value changes, but that's a second example where the ping's schedule _is_ under the data owner's control.

docs/user/adding-new-metrics.md
mdboom: It's a little mixed up, particularly since the lifetime section was moved as part of this change.  Maybe we can do another round of review all together, and if it gets stalled, I'll pick the two apart.

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+
+- **C. Application lifetime, set on change**: The value isn't included in Ping 1, because Glean doesn't know about it yet. After the value is changed, it is included in Pings 2 and 3, but then due to application restart it is cleared, so it is not included until the value is manually toggled again.
+
+- **D. Application, set on init and change**: The default value is included in Ping 1, and the changed value is included in Pings 2 and 3. Even though the application startup causes it to be cleared, it is set again, and all subsequent pings also have the value.
 
-Metric identifiers (the combination of a metric's category and name) must be unique across all metrics that are sent by an application.
+- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
+
+- **F. User, set on change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
```
travis79:
```suggestion
- **F. User, set on init and change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
```

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+At that time, the metric should be removed, which involves removing it from the `metrics.yaml` file and removing uses of it in the source code.
+Removing a metric does not affect the availability of data already collected by the pipeline.
+
+If the metric is still needed after its expiration date, it should go back for [another round of data review](https://wiki.mozilla.org/Firefox/Data_Collection) to have its expiration date extended.
+
+### When should Glean automatically clear the measurement?
+
+The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available:
+
+- `ping` (default): The metric is cleared each time it is sent in the ping.
```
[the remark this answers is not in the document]
mdboom: Very good point.  Let's use "submit" and not "collect" to be consistent with the APIs on that, though.

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+
+### When should Glean automatically clear the measurement?
+
+The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available:
+
+- `ping` (default): The metric is cleared each time it is submitted in the ping.
   This is the most common case, and should be used for metrics that are highly dynamic, such as things computed in response to the user's interaction with the application.
-- `application`: The metric is related to an application run, and is reset only when the application restarts.
+  
+- `application`: The metric is related to an application run, and is cleared every time the application restarts.
```
Dexterp37:
```suggestion
- `application`: The metric is related to an application run, and is cleared after the application restarts and any Glean-owned ping, due at startup, is submitted.
```

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+
+- **C. Application lifetime, set on change**: The value isn't included in Ping 1, because Glean doesn't know about it yet. After the value is changed, it is included in Pings 2 and 3, but then due to application restart it is cleared, so it is not included until the value is manually toggled again.
+
+- **D. Application, set on init and change**: The default value is included in Ping 1, and the changed value is included in Pings 2 and 3. Even though the application startup causes it to be cleared, it is set again, and all subsequent pings also have the value.
 
-Metric identifiers (the combination of a metric's category and name) must be unique across all metrics that are sent by an application.
+- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all subsequent pings.
```
Dexterp37:
```suggestion
- **E. User, set on change**: The default value is missing from Ping 1, but since `user` lifetime metrics aren't cleared unless the user profile is reset (e.g. on Android, when the product is uninstalled), it is included in all subsequent pings.
```

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+
+- **F. User, set on init and change**: Since `user` lifetime metrics aren't cleared unless the user profile is reset, it is included in all pings.
```
Dexterp37: Would you consider adding a note mentioning that `user` lifetime metrics will be kept on sending even if the metric was set in Run 1 but never in Run 2?

docs/user/adding-new-metrics.md
```diff
@@ -1,97 +1,184 @@
...
+
+#### What if none of these lifetimes are appropriate?
+
+If the timing at which the metric is sent in the ping needs to closely match the timing of the metrics value, the best option is to use a [custom ping](pings/custom.html) to manually control when pings are sent.
+
+### What should this new metric be called?
+
+#### Reuse names from other applications
+
+There's a lot of value using the same name for analogous metrics collected across different products. For example, BigQuery makes it simple to join columns with the same name across multiple tables. Therefore, we encourage you to investigate if a similar metric is already being collected by another product and use the same naming whenever feasible.
```
Dexterp37: Thinking again of this, if some metric is really shared across products of the same platform, maybe devs should consider writing a component/library using glean?

docs/user/adding-new-metrics.md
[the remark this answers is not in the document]
travis79: That's even better!

docs/user/adding-new-metrics.md
mdboom: I think it's a good idea to add a note about that, but it may not always be feasible.  Think of "open new tab", for example.  That's handled by a completely different code in Desktop vs. Fenix that it probably doesn't make sense to merge, even though they both have that "concept" of "open new tab".

I think maybe we say something like:

"Therefore, we encourage you to investigate if a similar metric is already being collected by another product.  If it is, there may be an opportunity for code reuse across these products.  But even if that's not feasible, use the same metric name for similar actions and concepts whenever possible."

Dexterp37: Good enough! Thank you for your patience!

PR #751, opened 2020-03-03
Update Python SDK documentation for logging pings 

hackebrot:
My application used `logging.basicConfig(level=logging.INFO)` to configure logging and I realized that pings weren't logged even though I set the according configuration option `log_pings` to `True` (see [documentation](https://mozilla.github.io/glean/book/user/debugging/python.html#logging-pings)). 📝  

This PR updates the Python SDK documentation to point out that the `glean` logger uses level `DEBUG` and consumers need to configure logging accordingly, if they want pings to be logged.

### DIFF
--- a/docs/user/debugging/python.md
+++ b/docs/user/debugging/python.md
@@ -25,6 +25,15 @@ After doing so, something like `pings.custom_ping.submit()` will send the custom
 
 ## Logging pings
 
-If the `Glean.configuration.log_pings` property is set to `True`, pings are logged to the console whenever they are submitted.
+If the `Glean.configuration.log_pings` property is set to `True`, pings are
+logged to the console on `DEBUG` level whenever they are submitted. You can set
+this property in a similar way as the `ping_tag` property above.
 
-You can set this property in a similar way as the `ping_tag` property above.
+Make sure that when you configure logging in your application, you set the
+level for the `glean` logger to `DEBUG` or higher. Otherwise pings won't be
+logged even if `log_pings` is set to `True`.
+
+See the [Python logging documentation][python-logging] for more information.
+
+[python-logging]: https://docs.python.org/3.8/library/logging.html
+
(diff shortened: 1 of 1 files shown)


badboy:
> Both of these would bring Python more in line with the other platforms.

Other platforms _also_ rely on the platform's logging mechanism and use the `debug` level.

hackebrot:
No worries @mdboom! 🙂 

As a consumer of the Python SDK, I would probably prefer to configure this through Python's logging capabilities rather than a configuration option in Glean (`log_pings`):

I would expect Glean `DEBUG` logs to show for `logging.basicConfig(level=logging.DEBUG)`, but no logs if my application does not configure logging. And if I wish to suppress any logs from Glean even though I use `DEBUG` level for my application, I could always set this explicitely by doing the following:

```python
import logging

logging.getLogger("glean").setLevel(logging.WARNING)
```

https://docs.python.org/3/howto/logging.html#configuring-logging-for-a-library

mdboom:
> > Both of these would bring Python more in line with the other platforms.
> 
> Other platforms _also_ rely on the platform's logging mechanism and use the `debug` level.

Oh, right. Sorry I was wrong about that. I think on Android Studio I'm just used to the logging level being chattier by default.

In that case, I think this PR seems like the way to go, with one small in-line change request below.

docs/user/debugging/python.md
```diff
@@ -25,6 +25,15 @@ After doing so, something like `pings.custom_ping.submit()` will send the custom
...
-If the `Glean.configuration.log_pings` property is set to `True`, pings are logged to the console whenever they are submitted.
+If the `Glean.configuration.log_pings` property is set to `True`, pings are
+logged to the console on `DEBUG` level whenever they are submitted. You can set
+this property in a similar way as the `ping_tag` property above.
+
+Make sure that when you configure logging in your application, you set the
+level for the `glean` logger to `DEBUG` or higher. Otherwise pings won't be
+logged even if `log_pings` is set to `True`.
+
+See the [Python logging documentation][python-logging] for more information.
```
mdboom: Can we add a short example of turning this log level on (just for Glean)?  And still link to the full Python docs as we do here.

hackebrot: Done! ✅

PR #755, opened 2020-03-05
[doc only] Illustrate ping schedules and various timing-related metrics

mdboom:


docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The following diagram shows a typical timeline of a mobile application, when pings are sent and what timing-related information is included.
+
+![ping timeline diagram](ping_timing.svg)
+
+There are two distinct runs of the application, where the OS shutdown the application at the end of Run 1, and the user started it up again at the beginning of Run 2.
+
+There are three distinct foreground sessions, where the application was visible on the screen and the user was able to interact with it.
+
+The rectangles for the baseline and metrics pings represent the "ping lifetimes" of those pings, which always start exactly at the end of the preceding ping.  The `ping_info.start_time` and `ping_info.end_time` metrics included in these pings correspond to these ping lifetimes.
```
Dexterp37:
```suggestion
The rectangles for the `baseline` and `metrics` pings represent the measurement windows of those pings, which always start exactly at the end of the preceding ping.  The `ping_info.start_time` and `ping_info.end_time` metrics included in these pings correspond to these beginning and the end of their measurement windows.
```

We use "measurement window" as the term to describe "ping lifetime" in other docs. Does this make sense?

docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The `baseline.duration` metric (included only in baseline pings) corresponds to the session length, which, since ping lifetimes always extend to the next ping, is not always the same thing as the `baseline` ping's lifetime.
```
Dexterp37:
```suggestion
The `baseline.duration` metric (included only in `baseline` pings) corresponds to amount of time the application spent on the foreground, which, since measurement window always extend to the next ping, is not always the same thing as the `baseline` ping's measurement window.
```

docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The `submission_timestamp` is the time the ping was received at the telemetry endpoint.  It is not exactly the same as `ping_info.end_time`, since there may be various networking and system latencies (represented by the dotted horizontal line, not to scale).  Also of note is that `start_time`/`end_time` are measured using the client's real-time clock in its local timezone, which may include clock skew or completely incorrect values.
```
Dexterp37:
```suggestion
The `submission_timestamp` is the time the ping was received at the telemetry endpoint, added by the ingestion pipeline.  It is not exactly the same as `ping_info.end_time`, since there may be various networking and system latencies (represented by the dotted horizontal line, not to scale).  Also of note is that `start_time`/`end_time` are measured using the client's real-time clock in its local timezone, which may include clock skew or completely correct values.
```

Was "completely incorrect" intentional?
I'm also wondering if we should be a bit more formal and precise about the latencies we know exist. Chris [wrote about that](https://chuttenblog.wordpress.com/2016/11/10/data-science-is-hard-case-study-latency-of-firefox-crash-rates/) a long time ago (i.e. "latency = reporting delay + submission delay + aggregation delay + migration delay + query scheduling delay"), while we were reducing latency in Firefox. Maybe there's an opportunity to make it part of our docs and reference it here?

docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The "Baseline 2" ping illustrates an important corner case. When "Session 2" ended, the OS also shut down the entire process, and Glean did not have an opportunity to send a `baseline` ping immediately.  In this case, it is sent at the next available opportunity when the application starts up again in "Run 2".  This `baseline` ping is annotated with the reason code `dirty_startup`.
```
Dexterp37:
```suggestion
The "Baseline 2" ping illustrates an important corner case. When "Session 2" ended, the OS also shut down the entire process, and the Glean SDK did not have an opportunity to send a `baseline` ping immediately.  In this case, it is sent at the next available opportunity when the application starts up again in "Run 2".  This `baseline` ping is annotated with the reason code `dirty_startup`.
```

docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The "Metrics 2" ping likewise illustrates another important corner case. "Metrics 1" was able to be sent at the target time of 04:00 because the application was currently running.  However, the next time 04:00 came around, the application was not active, so Glean was unable to send a metrics ping.  It is sent at the next available opportunity, when the application starts up again in "Run 2".  This `metrics` ping is annotated with the reason code `overdue`.
```
Dexterp37:
```suggestion
The "Metrics 2" ping likewise illustrates another important corner case. "Metrics 1" was able to be sent at the target time of 04:00 (local device time) because the application was currently running.  However, the next time 04:00 came around, the application was not active, so the Glean SDK was unable to send a `metrics` ping.  It is sent at the next available opportunity, when the application starts up again in "Run 2".  This `metrics` ping is annotated with the reason code `overdue`.
```

docs/user/pings/ping-schedules-and-timings.md
mdboom: Totally makes sense.  Thanks.

docs/user/pings/ping-schedules-and-timings.md
mdboom: That's a good idea.  Since I think this is the only place in the docs that talk about submission_timestamp, we might as well do that here.

mdboom: My intention here is to say that the local clock may have a small amount of skew from universal time, or it may be wildly off -- as we see in the number of pings that are still in the 1970's.  But maybe that distinction doesn't matter -- we just need to say it's generally not 100% unreliable.

mdboom:
Marking this as WIP, as I'd like to get mgorlick's review before merging.  (?ni on [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1615663))

docs/user/pings/ping-schedules-and-timings.md
Dexterp37: > 
> 
> My intention here is to say that the local clock may have a small amount of skew from universal time, or it may be wildly off 

Oh! I didn't understand that, it's clearer now. Maybe change this to "or be completely incorrect?"? Or even leave it as it is, it might be just me not understanding this on a first read :)

mdboom: I decided to say "it's not a fully reliable source of time."  The details of why is probably beyond the scope of this document.

As an aside, we may still want to do something like this someday: https://www.chromium.org/developers/design-documents/sane-time

mdboom:
Making this non-WIP again.  Let's just review one last time within the team and merge.

docs/user/pings/ping-schedules-and-timings.md
```diff
@@ -0,0 +1,21 @@
...
+
+The `baseline.duration` metric (included only in `baseline` pings) corresponds to amount of time the application spent on the foreground, which, since measurement window always extend to the next ping, is not always the same thing as the `baseline` ping's measurement window.
+
+The `submission_timestamp` is the time the ping was received at the telemetry endpoint, added by the ingestion pipeline.  It is not exactly the same as `ping_info.end_time`, since there may be various networking and system latencies both on the client and in the ingestion pipeline (represented by the dotted horizontal line, not to scale).  Also of note is that `start_time`/`end_time` are measured using the client's real-time clock in its local timezone, which is not a fully reliable source of time.
+
+The "Baseline 2" ping illustrates an important corner case. When "Session 2" ended, the OS also shut down the entire process, and the Glean SDK did not have an opportunity to send a `baseline` ping immediately.  In this case, it is sent at the next available opportunity when the application starts up again in "Run 2".  This `baseline` ping is annotated with the reason code `dirty_startup`.
```
magorlick, line 19: Thank you for this thoughtful write up and diagram.  I think this document will make the interpretation much clearer.  

For these last two points about 'corner cases' -  do we have any reports that allow us to monitor these effects on an ongoing basis?  It would be nice to link out to some data points on how big of a corner case this is.

mdboom: We don't have anything simple to track this, but it would be good to create that.

It won't be possible to link to them from these docs, though, which are public.

PR #756, opened 2020-03-05
Bug 1620320: Add iOS-specific background state definition [doc-only]

travis79:


docs/user/pings/index.md
```diff
@@ -104,6 +104,34 @@ A pre-defined set of headers is additionally sent along with the submitted ping:
 ## Defining background state
 
 These docs refer to application 'background' state in several places.
-This specifically means when the activity is no longer visible to the user, it has entered the Stopped state, and the system invokes the [`onStop()`](https://developer.android.com/reference/android/app/Activity.html#onStop()) callback.
-This may occur, for example, when a newly launched activity covers the entire screen.
+
+{{#include ../../tab_header.md}}
+
+<div data-lang="Android" class="tab">
```
mdboom: It's sort of not-quite-the-right-thing, but on other pages we've used `Kotlin` and `Swift` rather than `Android` and `iOS`.  The reason it matters is that the user's preference is stored in local storage and follows them from page to page.

As part of the larger discussion about we better talk about languages vs. platforms, we could address that then...

travis79: Yeah, it felt a little _too_ specific to use the languages, but the preference trumps the tab title, I agree.

docs/user/pings/index.md
```diff
@@ -104,6 +104,34 @@ A pre-defined set of headers is additionally sent along with the submitted ping:
...
-This may occur, for example, when a newly launched activity covers the entire screen.
+
+{{#include ../../tab_header.md}}
+
+<div data-lang="Kotlin" class="tab">
+
+For Android, this specifically means when the activity is no longer visible to the user, it has entered the
+Stopped state, and the system invokes the [`onStop()`](https://developer.android.com/reference/android/app/Activity.html#onStop()) callback.
+
+This may occur, if the user uses `Overview` button to change to another, the user presses the `Back` button and
```
travis79:
```suggestion
This may occur, if the user uses `Overview` button to change to another app, the user presses the `Back` button and
```

PR #758, opened 2020-03-05
Bug 1594036: Document that `initialize` must always be called [doc-only]

travis79:


docs/user/general-api.md
```diff
@@ -27,6 +27,13 @@ The Glean SDK provides a general API that supports the following operations. See
 
 The following steps are required for applications using the Glean SDK, but not libraries.
 
+> **Note**: The `initialize` function _must_ be called, even if telemetry upload is disabled.
+> Glean needs to perform maintenance tasks even when telemetry is disabled, and because Glean
+> does this as part of it's initialization, it is _required_ to always call the `initialize`
```
mdboom:
```suggestion
> does this as part of its initialization, it is _required_ to always call the `initialize`
```

docs/user/general-api.md
```diff
@@ -27,6 +27,13 @@ The Glean SDK provides a general API that supports the following operations. See
...
+> function. Otherwise, Glean won't be able to clean up collected data, disable queuing of pre-init
+> tasks, or reset the user lifetime metrics and trigger a deletion request ping in the event that
```
mdboom: Do we want to mention "reset the user lifetime metrics", since it's still at the proposal stage?  Maybe you meant "application lifetime metrics"?

travis79: Makes sense, I was thinking more along the lines that we clear metrics as a side effect of disabling telemetry.  I'll remove this specific reference and make the statement more general.

PR #762, opened 2020-03-06
Python: Implement timing distribution

mdboom:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
```diff
@@ -76,7 +76,7 @@ class TimingDistributionMetricType internal constructor(
             return null
         }
 
-        // The Rust code for [stopAndAccumulate] runs async and we need to use the same clock for start and stop.
+        // The Rust code for [start] runs async and we need to use the same clock for start and stop.
```
badboy: Woops, good catch

glean-core/python/glean/metrics/distribution_data.py
```diff
@@ -0,0 +1,59 @@
...
+    @property
+    def sum(self) -> int:
+        return self._sum
+
+    @property
+    def count(self) -> int:
+        return sum(self._values.values())
+
+    @classmethod
+    def from_json_string(cls, json_string: str) -> "DistributionData":
```
badboy: Forgive my ignorance, is `cls` just as special as `self`, turning this into a static class method?

glean-core/python/glean/metrics/timing_distribution.py
```diff
@@ -0,0 +1,190 @@
...
+
+
+from .distribution_data import DistributionData
+from .lifetime import Lifetime
+from .timeunit import TimeUnit
+
+
+class TimingDistributionMetricType:
+    """
+    This implements the developer facing API for recording timingdistribution metrics.
```
badboy:
```suggestion
    This implements the developer facing API for recording timing distribution metrics.
```

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
badboy: ooooh wait, the comment was correct, just not super clear.

`start` _does_ run sync (there's no dispatcher in use).
But `stop` runs async (we don't care about its return value) and THAT's why we need to use `getElapsedTimeNanos` in both because we need to use the same clock source.
I'm all for extending the comment, but the original was already more correct than after your change now.

glean-core/python/glean/metrics/distribution_data.py
badboy: Raphael explained it to me. Naming doesn't _really_ matter, the decorator does though.

glean-core/python/glean/metrics/timing_distribution.py
```diff
@@ -0,0 +1,190 @@
...
+        `stop_and_ccumulate`): in that case the original start time will be
+        preserved.
+
+        Returns:
+            timer_id: The object to associate with this timing.
+        """
+        if self._disabled:
+            return None
+
+        # The Rust code for `start` runs async and we need to use the same
```
badboy: As in the Kotlin equivalent this comment is now incorrect.

docs/user/metrics/timing_distribution.md
```diff
@@ -172,12 +172,59 @@ XCTAssertEqual(1, pages.pageLoad.testGetNumRecordedErrors(.invalidValue))
...
+        self.timer_id = metrics.pages.page_load.start()
+
+    def on_page_loaded(self, event):
+        # ...
+        metrics.pages.page_load.store_and_accumulate(self.timer_id)
+```
+
+There are test APIs available too.  For convenience, properties `sum` and `count` are exposed to facilitate validating that data was recorded correctly.
+
+Continuing the `page_load` example above, at this point the metric should have a `sum == 11` and a `count == 2`:
```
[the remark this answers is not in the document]
mdboom: Makes sense.  We can handle updating all of the language examples at once as part of that bug.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
mdboom: Ok -- I'll update this to be *both* correct *and* clear.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
```diff
@@ -76,8 +76,10 @@ class TimingDistributionMetricType internal constructor(
             return null
         }
 
-        // The Rust code for [stopAndAccumulate] runs async and we need to use the same clock for start and stop.
-        // Therefore we take the time on the Kotlin side.
+        // Even though the Rust code for [start] runs syncronously, the Rust
```
badboy:
```suggestion
        // Even though the Rust code for [start] runs synchronously, the Rust
```

glean-core/python/glean/metrics/timing_distribution.py
```diff
@@ -0,0 +1,192 @@
...
+        `stop_and_ccumulate`): in that case the original start time will be
+        preserved.
+
+        Returns:
+            timer_id: The object to associate with this timing.
+        """
+        if self._disabled:
+            return None
+
+        # Even though the Rust code for [start] runs syncronously, the Rust
```
badboy:
```suggestion
        # Even though the Rust code for [start] runs synchronously, the Rust
```

PR #764, opened 2020-03-09
Add missing Python docs and make some things private

mdboom:
This was just an attempt to clean up the Python API docs.  Discovered in the process that some things that should be private were public (and thus being documented).  Technically a breaking change, but no one should have been relying on it.


mdboom:
It's really only a convention, but it's enforced by some linter tools and important for this case, our documentation tool knows not to document them.

PR #765, opened 2020-03-09
Python: Add string list metric

mdboom:


glean-core/android/src/test/java/mozilla/telemetry/glean/private/StringListMetricTypeTest.kt
```diff
@@ -97,31 +97,6 @@ class StringListMetricTypeTest {
         assertEquals("added1", snapshot2[3])
     }
 
-    @Test
-    fun `lists with no lifetime must not record data`() {
```
Dexterp37: Just to make sure: this is being removed because it's already covered by Rust tests, right?

mdboom: Sorry to not mention.  It's actually being removed because it's identical to the test below it.

PR #775, opened 2020-03-20
Add testing for python 3.8

wlach:


wlach:
Would be desirable to also test this on Windows and Mac (for each python variant), but this is a start.

PR #776, opened 2020-03-20
Python: Windows x86 support

mdboom:
This addresses the following bugs to make 32-bit Windows builds pass the
unit tests.

1622909: Release 32-bit wheels for Windows
1623335: Functional bucketing fixes
1623587: Unable to detect locale on Windows

Also relevant (but a follow-on) is:

1623307: Add Windows tests to CI

### DIFF
--- a/docs/user/collected-metrics/metrics.md
+++ b/docs/user/collected-metrics/metrics.md
@@ -60,7 +60,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.baseline.duration |[timespan](https://mozilla.github.io/glean/book/user/metrics/timespan.html) |The duration of the last foreground session.  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
-| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES"). If the locale can't be determined on the system, the value is "C", which is the POSIX value for "default locale".  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
+| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES"). If the locale can't be determined on the system, the value is "und", to indicate "undetermined".  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
 
 ## deletion-request
 
--- a/docs/user/collected-metrics/metrics.md
+++ b/docs/user/collected-metrics/metrics.md
@@ -60,7 +60,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.baseline.duration |[timespan](https://mozilla.github.io/glean/book/user/metrics/timespan.html) |The duration of the last foreground session.  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
-| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES").  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
+| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES"). If the locale can't be determined on the system, the value is "C", which is the POSIX value for "default locale".  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
 
 ## deletion-request
 
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -32,8 +32,8 @@ glean.baseline:
       - baseline
     description: |
       The locale of the application (e.g. "es-ES").
-      If the locale can't be determined on the system, the value is "C",
-      which is the POSIX value for "default locale".
+      If the locale can't be determined on the system, the value is "und",
+      to indicate "undetermined".
     bugs:
       - https://bugzilla.mozilla.org/1512938
       - https://bugzilla.mozilla.org/1525540
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -32,6 +32,7 @@ glean.baseline:
       - baseline
     description: |
       The locale of the application (e.g. "es-ES").
+      If the locale can't be determined on the system, the value is "C".
     bugs:
       - https://bugzilla.mozilla.org/1512938
       - https://bugzilla.mozilla.org/1525540
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -32,7 +32,8 @@ glean.baseline:
       - baseline
     description: |
       The locale of the application (e.g. "es-ES").
-      If the locale can't be determined on the system, the value is "C".
+      If the locale can't be determined on the system, the value is "C",
+      which is the POSIX value for "default locale".
     bugs:
       - https://bugzilla.mozilla.org/1512938
       - https://bugzilla.mozilla.org/1525540
--- a/glean-core/python/setup.py
+++ b/glean-core/python/setup.py
@@ -14,9 +14,16 @@ import wheel.bdist_wheel
 
 
 platform = sys.platform
-if os.environ.get("GLEAN_PYTHON_MINGW_BUILD") or platform.startswith("win"):
-    platform = "windows"
 
+if os.environ.get("GLEAN_PYTHON_MINGW_I686_BUILD"):
+    mingw_arch = "i686"
+elif os.environ.get("GLEAN_PYTHON_MINGW_X86_64_BUILD"):
+    mingw_arch = "x86_64"
+else:
+    mingw_arch = None
+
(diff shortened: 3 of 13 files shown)


glean-core/src/util.rs
```diff
@@ -111,6 +111,52 @@ pub(crate) fn truncate_string_at_boundary_with_error<S: Into<String>>(
     }
 }
 
+#[cfg(all(target_arch = "x86", target_os = "windows"))]
+pub mod floating_point_context {
```
badboy: Any documentation on what this does and why we need it?

glean-core/src/histogram/functional.rs
```diff
@@ -24,6 +26,8 @@ pub struct Functional {
 impl Functional {
     /// Instantiate a new functional bucketing.
     fn new(log_base: f64, buckets_per_magnitude: f64) -> Functional {
+        let _ = FloatingPointContext::new();
```
badboy: Assigning to `_` will drop immediately. If you want to drop it at the end of the scope you need to give it a name:

```suggestion
        let _fpc = FloatingPointContext::new();
```

glean-core/src/histogram/functional.rs
```diff
@@ -34,11 +38,13 @@ impl Functional {
     /// mathematical concept, even though the internal representation is stored and
     /// sent using the minimum value in each bucket.
     fn sample_to_bucket_index(&self, sample: u64) -> u64 {
+        let _ = FloatingPointContext::new();
```
badboy: As above

glean-core/src/histogram/functional.rs
```diff
@@ -34,11 +38,13 @@ impl Functional {
...
         ((sample + 1) as f64).log(self.exponent) as u64
     }
 
     /// Determines the minimum value of a bucket, given a bucket index.
     fn bucket_index_to_bucket_minimum(&self, index: u64) -> u64 {
+        let _ = FloatingPointContext::new();
```
badboy: As above

glean-core/metrics.yaml
```diff
@@ -32,6 +32,7 @@ glean.baseline:
       - baseline
     description: |
       The locale of the application (e.g. "es-ES").
+      If the locale can't be determined on the system, the value is "C".
```
Dexterp37: Can we add a small comment explaining what's "C" and why it is a meaningful value?

glean-core/src/util.rs
```diff
@@ -111,6 +111,52 @@ pub(crate) fn truncate_string_at_boundary_with_error<S: Into<String>>(
...
+    use libc::size_t;
+
+    #[link(name = "m")]
+    extern "C" {
+        fn _controlfp_s(current: *mut size_t, new: size_t, mask: size_t) -> size_t;
+    }
+
+    const MCW_PC: size_t = 0x00030000;
```
Dexterp37: nit: can we add some comments about these flags or link to the articles were we found them?

glean-core/src/util.rs
mdboom: See https://bugzilla.mozilla.org/show_bug.cgi?id=1623335.  Should probably add a comment with a link to the bug.

glean-core/metrics.yaml
Dexterp37: Hey @mdboom , this was not updated - was this intentional?

mdboom: Oops.  Should have updated this comment as well.

glean-core/src/util.rs
```diff
@@ -111,6 +111,64 @@ pub(crate) fn truncate_string_at_boundary_with_error<S: Into<String>>(
...
 }
 
+// On i686 on Windows, the CPython interpreter sets the FPU precision control
+// flag to 53 bits of precision, rather than the 64 bit default. This causes
+// different floating point results than on other architectures. This context
+// manager makes it easy to set the precision to 64-bits around floating point
+// operations.
+//
+// See https://bugzilla.mozilla.org/show_bug.cgi?id=1623335 for additional context.
+#[cfg(all(target_arch = "x86", target_os = "windows"))]
```
badboy: Should this be limited to `target_env = "mscv"` additionally?
See [docs](https://doc.rust-lang.org/reference/conditional-compilation.html#target_env).

mdboom:
I don't think so.  The problem exists whether the target is gnu or msvc...
The problem really lies in the CPython build.

docs/user/collected-metrics/metrics.md
```diff
@@ -60,7 +60,7 @@ The following metrics are added to the ping:
 | Name | Type | Description | Data reviews | Extras | Expiration |
 | --- | --- | --- | --- | --- | --- |
 | glean.baseline.duration |[timespan](https://mozilla.github.io/glean/book/user/metrics/timespan.html) |The duration of the last foreground session.  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
-| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES").  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
+| glean.baseline.locale |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |The locale of the application (e.g. "es-ES"). If the locale can't be determined on the system, the value is "C", which is the POSIX value for "default locale".  |[1](https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3)||never |
```
Dexterp37: In light of [this](https://bugzilla.mozilla.org/show_bug.cgi?id=1623587#c6) I think we should probably change it.

badboy:
> I don't think so. The problem exists whether the target is gnu or msvc... The problem really lies in the CPython build.

I see. Windows it is then!

docs/user/collected-metrics/metrics.md
mdboom: Oh, cool.  I couldn't find that -- but I was looking in the POSIX standards, not the Unicode Locale Database.  I suppose the latter is more relevant here anyway.  Let's use "und" -- I still think that's better than "en-US" as a fall back.

glean-core/python/setup.py
```diff
@@ -48,9 +55,12 @@
...
 setup_requirements = []
 
-shared_object_build_dir = "../../target/debug/"
-if os.environ.get("GLEAN_PYTHON_MINGW_BUILD"):
+if mingw_arch == "i686":
+    shared_object_build_dir = "../../target/i686-pc-windows-gnu/debug/"
+elif mingw_arch == "x86_64":
     shared_object_build_dir = "../../target/x86_64-pc-windows-gnu/debug/"
+else:
+    shared_object_build_dir = "../../target/debug/"
```
badboy: I did not realize we hard-code the output directory.
This is ... not optimal I assume.

1. We probably want release builds eventually
2. The sub directory of the target is only like that if cross-compiled.

We should file a bug and fix this up later.

glean-core/src/histogram/functional.rs
```diff
@@ -34,11 +38,13 @@ impl Functional {
     /// mathematical concept, even though the internal representation is stored and
     /// sent using the minimum value in each bucket.
     fn sample_to_bucket_index(&self, sample: u64) -> u64 {
+        let _fpc = FloatingPointContext::new();
```
badboy: Hm, reading this again, I wonder if we should explicitly call out that we rely on the `Drop` behavior to reset the behavior?
In Rust _not_ calling Drop is safe. The way we use it it's guaranteed it's called, but we might call it out.
The alternatives would be:

* Wrap it in a closure, that way we set and re-set on entering and exiting.
* or explicitly call `mem::drop(fpc)` to make it clear.

I don't want to block on this, it's safe behavior for now.

glean-core/python/setup.py
mdboom: Yeah -- we've had a bug for this for a while: https://bugzilla.mozilla.org/show_bug.cgi?id=1597714

The cross compiling subdirectories are only used when cross compiling is indicated using the `GLEAN_PYTHON_` env variables, and that's really just for making the official builds in CircleCI.  It's not really "supported" for developer use, as indicated by the docs:  https://mozilla.github.io/glean/book/dev/python/setting-up-python-build-environment.html#building-wheels-for-windows

badboy: 👍 so that's how few things I know about our Python package :D

glean-core/src/histogram/functional.rs
mdboom: I'm happy to clean this up to make it more obvious / safe.

Just curious though -- where is it documented that `Drop` might not be called?  The [Drop docs](https://doc.rust-lang.org/std/ops/trait.Drop.html) and the [Drop chapter](https://doc.rust-lang.org/book/ch15-03-drop.html) say it runs when the variable goes out of scope.  The goal here is certainly to minimize the scope to just around the floating point operation, and I wouldn't want this to have effect any longer.

badboy: Sorry, should have been more specific:

Drop _is_ called at the end of the scope, that part is guaranteed, especially in our case.

However, in general, it _is_ safe* as per language definition for `Drop` not to be called (it can happen through reference-count cycles, explicit `mem::forget`s or other things). Lots of stuff relies on drop for correct behavior (e.g. unlocking a mutex on drop), but not for safe behavior.

(\* safe here is the Rust definition of safety as "memory-safety")

I just want to save ourselves from forgetting what we rely on in this piece of code when we read it again in 6 weeks.

mdboom: Ah, that makes sense.  I think maybe the best course of action is to comment all of these call sites -- at a minimum it's not clear why this unused variable is there and someone might "remove the cruft" (they would break the unit tests on Windows x86 that we don't yet have in CI).

PR #777, opened 2020-03-23
Bug 1598993 - Add convenience `measure` methods to TimingDistribution and Timespan metrics

travis79:
This adds convenience methods to simplify measuring a function or block of code to both Android and iOS.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimespanMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimespanMetricType.kt
@@ -98,6 +98,18 @@ class TimespanMetricType internal constructor(
     }
 
     /**
+     * Convenience method to simplify measuring a function or block of code
+     */
+    inline fun <U> measure(funcToMeasure: () -> U): U {
+        start()
+        try {
+            return funcToMeasure()
+        } finally {
+            stop()
+        }
+    }
+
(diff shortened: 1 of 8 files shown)


travis79:
This passes the Swift tests per `make test-swift`, I don't see any reason to run them again until the merge with master.

mdboom:
Should we file bugs for adding this API to the other languages -- I don't think there's any reason we couldn't also do this there.

travis79:
> Should we file bugs for adding this API to the other languages -- I don't think there's any reason we couldn't also do this there.

This adds it to both Kotlin and Swift. I wasn't sure how to accomplish this in Python, this relates to the 'higher order functions' question I was ask you about.  I'm happy to file bugs for the other languages, or to take a stab at adding them to this PR, whichever you think.

mdboom:
Ah -- sorry I missed the Kotlin.  Maybe just file a bug for Python -- it's totally orthogonal so doesn't need to hold this up.

travis79:
> Ah -- sorry I missed the Kotlin. Maybe just file a bug for Python -- it's totally orthogonal so doesn't need to hold this up.

[Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1624626), and I'd still like to learn how something like this is done in Python, so maybe you can point me in the right direction when we have time to chat.

PR #779, opened 2020-03-24
Python Windows testing

mdboom:
1623307: Add Windows testing to CI
1624677: Make floating point calculations match on x86_64 Windows within Python

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -108,6 +108,57 @@ commands:
           name: Python tests
           command: make test-python
 
+  build-windows-x86_64-wheel:
+    steps:
+      - install-rustup
+      - setup-rust-toolchain
+      - install-mingw
+      - run:
+          name: Install Python development tools for host
+          command:
+            make python-setup
+      - run:
+          name: Build Windows glean_ffi.dll
+          command:
+            cargo build --target x86_64-pc-windows-gnu
+      - run:
+          name: Build Windows wheel
+          command: |
+            cd glean-core/python
+            GLEAN_PYTHON_MINGW_X86_64_BUILD=1 .venv/bin/python3 setup.py bdist_wheel
+
+  build-windows-i686-wheel:
+    steps:
+      - install-rustup
+      - setup-rust-toolchain
+      - install-mingw
+      - run:
+          name: Install Python development tools for host
+          command:
+            make python-setup
+      - run:
+          name: Build Windows glean_ffi.dll
+          command:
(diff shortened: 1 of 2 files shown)


glean-core/src/util.rs
```diff
@@ -142,7 +146,7 @@ pub mod floating_point_context {
     impl FloatingPointContext {
         pub fn new() -> Self {
             let mut current: size_t = 0;
-            let _err = unsafe { _controlfp_s(&mut current, PC_64, MCW_PC) };
+            let _err = unsafe { _controlfp_s(&mut current, PC_64 | RC_CHOP, MCW_PC | MCW_RC) };
```
Dexterp37: Oh, interesting. Is python changing this flag internally as well? Can you update the comment at the top of this class mentioning that we set rounding in addition to the precision?

.circleci/config.yml
```diff
@@ -134,7 +185,7 @@ commands:
             # https://github.com/rust-lang/rust/issues/49078
             # https://wiki.archlinux.org/index.php/Rust#Windows
             for lib in crt2.o dllcrt2.o libmsvcrt.a; do cp -v /usr/x86_64-w64-mingw32/lib/$lib ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/x86_64-pc-windows-gnu/lib/; done
-            for lib in crt2.o dllcrt2.o libmsvcrt.a; do cp -v /usr/i686-w64-mingw32/lib/$lib ~/.rustup/toolchains/stable-i686-unknown-linux-gnu/lib/rustlib/i686-pc-windows-gnu/lib/; done
+            for lib in crt2.o dllcrt2.o libmsvcrt.a; do cp -v /usr/i686-w64-mingw32/lib/$lib ~/.rustup/toolchains/stable-x86_64-unknown-linux-gnu/lib/rustlib/i686-pc-windows-gnu/lib/; done
```
Dexterp37: Why are we changing this line partially?

mdboom: It was wrong when it was first added in #776, but it never ran in CI since it used to only be used for deploying releases.  The first part of that path is the host architecture (which is always 64-bit), and the second is the target (which is 32-bit).

glean-core/src/util.rs
mdboom: Yes -- the rounding was changing on 64-bit, the precision was changing on 32-bit.  I'll update the comment.

.circleci/config.yml
Dexterp37: Thanks for clarifying!

PR #783, opened 2020-03-26
1604191: Python: run Glean work on a worker thread

mdboom:
This runs most of Glean's work (and its blocking I/O) on a separate worker thread.

This removes the need for users to do this themselves for performance reasons.  It doesn't really address @wlach's issue of wanting to fire off pings completely independently as a library, which still requires more discussion (see https://bugzilla.mozilla.org/show_bug.cgi?id=1625166).

### DIFF
--- a/docs/user/adding-glean-to-your-project.md
+++ b/docs/user/adding-glean-to-your-project.md
@@ -266,6 +266,10 @@ Please refer to the [custom pings documentation](pings/custom.md).
 > **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
 > This is also required for any new metric automatically collected by the Glean SDK.
 
+### Parallelism
+
+All of Glean's target languages use a separate worker thread to do most of Glean's work, including any I/O. This thread is fully managed by Glean as an implementation detail. Therefore, users should be free to use the Glean API wherever it is most convenient, without worrying about the performance impact of updating metrics and sending pings.
+
 ### Testing metrics
 
 In order to make testing metrics easier 'out of the box', all metrics include a set of test API functions in order to facilitate unit testing.  These include functions to test whether a value has been stored, and functions to retrieve the stored value for validation.  For more information, please refer to [Unit testing Glean metrics](testing-metrics.md).
--- a/glean-core/python/glean/_dispatcher.py
+++ b/glean-core/python/glean/_dispatcher.py
@@ -45,7 +45,7 @@ class _ThreadWorker:
         # that importing Glean alone does not start an unnecessary thread.
         self._started = False
 
-    def _add_task(self, sync: bool, task: Callable, *args, **kwargs):
+    def add_task(self, sync: bool, task: Callable, *args, **kwargs):
         """
         Add a task to the worker queue.
 
@@ -68,18 +68,6 @@ class _ThreadWorker:
             self._queue.put((task, args, kwargs))
             if sync:
                 self._queue.join()
-
-    def add_task(self, sync: bool, task: Callable, *args, **kwargs):
-        """
-        Add a task to the worker queue.
-
-        Args:
-            sync (bool): If `True`, block until the task is complete.
-            task (Callable): The task to run.
-
-        Additional arguments are passed to the task.
-        """
-        self._add_task(sync, task, *args, **kwargs)
 
     def _start_worker(self):
         """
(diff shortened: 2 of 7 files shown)


glean-core/python/tests/test_dispatcher.py
```diff
@@ -18,13 +18,13 @@ def update():
     for i in range(3):
```
Dexterp37: Since we're adding threading to the mix, we should probably port over some of the related tests from [DispatchersTest.kt](https://searchfox.org/glean/source/glean-core/android/src/test/java/mozilla/telemetry/glean/DispatchersTest.kt). I'm specifically thinking off:

- verify that tasks are dispatched on a separate thread when we're not in test mode;
- verify that, when running off the main thread, tasks are executed in the order they are received;
- exceptions thrown in the dispatched tasks do not break glean.

glean-core/python/glean/_dispatcher.py
```diff
@@ -18,22 +135,36 @@ class Dispatcher:
...
         Reset the dispatcher so the queue is cleared, and it is reset into
         queueing mode.
         """
         cls._queue_initial_tasks = True
-        cls._task_queue = []
+        cls._preinit_task_queue = []
         cls._overflow_count = 0
 
+    @classmethod
+    def _execute_task(cls, func: Callable, *args, **kwargs):
```
Dexterp37: Since we're now supporting priorities, we should also make sure to add test coverage for [this case](https://searchfox.org/glean/rev/55bba00886c39f2291906467ea519c6457f3a552/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt#447).

It would also be relevant to test (I swear there's a test for that in Kotlin, but can't find it!) that 'application' lifetime metrics are cleared *after* the Glean-owned pings are generated.

glean-core/python/glean/_dispatcher.py
```diff
@@ -2,11 +2,113 @@
...
+        """
+        Implements the worker thread. Takes tasks off of the queue and runs
+        them.
+        """
+        while True:
+            task, args, kwargs = self._queue.get()
+            try:
+                task(*args, **kwargs)
+            except Exception as e:
+                log.error(str(e))
```
wlach:
```suggestion
                log.exception("Glean error")
```

See: https://www.loggly.com/blog/exceptional-logging-of-exceptions-in-python/

glean-core/python/glean/_dispatcher.py
```diff
@@ -2,11 +2,113 @@
...
+        Additional arguments are passed to the task.
+        """
+        self._add_task(sync, task, *args, **kwargs)
+
+    def _start_worker(self):
+        """
+        Starts the worker thread.
+
+        It is started in daemon mode so it will run as long as the parent
+        thread.
```
wlach: It feels like this implementation detail would best be put inside the top of this file as well as the documentation, it's something I would be concerned about as a user of this library.

glean-core/python/glean/_dispatcher.py
mdboom: TIL.  That's really cool.

docs/user/adding-glean-to-your-project.md
```diff
@@ -266,6 +266,10 @@ Please refer to the [custom pings documentation](pings/custom.md).
 > **Important**: as stated [before](adding-glean-to-your-project.md#before-using-glean), any new data collection requires documentation and data-review.
 > This is also required for any new metric automatically collected by the Glean SDK.
 
+### Parallelism
+
+All of Glean's target languages use a separate worker thread to do most of Glean's work, including any I/O. This thread is fully managed by Glean as an implementation detail. Therefore, users should be free to use the Glean API wherever it is most convenient, without worrying about the performance impact of updating metrics and sending pings.
```
Dexterp37: I wonder if that's the best place to put it. I have no better suggestion though, as it looks kind of off even in the "The General API" page.

mdboom: Yeah, I struggled with where to put it myself.  It's sort of a "general design principles" thing, but we don't have that, and I don't think there's much that's user-relevant besides this...

Dexterp37: Yes, I agree :-\ Let's leave it where it is for now, and revisit in the future, in case we add something else!

PR #784, opened 2020-03-26
Remove all finalizers from metric type implementations in Kotlin

badboy:
They cause us more trouble than they are worth.
This is safe.
The only thing we did do in finalizers is cleaning up the allocated
object on the native Rust side.
Without these we now potentially leak memory.
_But_ metric types were already only used for metrics, and those are
statics, so they live for the entire application runtime.
Once the application is done and exits the Glean library will also be
unloaded and thus all its data be deallocated.

The only place where we create metrics at runtime is in tests.
So we're now leaking those in tests, but that should be fine.

As a side note: Java 9 deprecates finalizers in the language, Android
itself doesn't support Java 9 (yet) and has its own implementation of it
anyway, but that's nonetheless a good hint that finalizers might not be
the best idea afterall.

### DIFF
--- a/docs/dev/core/new-metric-type/kotlin.md
+++ b/docs/dev/core/new-metric-type/kotlin.md
@@ -42,12 +42,6 @@ class CounterMetricType(
                 disabled = disabled.toByte())
     }
 
-    protected fun finalize() {
-        if (this.handle != 0L) {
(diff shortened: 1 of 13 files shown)


badboy:
Using my admin powers to merge, the new Python tests weren't part of the codebase when I submitted this, so they won't run on this PR (but will run on merge)

PR #785, opened 2020-03-26
Warn against the use of `user` lifetime in the docs [doc only]

Dexterp37:
This is attempting to de-emphasize the `user` lifetime in our docs, explicitly warning users. I'm flagging the three of you given our last discussion on the topic.

**Note**: this additionally de-duplicates the lifetime docs, making sure it only lives in one place.


docs/user/adding-new-metrics.md
```diff
@@ -67,19 +67,9 @@ If the metric is still needed after its expiration date, it should go back for [
 
 ### When should Glean automatically clear the measurement?
 
-The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available:
+The `lifetime` parameter of a metric defines when its value will be cleared. There are three lifetime options available, [`ping`, `application` and `user`](metrics-parameters.html).
```
mdboom: While generally I don't like duplication in the docs, on this one I'm kind of :/  The reader really needs to click on that link and come back to understand anything about what immediately follows.  I think I'd prefer to have the long discussion here and shorten the reference documentation to just the bare bones and link back here for the details.

chutten: I agree. Is there a server-side inlining step we could use to embed the same content in both places, perhaps?

badboy: There is:

```
{{#include ../path/to/file.md}}
```

Care needs to be taken for relative links in the included file though.

Dexterp37: Oh, nice. Let me try this, thanks @badboy

docs/user/lifetimes-parameters.md
```diff
@@ -0,0 +1,3 @@
...
+  - `application`: The metric is related to an application run, and is cleared after the application restarts and any Glean-owned ping, due at startup, is submitted. This should be used for things that are constant during the run of an application, such as the operating system version. In practice, these metrics are generally set during application startup. A common mistake---using the ping lifetime for these type of metrics---means that they will only be included in the first ping sent during a particular run of the application.
+  - `user`: **Reach out to the Glean team before using this.**. The metric is part of the user's profile and will live as long as the profile lives. This is often not the best choice unless the metric records a value that _really_ needs to be persisted for the full lifetime of the user profile, e.g. an identifier like the `clientId`, the day the product was first executed. It is rare to use this lifetime outside of some metrics that are built in to the Glean SDK.
```
mdboom:
```suggestion
  - `user`: **Reach out to the Glean team before using this.**. The metric is part of the user's profile and will live as long as the profile lives. This is often not the best choice unless the metric records a value that _really_ needs to be persisted for the full lifetime of the user profile, e.g. an identifier like the `client_id`, the day the product was first executed. It is rare to use this lifetime outside of some metrics that are built in to the Glean SDK.
```

Let's name is as it appears in the ping and elsewhere in the docs...

.dictionary
```diff
@@ -115,6 +115,7 @@ linux
 logcat
 macOS
 manylinux
+md
```
mdboom: Weird -- I can't see where this is actually used, but probably no harm...

badboy: In the `include` statements in file names. Looks like aspell isn't excluding those. I advised Dexter to just add it to the dictionary for now.

Dexterp37:
Using my admin privileges due to the Python tests.

PR #786, opened 2020-03-26
WIP: 1624695: Act on changes in upload enable state outside of application

mdboom:
This is trickier than it needs to be because the handling of actions
based on the change in upload state is handled both on the Rust and
language wrapper side.

The solution I arrived at is to no longer pass in an initial value
for upload enabled when initializing the Rust side, but to instead let
Glean determine the upload enable state from the database -- it is
disabled if either there is no client_id or client_id == c0ffee.

Then the language bindings immediately change the upload state based
on the what the user has passed in, allowing the usual state flipping
operations to run as usual.

Even though the FfiConfiguration object loses an uploadEnabled field as
part of this change, that is purely internal API and the public API doesn't
change at all here.

Before merging:

- [ ] Port new tests to Swift

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -153,7 +153,6 @@ open class GleanInternalAPI internal constructor () {
             val cfg = FfiConfiguration(
                 dataDir = gleanDataDir.path,
                 packageName = applicationContext.packageName,
-                uploadEnabled = uploadEnabled,
                 maxEvents = configuration.maxEvents,
                 delayPingLifetimeIO = false
             )
@@ -164,6 +163,8 @@ open class GleanInternalAPI internal constructor () {
             if (!initialized) {
                 return@executeTask
             }
+
+            setUploadEnabled(uploadEnabled)
 
             // If any pings were registered before initializing, do so now.
             // We're not clearing this queue in case Glean is reset by tests.
@@ -261,7 +262,7 @@ open class GleanInternalAPI internal constructor () {
             val originalEnabled = getUploadEnabled()
 
             @Suppress("EXPERIMENTAL_API_USAGE")
-            Dispatchers.API.launch {
+            Dispatchers.API.executeTask {
                 // glean_set_upload_enabled might delete all of the queued pings. We
                 // therefore need to obtain the lock from the PingUploader so that
                 // iterating over and deleting the pings doesn't happen at the same time.
@@ -593,7 +594,8 @@ open class GleanInternalAPI internal constructor () {
     internal fun resetGlean(
         context: Context,
         config: Configuration,
-        clearStores: Boolean
+        clearStores: Boolean,
+        uploadEnabled: Boolean = true
     ) {
         Glean.enableTestingMode()
 
@@ -606,7 +608,7 @@ open class GleanInternalAPI internal constructor () {
 
         // Init Glean.
         Glean.testDestroyGleanHandle()
-        Glean.initialize(context, true, config)
+        Glean.initialize(context, uploadEnabled, config)
     }
 
     /**
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
@@ -689,6 +689,11 @@ class GleanTest {
             uploadEnabled = false
         )
 
+        // Now trigger it to upload
+        triggerWorkManager(context, DeletionPingUploadWorker.PING_WORKER_TAG)
+
+        assertEquals(1, server.requestCount)
+
         resetGlean(
             context,
             Glean.configuration.copy(
@@ -702,6 +707,6 @@ class GleanTest {
         // Now trigger it to upload
         triggerWorkManager(context, DeletionPingUploadWorker.PING_WORKER_TAG)
 
-        assertEquals(0, server.requestCount)
+        assertEquals(1, server.requestCount)
     }
 }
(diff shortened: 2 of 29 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -646,4 +646,62 @@ class GleanTest {
...
+
+        val request = server.takeRequest(20L, TimeUnit.SECONDS)
+        val docType = request.path.split("/")[3]
+        assertEquals("deletion-request", docType)
+    }
+
+    @Test
+    fun `no sending of deletion ping if unchanged outside of run`() {
+        val server = getMockWebServer()
+        resetGlean(
```
Dexterp37: Ideally we'd also want to check that we're not mistakenly resetting the client id for extra safety. Unless that's done in some existing test already

glean-core/ios/Glean/Glean.swift
```diff
@@ -199,7 +200,7 @@ public class Glean {
         if isInitialized() {
             let originalEnabled = getUploadEnabled()
 
-            Dispatchers.shared.launchAPI {
+            Dispatchers.shared.launchConcurrent {
```
Dexterp37: Same concern here that I had from the Android version

glean-core/preview/src/configuration.rs
```diff
@@ -7,8 +7,6 @@
 /// Optional values will be filled in with default values.
 #[derive(Debug, Clone)]
 pub struct Configuration {
-    /// Whether upload should be enabled.
-    pub upload_enabled: bool,
```
Dexterp37: Are you sure this is actually required, in this case? This is basically an external package/API, we're changing internals in other places.

glean-core/preview/src/lib.rs
```diff
@@ -22,12 +22,12 @@
 //! let cfg = Configuration {
 //!     data_path: "/tmp/data".into(),
 //!     application_id: "org.mozilla.glean_core.example".into(),
-//!     upload_enabled: true,
 //!     max_events: None,
 //!     delay_ping_lifetime_io: false,
 //!     channel: None,
 //! };
 //! glean_preview::initialize(cfg, ClientInfoMetrics::unknown())?;
+//! glean_preview::set_upload_enabled(true);
```
Dexterp37: I think that `glen_preview::initialize`, in this case, should take care of calling the right thing internally.

glean-core/src/lib.rs
```diff
@@ -181,7 +179,7 @@ impl Glean {
         let event_data_store = EventDatabase::new(&cfg.data_path)?;
 
         let mut glean = Self {
-            upload_enabled: cfg.upload_enabled,
+            upload_enabled: false,
```
Dexterp37: nit: let's add a comment to not scare our future selves out. This _should not_ trigger a wipeout of the stored client id.

glean-core/src/lib.rs
```diff
@@ -195,21 +193,35 @@ impl Glean {
...
+        // initialization. This allows for the detection of changes to this
+        // preference outside of the run of the applicaton, such as in a
+        // configuration file and allows changes such as clearing the database
+        // and sending a deletion ping to happen. See bug 1624695.
+        glean.upload_enabled = match glean
+            .core_metrics
+            .client_id
+            .get_value(&glean, "glean_client_info")
+        {
+            None => true,
```
Dexterp37: What if we keep all the existing code and change this part slightly: let's only find if we have an existing non coffee client id. If so, forcefully trigger the deletion ping here, as done in [there](https://github.com/mozilla/glean/blob/2b5ea921f814c4f145793bbed061b58cc573685e/glean-core/src/lib.rs#L302-L306). This would require us to deteramine is_first_run a bit earlier.

Moreover, wouldn't the current approach generate a new deletion ping the first time a new profile starts the product with telemetry disabled?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -261,7 +262,7 @@ open class GleanInternalAPI internal constructor () {
             val originalEnabled = getUploadEnabled()
 
             @Suppress("EXPERIMENTAL_API_USAGE")
-            Dispatchers.API.launch {
+            Dispatchers.API.executeTask {
```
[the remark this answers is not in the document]
mdboom: This needs to happen before any of the preinit tasks run.  `executeTask` merely adds to the "real" queue rather than the preinit queue.  After preinit time, it's basically identical to `launch`.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
mdboom: It might be, but it's not a bad idea to do it here as well.

glean-core/preview/src/lib.rs
mdboom: Good point.

glean-core/src/lib.rs
mdboom: > What if we keep all the existing code and change this part slightly: let's only find if we have an existing non coffee client id. If so, forcefully trigger the deletion ping here, as done in there. This would require us to deteramine is_first_run a bit earlier.

I tried that initially, but the problem is that we need to *both* submit the deletion ping (which happens in Rust) *and* trigger the deletion ping uploader (from the language bindings).

I have some thoughts about refactoring this further to make the line between the Rust and language binding side less weird here.

> Moreover, wouldn't the current approach generate a new deletion ping the first time a new profile starts the product with telemetry disabled?

Yes, that's true.  Should work on that case as well, I guess.

mdboom:
Found a much simpler approach in #791

PR #790, opened 2020-03-30
WIP: Handle upload enable change at startup using callbacks

mdboom:
(Kotlin-only implementation for now).

This is an alternative to #786.  It avoids the weird flow control by having the Rust core be completely in charge of what to do when the upload state changes (and fixes a bug where a deletion ping is sent on first run, since the startup logic for first run can be made slightly different).

However, it is fundamentally broken since it deadlocks.  When the callback is called, it does things involving the global Glean singleton, and since the lock on that is already held, it stalls waiting for the lock.  There might be ways of rewriting the FFI wrapper to drop the glean lock and *then* call the callback, but since the callback is a member of the Glean singleton, I couldn't find a way to do it.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 160 utf-8
+personal_ws-1.1 en 161 utf-8
 AAR
 AARs
 APIs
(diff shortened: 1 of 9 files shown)


mdboom:
Found a much simpler approach in #791.

PR #791, opened 2020-03-30
1624695: Act on changes in upload enable state outside of application

mdboom:
I found a much simpler way forward here that really only adds some slightly different logic in `Glean::new` and a bit of a refactor.  I'm much happier with this than the earlier two approaches.

TODO:
- [x] Port new unit tests to Swift (I don't currently have access to a Mac -- volunteers?)

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -195,7 +195,38 @@ impl Glean {
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
         };
-        glean.on_change_upload_enabled(cfg.upload_enabled);
+
+        // The upload enabled flag may have changed since the last run, for
+        // example by the changing of a config file.
+        if cfg.upload_enabled {
+            // If upload is enabled, just follow the normal code path to
+            // instantiate the core metrics.
+            glean.on_upload_enabled();
+        } else {
+            // If upload is disabled, and we've never run before, only set the
+            // client_id to KNOWN_CLIENT_ID, but do not send a deletion request
+            // ping.
+            // If we have run before, and if the client_id is not equal to
+            // the KNOWN_CLIENT_ID, do the full upload disabled operations to
+            // clear metrics, set the client_id to KNOWN_CLIENT_ID, and send a
+            // deletion request ping.
+            match glean
+                .core_metrics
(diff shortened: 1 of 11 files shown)


glean-core/src/lib.rs
```diff
@@ -308,22 +334,27 @@ impl Glean {
...
     }
 
-    /// Handles the changing of state when upload_enabled changes.
+    /// Handles the changing of state from upload disabled to enabled.
     ///
     /// Should only be called when the state actually changes.
-    /// When disabling, all pending metrics, events and queued pings are cleared.
     ///
-    /// When enabling, the core Glean metrics are recreated.
+    /// The core Glean metrics are recreated.
```
badboy: the docs should call out that the flag is not yet set to `true` when this message should be called.

glean-core/src/lib.rs
```diff
@@ -308,22 +334,27 @@ impl Glean {
...
-    /// # Arguments
+    /// Should only be called when the state actually changes.
     ///
-    /// * `flag` - When true, enable metric collection.
-    fn on_change_upload_enabled(&mut self, flag: bool) {
-        if flag {
-            self.initialize_core_metrics();
-        } else {
-            self.clear_metrics();
+    /// All pending metrics, events and queued pings are cleared.
```
badboy: The docs should call out that it requires the flag to still be true for a deletion request ping to be sent (IMO that could also be a comment on the `if let Err` line below, given that we might not want to document the exact details of the function in its  documentation)

glean-core/src/lib.rs
mdboom: But that's not strictly true -- it might be set to true already if this is called during initialization.

badboy: Ah right.
But it is strictly true that afterwards it's definitely set to true.

mdboom: Got it.  Adding this to the comment (and `on_upload_disabled`, too)

mdboom:
This includes a bug fix for the metrics ping scheduler sending on the last day of the month.  Does anyone feel strongly about separating that out?

badboy:
> This includes a bug fix for the metrics ping scheduler sending on the last day of the month. Does anyone feel strongly about separating that out?

As long as it is its own commit, I'm fine.

PR #793, opened 2020-03-31
Expire `user` lifetime metrics

Dexterp37:
This adds a mechanism to check any 'user' lifetime metric against an internal list of metric ids and
expiration dates. The expiration date for each metric is checked against the date the application was started, not the build date. This is a major different compared to our build-time expiration mechanism for other lifetimes. We may live with this, since we don't expect 'user' lifetimes to have expiration in the long run.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -4,10 +4,13 @@
 
 use std::collections::btree_map::Entry;
 use std::collections::BTreeMap;
+use std::collections::HashMap;
 use std::fs;
 use std::str;
 use std::sync::RwLock;
 
+use chrono::{Local, NaiveDate};
+use lazy_static::lazy_static;
 use rkv::{Rkv, SingleStore, StoreOptions};
 
 use crate::metrics::Metric;
@@ -15,6 +18,19 @@ use crate::CommonMetricData;
 use crate::Glean;
 use crate::Lifetime;
 use crate::Result;
+
+lazy_static! {
+    // A map of all the known 'user' lifetime metrics along with their expiration
+    // dates. See bug 1604854 for more context on why this is needed.
+    static ref USER_LIFETIME_EXPIRATION_MAP: HashMap<&'static str, NaiveDate> = vec![
+        ("glean_internal_test.user_metric_expired", NaiveDate::from_ymd(2015, 3, 14)),
+    ].into_iter().collect();
+
+    // Get the current date at runtime. While we should use the build-date to be
+    // consistent with other metric types, given the edge case this already is,
+    // it might be enough to go with a run-time date.
+    static ref STARTUP_LOCAL_DATE: NaiveDate = Local::now().naive_utc().date();
+}
 
 pub struct Database {
     /// Handle to the database environment.
@@ -73,6 +89,9 @@ impl Database {
             application_store,
             ping_lifetime_data,
         };
+
+        // Force the startup date to be initialized.
+        log::debug!("Startup UTC date is: {:?}", *STARTUP_LOCAL_DATE);
 
         db.load_ping_lifetime_data();
 
@@ -147,6 +166,60 @@ impl Database {
         }
     }
 
+    /// Checks if the 'user' lifetime metric is expired and, if so, removes it.
+    ///
+    /// ## Arguments
+    ///
+    /// * `storage_name`: The storage name to iterate over.
+    /// * `metric_id_slice`: The metric id to check.
+    ///
+    /// ## Return value
+    ///
+    /// Returns `true` if the metric is expired and should not be reported,
+    /// `false` otherwise.
+    ///
+    /// ## Panics
+    ///
+    /// This function will **not** panic on database errors.
+    fn check_and_remove_expired_user_lifetime(
+        &self,
+        storage_name: &str,
+        metric_id_slice: &[u8],
+    ) -> bool {
+        let metric_name = match str::from_utf8(metric_id_slice) {
+            Ok(metric_name) => metric_name.to_string(),
+            _ => return false,
+        };
+
+        match USER_LIFETIME_EXPIRATION_MAP.get(metric_name.as_str()) {
+            Some(&expiration) => {
+                let is_expired = *STARTUP_LOCAL_DATE > expiration;
+                if is_expired {
+                    // This should be fine to delete now, even though we may be iterating
+                    // through the database entries.
+                    if let Err(e) =
+                        self.remove_single_metric(Lifetime::User, storage_name, &metric_name)
+                    {
+                        log::error!(
+                            "Failed to remove  expired {} from {}: {:?}",
+                            metric_name,
+                            storage_name,
+                            e
+                        );
+                    } else {
+                        log::debug!("Removed expired {} from {}", metric_name, storage_name);
+                    }
+                }
+                // While we might not be able to remove it from the database, we can still
+                // make sure to not report it.
+                true
+            }
+            // If we can't find the expiration in our table, that's fine. We might
+            // not know about it, or it's really not meant to expire.
+            _ => false,
+        }
+    }
+
     /// Iterates with the provided transaction function over the requested data
     /// from the given storage.
     ///
@@ -210,6 +283,16 @@ impl Database {
             }
 
             let metric_name = &metric_name[len..];
+
+            // Don't report the 'user' lifetiem metric if it expired and additionally
+            // mark it for removal. We need to do it now as we don't know the storage
+            // names beforehand.
+            if lifetime == Lifetime::User
+                && self.check_and_remove_expired_user_lifetime(storage_name, metric_name)
+            {
+                continue;
+            }
+
             let metric: Metric = match value.expect("Value missing in iteration") {
                 rkv::Value::Blob(blob) => unwrap_or!(bincode::deserialize(blob), continue),
                 _ => continue,
@@ -1026,5 +1109,68 @@ mod test {
                 .unwrap_or(None)
                 .is_some());
         }
+    }
+
+    #[test]
+    fn test_purge_expired_user_lifetime_metrics() {
+        // Init the database in a temporary directory.
+        let dir = tempdir().unwrap();
+        let str_dir = dir.path().display().to_string();
+        let db = Database::new(&str_dir, false).unwrap();
+
+        let test_storage = "test-storage-expired-user";
+        let metric_id_pattern = "glean_internal_test.user_metric";
+
+        // Write sample metrics to the database. We write all lifetimes to
+        // make sure we don't mess with anything else that's already stored.
+        let lifetimes = vec![Lifetime::User, Lifetime::Ping, Lifetime::Application];
+
+        for lifetime in lifetimes.iter() {
+            db.record_per_lifetime(
+                *lifetime,
+                test_storage,
+                &format!("{}_retain", metric_id_pattern),
+                &Metric::String("retain".to_string()),
+            )
+            .unwrap();
+        }
+
+        // Save a 'user'-lifetime metric that will be expired.
+        db.record_per_lifetime(
+            Lifetime::User,
+            test_storage,
+            &format!("{}_expired", metric_id_pattern),
+            &Metric::String("expired".to_string()),
+        )
+        .unwrap();
+
+        // Verify that "telemetry_test.single_metric_retain" is still around for all lifetimes.
+        let mut total_found_metrics = 0;
+        for lifetime in lifetimes.iter() {
+            let mut found_metrics = 0;
+            let mut snapshotter = |metric_name: &[u8], metric: &Metric| {
+                found_metrics += 1;
+                let metric_id = String::from_utf8_lossy(metric_name).into_owned();
+                assert_eq!(format!("{}_retain", metric_id_pattern), metric_id);
+                match metric {
+                    Metric::String(s) => assert_eq!("retain", s),
+                    _ => panic!("Unexpected data found"),
+                }
+            };
+
+            // Check the count of the metrics that were found for this lifetime.
+            db.iter_store_from(*lifetime, test_storage, None, &mut snapshotter);
+            assert_eq!(
+                1, found_metrics,
+                "We only expect 1 metric for this lifetime."
+            );
+            // Increment the total count as well.
+            total_found_metrics += found_metrics;
+        }
+
+        assert_eq!(
+            3, total_found_metrics,
+            "We only expect 3 metrics to be reported across all lifetimes."
+        );
     }
 }


Dexterp37:
This is blocked on #792.

glean-core/src/database/mod.rs
```diff
@@ -16,6 +19,21 @@ use crate::Glean;
...
+    static ref USER_LIFETIME_EXPIRATION_MAP: HashMap<&'static str, NaiveDate> = vec![
+        ("glean_internal_test.user_metric_expired", NaiveDate::from_ymd(2015, 3, 14)),
+    ].into_iter().collect();
+}
+
+lazy_static!{
+    // Get the current date at runtime. While we should use the build-date to be
+    // consistent with other metric types, given the edge case this already is,
+    // it might be enough to go with a run-time date.
+    static ref STARTUP_LOCAL_DATE: NaiveDate = Local::now().naive_local().date();
```
Dexterp37: @badboy @mdboom what's your take on this simplification? If we were to go for the "build time" date, we'd need to change the FFI to pass in a date from e.g. the Android build system. Given the fact that this is an edge case already, we might as well simplify our lives.

glean-core/src/database/mod.rs
```diff
@@ -16,6 +19,21 @@ use crate::Glean;
 use crate::Lifetime;
 use crate::Result;
 
+lazy_static!{
+    // A map of all the known 'user' lifetime metrics along with their expiration
+    // dates. See bug 1604854 for more context on why this is needed.
+    static ref USER_LIFETIME_EXPIRATION_MAP: HashMap<&'static str, NaiveDate> = vec![
+        ("glean_internal_test.user_metric_expired", NaiveDate::from_ymd(2015, 3, 14)),
```
Dexterp37: After the design questions are settled, I'll add the metrics as a separate PR.

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,49 @@ impl Database {
...
+        };
+
+        match USER_LIFETIME_EXPIRATION_MAP.get(metric_name_str.as_str()) {
+            Some(&expiration) => 
+            {
+                let is_expired = *STARTUP_LOCAL_DATE > expiration;
+                if is_expired {
+                    // This should be fine to delete now, even though we may be iterating
+                    // through the database entries.
+                    if let Err(e) = self.remove_single_metric(Lifetime::User, storage_name, &metric_name_str) {
```
Dexterp37: @badboy this seems to be safe, no test is complaining. I was a bit surprised it worked, but then I realized that LMDB can probably handle reading vs writing while iterating. A different strategy would be to move this code to `iter_store_from`, store the expired ids in a vector and then remove them outside of the `while`.

glean-core/src/database/mod.rs
mdboom: I think this is fine, but maybe either (1) use UTC here (which is what metric expiration at build time is based on), or (2) add 36 hours so we would never remove metrics before the expected build time.  (2) may actually be better to account for clock skew.

glean-core/src/database/mod.rs
```diff
@@ -210,6 +281,16 @@ impl Database {
             }
 
             let metric_name = &metric_name[len..];
+
+            // Don't report the 'user' lifetiem metric if it expired and additionally
+            // mark it for removal. We need to do it know as we don't know the storage
```
mdboom:
```suggestion
            // mark it for removal. We need to do it now as we don't know the storage
```

glean-core/src/database/mod.rs
```diff
@@ -2,8 +2,11 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use chrono::{Local, NaiveDate};
+use lazy_static::lazy_static;
```
badboy: I personally prefer to split out crate uses from std uses, so I would suggest moving that down to line 14 (old line 11)

glean-core/src/database/mod.rs
```diff
@@ -16,6 +19,21 @@ use crate::Glean;
...
 
+lazy_static! {
+    // A map of all the known 'user' lifetime metrics along with their expiration
+    // dates. See bug 1604854 for more context on why this is needed.
+    static ref USER_LIFETIME_EXPIRATION_MAP: HashMap<&'static str, NaiveDate> = vec![
+        ("glean_internal_test.user_metric_expired", NaiveDate::from_ymd(2015, 3, 14)),
+    ].into_iter().collect();
+}
+
+lazy_static! {
```
badboy: `lazy_static!` allows multiple statics to be in a single block.

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,59 @@ impl Database {
...
+    /// Returns a `true` if the metric is expired and should not be reported,
+    /// `false` otherwise.
+    ///
+    /// ## Panics
+    ///
+    /// This function will **not** panic on database errors.
+    fn check_and_remove_expired_user_lifetime(
+        &self,
+        storage_name: &str,
+        metric_id_slice: &[u8],
```
badboy: Why the `slice` in the name?

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,59 @@ impl Database {
...
 
+    /// Checks if the 'user' lifetime metric is expired and, if so, removes it.
+    ///
+    /// ## Arguments
+    ///
+    /// * `storage_name`: The storage name to iterate over.
+    /// * `metric_id_slice`: The metric id to check.
+    /// ## Return value
+    ///
+    /// Returns a `true` if the metric is expired and should not be reported,
```
badboy:
```suggestion
    /// Returns `true` if the metric is expired and should not be reported,
```

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,59 @@ impl Database {
         }
     }
 
+    /// Checks if the 'user' lifetime metric is expired and, if so, removes it.
+    ///
+    /// ## Arguments
+    ///
+    /// * `storage_name`: The storage name to iterate over.
+    /// * `metric_id_slice`: The metric id to check.
+    /// ## Return value
```
badboy: Newline before this.
```suggestion

    /// ## Return value
```

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,59 @@ impl Database {
...
+                        );
+                    } else {
+                        log::debug!("Removed expired {} from {}", metric_name_str, storage_name);
+                    }
+                }
+                // While we might not be able to remove it from the database, we can still
+                // make sure to not report it.
+                return true;
+            }
+            // We if we can't find the expiration in our table, that's fine. We might
```
badboy:
```suggestion
            // If we can't find the expiration in our table, that's fine. We might
```

glean-core/src/database/mod.rs
badboy: That is ... certainly a change in what we communicated previously. I'm fine with that, but we should make that clear.
Multiple questions I have:

* This date is only set on first access. Should we make sure to init it in initialize?
* Do we already have a list of expired/expiring user lifetime metrics?
	* That would allow us to evaluate the following: Can we just build Glean at a point where all those are expired and therefore hard-code their deletion?
	* Are we going to disallow expiration of user-lifetime metrics via glean_parser?

mdboom: > That is ... certainly a change in what we communicated previously.

You mean the UTC part?  I think that's been the definition from the beginning (so that builds built in multiple locations wouldn't be different)  https://mozilla.github.io/glean_parser/metrics-yaml.html#expires

badboy: No the "datetime at runtime" vs "build time"

Dexterp37: > 
> 
> That is ... certainly a change in what we communicated previously. I'm fine with that, but we should make that clear.
> Multiple questions I have:
> 
>     * This date is only set on first access. Should we make sure to init it in initialize?

Good point.

>     * Do we already have a list of expired/expiring user lifetime metrics?

I'm building this list, it will come as a separate commit.

>       * That would allow us to evaluate the following: Can we just build Glean at a point where all those are expired and therefore hard-code their deletion?

To make sure I understand, @badboy , do you mean postponing this change to a "future date" in which we'll just release the code to clear them all up?

>       * Are we going to disallow expiration of user-lifetime metrics via glean_parser?

I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1626516) for that.

glean-core/src/database/mod.rs
Dexterp37: @badboy any comment on this?

glean-core/src/database/mod.rs
```diff
@@ -147,6 +165,59 @@ impl Database {
...
+    ///
+    /// ## Panics
+    ///
+    /// This function will **not** panic on database errors.
+    fn check_and_remove_expired_user_lifetime(
+        &self,
+        storage_name: &str,
+        metric_id_slice: &[u8],
+    ) -> bool {
+        let metric_name_str = match str::from_utf8(metric_id_slice) {
```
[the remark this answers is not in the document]
Dexterp37: Yes, I noticed that throughout the code we're not consistent: we use `metric_name` when we really mean `metric_Id` (category + name). I'll file a bug!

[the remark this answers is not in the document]
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1626542

glean-core/src/database/mod.rs
Dexterp37: @badboy is there a bug for this, already?

glean-core/src/database/mod.rs
Dexterp37: @mdboom what will the 36 hours offset give us?

badboy: > do you mean postponing this change to a "future date" in which we'll just release the code to clear them all up?

Yeah. If we know we have expired ones, why not clean only those up when we know that's the case. We don't allow expiry in the future so we're good.

glean-core/src/database/mod.rs
Dexterp37: Mh, that's a good question. I feel like a coward, but that's how I got the Rust compiler to stop screaming at me. Otherwise, it was complaining about `&str` not being allowed when attempting to look up items in the map.

glean-core/src/database/mod.rs
Dexterp37: > 
> 
> > do you mean postponing this change to a "future date" in which we'll just release the code to clear them all up?
> 
> Yeah. If we know we have expired ones, why not clean only those up when we know that's the case. We don't allow expiry in the future so we're good.

That's another strategy, I'd be fine with that. It puts the burden of remembering when things expire on us, though.

glean-core/src/database/mod.rs
badboy: No, but I can file one.

glean-core/src/database/mod.rs
badboy: It also requires Fenix (or similar) to upgrade Glean in a timely manner after the expiration.
Still I think it's a worthwhile way to discuss.

glean-core/src/database/mod.rs
badboy: filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1626558

glean-core/src/database/mod.rs
badboy: hm ... I need to re-read that code and also rkv write vs read mode, just to make sure.
I'll get back to you.

glean-core/src/database/mod.rs
mdboom: > @mdboom what will the 36 hours offset give us?

Given clock skew and clocks set in the entirely wrong timezones, an assurance that user metrics are never deleted early.  But on second thought, this is just too much of a fudge.  Using UTC rather than local seems like low-hanging fruit, but let's not add any additional weirdness on top.

Dexterp37: @mdboom , what's your take on what @badboy suggested a couple of comments above?

mdboom: If we could take all date parsing at runtime out and just assume metrics don't go in the remove list until they are already expired, it does simplify things.  We will have to remember to do it, but maybe we set a calendar reminder (for ourselves) to update that list when we know they are going to expire?

Dexterp37: Apologies for keeping the conversation going, but I still have a few doubts about the next step. As far as I understood, the discussion above distills down to a few options.

1. Merge this PR as it is, add the expired metrics later when they are about to expire. We'd still need to coordinate and cut a release before metrics expire.
2. Drop this PR, resume it *after* the metrics we're looking to expire are actually expired, then cut a new release of the SDK that will purge that data in a way similar to the one in this PR.
3. Merge this PR together with a commit with the expiration dates.
4. Do nothing, postpone decision to when metrics are expired.

@badboy @mdboom I don't think (1) gives us any advantage over (4). I'd be happy with either (2) or (3) happening. What's your take?

badboy: So ... I did take a brief look into Fenix and GeckoView for expiring user-lifetime metrics:

* `activation.activation_id` expires 2020-09-01
* a bunch of metrics in m-c: expire 2021-02-01

Am I missing other metrics we know about?
That at least gives us ample time to tackle this, obviously I'd like this to move sooner.

The code approach here looks good to me.

I think the only thing I'd like to see a decision on is whether we base this on a runtime date as the PR currently does or on the build-time date of Glean (with a reminder for us to push releases on those two dates above).
(I'm explicitly ignoring a "pass in build time of app" approach, that's more hassle than it's worth).

Does that make sense?

Dexterp37: > 
> 
> So ... I did take a brief look into Fenix and GeckoView for expiring user-lifetime metrics:
> 
>     * `activation.activation_id` expires 2020-09-01
> 
>     * a bunch of metrics in m-c: expire 2021-02-01
> 
> 
> Am I missing other metrics we know about?

Not that I'm aware of, no.

> I think the only thing I'd like to see a decision on is whether we base this on a runtime date as the PR currently does or on the build-time date of Glean (with a reminder for us to push releases on those two dates above).

I have a slight preference on "runtime date". While I know this makes it less deterministic, as it will heavily depend on the time the application runs, it gives us a guaranteed lower bound date (the expiration date) before which nothing gets removed. After that date we should not rely on the data anyway, so users of the data will be aware.

Using the SDK build date will provide us with a strict deterministic date for expiration, but given the time it will take for the SDK to propagate to A-C and then Fenix, I think the result will be pretty much the same.

glean-core/src/database/mod.rs
```diff
@@ -147,6 +166,60 @@ impl Database {
...
+        };
+
+        match USER_LIFETIME_EXPIRATION_MAP.get(metric_name.as_str()) {
+            Some(&expiration) => {
+                let is_expired = *STARTUP_LOCAL_DATE > expiration;
+                if is_expired {
+                    // This should be fine to delete now, even though we may be iterating
+                    // through the database entries.
+                    if let Err(e) =
+                        self.remove_single_metric(Lifetime::User, storage_name, &metric_name)
```
badboy, line 201: I think I know now why this is working fine: transactions!

Both the reading and iterating of the outside function as well as the deletion are running in their own transactions.
While the `remove_single_metric` immediately commits and thus removes data, the reader is in a separate transaction and thus will not see that changed data (= the fact that data is deleted).
That's why this is all safe and working

(rkv allows multiple open reader transactions, but only a single writer transaction)

mdboom:
I think we should defer this until the metrics expire and then not use date logic to expire them at runtime when we are ready.

Dexterp37:
> 
> 
> I think we should defer this until the metrics expire and then not use date logic to expire them at runtime when we are ready.

👍 Closing this PR, I tracked this discussion [in the related bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1604854#c7).

PR #794, opened 2020-03-31
Bug 1625916 - Detect the operating system at compile time and set the metric

badboy:


glean-core/src/system.rs
```diff
@@ -0,0 +1,81 @@
...
+/// `target_os` when building this crate: `ios`
+pub const OS: &str = "iOS";
+
+#[cfg(target_os = "linux")]
+/// `target_os` when building this crate: `linux`
+pub const OS: &str = "Linux";
+
+#[cfg(target_os = "macos")]
+/// `target_os` when building this crate: `macos`
+pub const OS: &str = "Darwin";
```
badboy: cc @chutten: I guess `Darwin` is the right thing. that's what we currently have in Telemetry as well.

Dexterp37: @badboy I think we should talk to @fbertsch about this, too. The `telemetry.main` dataset, for example, has the `normalized_os` field that reports `"Mac"`. Maybe we can do the proper thing here by default?

glean-core/src/lib.rs
```diff
@@ -611,6 +614,11 @@ impl Glean {
         Ok(())
     }
 
+    /// Set internally-handled application lifetime metrics.
+    fn set_application_lifetime_metrics(&self) {
```
```suggestion
    fn set_application_lifetime_core_metrics(&self) {
```
To make it clear that this is about setting core stuff.

glean-core/src/system.rs
chutten: Yeah, if we were interested in keeping what Telemetry had, then windows should be Windows_NT : |

Let's not slow down too much with consultation, though, because we're not going to be sending any data with this info for a while. If we don't get an answer, say, today then we can check in a "we'd be okay with this if it shipped" solution and file a follow-up to consult on the exact value DE and DS might want.

badboy: I'll informally ask Frank for a quick answer.

Dexterp37: I'm fine with not blocking on this, now, but please do file a follow-up bug for understanding what's the best value we can offer here.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1626301

PR #795, opened 2020-03-31
Bug 1625207 - device manufacturer & model are now optional

badboy:
Needs upstream schema to be deployed.

I also intend to update the schema in this repository once that is done.

### DIFF
--- a/docs/user/pings/index.md
+++ b/docs/user/pings/index.md
@@ -46,8 +46,8 @@ Optional fields are marked accordingly.
 | `app_display_version` | String | The user-visible version string (e.g. "1.0.3") |
 | `architecture` | String | The architecture of the device (e.g. "arm", "x86") |
 | `client_id` | UUID |  *Optional* A UUID identifying a profile and allowing user-oriented correlation of data |
-| `device_manufacturer` | String | The manufacturer of the device |
-| `device_model` | String | The model name of the device. On Android, this is [`Build.MODEL`], the user-visible name of the device. |
+| `device_manufacturer` | String | *Optional* The manufacturer of the device |
+| `device_model` | String | *Optional* The model name of the device. On Android, this is [`Build.MODEL`], the user-visible name of the device. |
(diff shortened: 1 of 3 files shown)


docs/user/pings/index.md
```diff
@@ -46,8 +46,8 @@ Optional fields are marked accordingly.
 | `app_display_version` | String | The user-visible version string (e.g. "1.0.3") |
 | `architecture` | String | The architecture of the device (e.g. "arm", "x86") |
 | `client_id` | UUID |  *Optional* A UUID identifying a profile and allowing user-oriented correlation of data |
-| `device_manufacturer` | String | The manufacturer of the device |
-| `device_model` | String | The model name of the device. On Android, this is [`Build.MODEL`], the user-visible name of the device. |
+| `device_manufacturer` | String | *Optional* The manufacturer of the device |
```
Dexterp37: Would you kindly also mention the new behaviour [here](https://github.com/mozilla/glean/blob/4311592bf451259d0c139fa64cf0eec3945de091/glean-core/metrics.yaml#L77-L107)?

glean-core/metrics.yaml
```diff
@@ -81,6 +81,7 @@ glean.internal.metrics:
       - glean_client_info
     description: |
       The manufacturer of the device the application is running on.
+      Not set if the device manufacturer can't be determined (e.g. on Desktop).
```
mdboom: pebble: It's not that it *can't*, it's that it's not terribly useful or accurate.

badboy: He, I left that vague enough in case we ever reconsider it :D

But I'm happy to change it to some more appropriate.

badboy:
> Should we also remove this from the Python bindings -- either here or in a follow-on bug? If the determination is that this isn't very useful on Desktop and therefore we don't collect it (I agree), we probably shouldn't collect it on _any_ Desktop platform.

I file a bug for that.

badboy:
Done: https://bugzilla.mozilla.org/show_bug.cgi?id=1626303

PR #797, opened 2020-03-31
bug 1604177: Add reason code to the events ping

mdboom:
TODO:

- [x] Swift unit tests

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -327,8 +327,8 @@ public class Glean {
 
     /// Handle background event and submit appropriate pings
     func handleBackgroundEvent() {
-        Pings.shared.baseline.submit(Pings.shared.baselineReasonCodes.background)
-        Pings.shared.events.submit(Pings.shared.eventsReasonCodes.background)
+        Pings.shared.baseline.submit(reason: .background)
+        Pings.shared.events.submit(reason: .background)
     }
 
     /// Collect and submit a ping by name for eventual uploading
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -327,8 +327,8 @@ public class Glean {
 
     /// Handle background event and submit appropriate pings
     func handleBackgroundEvent() {
-        Pings.shared.baseline.submit()
-        Pings.shared.events.submit()
+        Pings.shared.baseline.submit(Pings.shared.baselineReasonCodes.background)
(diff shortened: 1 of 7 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -328,8 +328,8 @@ public class Glean {
 
     /// Handle background event and submit appropriate pings
     func handleBackgroundEvent() {
-        Pings.shared.baseline.submit()
-        Pings.shared.events.submit()
+        Pings.shared.baseline.submit(Pings.shared.baselineReasonCodes.background)
```
travis79: Due to the way that the `submit` function is defined in the Swift bindings, you cannot omit the `reason:` label. You can also make this more Swifty by just using `.background` and letting Swift infer the rest.  Like this:
```Swift
Pings.shared.baseline.submit(reason: .background)
Pings.shared.events.submit(reason: .background)
```

travis79: Also, thanks for catching the missing baseline reason code :)

mdboom: > Also, thanks for catching the missing baseline reason code :)

Well, in fairness, there's only one "reason" for the baseline ping in Swift since it's more predictable, but I think it's nice to be consistent here.

travis79: I think there are still ways to get the `.dirty` reason for the baseline ping, if the app crashed, for instance.

mdboom: Oh, indeed.  I guess I didn't realize that...  Good to know.

glean-core/pings.yaml
```diff
@@ -83,6 +83,15 @@ events:
     - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
   notification_emails:
     - glean-team@mozilla.com
+  reasons:
+    startup: |
+      The ping was submitted at startup. The events ping is always sent at
```
badboy, line 88: This should mention that it is submitted on startup if it contains pending events.

mdboom:
@travis79 , @badboy : Any takers to port the new unit tests to Swift?

travis79:
> @travis79 , @badboy : Any takers to port the new unit tests to Swift?

I'll do it

travis79:
> > @travis79 , @badboy : Any takers to port the new unit tests to Swift?
> 
> I'll do it

All done, commit pushed to your PR branch.

PR #798, opened 2020-03-31
Bug 1626086: Wait for tasks to complete before shutting down

mdboom:


glean-core/python/glean/_dispatcher.py
```diff
@@ -73,12 +76,13 @@ def _start_worker(self):
...
+        self._thread = threading.Thread(target=self._worker)
         # Start the thread in daemon mode.
-        t.daemon = True
-        t.start()
+        self._thread.daemon = True
+        self._thread.start()
         self._started = True
-        self._ident = t.ident
+        self._ident = self._thread.ident
+        atexit.register(self._shutdown_thread)
```
wlach: Can you add a comment here describing what we're doing?

mdboom:
> Looks pretty good to me. Should we add a note about this behaviour to the SDK documentation?

Yeah.  On some level, this is like a bugfix on an implementation detail that's documented here: https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html#parallelism

But it's just tricky/weird enough that Python developers probably want to know this is happening.  Let me add a little bit in the Python-specific documentation.

docs/user/adding-glean-to-your-project.md
```diff
@@ -270,6 +270,19 @@ Please refer to the [custom pings documentation](pings/custom.md).
...
 
+{{#include ../tab_header.md}}
+
+<div data-lang="Python" class="tab">
+When using the Python bindings, Glean's work is done on a separate thread, managed by Glean itself.
+Glean releases the Global Interpreter Lock (GIL), therefore your application's threads should not be in contention with Glean's thread.
+
+Glean installs an [`atexit` handler](https://docs.python.org/3/library/atexit.html) so the Glean thread can attempt to cleanly shut down when your application exits.
+This handler will wait up to 1 second for any pending work to complete.
+If that times out, some Glean work may be lost.
```
Dexterp37, line 281: @mdboom , on other platforms I/O and networking happen on a thread that's separate from the "dispatcher" queue for API calls.

If we did something similar here, thus being consistent with Kotlin/Swift, we'd mitigate data loss: API/ping serialization are _very likely_ to finish < 1s, networking operations are more likely to take more time on weird network conditions.

mdboom: Yep -- I have this bug for the follow-up: https://bugzilla.mozilla.org/show_bug.cgi?id=1626403

In the case of Python, I think we'll use a separate process (rather than a thread) so that networking can continue even if the main process finishes, but it's the same concept.  My only hesitation on the follow-up is whether to do it before or after the Rust networking refactor lands.

wlach: Will the work actually be lost? Or would Glean be smart enough to resend it the next time the command / application is invoked?

mdboom: Oh, yeah, of course.  If networking is off, say, the ping file won't be deleted and a reattempt will be made next time.  So pretty low risk.

wlach: I'd suggest mentioning that, since the current wording is a little scary.

Dexterp37: In general, right now, I think there's a _minor_ chance of loosing a small fraction of data: since I/O and networking are happening on the same dispatcher, if any work is queued _after_ this lengthy operations **and** the timeout is hit, the work queued after is lost, I think.

If B is a lengthy task on which we hit the 1s timeout, then the following can happen:

Queued tasks: A-B (timeout) - C (lost)- D (lost) - E (lost)

mdboom:
I went ahead and made uploading happen in a separate process.  I think the pending Rust refactor on that won't impact this too much after all.

That way, I don't think we need the scary warning, since it's become a really unlikely corner case.

mdboom:
> As I mentioned [here](https://bugzilla.mozilla.org/show_bug.cgi?id=1626403#c1), I'm worried that multiprocessing might have issues with the setup I'm using for mozregression on Windows. See this very long thread on stackoverflow, for example: https://stackoverflow.com/questions/24944558/pyinstaller-built-windows-exe-fails-with-multiprocessing
> 
> Can you make it optional?

Gah.  That's horrifying, but not surprising when you think it all through.  It's even failing our Windows tests here (which aren't doing any PyInstaller magic).

Assuming I can figure out what's going on with the Windows unit tests, making it optional through a `allow_multiprocessing` config flag or something could work.

If not, it's easy enough for me to take this last commit out and merge this with the multithreading, without the multiprocessing.  That's at least an 80% solution.

Once the refactor to move a lot of the upload processing to Rust is complete, it will be pretty easy to write a pure-Rust ping uploader, which might avoid some of these issues (assuming we can run a binary executable inside of a PyInstaller archive).

mdboom:
Fixed the basic Windows failures, and added a configuration option to make the multiprocessing optional.

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -31,81 +32,143 @@ def storage_directory(cls) -> Path:
...
 
-        if processed:
+    @classmethod
+    def _do_process_sync(cls) -> bool:
+        from .. import Glean
+
+        return _process(cls.storage_directory(), Glean._configuration)
+
+    @classmethod
+    def _test_process_sync(cls) -> bool:
```
wlach, line 74: This feels like something that should live in a test or test fixture, rather than in the implementation. In this case I'd personally just create a small helper function inside the test file itself.

mdboom: This needs to be here because it needs to run when the global `_testing_mode` flag is set to `True`.  `_testing_mode` is required even for Glean's users writing their own unit tests involving Glean, so it has to be in code shipped in the library.

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -31,81 +32,143 @@ def storage_directory(cls) -> Path:
...
+            cls._do_process_async()
+        else:
+            cls._do_process_sync()
 
     @classmethod
-    def _process_file(cls, path: Path) -> bool:
-        """
-        Processes a single ping file.
-        """
+    def _do_process_async(cls) -> "multiprocessing.Process":
```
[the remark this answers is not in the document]
mdboom: Maybe `_process_pings_multiprocessing`?

PR #800, opened 2020-04-01
Bug 1626258 - Better handling of global object in a threadsafe manner.

badboy:
This has two things:

* Getting the global object now gives an option and thus the FFI layer can return an error instead of panicking.
* Better docs on why the other code is ok, with the addition of logging an error instead of panicking.

Again: it is unlikely to ever hit any of these cases for our current wrappers, as they only initialize once.

### DIFF
--- a/glean-core/src/error.rs
+++ b/glean-core/src/error.rs
@@ -54,6 +54,9 @@ pub enum ErrorKind {
     /// Unknown error
     Utf8Error,
 
+    /// Glean not initialized
+    NotInitialized,
+
     #[doc(hidden)]
     __NonExhaustive,
 }
@@ -75,6 +78,13 @@ impl Error {
             kind: ErrorKind::Utf8Error,
         }
     }
+
+    /// Indicates an error that no global Glean object is initialized
+    pub fn not_initialized() -> Error {
+        Error {
+            kind: ErrorKind::NotInitialized,
+        }
+    }
 }
 
 impl std::error::Error for Error {}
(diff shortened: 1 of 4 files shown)


badboy:
(I'd like two pairs of eyes on this, just to be on the safe side)

glean-core/src/lib.rs
```diff
@@ -81,15 +81,31 @@ static GLEAN: OnceCell<Mutex<Glean>> = OnceCell::new();
 /// Get a reference to the global Glean object.
 ///
 /// Panics if no global Glean object was set.
```
travis79, line 83: Is this comment still true after the protections you added?

badboy: Of course not, missed that. thanks!

glean-core/src/error.rs
```diff
@@ -54,6 +54,9 @@ pub enum ErrorKind {
     /// Unknown error
     Utf8Error,
 
+    /// Glean not initialized
+    Missing,
```
Dexterp37, line 58: nit: what about `NotInitialized`?

badboy: Better wording indeed.

badboy:
> I wonder if this will have any impact on the obscure errors we're seeing on Android through sentry, where kotlin is attempting to call functions on uninitialized objects. Maybe this might help bisect, if the problem happens because of Rust panicking?

So far these calls have been on the Kotlin object being null though.

Dexterp37:
> 
> 
> > I wonder if this will have any impact on the obscure errors we're seeing on Android through sentry, where kotlin is attempting to call functions on uninitialized objects. Maybe this might help bisect, if the problem happens because of Rust panicking?
> 
> So far these calls have been on the Kotlin object being null though.

Yes, but what if the object was null because, for some reason, Rust panicked and Kotlin failed to "init the object"? I mean, I have no idea on how this would even be possible, but maybe it is :D

PR #801, opened 2020-04-02
Bug 1624834 - Updates CircleCI config to use Xcode 11.4.0

travis79:


travis79:
FYI: I'm going to need to kick off the iOS task for this in order to test the new configuration.

travis79:
@badboy Just to fill you in, we noticed that the swift-docs task was failing on CI most likely due to an update to jazzy (after updating locally I was able to reproduce the issue). Along with the update to Xcode 11.4.0, this PR also includes commits that address the failing CI task.

PR #803, opened 2020-04-03
Return the undefined locale for Swift if it can't be otherwise determined

badboy:
this also updates the metrics.yaml in one place, as right now we're still sending that metric.

I looked at the Kotlin code and it seems it "can, but never should be" an empty string, instead of our now "und" string.
Want me to also catch that?

See https://github.com/mozilla/glean/blob/undefined-locale-swift/glean-core/android/src/main/java/mozilla/telemetry/glean/utils/LocaleUtils.kt#L39

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -32,7 +32,8 @@ glean.baseline:
       - baseline
     description: |
       The locale of the application (e.g. "es-ES").
-      If the locale can't be determined on the system, the value is "und",
+      If the locale can't be determined on the system, the value is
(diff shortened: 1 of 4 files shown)


glean-core/metrics.yaml
```diff
@@ -209,6 +209,8 @@ glean.internal.metrics:
       - glean_client_info
     description: |
       The locale of the application (e.g. "es-ES").
+      If the locale can't be determined on the system, the value is "und",
```
Dexterp37, line 212: oh, good catch, this slipped through my last review.

PR #804, opened 2020-04-03
Bug 1626969 - Call out that locale is determined once at initialization

badboy:
Based on #803.


glean-core/metrics.yaml
```diff
@@ -31,7 +31,7 @@ glean.baseline:
     send_in_pings:
       - baseline
     description: |
-      The locale of the application (e.g. "es-ES").
+      The locale of the application during initialization (e.g. "es-ES").
       If the locale can't be determined on the system, the value is "und",
```
Dexterp37, line 35:
```suggestion
      If the locale can't be determined on the system, the value is ["und"](https://unicode.org/reports/tr35/#Unknown_or_Invalid_Identifiers),
```

glean-core/metrics.yaml
```diff
@@ -208,7 +208,7 @@ glean.internal.metrics:
     send_in_pings:
       - glean_client_info
     description: |
-      The locale of the application (e.g. "es-ES").
+      The locale of the application during initialization (e.g. "es-ES").
       If the locale can't be determined on the system, the value is "und",
```
Dexterp37, line 212:
```suggestion
      If the locale can't be determined on the system, the value is ["und"](https://unicode.org/reports/tr35/#Unknown_or_Invalid_Identifiers),
```

glean-core/metrics.yaml
badboy: Taking those in the other PR (#803)

PR #805, opened 2020-04-06
Bug 1627312 - Fix undocumented pieces in Swift

badboy:
```
100% documentation coverage with 0 undocumented symbols
```

### DIFF
--- a/glean-core/ios/Glean/Metrics/DistributionData.swift
+++ b/glean-core/ios/Glean/Metrics/DistributionData.swift
@@ -14,6 +14,9 @@ public class DistributionData {
     let values: [UInt64: UInt64]
     let sum: UInt64
 
+    /// Parse the distribution data from the given JSON string.
+    ///
+    /// If the string is not valid JSON or data is missing `nil` is returned.
     public init?(fromJson json: String) {
         do {
             let data = json.data(using: .utf8)!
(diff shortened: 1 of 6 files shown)


glean-core/ios/Glean/Metrics/DistributionData.swift
```diff
@@ -14,6 +14,9 @@ public class DistributionData {
     let values: [UInt64: UInt64]
     let sum: UInt64
 
+    /// Parse the distribution data from the given JSON string.
+    ///
+    /// If the string is not valid JSON or data is missing `nil` is returned.
```
travis79, line 19: I'm not sure it's entirely necessary for all functions but you omitted the `- parameters` section for this function.

glean-core/ios/Glean/Metrics/Ping.swift
```diff
@@ -25,6 +29,7 @@ public class PingBase {
     var handle: UInt64
     var name: String
 
+    /// Create a new Ping with the given name.
```
travis79, line 32: Again, not sure if we really need it here but this is also missing `- parameters` section.

glean-core/ios/Glean/Metrics/DistributionData.swift
badboy: yeah I was like "do I really need to document a very well named parameter?"

travis79: Good enough explanation for me :))

PR #806, opened 2020-04-06
Bug 1627286: Move baseline ping from background to foreground

mdboom:


mdboom:
Closing in favor of #807

PR #807, opened 2020-04-06
Bug 1627286: Add a baseline ping on foreground

mdboom:
This is an alternative to #806, which *adds* a baseline ping on foreground rather than *replacing* the baseline ping on background with one on foreground.  This should at least provide some context when comparing the two options.

TODO:
- [x] Update the `ping_schedules_and_timings.md` doc to reflect the new behavior.
- [x] Port the changes to Swift
- [ ] Consider adding a foreground baseline ping when re-enabling upload


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -197,19 +197,21 @@ open class GleanInternalAPI internal constructor () {
...
+            // it is always sent before any baseline ping with reason "foreground".
             if (!isFirstRun && LibGleanFFI.INSTANCE.glean_is_dirty_flag_set().toBoolean()) {
                 submitPingByNameSync("baseline", "dirty_startup")
                 // Note: while in theory we should set the "dirty flag" to true
                 // here, in practice it's not needed: if it hits this branch, it
                 // means the value was `true` and nothing needs to be done.
             }
 
+            // Signal Dispatcher that init is complete
+            Dispatchers.API.flushQueuedInitialTasks()
```
Dexterp37, line 213: This is basically matching what you're fixing with #810 , rigth?

CHANGELOG.md
```diff
@@ -3,6 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v26.0.0...master)
 
 * General:
+  * The baseline ping is now sent when the application goes to foreground, in addition to background and dirty-startup.
```
Dexterp37, line 6: Will you file bugs for the other language bindings?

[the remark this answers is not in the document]
Dexterp37: This PR seems to be Android only. Should we move it to the relevant section?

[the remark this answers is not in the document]
travis79: > This PR seems to be Android only. Should we move it to the relevant section?

I will be adding the iOS bits to this PR, I think @mdboom was just waiting to see if there were any implementation changes requested before doing anything with iOS.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: No, slightly different.  We need to send the `dirty_startup` ping before the `foreground` ping, so that any ping-lifetime metrics on the baseline ping are sent correctly in the `dirty_startup` ping.

CHANGELOG.md
mdboom: I thought we would handle iOS here -- and Python doesn't have these pings.

mdboom:
Before merging this, I wouldn't to come back around to @travis79 's suggestion (in #806) that we also send a baseline ping when going from upload disabled to enabled.  It does seem more complete that way -- but is that for here, or a follow-up bug?

Dexterp37:
> 
> 
> Before merging this, I wouldn't to come back around to @travis79 's suggestion (in #806) that we also send a baseline ping when going from upload disabled to enabled. It does seem more complete that way -- but is that for here, or a follow-up bug?

@mdboom I'd leave it as it is for now, since we're still trying to figure out how this can be helpful. I'd not overcharge the 'baseline' ping with behaviours to tackle each single edge case. We can deal with that once needed.

mdboom:
> I'd not overcharge the 'baseline' ping with behaviours to tackle each single edge case.

Yeah -- it's pretty edge.  I doubt users toggle their telemetry very often.  I'm not sure we know how often, but we could probably get a rough idea by finding `seq==0` pings where `first_run_date` is really far in the past.

travis79:
> > I'd not overcharge the 'baseline' ping with behaviours to tackle each single edge case.
> 
> Yeah -- it's pretty edge. I doubt users toggle their telemetry very often. I'm not sure we know how often, but we could probably get a rough idea by finding `seq==0` pings where `first_run_date` is really far in the past.

Just my two-cents, but I would think that toggling telemetry would be its own reason code, `uploadEnabled` or something...

mdboom:
> > > I'd not overcharge the 'baseline' ping with behaviours to tackle each single edge case.
> > 
> > 
> > Yeah -- it's pretty edge. I doubt users toggle their telemetry very often. I'm not sure we know how often, but we could probably get a rough idea by finding `seq==0` pings where `first_run_date` is really far in the past.
> 
> Just my two-cents, but I would think that toggling telemetry would be its own reason code, `uploadEnabled` or something...

Yeah -- I agree about it's own reason code.  But I also think that could be a follow-on to this PR if we determine it's necessary..

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -215,31 +216,40 @@ class GleanTest {
...
+                if (docType == "events") {
+                    assertEquals(1, json.getJSONArray("events").length())
+                } else if (docType == "baseline") {
+                    val seq = json.getJSONObject("ping_info").getInt("seq")
+                    val baselineMetricsObject = json.getJSONObject("metrics")
+                    val baselineStringMetrics = baselineMetricsObject.getJSONObject("string")
+                    assertEquals(1, baselineStringMetrics.length())
+                    assertNotNull(baselineStringMetrics.get("glean.baseline.locale"))
+
+                    // There are three baseline pings -- the first has no duration, the second does
```
badboy, line 239: and the third? :P

glean-core/pings.yaml
```diff
@@ -34,6 +34,11 @@ baseline:
       *Note*: this ping will not contain the `glean.baseline.duration` metric.
     background: |
       The ping was submitted before going to background.
+    foreground: |
+      The ping was submitted when the application went to foreground, which
+      includes when the application starts.
+
+      The `glean.baseline.duration` metric is never sent in these pings.
```
badboy, line 41: Should this just be the same as for `dirty_startup`? Makes it easier to see the similarity.

```suggestion
      *Note*: this ping will not contain the `glean.baseline.duration` metric.
```

CHANGELOG.md
```diff
@@ -20,7 +22,6 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v26.0.0...v27.0.0)
 
-* General:
```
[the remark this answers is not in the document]
badboy: Committed that direct to master just now.

PR #808, opened 2020-04-07
Bug 1627841 - Cancel on exceptions in wrapped measurement function

badboy:
Needs iOS tests before merge.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimespanMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimespanMetricType.kt
@@ -90,14 +90,22 @@ class TimespanMetricType internal constructor(
 
     /**
      * Convenience method to simplify measuring a function or block of code
+     *
+     * If the measured function throws, the measurement is canceled and the exception rethrown.
      */
+    @Suppress("TooGenericExceptionCaught")
     inline fun <U> measure(funcToMeasure: () -> U): U {
         start()
-        try {
-            return funcToMeasure()
-        } finally {
-            stop()
+
+        val returnValue = try {
+            funcToMeasure()
+        } catch (e: Exception) {
+            cancel()
+            throw e
         }
+
+        stop()
+        return returnValue
     }
 
     /**
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
@@ -110,15 +110,23 @@ class TimingDistributionMetricType internal constructor(
     }
 
     /**
-     * Convenience method to simplify measuring a function or block of code
+     * Convenience method to simplify measuring a function or block of code.
+     *
+     * If the measured function throws, the measurement is canceled and the exception rethrown.
      */
+    @Suppress("TooGenericExceptionCaught")
     inline fun <U> measure(funcToMeasure: () -> U): U {
         val timerId = start()
-        try {
-            return funcToMeasure()
-        } finally {
-            stopAndAccumulate(timerId)
+
+        val returnValue = try {
+            funcToMeasure()
+        } catch (e: Exception) {
+            cancel(timerId)
+            throw e
         }
+
+        stopAndAccumulate(timerId)
+        return returnValue
     }
 
     /**
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimespanMetricTypeTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimespanMetricTypeTest.kt
@@ -289,7 +289,7 @@ class TimespanMetricTypeTest {
     }
 
     @Test
-    fun `measure function still measures and allows exceptions to bubble up`() {
+    fun `measure function bubbles up exceptions and timing is canceled`() {
         // Define a timespan metric, which will be stored in "store1"
         val metric = TimespanMetricType(
             disabled = false,
@@ -306,8 +306,7 @@ class TimespanMetricTypeTest {
         }
 
         // Attempt to measure the function that will throw an exception.  The `measure` function
-        // should allow the exception to bubble up while still measuring and recording to the
-        // metric.
+        // should allow the exception to bubble up, the timespan measurement is canceled.
         try {
             metric.measure {
                 testFunc()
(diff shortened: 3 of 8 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimespanMetricType.kt
```diff
@@ -91,13 +91,19 @@ class TimespanMetricType internal constructor(
...
-        try {
-            return funcToMeasure()
-        } finally {
-            stop()
+
+        val returnValue = try {
+            funcToMeasure()
+        } catch (e: Exception) {
+            cancel()
+            throw e
```
Dexterp37, line 102: I don't think we should throw exceptions from our APIs. We never do, our APIs are execption safe. We should instead document that we swallow exception and if a different behaviour is required then the product should not use this method.

badboy: Hm, I kinda disagree with that. We're handling user-provided functions (for convenience) and we're merely passing on the behavior of their code (it's not that different from returning their return value).

Dexterp37: This would be the only API function behaving different, which is something I don't think we should do (... that would be an exception ). We can handle user-defined function, but it doesn't mean it should break consistency with our other APIs.

This doesn't block any other use-case, since the other APIs can still be used.

badboy: IMO it doesn't behave "differently". Glean itself still doesn't produce any exceptions (I see what you did there!).

If users have potentially-throwing functions they need to measure they again need to come up with the exact same wrappper I wrote here instead of using our `measure` utility.

Dexterp37: Functions with exceptions can still be measured. To me this behaves differently: we guarantee that our APIs do not throw exceptions. Having one API suddenly behaving differently (even if the exception originates from a user defined function) can be confusing and a source of subtle bugs. Consumers of the API have two (with no exception) have 3 choices:

(1) Catch any exception in the function they need to measure.
(2) Live with the fact that our APIs swallow exceptions.
(3) Not use this utility wrapper.

I'd rather have consistency with the fact that users can expect no exception to come through our APIs than this added flexibility

travis79: I'm kinda torn over this one...

@Dexterp37 regardless of whether or not we catch the exception, @badboy is re-throwing them so they will make it back to the calling application and won't get swallowed. The only thing different here is that he is cancelling recording the timer metric so that we Glean doesn't record a value if the measured function throws.

If they were worried about a function throwing before, they would probably have already wrapped it in try/catch and thus would catch the exception thrown by a function wrapped by the `measure` function anyway. If we make this change, then it is not possible for us to measure the function under all circumstances (including when it throws). I think that being able to measure the function under this case may be useful, because it may help to indicate where it threw (maybe).  I agree that better documenting what is going on is probably a good idea, but I'd like to be sure that cancelling when a measured exception is thrown is the right thing to do.

An example of what I was thinking if their existing code looked like this:
```
try {
  someFunctionThatThrows()
} catch(e: Exception)
  // do something about it
}
```
Then they could do something like this and both know that the function threw and about how long it ran before that happened:
```
try {
  metric.measure { someFunctionThatThrows() }
} catch(e: Exception) {
  // record if the function threw for analysis, and we have a time recorded
  metricFuncDidThrow.set(true)
}
```

travis79: What if we added a parameter to the `measure` function that allowed the caller to select the behavior they wanted? i.e. `measure(cancelOnThrow: true) { functionToMeasure() }`

Dexterp37: @travis79 I do like the idea of allowing choosing a behaviour, but I think we should always cancel on exception and only allow to choose whether or not to propagate the exception :)

travis79: > @travis79 I do like the idea of allowing choosing a behaviour, but I think we should always cancel on exception and only allow to choose whether or not to propagate the exception :)

As long as we default to propagating the exception since that would be the behavior if the function weren't being measured, I am okay with making that configurable within the function. As for whether or not to cancel? My opinion on that is weakly held as I don't have a concrete use case for not cancelling, so I am fine with the cancel on exception behavior also.

Dexterp37: > As long as we default to propagating the exception since that would be the behavior if the function weren't being measured

I think we should definitely cancel, but I'm not sure we should default to "propagate exceptions". See my previous comments :)

travis79: > I think we should definitely cancel, but I'm not sure we should default to "propagate exceptions". See my previous comments :)

The _only_ problem is that we would be swallowing application exceptions and not _Glean_ exceptions, and I'm not comfortable with that.  If a consuming application expected a function to throw in certain cases and we swallow that inside of the measure function, then we are preventing the application from handling its own exceptions. I totally agree that Glean shouldn't cause exceptions in the application, but we shouldn't interfere with an exception raised by a function defined in the application.

Dexterp37: > 
> 
> > I think we should definitely cancel, but I'm not sure we should default to "propagate exceptions". See my previous comments :)
> 
> The _only_ problem is that we would be swallowing application exceptions and not _Glean_ exceptions, and I'm not comfortable with that. If a consuming application expected a function to throw in certain cases and we swallow that inside of the measure function, then we are preventing the application from handling its own exceptions. 

All right, while I don't agree with this default behaviour, your position is relatively similar to @badboy 's one.

So I'd be fine with taking the initial solution by @badboy with documentation explicitly stating that this might throw if the function throws.

badboy: I realize that I forgot the docs on this function, but there is a sentence on the other function. Will adopt that and then land it.

PR #813, opened 2020-04-08
Remove ping uploading logic from Android bindings in favour of Rust implementation (again)

badboy:
This is taking over #759 (I can't push to Bea's fork).
It incorporates my last review comments in the last commits, otherwise it's all Bea's work.

I personally will give it another review look tomorrow, but I could use some eyes on that. Any takers?


Dexterp37:
> 
> 
> Tried to review -- the fact that this is a pull into the branch and not master confuses me since I don't have the context of the earlier changes. So I'm going to pass and leave this for others...

I'm happy to review, as I have context on this.

badboy:
> Tried to review -- the fact that this is a pull into the branch and not master confuses me since I don't have the context of the earlier changes. So I'm going to pass and leave this for others...

No problem! That's the downside of our current approach. That branch will then soon go away I hope.



> > Tried to review -- the fact that this is a pull into the branch and not master confuses me since I don't have the context of the earlier changes. So I'm going to pass and leave this for others...
> 
> I'm happy to review, as I have context on this.

Thanks! :)

glean-core/android/src/test/java/mozilla/telemetry/glean/net/HttpURLConnectionUploaderTest.kt
```diff
@@ -57,43 +55,7 @@ class HttpURLConnectionUploaderTest {
     }
 
     @Test
-    fun `Glean headers are correctly dispatched`() {
```
Dexterp37, line 60: Should we really remove this? It is checking that the Android object is reporting the same headers we're sending from Glean. Since this class is about testing the specific Android connection object, I'd leave it around

glean-core/android/src/test/java/mozilla/telemetry/glean/private/PingTypeTest.kt
```diff
@@ -201,17 +201,17 @@ class PingTypeTest {
         val server = getMockWebServer()
 
         val counter = CounterMetricType(
-            disabled = false,
-            category = "test",
-            lifetime = Lifetime.Ping,
-            name = "counter",
-            sendInPings = listOf("unknown")
+                disabled = false,
```
Dexterp37, line 204: Out of curiosity, why are we changing indentation here and in other places?

badboy: by accident.

glean-core/android/src/test/java/mozilla/telemetry/glean/net/HttpURLConnectionUploaderTest.kt
badboy: Test still passes, so no, we probably shouldn't, adding it back.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt
```diff
@@ -62,52 +61,16 @@ class HttpURLConnectionUploader : PingUploader {
...
             // There's nothing we can do to recover from this here. So let's just log an error and
             // notify the service that this job has been completed - even though we didn't upload
             // anything to the server.
             Log.e(LOG_TAG, "Could not upload telemetry due to malformed URL", e)
-            return true
+            return null
         } catch (e: IOException) {
             Log.w(LOG_TAG, "IOException while uploading ping", e)
-            return false
+            return null
```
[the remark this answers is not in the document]
badboy: Yeah, we should change that. an I/O exception could mean that the server connection failed or something I guess, in which case we want to try a re-upload later.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -192,10 +99,27 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
...
+                    // Upload the ping request.
+                    // If the status is `null` there was some kind of unrecoverable error
+                    // so we return a known unrecoverable error status code
+                    // which will ensure this gets treated as such.
+                    val status = Glean.httpClient.doUpload(
+                            action.request.path,
+                            action.request.body,
+                            action.request.headers,
+                            Glean.configuration
+                    ) ?: UNRECOVERABLE_ERROR_STATUS_CODE
```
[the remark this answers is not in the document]
badboy: So ... adressing that and the previous comment: I came up with a solution: Instead of bool or `Int?` we introduce a new type we can _properly_ match on.
See https://github.com/mozilla/glean/pull/813/commits/c68c2ec6242d14723bef042f29d3adc997ac2be1

This **is** a breaking change, becuase the PingUploader interface is public and implemented in android-components. The fix in a-c is simple.

I am still converting it to "HTPP status codes" on the way into Rust, but we could address that in a followup.

It's more code than just flipping around some minor types, but IMO it's easier to understand.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
```diff
@@ -10,6 +10,35 @@ package mozilla.telemetry.glean.net
...
+ * A HTTP response code.
+ *
+ * This can still indicate an error, depending on the status code.
+ */
+data class HttpResponse(val statusCode: Int) : UploadResult()
+
+/**
+ * An unrecoverable upload failure.
+ *
+ * The ping data is removed afterwards.
```
Dexterp37, line 30: nit: can you mention an example?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
```diff
@@ -10,6 +10,35 @@ package mozilla.telemetry.glean.net
...
+ */
+object UnrecoverableFailure : UploadResult()
+
+/**
+ * A recoverable failure.
+ *
+ * During upload something went wrong,
```
Dexterp37, line 37: nit: can you mention an example?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -107,12 +113,19 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
...
                             action.request.path,
                             action.request.body,
                             action.request.headers,
                             Glean.configuration
-                    ) ?: UNRECOVERABLE_ERROR_STATUS_CODE
+                    )
+
+                    // Translate the upload result into a status code the Rust side understands.
+                    val status = when (result) {
+                        is RecoverableFailure -> RECOVERABLE_ERROR_STATUS_CODE
```
Dexterp37, line 125: I'd really like this part to live in Rust, rather than here (we can still do this as a follow-up). The logic in Rust will need to make a decision, in the future, in order to potentially record statistics about this. I'd be happy for us to just file a bug for it, no need to do it right now.

badboy: I want to move it over as well, but I didn't want to blow up this PR even further.

badboy: done: https://bugzilla.mozilla.org/show_bug.cgi?id=1628700

badboy:
Everything addressed.

I just want to call out that this is mostly @brizental's work here, I merely took the last few things to push it over the finish line.

PR #814, opened 2020-04-08
Bug 1628301, Bug 1628297: Fix race condition deleting tempdir

mdboom:
`Glean._reset` is called:

1) as part of resetting every unit test, both within Glean and for our users
   writing Glean-aware unit tests,

2) atexit

Glean._reset deletes the Glean data directory if it's a temporary directory
(but not under normal circumstances where a permanent location is used).

There is a race condition between the ping uploader and Glean._reset if
the data directory is deleted while pings are still being uploaded. This
solves that by always firing off a single ping uploader process at a time,
and running both ping uploaders and the temporary directory deletion on that
same one-at-a-time process.

It does not use a Process pool to keep a single process around, because
that would add a lot of complexity to handle a daemon process, vs. this
"fire and forget" process.

This also makes Glean._reset private.

This does not address the broader question of multiple top-level application processes running at the same time and having a race condition between each of their ping uploaders.  See https://bugzilla.mozilla.org/show_bug.cgi?id=1628294 for information on that.

### DIFF
--- a/glean-core/python/glean/_process_dispatcher.py
+++ b/glean-core/python/glean/_process_dispatcher.py
@@ -1,0 +1,76 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+This module implements a class to dispatch work to another process.
+This work always happens one-at-a-time.  That is, it doesn't create
+another worker process until the previous worker process has completed.
+
+This is used only by the PingUploadWorker (and the test-only feature to delete
+temporary data directories which has a potential race condition with the
+PingUploadWorker).
+"""
+
+import sys
+from typing import Optional, TYPE_CHECKING, Union
+
+
+if TYPE_CHECKING:
+    import multiprocessing  # noqa
+
+
+def _work_wrapper(func, args):
+    """
+    A wrapper to call a function and convert its boolean success value into a
+    process exitcode.
+    """
+    success = func(*args)
+
+    if success:
+        sys.exit(0)
+    else:
+        sys.exit(1)
+
+
+class ProcessDispatcher:
+    """
+    A class to dispatch work to another process. This work always happens
+    one-at-a-time. That is, it doesn't create another worker process until the
+    previous worker process has completed.
+
+    Since running a second process might block, this should only be used from
+    the worker thread in `_dispatcher.Dispatcher.
+    """
+
+    # Store the last run process object so we can `join` it when starting
+    # another process.
+    _last_process = None  # type: Optional[multiprocessing.Process]
+
+    @classmethod
+    def dispatch(cls, func, args) -> Union[bool, "multiprocessing.Process"]:
+        from . import Glean
+
+        if Glean._configuration._allow_multiprocessing:
+            # Only import the multiprocessing library if it's actually needed
+            import multiprocessing  # noqa
+
+            # We only want one of these processes running at a time, so if there's
+            # already one, join on it. PingUploadWorker.process is only triggered
+            # from a worker thread anyway, so this blocking will not block the main
+            # user thread.
+            if cls._last_process is not None:
+                cls._last_process.join()
+                cls._last_process = None
+
+            p = multiprocessing.Process(target=_work_wrapper, args=(func, args))
+            p.start()
+
+            cls._last_process = p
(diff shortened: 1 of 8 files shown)


glean-core/python/glean/_process_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+temporary data directories which has a potential race condition with the
+PingUploadWorker).
+"""
+
+import sys
+from typing import Optional, TYPE_CHECKING, Union
+
+
+if TYPE_CHECKING:
+    import multiprocessing  # noqa
```
badboy, line 20: Probably some python thingies, but why is this behind an if?

glean-core/python/glean/_process_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+    # Store the last run process object so we can `join` it when starting
+    # another process.
+    _last_process = None  # type: Optional[multiprocessing.Process]
+
+    @classmethod
+    def dispatch(cls, func, args) -> Union[bool, "multiprocessing.Process"]:
+        from . import Glean
+
+        if Glean._configuration._allow_multiprocessing:
+            # Only import the multiprocessing library if it's actually needed
```
badboy, line 55: Ah, that explains my question from above.

glean-core/python/glean/_process_dispatcher.py
```diff
@@ -0,0 +1,76 @@
...
+            import multiprocessing  # noqa
+
+            # We only want one of these processes running at a time, so if there's
+            # already one, join on it. PingUploadWorker.process is only triggered
+            # from a worker thread anyway, so this blocking will not block the main
```
badboy, line 60: This is provably wrong: it's called in `atexit`.
So I think adjusting this slightly to say the same as the class doc: it might block, better run it off-the-main-thread

glean-core/python/glean/glean.py
```diff
@@ -169,17 +179,26 @@ def configuration(cls):
...
+        # is called atexit, which also waits for the Dispatcher queue to
+        # complete.
+
         Dispatcher.reset()
         if cls._initialized:
             _ffi.lib.glean_destroy_glean()
         cls._initialized = False
         if cls._destroy_data_dir and cls._data_dir.exists():
-            shutil.rmtree(str(cls._data_dir))
+            # This needs to be run in the same one-at-a-time process as the
```
badboy, line 197: I might be misunderstanding this, but the upload worker is notably a _different_ process, it's just that these processes are run one-by-one, right?

glean-core/python/glean/glean.py
```diff
@@ -169,17 +179,26 @@ def configuration(cls):
...
+            # PingUploadWorker to avoid a race condition.
+            p = ProcessDispatcher.dispatch(_rmtree, (str(cls._data_dir),))
+            if not isinstance(p, bool):
```
badboy, line 200: Ugh, different return types /o\
But I guess I'm fine with it, the alternative would be to return a type that's joinable (and throws an exception on `join` if it wraps `false`, and does nothing otherwise)

glean-core/python/glean/_process_dispatcher.py
mdboom: This is the nexus of a bunch of half solutions in Python ;)  Importing the multiprocessing module could fail in contexts where it won't work (the one we care about is PyInstaller which mozregression uses to create a single-file distribution).  So we put it behind a config flag (`allow_multiprocessing`), but we still want to do type-checking (at build time) on things that involve multiprocessing, so we have to unconditionally import it so the type checking tool (mypy) can do it's thing.

glean-core/python/glean/_process_dispatcher.py
mdboom: That's a good point.  At least the `atexit` thing will only block in testing mode (when using temporary directories), so I'm not concerned about the implications of that.  But I'll fix the comment to be more clear.

glean-core/python/glean/glean.py
mdboom: That's correct.  I'll try to clarify this.  It's not *the* upload worker process (because we fire off a new one for each upload task), but only one runs at a time.

glean-core/python/glean/glean.py
mdboom: Yeah -- I had thought about this.  It's probably clearer from the outside / messier on the inside ;)

PR #815, opened 2020-04-08
Bug 1628323 - Change Glean iOS to use Application Support directory

travis79:
This updates the Glean-iOS bindings to use the Application Support directory rather than the Documents directory.  It also renames the `getDocumentsDirectory` utility function to be more appropriately named.

### DIFF
--- a/glean-core/ios/Glean/Utils/Utils.swift
+++ b/glean-core/ios/Glean/Utils/Utils.swift
@@ -98,11 +98,11 @@ extension String {
     }
 }
 
-/// Helper function to retrive the application's Documents directory for persistent file storage
+/// Helper function to retrive the application's Application Support directory for persistent file storage
 ///
-/// - returns: `String` representation of the path to the Documents directory
-func getDocumentsDirectory() -> URL {
-    let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+/// - returns: `URL` of the Application Support directory
+func getGleanDirectory() -> URL {
+    let paths = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)
     let documentsDirectory = paths[0]
     return documentsDirectory.appendingPathComponent("glean_data")
 }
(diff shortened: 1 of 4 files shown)


glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -98,11 +98,11 @@ extension String {
...
 
-/// Helper function to retrive the application's Documents directory for persistent file storage
+/// Helper function to retrive the application's Application Support directory for persistent file storage
 ///
-/// - returns: `String` representation of the path to the Documents directory
-func getDocumentsDirectory() -> URL {
-    let paths = FileManager.default.urls(for: .documentDirectory, in: .userDomainMask)
+/// - returns: `URL` of the Application Support directory
+func getGleanDirectory() -> URL {
+    let paths = FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)
```
badboy, line 105: That was ... easy.

travis79:
> Should we add migration code, that is: copying the stuff from the old place to the new one if there is any?

I discussed this with @Dexterp37 and I don't think it's worth the effort since Lockwise is our only iOS client currently and since they are only fixing bugs and not adding features or upgrades at this time, it should be okay until they want to update to this version, at which point we can probably put the migration code in Lockwise rather than Glean (or Glean if that's easier).

It amounts to copying a folder from one directory to another so it's pretty simple to implement, if you think it's worth it to go ahead and have migration code in this PR.

I'm not opposed to either direction we go, I just went this way as I expect Lockwise the app may never update to a version of Glean that matters.

badboy:
> It amounts to copying a folder from one directory to another so it's pretty simple to implement, if you think it's worth it to go ahead and have migration code in this PR.

I'm even fine with leaving it out. Given that we don't have users that depend on it just yet, it's better to do it that way and not care.
But please do keep track _if_ there's any movement on Lockwise ever.

travis79:
> > It amounts to copying a folder from one directory to another so it's pretty simple to implement, if you think it's worth it to go ahead and have migration code in this PR.
> 
> I'm even fine with leaving it out. Given that we don't have users that depend on it just yet, it's better to do it that way and not care.
> But please do keep track _if_ there's any movement on Lockwise ever.

[Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1628684) to remind us what to do if Lockwise ever upgrades.

PR #816, opened 2020-04-08
Bug 1628310: Don't raise HTTPException or socket.gaierror

mdboom:
These exceptions occur when the network is down, the URL doesn't resolve,
and a bunch of other things. If any of these fail, we should just log and return
False so the ping uploader can move on to other pings or try again next time.
This is preferable to displaying a long (but ultimately harmless) traceback.

### DIFF
--- a/glean-core/python/glean/net/http_client.py
+++ b/glean-core/python/glean/net/http_client.py
@@ -60,10 +60,10 @@ class HttpClientUploader(base_uploader.BaseUploader):
             )
             response = conn.getresponse()
         except http.client.HTTPException as e:
-            log.debug("http.client.HTTPException: {}".format(e))
+            log.error("http.client.HTTPException: {}".format(e))
             return False
         except socket.gaierror as e:
-            log.debug("socket.gaierror: {}".format(e))
+            log.error("socket.gaierror: {}".format(e))
             return False
 
         log.debug("Ping upload: {}".format(response.status))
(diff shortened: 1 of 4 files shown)


glean-core/python/tests/test_network.py
```diff
@@ -81,3 +82,8 @@ def test_unknown_scheme():
         fd.write(b"{}\n")
 
     assert False is PingUploadWorker._test_process_sync()
+
+
+def test_unknown_url_no_exception():
+    # Shouldn't leak any socket or HTTPExceptions
+    assert not HttpClientUploader.upload("http://nowhere.mozilla.org", "{}", [])
```
Dexterp37, line 89:
```suggestion
    assert not HttpClientUploader.upload("http://nowhere.example.com", "{}", [])
```

glean-core/python/glean/net/http_client.py
```diff
@@ -50,10 +51,20 @@ def upload(cls, url: str, data: str, headers: List[Tuple[str, str]]) -> bool:
         else:
             raise ValueError("Unknown URL scheme {}".format(parsed_url.scheme))
 
-        conn.request(
-            "POST", parsed_url.path, body=data.encode("utf-8"), headers=dict(headers),
-        )
-        response = conn.getresponse()
+        try:
```
Dexterp37, line 54: Please kindly add a changelog entry for this :)

PR #818, opened 2020-04-09
Fix indentation

badboy:
[doc only]

Forgot to add them back to the just-merged PR, so here they come now. No need to run tests.


badboy:
Landing without waiting for review, whitespace changes only.

PR #819, opened 2020-04-09
Bug 1628885: Use subprocess rather than multiprocessing for ping uploading

mdboom:
This avoids a number of features that are tricky on Windows, especially on MSYS, such as
duplicating file handles to a spawned subprocess.

### DIFF
--- a/glean-core/python/glean/_process_dispatcher.py
+++ b/glean-core/python/glean/_process_dispatcher.py
@@ -10,52 +10,46 @@ another worker process until the previous worker process has completed.
 This is used only by the PingUploadWorker (and the test-only feature to delete
 temporary data directories which has a potential race condition with the
 PingUploadWorker).
+
+This uses the lower level `subprocess` library rather than `multiprocessing` to
+avoid the unnecessary complexity and not place burdens on Glean's users to
+architect their application startup to be `multiprocessing` compatible on
+Windows.
 """
 
+import base64
+import pickle
+import subprocess
 import sys
-from typing import Optional, TYPE_CHECKING, Union
+from typing import Optional, Union
 
 
-if TYPE_CHECKING:
-    import multiprocessing  # noqa
-
-
-def _work_wrapper(func, args):
-    """
-    A wrapper to call a function in another process and convert its boolean
-    success value into a process exitcode.
-    """
-    success = func(*args)
-
-    if success:
-        sys.exit(0)
-    else:
-        sys.exit(1)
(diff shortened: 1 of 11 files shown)


glean-core/python/glean/_process_dispatcher.py
```diff
@@ -10,52 +10,46 @@
...
 
-    def join(self):
-        self._joined = True
+    def wait(self):
+        self._waited = True
 
     @property
-    def exitcode(self):
+    def returncode(self):
         if not self._joined:
```
badboy, line 51:
```suggestion
        if not self._waited:
```

glean-core/python/glean/_process_dispatcher.py
```diff
@@ -10,52 +10,46 @@
...
 
 class _SyncWorkWrapper:
     """
     A wrapper to synchronously call a function in the current process, but make
-    it have the same (limited) interface as if it were a
-    multiprocessing.Process:
+    it have the same (limited) interface as if it were a `subprocess.Popen`
+    object:
 
         >>> p = ProcessDispatcher.dispatch(myfunc, ())
```
badboy, line 36: Why do the docs of `_SyncWorkWrapper` talk about `ProcessDispatcher`?

badboy: ooooooh, this whole thing only exists so we can swap out the async vs sync worker.
That wasn't super-clear to me. Maybe worth calling out a bit in the docs?

glean-core/python/glean/_process_dispatcher.py
```diff
@@ -75,25 +69,27 @@ class ProcessDispatcher:
...
+                cls._last_process.wait()
                 cls._last_process = None
 
-            p = multiprocessing.Process(target=_work_wrapper, args=(func, args))
-            p.start()
+            p = subprocess.Popen(
+                [
+                    sys.executable,
+                    _process_dispatcher_helper.__file__,
+                    base64.b64encode(pickle.dumps((func, args))).decode("ascii"),
```
badboy, line 90: Are there any limitations with sending base64-encoded and pickled data over?
I know there are arg length limit on different platforms, do we need to take care of that?

I know we probably control all code paths calling this, but we should still document the limitations (if any) for our future selfs.

glean-core/python/glean/_process_dispatcher.py
mdboom: This also highlights a lack of test coverage here, which I'll also address.

PR #825, opened 2020-04-15
Bug 1628747 - Bring back deletion request pings

badboy:
As it turns out removing those tests indeed made us _miss_ some of the details of the deletion-request ping, but not for long.

I brought back the tests, rewrote them to work against the new mechanism and extended them slightly, just for good measure.

I also added some additional trace logging. It's only visible if `RUST_LOG=trace` is set, plus some cleanup.

### DIFF
--- a/glean-core/src/upload/mod.rs
+++ b/glean-core/src/upload/mod.rs
@@ -121,7 +121,7 @@ impl PingUploadManager {
 
         queue.retain(|ping| ping.is_deletion_request());
         log::trace!(
-            "{} pings left in the queue (only deletion-request)",
+            "{} pings left in the queue (only deletion-request expected)",
             queue.len()
         );
         queue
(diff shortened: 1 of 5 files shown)


badboy:
Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1630273, rekicking tests.

glean-core/src/upload/mod.rs
```diff
@@ -111,12 +113,17 @@ impl PingUploadManager {
...
     pub fn clear_ping_queue(&self) -> RwLockWriteGuard<'_, VecDeque<PingRequest>> {
+        log::trace!("Clearing ping queue");
         let mut queue = self
             .queue
             .write()
             .expect("Can't write to pending pings queue.");
 
         queue.retain(|ping| ping.is_deletion_request());
+        log::trace!(
+            "{} pings left in the queue (only deletion-request)",
```
Dexterp37, line 124:
```suggestion
            "{} pings left in the queue (only deletion-request expected)",
```

PR #828, opened 2020-04-15
Bug 1626558 - Replace use of lazy_static with once_cell

travis79:
This seemed like pretty low hanging fruit for a Rust noob like me.  I expect that there will be things to change to make this correct but I am using it to learn more Rust so bring it on!

### DIFF
--- a/glean-core/ffi/src/labeled.rs
+++ b/glean-core/ffi/src/labeled.rs
@@ -2,7 +2,6 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
-use lazy_static::lazy_static;
 use std::convert::TryFrom;
 
 use glean_core::{metrics::*, CommonMetricData, Lifetime};
@@ -30,10 +29,8 @@ use crate::*;
 /// * `get_name` - Function name to get a new instance of the underlying metric.
 macro_rules! impl_labeled_metric {
     ($metric:ty, $global:ident, $metric_global:ident, $new_name:ident, $destroy_name:ident, $get_name:ident, $test_get_num_recorded_errors:ident) => {
-        lazy_static! {
-            static ref $global: ConcurrentHandleMap<LabeledMetric<$metric>> =
-                ConcurrentHandleMap::new();
-        }
+        static $global: once_cell::sync::Lazy<ConcurrentHandleMap<LabeledMetric<$metric>>> =
+            once_cell::sync::Lazy::new(ConcurrentHandleMap::new);
         $crate::define_infallible_handle_map_deleter!($global, $destroy_name);
 
         /// Create a new labeled metric.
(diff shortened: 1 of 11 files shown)


glean-core/ffi/src/labeled.rs
```diff
@@ -30,10 +32,8 @@ use crate::*;
 /// * `get_name` - Function name to get a new instance of the underlying metric.
 macro_rules! impl_labeled_metric {
     ($metric:ty, $global:ident, $metric_global:ident, $new_name:ident, $destroy_name:ident, $get_name:ident, $test_get_num_recorded_errors:ident) => {
-        lazy_static! {
-            static ref $global: ConcurrentHandleMap<LabeledMetric<$metric>> =
-                ConcurrentHandleMap::new();
-        }
+        static $global: Lazy<ConcurrentHandleMap<LabeledMetric<$metric>>> =
+            Lazy::new(|| ConcurrentHandleMap::new());
```
badboy, line 36: If we use the full type here we won't need the import in all other files.
We can also avoid allowing that lint.

```suggestion
        static $global: once_cell::sync::Lazy<ConcurrentHandleMap<LabeledMetric<$metric>>> =
            once_cell::sync::Lazy::new(ConcurrentHandleMap::new);
```

glean-core/ffi/src/macros.rs
```diff
@@ -43,9 +44,7 @@ macro_rules! define_metric {
             $op:ident -> $op_fn:ident($($op_argname:ident: $op_argtyp:ty),* $(,)*)
         ),* $(,)*
     }) => {
-        lazy_static::lazy_static! {
-            pub static ref $metric_map: ffi_support::ConcurrentHandleMap<glean_core::metrics::$metric_type> = ffi_support::ConcurrentHandleMap::new();
-        }
+        pub static $metric_map: Lazy<ffi_support::ConcurrentHandleMap<glean_core::metrics::$metric_type>> = Lazy::new(|| ffi_support::ConcurrentHandleMap::new());
```
badboy, line 47: Same here.

glean-core/ffi/src/ping_type.rs
```diff
@@ -2,18 +2,19 @@
...
 use glean_core::metrics::PingType;
 
 use crate::ffi_string_ext::FallibleToString;
 use crate::handlemap_ext::HandleMapExtension;
 use crate::{from_raw_string_array, with_glean_value, with_glean_value_mut, RawStringArray};
 
-lazy_static! {
-    pub(crate) static ref PING_TYPES: ConcurrentHandleMap<PingType> = ConcurrentHandleMap::new();
-}
+pub(crate) static PING_TYPES: Lazy<ConcurrentHandleMap<PingType>> =
```
badboy, line 16: Same here for avoiding hitting that lint.
It's not in a macro, so importing the type and using the short name is better.

travis79: Thanks, I knew there had to be an easier way :)

PR #830, opened 2020-04-16
Bug 1626542 - Rename `metric_name` to `metric_id` for clarification

travis79:
This renames usage of `metric_name` to `metric_id` to be more clear about what is being used.

### DIFF
--- a/glean-core/src/storage/mod.rs
+++ b/glean-core/src/storage/mod.rs
@@ -31,7 +31,7 @@ fn snapshot_labeled_metrics(
     let map = snapshot.entry(ping_section).or_insert_with(HashMap::new);
 
     let mut s = metric_id.splitn(2, '/');
-    let metric_id = s.next().unwrap(); // Safe unwrap, the function is only called when the name does contain a '/'
+    let metric_id = s.next().unwrap(); // Safe unwrap, the function is only called when the id does contain a '/'
     let label = s.next().unwrap(); // Safe unwrap, the function is only called when the name does contain a '/'
 
     let obj = map.entry(metric_id.into()).or_insert_with(|| json!({}));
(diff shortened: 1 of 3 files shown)


glean-core/src/storage/mod.rs
```diff
@@ -132,9 +132,9 @@ impl StorageManager {
     ) -> Option<Metric> {
         let mut snapshot: Option<Metric> = None;
 
-        let mut snapshotter = |metric_name: &[u8], metric: &Metric| {
-            let metric_name = String::from_utf8_lossy(metric_name).into_owned();
-            if metric_name == metric_id {
+        let mut snapshotter = |id: &[u8], metric: &Metric| {
+            let id = String::from_utf8_lossy(id).into_owned();
```
travis79, line 136: I was afraid that this would shadow the function parameter with the same name, so, considering that this is contained within the function, I opted for `id` here instead of `metric_id`.

glean-core/src/storage/mod.rs
```diff
@@ -17,24 +17,24 @@ use crate::Lifetime;
...
+    metric_id: &str,
     metric: &Metric,
 ) {
     let ping_section = format!("labeled_{}", metric.ping_section());
     let map = snapshot.entry(ping_section).or_insert_with(HashMap::new);
 
-    let mut s = metric_name.splitn(2, '/');
-    let metric_name = s.next().unwrap(); // Safe unwrap, the function is only called when the name does contain a '/'
+    let mut s = metric_id.splitn(2, '/');
+    let metric_id = s.next().unwrap(); // Safe unwrap, the function is only called when the name does contain a '/'
```
Dexterp37, line 34:
```suggestion
    let metric_id = s.next().unwrap(); // Safe unwrap, the function is only called when the id does contain a '/'
```

PR #831, opened 2020-04-16
Bug 1609251 - Rename "ping-uuid" to "document_id" for clarity

travis79:


badboy:
uh yes please. Let me just quickly check if we can merge that cleanly into the feature branch to avoid breaking it later.

PR #832, opened 2020-04-17
Bug 1620682 - Use the same upload result type across FFI

badboy:


glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
```diff
@@ -15,22 +17,34 @@ typealias HeadersList = List<Pair<String, String>>
...
 
 /**
  * A HTTP response code.
  *
  * This can still indicate an error, depending on the status code.
  */
-data class HttpResponse(val statusCode: Int) : UploadResult()
+data class HttpResponse(val statusCode: Int) : UploadResult() {
+    override fun toFfi(): Int {
+        return Constants.UPLOAD_RESULT_HTTP_STATUS or statusCode
```
travis79, line 33: Having come from a C background, the fact that Kotlin uses `and` and `or` rather than `&` and `|` is kinda weird.  It's definitely more readable but I have a tendency to look at the words as the logical varieties and not the bitwise ones.

glean-core/ffi/glean.h
```diff
@@ -16,6 +16,23 @@
 #include <stdint.h>
 #include <stdlib.h>
 
+/**
+ * A HTTP response code.
+ *
+ * The actual response code is encoded in the lower bits.
+ */
+#define UPLOAD_RESULT_HTTP_STATUS 32768
```
travis79, line 24: I guess cbindgen has something against hex notation?

glean-core/src/upload/result.rs
```diff
@@ -0,0 +1,66 @@
...
+    /// This can still indicate an error, depending on the status code.
+    HttpStatus(u32),
+}
+
+impl From<u32> for UploadResult {
+    fn from(status: u32) -> Self {
+        match status {
+            status if (status & UPLOAD_RESULT_HTTP_STATUS) == UPLOAD_RESULT_HTTP_STATUS => {
+                // Extract the status code from the lower bits.
+                let http_status = status & !UPLOAD_RESULT_HTTP_STATUS;
```
travis79, line 56: Also weird to see Rust deviate from the C family of bitwise operators and not go with the `~` for the logical NOT operation.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
badboy: I wrote this first with `|` just for the compiler to shout at me.
I was pleasantly surprised that it uses words.

glean-core/ffi/glean.h
badboy: I'm sure internally 0x1 and 1 are just stored as integers, so they loose any formatting information.
It's auto-generated so I don't care.

glean-core/src/upload/result.rs
badboy: As above, I wrote this with `~` first.
Then the compiler was helpful:

```
error: `~` cannot be used as a unary operator
 --> src/main.rs:3:17
  |
3 |     let x: u8 = ~0xf;
  |                 ^ help: use `!` to perform bitwise not
```	

`!` is indeed both bitwise and logical negation, and it's all driven by [a trait: `Not`](https://doc.rust-lang.org/std/ops/trait.Not.html).
But yes, it differs from C.

glean-core/ffi/glean.h
travis79: I wasn't worried that it would do the wrong thing at all, I was just curious about how cbindgen worked.

glean-core/src/upload/result.rs
Dexterp37: Yup, I concur. A bit surprising, but I admit I like this more ... it feels consistent :)

PR #836, opened 2020-04-20
[doc only] 1615672: Elaborate on why custom pings are useful

mdboom:


docs/user/adding-new-metrics.md
```diff
@@ -99,6 +99,8 @@ Note that for all of the metric configurations, the toggle of the preference off
 
 If the timing at which the metric is sent in the ping needs to closely match the timing of the metrics value, the best option is to use a [custom ping](pings/custom.html) to manually control when pings are sent.
 
+This is especially useful when metrics need to be tightly related to one another, for example when Metric A had value X during the exact same time frame as Metric B had value Y.  If these metrics were in different pings, with different measurement windows, it is much harder to do that kind of reasoning with much certainty.
```
badboy, line 102: I wonder if we should make "Metric A" and "X" more concrete here as examples.
To me "for example" with very abstract values is less explanatory.

Like actually using the render example we know about (or an abstraction from that).

"selected renderer" has "webrender" and "render time" is "50ms" (or whatever we have there actually).

docs/user/pings/custom.md
```diff
@@ -2,6 +2,12 @@
 
 Applications can define metrics that are sent in custom pings. Unlike the built-in pings, custom pings are sent explicitly by the application.
 
+This is useful when the scheduling of the built-in pings ([metrics](metrics.html), [baseline](baseline.html) and [events](events.html)) are not appropriate for your data.  Since the timing of the submission of custom pings is handled by the application, the measurement window is under the application's control.
+
+This is especially useful when metrics need to be tightly related to one another, for example when you need to know that Metric A had value X during the exact same time frame as Metric B had value Y.  If these metrics were in different pings, with different measurement windows, it is much harder to do that kind of analysis with complete certainty.
```
badboy, line 7: as above.

docs/user/pings/custom.md
```diff
@@ -2,6 +2,12 @@
...
+
+A real-world example of this is: when the WebRender backend is set to a particular value, what is the distribution of frame-rendering times?
```
badboy, line 9: See, here we use a very concrete example.

PR #837, opened 2020-04-21
[wip] 1626874: Enable the use of the testing APIs without Robolectric

Dexterp37:
This is a PoC to show case the approach proposed by @mcomella in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1626874).

This works kind of great: the reference test we're using goes from taking ~4s to run to about 380ms.

There's a bunch of things that still need figuring out:

- [ ] This mode of execution exclusively works with _real_ unit tests. This means we should not attempt to use `ProcessLifecycleOwner` or `WorkManager`, otherwise it will crash. Are we fine with allowing this? This should be safe when testing metrics APIs, but it's definitely unsafe if attempting to trigger pings or manually init glean.
- [ ] It requires stubbing & mocking quite a few things from Android. I did my best to keep the code the same, trying to inject dependencies as needed. This works poorly, though, due to `setUploadEnabled`. Not too much a big deal, I guess.
- [ ]  We should clearly document what's testable this way and what's not. Integration tests are a no-no, some classes of unit tests are ok.
- [ ] We should make sure there's no interaction between tests and that the "database isolation" using temporary directories work as expected.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -8,7 +8,6 @@ import android.app.ActivityManager
 import android.util.Log
 import android.content.Context
 import android.content.pm.PackageManager
-import android.os.Build
 import android.os.Process
 import androidx.annotation.MainThread
 import androidx.annotation.VisibleForTesting
@@ -35,6 +34,8 @@ import mozilla.telemetry.glean.scheduler.GleanLifecycleObserver
 import mozilla.telemetry.glean.scheduler.DeletionPingUploadWorker
 import mozilla.telemetry.glean.scheduler.PingUploadWorker
 import mozilla.telemetry.glean.scheduler.MetricsPingScheduler
+import mozilla.telemetry.glean.utils.AndroidBuildInfo
+import mozilla.telemetry.glean.utils.RealBuildInfo
 import mozilla.telemetry.glean.utils.ThreadUtils
 import org.json.JSONObject
 
@@ -90,6 +91,11 @@ open class GleanInternalAPI internal constructor () {
     // case pings are to be immediately submitted by the WorkManager.
     internal var isSendingToTestEndpoint: Boolean = false
 
+    // We need to store this whenever we call `Glean.initialize` in order to
+    // make sure tests have a consistent behaviour due to `setUploadEnabled`.
+    // This is not really required outside of testing.
+    private var buildInfoProvider: AndroidBuildInfo = RealBuildInfo()
+
     /**
      * Initialize the Glean SDK.
      *
@@ -132,11 +138,34 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
+        internalInitialize(
+            applicationContext,
+            uploadEnabled,
+            configuration,
+            RealBuildInfo(),
+            // Lifecycle registration must only be skipped in unit tests.
+            skipLifecycleRegistration = false
+        )
+    }
+
+    /**
+     * Internal-only function to perform the initialization. In addition to being used by
+     * the public API surface, this is used by the testing rules to stub out device info
+     * without using any third party package.
+     */
+    private fun internalInitialize(
+        applicationContext: Context,
+        uploadEnabled: Boolean,
+        configuration: Configuration = Configuration(),
+        buildInfo: AndroidBuildInfo,
+        skipLifecycleRegistration: Boolean
+    ) {
         if (isInitialized()) {
             Log.e(LOG_TAG, "Glean should not be initialized multiple times")
             return
         }
 
+        this.buildInfoProvider = buildInfo
         this.applicationContext = applicationContext
 
         this.configuration = configuration
@@ -174,7 +203,7 @@ open class GleanInternalAPI internal constructor () {
             // The next times we start, we would have them around already.
             val isFirstRun = LibGleanFFI.INSTANCE.glean_is_first_run().toBoolean()
             if (isFirstRun) {
-                initializeCoreMetrics(applicationContext)
+                initializeCoreMetrics(applicationContext, buildInfo)
             }
 
             // Deal with any pending events so we can start recording new ones
@@ -204,7 +233,7 @@ open class GleanInternalAPI internal constructor () {
             // Any new value will be sent in newly generated pings after startup.
             if (!isFirstRun) {
                 LibGleanFFI.INSTANCE.glean_clear_application_lifetime_metrics()
-                initializeCoreMetrics(applicationContext)
+                initializeCoreMetrics(applicationContext, buildInfo)
             }
 
             // Signal Dispatcher that init is complete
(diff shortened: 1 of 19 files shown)


badboy:
> This works kind of great: the reference test we're using goes from taking ~4s to run to about 380ms.

that is ... impressive.

> This mode of execution exclusively works with _real_ unit tests. This means we should not attempt to use `ProcessLifecycleOwner` or `WorkManager`, otherwise it will crash. Are we fine with allowing this? This should be safe when testing metrics APIs, but it's definitely unsafe if attempting to trigger pings or manually init glean.

Only allowing _real_ unit tests sounds in ... you know ... unit tests would be great.
But given how easy it is to trigger pings, that leading to crashes would be very unfortunate. 

> It requires stubbing & mocking quite a few things from Android. I did my best to keep the code the same, trying to inject dependencies as needed. This works poorly, though, due to `setUploadEnabled`. Not too much a big deal, I guess.

(haven't checked the implementation yet) As long as it's mostly static data we're mocking I'd be fine with it.

>  We should clearly document what's testable this way and what's not. Integration tests are a no-no, some classes of unit tests are ok.

Any chance we could _somehow_ detect what mode we're in and therefore also improve error messages/crashes when someone uses something they shouldn't?

Dexterp37:
> But given how easy it is to trigger pings, that leading to crashes would be very unfortunate.

Yes, I agree. I'm not sure there's anything we could do, though, other than documenting. Unless we really want to change internals for testing purposes :(

> > It requires stubbing & mocking quite a few things from Android. I did my best to keep the code the same, trying to inject dependencies as needed. This works poorly, though, due to `setUploadEnabled`. Not too much a big deal, I guess.
> 
> (haven't checked the implementation yet) As long as it's mostly static data we're mocking I'd be fine with it.

Yup, mostly static.

> > We should clearly document what's testable this way and what's not. Integration tests are a no-no, some classes of unit tests are ok.
> 
> Any chance we could _somehow_ detect what mode we're in and therefore also improve error messages/crashes when someone uses something they shouldn't?

I don't think we can, unless we decide to change our internals to accommodate tests, which may lead to other problems. But let me try a bit more tinkering.

Dexterp37:
> But given how easy it is to trigger pings, that leading to crashes would be very unfortunate.

Yes, I agree. I'm not sure there's anything we could do, though, other than documenting. Unless we really want to change internals for testing purposes :(

> > It requires stubbing & mocking quite a few things from Android. I did my best to keep the code the same, trying to inject dependencies as needed. This works poorly, though, due to `setUploadEnabled`. Not too much a big deal, I guess.
> 
> (haven't checked the implementation yet) As long as it's mostly static data we're mocking I'd be fine with it.

Yup, mostly static.

> > We should clearly document what's testable this way and what's not. Integration tests are a no-no, some classes of unit tests are ok.
> 
> Any chance we could _somehow_ detect what mode we're in and therefore also improve error messages/crashes when someone uses something they shouldn't?

I don't think we can, unless we decide to change our internals to accommodate tests, which may lead to other problems. But let me try a bit more tinkering.

Dexterp37:
I tried to extend the testing to our other tests and this uncovered a bunch of new issues that we might want to iron out, for example:

> java.lang.RuntimeException: Method length in org.json.JSONArray not mocked. See http://g.co/androidstudio/not-mocked for details.

	at org.json.JSONArray.length(JSONArray.java)
	at mozilla.telemetry.glean.utils.JsonUtilsKt.asSequence(JsonUtils.kt:42)
	at mozilla.telemetry.glean.utils.JsonUtilsKt.toList(JsonUtils.kt:39)
	at mozilla.telemetry.glean.private.StringListMetricType.testGetValue(StringListMetricType.kt:162)
	at mozilla.telemetry.glean.private.StringListMetricType.testGetValue$default(StringListMetricType.kt:145)
	at mozilla.telemetry.glean.private.StringListMetricTypeTest.The API saves to its storage engine 

There's similar failures in other places. The solution is to either re-implement `org.json.JSONArray` (eugh) or to not use JSON as a way to pass data back and forth in our FFI boundary.

I think we should go for the second option, if we wanted to keep pursuing this. @badboy @mdboom ?

badboy:
> There's similar failures in other places. The solution is to either re-implement `org.json.JSONArray` (eugh) or to not use JSON as a way to pass data back and forth in our FFI boundary.

Wait, why do we need to mock JSON? Mocking that is _certainly_ not an option.

Dexterp37:
> 
> 
> > There's similar failures in other places. The solution is to either re-implement `org.json.JSONArray` (eugh) or to not use JSON as a way to pass data back and forth in our FFI boundary.
> 
> Wait, why do we need to mock JSON? Mocking that is _certainly_ not an option.

Unfortunately Android fails hard in unit tests if you touch _anything_ that is not your code. @mcomella, any other thought with respect to this and the previous comment?

mcomella:
>> Wait, why do we need to mock JSON? Mocking that is certainly not an option.
>
> Unfortunately Android fails hard in unit tests if you touch anything that is not your code. @mcomella, any other thought with respect to this and the previous comment?

I honestly haven't solved this before but I found some seemingly decent solutions on SO:
1. You can import an external `org.json` implementation: https://stackoverflow.com/a/29684901/
1. There's a gradle plugin, "unmock", that allows you to pull real implementations for certain classes from robolectric (or maybe the real android.jar?) without adding a test runner. https://stackoverflow.com/a/30628007/
    1. Note: I'm sketched out about adding unvetted third party code to the project so I'd prefer solution 1) but it's interesting to consider this in the future, if there are many classes we need implementations for

> We have a similar problem with SystemClock.elapsedRealtimeNanos, but I think it's easier to mock that.

I typically pass this function in as a dependency and replace it with a test implementation:
```kotlin
class Example(
    val getElapsedRealtimeNanos: () -> Long = SystemClock::elapsedRealtimeNanos
) { // ...
```

For what this might look like in a test, see [this example](https://searchfox.org/mozilla-mobile/rev/7ce21a195d08b2358070d94ae4c1091a2db70f6f/fenix/app/src/test/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurementTest.kt#34) (and sorry that I mocked glean! 😬).

fwiw, we've found it's actually a best practice to mock anything non-deterministic in tests because it's entirely possible that sometimes those tests will fail. For example, we used a UUID function from Robolectric once and sometimes it would return `null` instead of a valid UUID (unlike the real implementation). While this is clearly due to a bug, in theory, since we compared two UUIDs and random in computers is not truly random, that test could have failed intermittently with very, very low probability: you can have more confidence in your tests if you return mocked or dummy values instead of using non-deterministic outputs.

mcomella:
> You can import an external org.json implementation: https://stackoverflow.com/a/29684901/

I tested this one locally and it seems to work correctly when you don't RunWith Robolectric and when you do.

Dexterp37:
Closing this for now, see the [motivations here](https://bugzilla.mozilla.org/show_bug.cgi?id=1626874#c13).

PR #842, opened 2020-04-23
Taskcluster updates

badboy:


.taskcluster.yml
```diff
@@ -10,7 +10,7 @@ tasks:
     - $let:
           taskgraph:
               branch: taskgraph
-              revision: f949fb56d96b9509b071ba55853f2d60e8dbbf57
+              revision: a8366f88fc8b35b284ad550358c89ec10133fb42
```
MihaiTabara, line 13: Thanks for updating this, it's good to have this up-to-date to benefit for all the under-the-hood goodies in taskgraph. I'll note though that this changes infrequently so shouldn't be concerned about pinning this constantly.

.taskcluster.yml
```diff
@@ -80,7 +80,9 @@ tasks:
                     then: ${event.action}
                     else: 'UNDEFINED'
             in:
-                $if: 'tasks_for == "github-pull-request" && pullRequestAction in ["opened", "reopened", "edited", "synchronize"]'
+                $if: >
+                  (tasks_for == "github-pull-request" && pullRequestAction in ["opened", "reopened", "synchronize"])
```
MihaiTabara, line 84: Thanks for removing "edited" from the list as it can be quite annoying.

PR #845, opened 2020-04-27
Add a new task to run Rust tests

badboy:
cc @MihaiTabara 

Just extending TaskCluster by 2 tasks:
* Rust tests
* Rust clippy (linting)

(meanwhile also fixes some clippy warnings)

Did this to play around with TaskCluster a bit more and understand how transforms work and how new tasks are added.

Contains `[ci skip]`, so CircleCI tasks won't run.


taskcluster/scripts/rust/rustup-setup.sh
```diff
@@ -0,0 +1,27 @@
+#!/bin/bash
```
Dexterp37, line 1: nit: missing license header

PR #848, opened 2020-04-27
[doc only] Bug 1632237 - Document how to send secondary ids in deletion request pings

travis79:
Augments the deletion request ping documentation to indicate how to send secondary ids in the deletion request ping.

### DIFF
--- a/docs/user/pings/deletion_request.md
+++ b/docs/user/pings/deletion_request.md
@@ -21,14 +21,7 @@ As such it attempts to send itself at the moment the user opts out of data colle
 >      A UUID uniquely identifying the legacy client.
 >    send_in_pings:
 >      - deletion_request
->    lifetime: user
->    bugs:
->      - <relevant bug url>
->    data_reviews:
->      - <relevant data review url>
->    notification_emails:
->      - product-team@example.com
->    expires: never
+>    ...
 > ```
 
 ## Scheduling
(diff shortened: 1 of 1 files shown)


docs/user/pings/deletion_request.md
```diff
@@ -9,14 +9,36 @@ This ping contains the client id.
...
+
 ## Scheduling
 
 The `deletion-request` ping is automatically submitted when upload is disabled in Glean.
 If upload fails, it is retried after Glean is initialized.
 
 ## Contents
 
-The `deletion-request` does not contain additional metrics.
+The `deletion-request` does not contain additional metrics aside from secondary ids that have been added.
```
Dexterp37, line 41: Instead of having this content (bound to be outdated!), what if we link to [our autogenerated docs](https://mozilla.github.io/glean/book/user/collected-metrics/metrics.html#deletion-request)?

docs/user/pings/deletion_request.md
```diff
@@ -9,14 +9,36 @@ This ping contains the client id.
...
+> entry for the id to be added with a `send_in_pings` value of `deletion_request`.
+>
+> An example `metrics.yaml` entry might look like this:
+> ```
+> legacy_client_id:
+>    type: uuid
+>    description:
+>      A UUID uniquely identifying the legacy client.
+>    send_in_pings:
+>      - deletion_request
```
Dexterp37, line 23: We maybe shorten this example by just keeping the `type` and `send_in_pings` properties in addition to adding a ` ...` :-) Since most of the fields are bogus anyway

docs/user/pings/deletion_request.md
travis79: Hmmm, sounds good to me.  It still says "This ping contains no metrics." which wouldn't necessarily be true if they added a legacy telemetry id to the ping.  I don't see that text in the `pings.yaml` file, how does it make it into the autogenerated docs?

Dexterp37: Mh, good point. Forget this one and leave your change as it is.

PR #850, opened 2020-04-27
Bug 1630997: Control min and max values in timing distribution

mdboom:


glean-core/src/metrics/memory_distribution.rs
```diff
@@ -171,7 +171,7 @@ impl MemoryDistributionMetric {
 
         if num_too_log_samples > 0 {
             let msg = format!(
-                "Accumulated {} samples longer than 10 minutes",
+                "Accumulated {} samples larger than 1TB",
```
travis79, line 174: Oops... good catch!

mdboom:
> Question: would it make sense to add tests for the other `time_units` at the platform layers, even though we are testing them in glean-core?

Yeah, that's a good idea.  It would at least test that the unit value is being passed in correctly.

mdboom:
@travis79: I added Kotlin and Python tests. Interested in doing a Swift one?

I also improved the Rust test to test things in more detail (which is hard to do from outside since we don't expose the raw histogram data across FFI).

glean-core/python/tests/metrics/test_timing_distribution.py
```diff
@@ -12,33 +12,28 @@
 from glean import _util
 
 
-def test_the_api_saves_to_its_storage_engine():
+def test_the_api_saves_to_its_storage_engine(monkeypatch):
```
mdboom, line 15: Not strictly necessary, but I wasn't aware of this feature in pytest until recently, and I think it improves some of the weirdness in the test.

travis79:
> @travis79: I added Kotlin and Python tests. Interested in doing a Swift one?
> 
> I also improved the Rust test to test things in more detail (which is hard to do from outside since we don't expose the raw histogram data across FFI).

Sure!  Let me check out your branch and I'll add a commit with Swift tests.

travis79:
Okay, I've added a Swift test to match the other platforms.  I ended up having to add a wrapper function to the `TimingDistributionMetricType` in order to be able to mock it so that I could provide arbitrary clock values in the test.  

Flagging @badboy for a r?

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -247,13 +265,15 @@ impl TimingDistributionMetric {
                 if sample < 0 {
                     num_negative_samples += 1;
                 } else {
-                    let sample = sample as u64;
-                    let mut sample = self.time_unit.as_nanos(sample);
+                    let mut sample = sample as u64;
+
                     if sample > MAX_SAMPLE_TIME {
```
badboy, line 270: No minimum check here?

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -247,13 +265,15 @@ impl TimingDistributionMetric {
                 if sample < 0 {
                     num_negative_samples += 1;
                 } else {
-                    let sample = sample as u64;
-                    let mut sample = self.time_unit.as_nanos(sample);
+                    let mut sample = sample as u64;
```
badboy, line 268: Maybe we should call it out that `sample` is already in `time_unit` and thus we can compare it directly to the `MAX_SAMPLE_TIME`

glean-core/src/metrics/timing_distribution.rs
badboy: It could still be `0`, which we don't allow for others.

mdboom:
> Code-wise this looks ok.
> 
> I'm just slightly worried about the fact that we already had a `time_unit`.
> The docs call this out:
> 
> > Note: Timing distributions have an optional `time_unit` parameter that is only used when samples are provided directly from an external tool in a unit other than nanoseconds.
> 
> (on https://mozilla.github.io/glean/book/user/metrics/timing_distribution.html#configuration)

I missed this in the docs.  Definitely needs an update.

> This is used in `accumulateSamples`.
> That behavior now changes, which we should at least call out, if not making sure we don't break anyone.

I don't think we can break anyone.  For underflowing, it is impossible for a 1ms sample being sent to a metric with unit of milliseconds to underflow (because integers).  At the upper end, it is possible that values that formerly were out of range will no longer be errors.

docs/user/metrics/timing_distribution.md
```diff
@@ -17,6 +17,15 @@ If the Glean upload is disabled at the time `stopAndAccumulate` is called, nothi
 
 Multiple concurrent timespans in different threads may be measured at the same time.
 
+Timings are always stored and sent in the payload as nanoseconds. However, the `time_unit` parameter
```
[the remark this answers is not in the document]
mdboom: The design proposal did go through a few revisions before we ended up here.  One of the outcomes of the post mortem is that wasn't captured very well.

docs/user/metrics/timing_distribution.md
```diff
@@ -17,6 +17,15 @@ If the Glean upload is disabled at the time `stopAndAccumulate` is called, nothi
...
+controls the minimum and maximum values that will recorded:
+
+  - `nanosecond`: 1ns < x < 10 minutes
+  - `microsecond`: 1μs < x < ~6.94 days
+  - `millisecond`: 1ms < x < ~19 years
```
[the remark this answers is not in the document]
mdboom: Indeed.  Thanks!

glean-core/src/metrics/timing_distribution.rs
mdboom: Good point

glean-core/src/metrics/timing_distribution.rs
mdboom: Good idea.  Added comment.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt
```diff
@@ -309,4 +309,41 @@ class TimingDistributionMetricTypeTest {
...
+                    maxSampleTime * 1000000L
+            )) {
+                `when`(metric.getElapsedTimeNanos()).thenReturn(0L)
+                val timerId = metric.start()
+                `when`(metric.getElapsedTimeNanos()).thenReturn(value)
+                metric.stopAndAccumulate(timerId)
+            }
+
+            val snapshot = metric.testGetValue()
+            assertTrue(snapshot.values.size < 318)
```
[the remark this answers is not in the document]
mdboom: I've added a comment.  It's a weird number, but it comes from working back from the original upper bound of 10 minutes.

mdboom:
Converting to a draft.  There are very minor changes to glean_parser in https://github.com/mozilla/glean_parser/pull/186 that we should upgrade to here as part of this PR.

daoshengmu:
Please let me know if it has included to A-C and which version. We wanna start to use this functionality.

mdboom:
This was released in Glean 29.1.0, which is currently only on android-components 41.0.0-SNAPSHOT: https://mozac.org/changelog/

PR #853, opened 2020-04-30
Allow calling `PingType.submit()` from Java without parameters

Dexterp37:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
```diff
@@ -72,6 +72,7 @@ class PingType<ReasonCodesEnum : Enum<ReasonCodesEnum>> (
      *
      * @param reason The reason the ping is being submitted.
      */
+    @JvmOverloads
```
mdboom, line 75: "I for one..."  oh, forget it.

PR #854, opened 2020-04-30
Bug 1634310: Python: Fix race condition in atexit

mdboom:
Fixes a race condition in the `atexit` handlers.

Glean currently has two atexit handlers: (a) to make sure the thread worked completes all of its tasks, and (b) that (among other things) deletes the data directory if it's a tmpdir. atexit handlers are run sequentially on the main thread, but the ordering is based on the order in which they are registered, which is somewhat non-deterministic in Glean.

If (b) runs before (a), the data directory is deleted, and then any operations that might be waiting the thread queue will fail with "Database not found".

The fix is to combine the atexit handlers into one, and join on the thread queue before deleting the tempdir.

This raised another issue in my mind that using a tempdir by default is probably not a good choice, and we are seeing this bug in burnham only because burnham doesn't override the data dir (as other "real" apps, such as mozregression have done). Changing the default to a retained directory probably makes sense, and I created bug 1634410 to track that work.

### DIFF
--- a/glean-core/python/glean/_dispatcher.py
+++ b/glean-core/python/glean/_dispatcher.py
@@ -7,7 +7,6 @@ This module implements a single-threaded (mostly FIFO) work queue on which
 most Glean work is done.
 """
 
-import atexit
 import functools
 import logging
 import queue
@@ -82,9 +81,6 @@ class _ThreadWorker:
         self._thread.start()
         self._started = True
         self._ident = self._thread.ident
-        # Register an atexit function to wait for the worker thread to
-        # complete.
-        atexit.register(self._shutdown_thread)
 
     def _worker(self):
         """
@@ -105,15 +101,20 @@ class _ThreadWorker:
 
     def _shutdown_thread(self):
         """
-        An atexit handler to tell the worker thread to shutdown and then wait
-        for 1 seconds for it to finish.
+        Tell the worker thread to shutdown and then wait for 1 seconds for it
+        to finish.
         """
+        if not self._started:
+            return
+
         # Send an END_MARKER to the worker thread to shut it down cleanly.
         self._queue.put((self.END_MARKER, (), {}))
         # Wait up to 1 second for the worker thread to complete.
         self._thread.join(1.0)
         if self._thread.is_alive():
             log.error("Timeout sending Glean telemetry")
+        self._started = False
+        self._thread = None
 
 
 class Dispatcher:
--- a/glean-core/python/glean/net/http_client.py
+++ b/glean-core/python/glean/net/http_client.py
@@ -68,6 +68,9 @@ class HttpClientUploader(base_uploader.BaseUploader):
         except OSError as e:
             log.error("OSError: '{}' {}".format(url, e))
             return False
+        except Exception as e:
+            log.error("Unknown Exception: '{}' {}".format(url, e))
+            return False
 
         log.debug("Ping upload: {}".format(response.status))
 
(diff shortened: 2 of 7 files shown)


mdboom:
Cc: @hackebrot

glean-core/python/glean/net/http_client.py
```diff
@@ -60,10 +60,13 @@ def upload(cls, url: str, data: str, headers: List[Tuple[str, str]]) -> bool:
...
             response = conn.getresponse()
         except http.client.HTTPException as e:
-            log.error("http.client.HTTPException: {}".format(e))
+            log.error("http.client.HTTPException: '{}' {}".format(url, e))
             return False
         except socket.gaierror as e:
-            log.error("socket.gaierror: {}".format(e))
+            log.error("socket.gaierror: '{}' {}".format(url, e))
+            return False
+        except OSError as e:
```
mdboom, line 68: Strictly speaking this is a separate bug -- we get this error if DNS lookup fails.  In typical Python fashion the exceptions that `conn.getresponse` can throw aren't documented, so we're playing whack-a-mole here.

glean-core/python/glean/_dispatcher.py
```diff
@@ -105,15 +101,20 @@ def _worker(self):
 
     def _shutdown_thread(self):
         """
-        An atexit handler to tell the worker thread to shutdown and then wait
-        for 1 seconds for it to finish.
+        Tell the worker thread to shutdown and then wait for 1 seconds for it
```
Dexterp37, line 104: In general, when going from mobile to the desktop world, we should try to be a bit more careful at shutdown (which is something we didn't have to pay much attention to, up to now!). We should probably:

- when shutdown is _initiated_, disallow all the recording APIs;
- terminate pending uploads;
- wait for ping I/O (if any).

So, in general, perform a clean telemetry shutdown as we currently do for legacy telemetry on Firefox desktop. This is not something that should be tackled in this bug, but it could be worthwhile filing a bug about this.

glean-core/python/glean/net/http_client.py
Dexterp37: Since we don't really want this to ever throw, is there any value in just changing this to `except Exception as e:`?

glean-core/python/glean/testing/__init__.py
```diff
@@ -36,14 +36,14 @@ def reset_glean(
...
 
-    Dispatcher._testing_mode = True
-
     data_dir = None  # type: Optional[Path]
     if not clear_stores:
         Glean._destroy_data_dir = False
         data_dir = Glean._data_dir
 
     Glean._reset()
+    Dispatcher._testing_mode = True
```
Dexterp37, line 45: Why was this moved?

glean-core/python/glean/net/http_client.py
mdboom: Yeah, I suppose that's fine.  Makes me feel icky in the event that anything truly unexpected *does* happen, but I should probably just get over that.

glean-core/python/glean/_dispatcher.py
mdboom: Yeah -- let's save that for another bug.  It's tricky, because CLI doesn't quite work the same as desktop.

glean-core/python/glean/testing/__init__.py
mdboom: `Glean._reset` waits for the worker thread to complete, and how that happens depends on the value of `_testing_mode`.  If `_testing_mode` changes before this happens, it won't wait on the worker thread, because it assumes all work is being done on the main thread.

I can add a comment to this effect here.

hackebrot:
Thank you for resolving this bug this quickly, @mdboom @Dexterp37! 👩‍🚀

PR #855, opened 2020-05-04
Bug 1634745 - Export ping request for use by Rust consumers

badboy:
Also includes minor other fixes.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
@@ -43,7 +43,6 @@ internal class UploadBody(
     }
 }
 
-@Structure.FieldOrder("tag", "upload")
 internal open class FfiPingUploadTask(
     // NOTE: We need to provide defaults here, so that JNA can create this object.
     @JvmField var tag: Byte = UploadTaskTag.Done.ordinal.toByte(),
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
@@ -51,6 +51,11 @@ internal open class FfiPingUploadTask(
 ) : Union() {
     class ByValue : FfiPingUploadTask(), Structure.ByValue
 
+    init {
+        // Initialize to be the `tag`-only variant
+        setType("tag")
+    }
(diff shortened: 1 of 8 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -43,14 +43,18 @@ internal class UploadBody(
     }
 }
 
-@Structure.FieldOrder("tag", "upload")
```
Dexterp37, line 46: Why was this removed?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -43,14 +43,18 @@ internal class UploadBody(
...
 internal open class FfiPingUploadTask(
     // NOTE: We need to provide defaults here, so that JNA can create this object.
     @JvmField var tag: Byte = UploadTaskTag.Done.ordinal.toByte(),
     @JvmField var upload: UploadBody = UploadBody()
 ) : Union() {
     class ByValue : FfiPingUploadTask(), Structure.ByValue
 
+    init {
+        // Initialize to be the `tag`-only variant
+        setType("tag")
```
Dexterp37, line 55: Interesting. Why was this working before?

badboy: Because that field is valid in both variants (it's also the first 8 bits in the `UploadBody`, therefore the first 8 bits of the union are always an integer).
This just makes it more explicit, which IMO helps us still understand this code in 2, 4 and 8 weeks.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
badboy: See the commit: a union hardly as an order, only one field is "active" at a time.
(The underlying `Union` inherits from `Structure` and just has a default implementation for `FieldOrder` to satisfy the class hierarchy)

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
Dexterp37: Interesting! Thanks for clarifying!

PR #860, opened 2020-05-06
Suggest needinfo'ing mreid when filing the bug to add a project to the pipeline

wlach:
My preference would be to simply do this as part of the template, but that does not seem to be
possible (bug flags seem not to be remembered as part of the bookmarkable template)

/cc @mreid-moz


wlach:
@mreid-moz gave his ok in PM, so gonna merge this. 😃

PR #861, opened 2020-05-06
Bug 1635865: Attempt to avoid a ConcurrentModificationException

mdboom:


mdboom:
The method is already `@Synchronized`, which handles the concurrent operations from user code.  

The issue here is that initialize iterates over `pingTypeQueue` while also adding to it -- which is concurrent *modification* which isn't allowed even in a single threaded world.  Additionally, this only happens on Android 5 where even if `add` does nothing (since it's already in the set), it throws the exception.  Later versions don't seem to throw the exception if it's already in the set, which is why we don't see this there.

(This is an argument, maybe, for running unit tests on old versions of Android, but I don't particularly want to build that CI ;)

Dexterp37:
> 
> 
> The method is already `@Synchronized`, which handles the concurrent operations from user code.
> 
> The issue here is that initialize iterates over `pingTypeQueue` while also adding to it -- which is concurrent _modification_ which isn't allowed even in a single threaded world. Additionally, this only happens on Android 5 where even if `add` does nothing (since it's already in the set), it throws the exception. Later versions don't seem to throw the exception if it's already in the set, which is why we don't see this there.

@mdboom do you mind adding this bit of information either as a comment in the code or in the bug itself (since you're referencing the bug in the code)?

travis79:
> (This is an argument, maybe, for running unit tests on old versions of Android, but I don't particularly want to build that CI ;)

You don't have to build that CI...  just add the annotation `@Config(emulateSdk=21)`  or whatever version that Android 5 uses and it will force the test to use that SDK version.

mdboom:
> > (This is an argument, maybe, for running unit tests on old versions of Android, but I don't particularly want to build that CI ;)
> 
> You don't have to build that CI... just add the annotation `@Config(sdk=21)` or whatever version that Android 5 uses and it will force the test to use that SDK version.
> (uses Robolectric, http://robolectric.org/configuring/)

Oh, thanks for that.  I had completely forgotten.  Let me do that here.  (I think just about any test that defines a few ping types would do...)

travis79:
> > > (This is an argument, maybe, for running unit tests on old versions of Android, but I don't particularly want to build that CI ;)
> > 
> > 
> > You don't have to build that CI... just add the annotation `@Config(sdk=21)` or whatever version that Android 5 uses and it will force the test to use that SDK version.
> > (uses Robolectric, [http://robolectric.org/configuring](http://robolectric.org/configuring/))
> 
> Oh, thanks for that. I had completely forgotten. Let me do that here. (I think just about any test that defines a few ping types would do...)

One thing to note, it _only_ tests the specified sdk version unless you use `minSdk` and/or `maxSdk`, and _then_ it tests all the specified versions.  So be aware that means it runs the test once for each sdk version so don't pick a long running test if you can help it.

badboy:
(Late to the party) Keep in mind that testing with `@Config(sdk=<old>)` will limit the APIs to that of the SDK, but the implementation is still the provided one, so I guess we wouldn't catch this problem while running on a new implementation (at least that's what I think)

PR #862, opened 2020-05-07
Bug 1634353 - Test upload behavior for pings in quick succession 

badboy:
Unfortunately we can't really test concurrent submissions, because that's essentially a race condition afterall:

1. If ping is submitted while old worker is not finished yet it will pick it up. The Rust test shows that. The upload mechanism keeps an in-memory list during runtime.
2. If ping is submitted and the old worker is already finished, the new worker will start.
3. If ping is submitted and the old worker is about to end, but the new worker still sees it, the ping won't get uploaded and the new worker will not do any work.

At worst we're acting slightly better than the old mechanism.
Case 3 is for such a small window, it's not worth tackling (and again: no data is loss, latest at the next submission or startup the pending ping gets submitted).

### DIFF
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
@@ -1,3 +1,7 @@
+/* This Source Code Form is subject to the terms of the Mozilla Public
+ * License, v. 2.0. If a copy of the MPL was not distributed with this
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
 package mozilla.telemetry.glean.scheduler
 
 import androidx.test.ext.junit.runners.AndroidJUnit4
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
@@ -1,0 +1,108 @@
+package mozilla.telemetry.glean.scheduler
+
+import androidx.test.ext.junit.runners.AndroidJUnit4
+import mozilla.telemetry.glean.Glean
+import mozilla.telemetry.glean.getContextWithMockedInfo
+import mozilla.telemetry.glean.getMockWebServer
+import mozilla.telemetry.glean.private.NoReasonCodes
+import mozilla.telemetry.glean.private.PingType
+import mozilla.telemetry.glean.resetGlean
+import mozilla.telemetry.glean.testing.GleanTestRule
+import mozilla.telemetry.glean.triggerWorkManager
+import mozilla.telemetry.glean.utils.tryGetLong
+import org.junit.Assert.assertEquals
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.json.JSONObject
+import java.util.concurrent.TimeUnit
+
+/**
+ * Testing behavior of custom pings.
+ *
+ * We already rely on the Rust side to test custom pings,
+ * but this enables us to test the upload mechanism specifically.
+ *
+ * Even if this seemingly duplicates some of the testing, this should be kept around.
+ */
+@RunWith(AndroidJUnit4::class)
+class CustomPingTest {
+    private val context = getContextWithMockedInfo()
+
+    @get:Rule
+    val gleanRule = GleanTestRule(context)
+
+    @Test
+    fun `sends empty custom ping`() {
+        // a smoke test for custom pings
+
+        val server = getMockWebServer()
+
+        resetGlean(context, Glean.configuration.copy(
+            serverEndpoint = "http://" + server.hostName + ":" + server.port,
+            logPings = true
+        ), clearStores = true, uploadEnabled = true)
+
+        // Define a new custom ping inline.
+        val customPing = PingType<NoReasonCodes>(
+            name = "custom-ping",
+            includeClientId = true,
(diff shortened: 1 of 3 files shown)


badboy:
spellcheck failed, because the fix from #857 is not in this branch yet.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        // This should launch one worker that handles all pending pings.
+        triggerWorkManager(context)
+
+        // Receive the first ping.
+        var request = server.takeRequest(2L, TimeUnit.SECONDS)
+        var docType = request.path.split("/")[3]
+        assertEquals("custom-ping", docType)
+
+        // Not much data in these pings,
+        // but order should be preserved, so we can check the sequence number.
```
mdboom, line 94: How is the order preserved?  Aren't they just filenames with a uuid on the filesystem?

badboy: 2 things:

1. While running they are added to the in-memory queue, thus preserving order
2. On startup we scan the directory and sort files by modified time.

badboy:
> This feels like the kind of stuff that may fail differently on Windows (where deleting open files is an error). Maybe nothing to do now, but worth looking at when this is ported to Python and FOG.

Huh? We don't keep open files around though. Data is in-memory or read from the file once. Unless we have a UUID conflict it should be fine.

mdboom:
> > This feels like the kind of stuff that may fail differently on Windows (where deleting open files is an error). Maybe nothing to do now, but worth looking at when this is ported to Python and FOG.
> 
> Huh? We don't keep open files around though. Data is in-memory or read from the file once. Unless we have a UUID conflict it should be fine.

Ok.  I forgot the file handling had changed so much on this branch.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -0,0 +1,109 @@
+package mozilla.telemetry.glean.scheduler
```
Dexterp37, line 1: nit: please add the license header

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -0,0 +1,109 @@
...
+ *
+ * We already rely on the Rust side to test custom pings,
+ * but this enables us to test the upload mechanism specifically.
+ *
+ * Even if this seemingly duplicates some of the testing, this should be kept around.
+ */
+@RunWith(AndroidJUnit4::class)
+class CustomPingTest {
+    @get:Rule
+    val gleanRule = GleanTestRule(ApplicationProvider.getApplicationContext())
```
Dexterp37, line 32: Can you use the [`GleanTestLocalServer`](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/testing/GleanTestLocalServer.kt) rule? here's [an example](https://searchfox.org/glean/source/samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/BaselinePingTest.kt#30).

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -0,0 +1,109 @@
...
+        // Define a new custom ping inline.
+        val customPing = PingType<NoReasonCodes>(
+            name = "custom-ping",
+            includeClientId = true,
+            sendIfEmpty = true,
+            reasonCodes = emptyList()
+        )
+
+        // Trigger the ping twice.
+        customPing.submit()
```
Dexterp37, line 81: Does it make sense to turn "test mode" off in these tests, to have async behaviour?

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
badboy: TIL.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
badboy: Actually ... it might. On the other hand it might lead to other problems. I'll run it a few times locally.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
badboy: wait, this is for instrumented tests. This is not one of these, this is a plain "unit" test inside Glean itself.

badboy: That won't work. I still need a glean around and I'd rather use the `GleanTestRule` for that. but that one also initializes work manager, and so does the TestLocalServer rule: https://github.com/mozilla/glean/blob/m16/concurrent-scheduling/glean-core/android/src/main/java/mozilla/telemetry/glean/testing/GleanTestLocalServer.kt#L48-L54

badboy:
> java.lang.AssertionError: A scheduled mozac_service_glean_ping_upload_worker must exist

Well, that would have been too easy I guess. Probably the async-ness makes it take longer to schedule than it takes to check.

PR #867, opened 2020-05-08
[doc only] Bug 1626574: Add info about generating docs for Python

mdboom:


docs/user/adding-glean-to-your-project.md
```diff
@@ -255,6 +255,20 @@ metrics = load_metrics(resource_filename(__name__, "metrics.yaml"))
...
+python3 -m glean_parser translate -f markdown -o docs metrics.yaml pings.yaml
+```
+
+To get more help about the commandline options:
+
+```sh
+python3 -m glean_parser translate --help
+```
+
+We recommend integrating this step into your project's documentation build. The details of that integration is left to you, since it depends on the documentation tool being used and how your project is set up.
```
Dexterp37, line 270: Do we want to make this a requirement for using the SDK or, at least, strongly recommend it and rephrase [this page](https://mozilla.github.io/glean/book/user/adding-glean-to-your-project.html) so that "add documentation for any new metric collected with the library in its repository" recommends using autogeneration tools?

docs/user/adding-glean-to-your-project.md
```diff
@@ -255,6 +255,20 @@ metrics = load_metrics(resource_filename(__name__, "metrics.yaml"))
 metrics.your_category.your_metric.set("value")
 ```
 
+Additionally, Glean provides a mechanism for automatically generating markdown docs from your `metrics.yaml` and `pings.yaml` files. To perform that translation, run `glean_parser`'s `translate` command:
```
badboy, line 258:
```suggestion
Additionally, Glean provides a mechanism for automatically generating markdown documentation from your `metrics.yaml` and `pings.yaml` files. To perform that translation, run `glean_parser`'s `translate` command:
```

docs/user/adding-glean-to-your-project.md
mdboom: I've added additional context that this is required, but needs an additional step for Python to integrate with user projects.  Does that address your concern, @Dexterp37 ?

Dexterp37: @mdboom partly: I think this section looks good now, but I think we should also change `add documentation for any new metric collected with the library in its repository (see an example);` to mention that it can/should be achieved by using our autodocs

mdboom: Ah, I see.  The "adding documentation" happens in the `metrics.yaml` file (and I think that's the intent of that sentence).  Including the `metrics.md` in the repo (and in the docs) is a separate issue, IMHO.  I'll see where it makes sense to add that.

Dexterp37: We can also take this to a separate bug/issue!

PR #869, opened 2020-05-11
Merge release changes back to master

badboy:


badboy:
I ... guess we disagree about the order of things now? oO

badboy:
> Eugh :\

It's a mismatch of version of the swiftformat tool. I intend to fix that in a followup now.

PR #870, opened 2020-05-11
Use latest lint tools on macOS

badboy:


badboy:
49 seconds extra runtime. IMO that's ... okayish.

PR #877, opened 2020-05-12
Enable uploading of gzipped ping payloads

Dexterp37:
This enables gzip ping upload on Android, by making the glean-core crate perform the compression.

Checklist:
- [x]  File a bug for the implementations for the other platforms ([1](https://bugzilla.mozilla.org/show_bug.cgi?id=1637613), [2](https://bugzilla.mozilla.org/show_bug.cgi?id=1637614))
- [ ] Perform QA
- [x] Evaluate the size impact of flate2 on the glean SDK (see [here](https://github.com/mozilla/glean/pull/877#issuecomment-628083471))


glean-core/ffi/src/upload.rs
```diff
@@ -147,7 +145,7 @@ impl Drop for FfiPingUploadTask {
             unsafe {
                 glean_str_free(*document_id);
                 glean_str_free(*path);
-                glean_str_free(*body);
+                //body.destroy(); // TODO FIXME: glean_str_free(*body);
```
Dexterp37, line 148: @badboy this is troubling me a bit - I don't seem to find a way to call `body.destroy()`, because `body` has been moved already. Clues?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -24,20 +24,25 @@ enum class UploadTaskTag {
...
-@Structure.FieldOrder("tag", "documentId", "path", "body", "headers")
+@Structure.FieldOrder("tag", "documentId", "path", "bodyLen", "body", "headers")
 internal class UploadBody(
     // NOTE: We need to provide defaults here, so that JNA can create this object.
     @JvmField val tag: Byte = UploadTaskTag.Done.ordinal.toByte(),
     @JvmField val documentId: Pointer? = null,
     @JvmField val path: Pointer? = null,
-    @JvmField val body: Pointer? = null,
+    // Note that the next two fields (`bodyLen` and `body`) are defined as a single
+    // structure, on the Rust side, called `ByteBuffer`. While the ideal would be to
```
badboy, line 34: Have you used the [`RustBuffer`](https://github.com/mozilla/application-services/blob/e3670c45716b496b550178cc5ed48fbdede816d3/components/support/android/src/main/java/mozilla/appservices/support/native/RustBuffer.kt#L44) structure from a-s?
You might also define this here as `body: RustBuffer.ByValue`

glean-core/ffi/src/upload.rs
badboy: Ah damn, yes.
I _think_ what we need to do is along the following lines ([full example with vec instead of ByteBuffer](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=dc33abca36e24dfb7543881047dbf582))

```rust
let body = mem::replace(body, ByteBuffer::new_with_size(0));
body.destroy();
``` 

What happens here is we take ownership of the `body`, but we need to put something in its place, so we just use an empty byte buffer (there needs to be _something_ because after this `drop` function runs, Rust will clean out all remaining stuff of memory and it can't know we already did that in parts). A byte buffer with size `0` creates a Vec with size 0, which does not allocate, so we don't create any new memory we would need to drop.

An alternative would be to have the `body` be an [`ManuallyDrop<ByteBuffer>`](https://doc.rust-lang.org/std/mem/struct.ManuallyDrop.html), so we can take it out. That's essentially the same (but will involve an `unsafe` call).

glean-core/src/upload/request.rs
```diff
@@ -57,6 +80,10 @@ impl PingRequest {
             "Content-Type",
             "application/json; charset=utf-8".to_string(),
         );
+        if is_gzipped {
+            // TODO: Do we need "Content-Length"?
```
badboy, line 84: We don't depend on it.
In pingsender we don't set it because we're effectively streaming the encoding, so we don't even know the size ahead of time.
As we do know it here, we could set it.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
Dexterp37: Yes, that's the first thing I tried, but still the same problem: when using `union` it fails because it can't determine the size of the struct.

Dexterp37:
@badboy mh, interesting thing with respect to the license check. Are we really not compatible with zlib?

badboy:
> @badboy mh, interesting thing with respect to the license check. Are we really not compatible with zlib?

I'm not sure. So far we didn't have to, so [we didn't enable it](https://github.com/mozilla/glean/blob/bdc6fa841382eb68a4439d7b52d0c070cd027702/deny.toml#L4-L9).
Worth checking [the licensing runbook](https://docs.google.com/document/d/1Oguqp43W4_ChyroJ9AJAzG1jSwkUWfKvBKVvrDxVsMg/edit)

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
badboy: Hm, that's ... weird and unfortunate.

Dexterp37:
Interesting. The vendored schema is now failing. It's querying:

> https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas//schemas/glean/glean/glean.1.schema.json

vs

> https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/master/schemas/glean/glean/glean.1.schema.json

Note that the `master` branch name is missing in the first url.

badboy:
> Interesting. The vendored schema is now failing. It's querying:
> 
> > [mozilla-services/mozilla-pipeline-schemas//schemas/glean/glean/glean.1.schema.json](https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas//schemas/glean/glean/glean.1.schema.json)
> 
> vs
> 
> > [mozilla-services/mozilla-pipeline-schemas:schemas/glean/glean/glean.1.schema.json@`master` (raw)](https://raw.githubusercontent.com/mozilla-services/mozilla-pipeline-schemas/master/schemas/glean/glean/glean.1.schema.json)
> 
> Note that the `master` branch name is missing in the first url.

Because it's failing earlier: `curl: (22) The requested URL returned error: 403 rate limit exceeded`

We just didn't bail out there.

badboy:
Why are we even trying to fetch the latest commit hash? We can just pull from master.

If you don't mind, apply this:

```patch
diff --git .circleci/config.yml .circleci/config.yml
index 6bb458dc..3b6a4817 100644
--- .circleci/config.yml
+++ .circleci/config.yml
@@ -232,7 +232,7 @@ jobs:
       - run:
           name: Check vendored schema for upstream updates
           command: |
-            bin/update-schema.sh latest
+            bin/update-schema.sh master
             if ! git diff --exit-code HEAD -- glean-core/preview/tests/glean.1.schema.json; then
               echo "===================================="
               echo "Latest schema from upstream changed."
```

badboy:
I can also fix that in a separate PR, if this PR is not quickly finished.

glean-core/ffi/src/upload.rs
```diff
@@ -147,9 +145,13 @@ impl Drop for FfiPingUploadTask {
             unsafe {
                 glean_str_free(*document_id);
                 glean_str_free(*path);
-                glean_str_free(*body);
                 glean_str_free(*headers);
             }
+            // Unfortunately, we cannot directly call `body.destroy();` as
+            // we're behind a mutable reference, so we have to manually take the
+            // ownership and drop.
```
badboy, line 152: nit: maybe mention that `ByteBuffer::new_with_size(0)` does not allocate, so we're not leaking memory.

glean-core/src/upload/request.rs
```diff
@@ -29,11 +34,18 @@ impl PingRequest {
     /// Automatically creates the default request headers.
     /// Clients may add more headers such as `userAgent` to this list.
     pub fn new(document_id: &str, path: &str, body: JsonValue) -> Self {
+        // We want uploads to be gzip'd. Instead of doing this for each platform
+        // we have language bindings for, apply compression here.
+        let gzipped_content = Self::gzip_content(path, body.to_string().as_bytes());
+        let add_gzip_header = gzipped_content.is_some();
+        let body = gzipped_content.unwrap_or_else(|| body.to_string().as_bytes().to_vec());
```
badboy, line 41: We can avoid one allocation:

```suggestion
        let body = gzipped_content.unwrap_or_else(|| body.to_string().into_bytes());
```

badboy: Actually, we could avoid another allocation by having

```
let body = body.to_string();
```

above and use that in both in the gzip_content call and then here.

glean-core/android/src/test/java/mozilla/telemetry/glean/utils/DateUtilsTest.kt
```diff
@@ -3,10 +3,10 @@
  * file, You can obtain one at http://mozilla.org/MPL/2.0/.
  */
 
+package mozilla.telemetry.glean.utils
```
badboy, line 6: huh, what happened in this file?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
badboy: Can we file a bug so we can look into this later again?

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/GzipUtils.kt
```diff
@@ -0,0 +1,22 @@
...
+ *
+ * @param data the gzipped [ByteArray] to decompress
+ * @return a [String] containing the uncompressed data.
+ */
+fun decompressGZIP(data: ByteArray): String {
+    val stringBuilder = StringBuilder()
+    GZIPInputStream(ByteArrayInputStream(data)).bufferedReader().useLines { lines ->
+        lines.forEach { stringBuilder.append(it) }
+    }
+    return stringBuilder.toString()
```
badboy, line 21: You should be able to read the whole thing at once, using [`readText`](https://kotlinlang.org/docs/reference/idioms.html#java-7s-try-with-resources).

```suggestion
    val content = GZIPInputStream(ByteArrayInputStream(data)).bufferedReader(UTF_8).use { it.readText() }
    return content
```

Maybe [even like this](https://stackoverflow.com/questions/41000584/best-way-to-use-bufferedreader-in-kotlin):

```suggestion
    val content = GZIPInputStream(ByteArrayInputStream(data)).bufferedReader().use(BufferedReader::readText)
    return content
```

glean-core/android/src/test/java/mozilla/telemetry/glean/TestUtil.kt
```diff
@@ -236,3 +238,19 @@ internal fun getMockWebServer(): MockWebServer {
     })
     return server
 }
+
+/**
+ * Convenience method to get the body of a request as a String.
+ * If the request body is gzipped, it will be decompressed. Otherwise
+ * the UTF8 representation will be reported.
```
badboy, line 245: This reads as if the decopmressed data isn't UTF-8.

How about:

```suggestion
 * The UTF8 representation of the request body will be returned.
 * If the request body is gzipped, it will be decompressed first. 
```

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
```diff
@@ -69,9 +69,10 @@ interface PingUploader {
      * @param url the URL path to upload the data to
      * @param data the serialized text data to send
      * @param headers a [HeadersList] containing the headers to add.
+     * @param isGzipped whether or not the payload is gzipped
      *
      * @return return the status code of the upload response,
      *         or null in case upload could not be attempted at all.
      */
-    fun upload(url: String, data: String, headers: HeadersList): UploadResult
+    fun upload(url: String, data: ByteArray, headers: HeadersList, isGzipped: Boolean): UploadResult
```
badboy, line 77: This is public API. Do we really want to break it _again_?
Why does the `upload` function need to know this?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt
```diff
@@ -104,9 +106,16 @@ class HttpURLConnectionUploader : PingUploader {
     }
 
     @Throws(IOException::class)
-    internal fun doUpload(connection: HttpURLConnection, data: String): Int {
-        connection.outputStream.bufferedWriter().use {
-            it.write(data)
+    internal fun doUpload(connection: HttpURLConnection, data: ByteArray, isGzipped: Boolean): Int {
+        connection.outputStream.use {
+            if (isGzipped) {
```
badboy, line 111: There's no need for this. Even if it is plaintext we should be able to write the ByteArray as is. Therefore not requiring the `isGzipped` flag at all.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt
```diff
@@ -74,12 +74,6 @@ class BaseUploader(d: PingUploader) : PingUploader by d {
         }
     }
 
-    /**
-     * TEST-ONLY. Allows to set specific dates for testing.
-     */
-    @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
-    internal fun getCalendarInstance(): Calendar { return Calendar.getInstance() }
```
badboy, line 81: oh hu, guess we weren't using this and just forgot to remote it?

Dexterp37: Yes :-)

glean-core/android/src/test/java/mozilla/telemetry/glean/utils/DateUtilsTest.kt
Dexterp37: It was missing the package declaration and Android Studio reported a warning, so I fixed it (see the related "BONUS:" commit message

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/GzipUtils.kt
Dexterp37: That's.. way better!

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
Dexterp37: Do you mean `isGzipped`? We're already (unfortunately) breaking it because of the signature change. This should not be too scary, we own the concept-fetch bridge in a-c

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1637611

glean-core/android/src/main/java/mozilla/telemetry/glean/net/PingUploader.kt
badboy: Ah right, we already break it.
Still, don't see the point of `isGzipped` in that function

Dexterp37: All right, dropped the gzip part :)

glean-core/android/src/main/java/mozilla/telemetry/glean/net/HttpURLConnectionUploader.kt
Dexterp37: I... wow, you're right.

Dexterp37:
> Can you also add this to the changelog already? While not affecting usage for users it should still be in there IMO.

Whoops, yes, sorry that slipped through.

> Want me to tackle the iOS implementation? As it stands the Swift tests won't pass.

That would be great, yes, I had flagged you on the related bug on bugzilla already,

badboy:
Note: **DO NOT MERGE**

I'm taking over for the iOS part.

Dexterp37:
Looks like the impact of `flate2` and its dependencies is negligible.

> File  .text     Size Crate
>  5.6%  24.3% 270.4KiB glean_core
>  5.1%  22.2% 247.7KiB glean_ffi
>  3.5%  15.1% 168.2KiB std
>  2.6%  11.2% 124.7KiB regex_syntax
>  1.2%   5.2%  58.4KiB [Unknown]
>  1.2%   5.1%  57.0KiB regex
>  0.8%   3.5%  38.5KiB serde_json
>  0.6%   2.5%  28.4KiB chrono
>  0.4%   1.7%  19.3KiB env_logger
>  0.3%   1.5%  16.4KiB miniz_oxide
>  0.3%   1.3%  14.1KiB lmdb_sys
>  0.2%   1.0%  10.7KiB rkv
>  0.2%   0.8%   9.1KiB c2_chacha
>  0.2%   0.7%   7.6KiB once_cell
>  0.1%   0.5%   5.0KiB termcolor
>  0.1%   0.3%   3.6KiB ffi_support
>  0.1%   0.3%   3.5KiB flate2
>  0.1%   0.3%   3.2KiB lmdb
>  0.1%   0.3%   3.2KiB uuid
>  0.1%   0.2%   2.5KiB ryu
>  0.4%   1.6%  17.4KiB And 19 more crates. Use -n N to show more.
> 23.1% 100.0%   1.1MiB .text section size, the file size is 4.7MiB

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v30.0.0...master)
 
+* Android
```
mdboom, line 5: Android and iOS?

Dexterp37:
(flagged @travis79 for the iOS part)

glean-core/ios/Glean/GleanFfi.h
```diff
@@ -86,6 +86,64 @@ typedef const int32_t *RawIntArray;
...
+ * this wrong on the other side of the FFI will cause memory corruption and crashes.
+ * `i64` is used for the length instead of `u64` and `usize` because JNA has interop
+ * issues with both these types.
+ *
+ * ByteBuffer does not implement Drop. This is intentional. Memory passed into it will
+ * be leaked if it is not explicitly destroyed by calling [`ByteBuffer::destroy`]. This
+ * is because in the future, we may allow it's use for passing data into Rust code.
+ * ByteBuffer assuming ownership of the data would make this a problem.
+ *
+ * Note that alling `destroy` manually is not typically needed or recommended,
```
travis79, line 108:
```suggestion
 * Note that calling `destroy` manually is not typically needed or recommended,
```

glean-core/ffi/glean.h
```diff
@@ -86,6 +86,64 @@ typedef const int32_t *RawIntArray;
...
+ * this wrong on the other side of the FFI will cause memory corruption and crashes.
+ * `i64` is used for the length instead of `u64` and `usize` because JNA has interop
+ * issues with both these types.
+ *
+ * ByteBuffer does not implement Drop. This is intentional. Memory passed into it will
+ * be leaked if it is not explicitly destroyed by calling [`ByteBuffer::destroy`]. This
+ * is because in the future, we may allow it's use for passing data into Rust code.
+ * ByteBuffer assuming ownership of the data would make this a problem.
+ *
+ * Note that alling `destroy` manually is not typically needed or recommended,
```
travis79, line 108:
```suggestion
 * Note that calling `destroy` manually is not typically needed or recommended,
```

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -73,17 +81,30 @@ public class HttpPingUploader {
...
+            // However in tests we're using OHHTTPStubs to stub out the HTTP upload.
+            // It has the known limitation that it doesn't simulate data upload,
+            // because the underlying protocol doesn't expose a hook for that.
+            // By setting `httpBody` here the data is still attached to the request,
+            // so OHHTTPStubs sees it.
+            // It shouldn't be too bad memory-wise and not duplicate the data in memory.
+            // This should only be a reference and Swift keeps track of all the places it's needed.
+            //
+            // See https://github.com/AliSoftware/OHHTTPStubs#known-limitations.
+            request.httpBody = data
```
travis79, line 106: Sounds like you've been having "fun"

glean-core/ios/GleanTests/TestUtils.swift
```diff
@@ -0,0 +1,43 @@
...
+/// When receiving a request, it extracts the ping type from the URL
+/// according to the endpoint URL format.
+///
+/// It assumes the request body is a JSON document and tries to decode it.
+/// If necessary, it decompresses the request body first.
+///
+/// - parameters
+///       * callback: A callback to validate the incoming request.
+///                   It receives a `pingType` and the ping's JSON-decoded `payload`.
+func stubServerReceive(callback: @escaping (String, [String: Any]?) -> Void) {
```
travis79, line 20: This is nice and cleans up a lot!

glean-core/ios/Glean/Net/HttpPingUploader.swift
badboy: So much, yeah.

glean-core/ios/GleanTests/TestUtils.swift
badboy: Eeeexactly

glean-core/ios/Glean/GleanFfi.h
badboy: These are in the generated headers and are coming from ffi-support.
We shouldn't fix that on our end, but rather upstream: https://docs.rs/ffi-support/0.4.0/ffi_support/struct.ByteBuffer.html

badboy:
@travis79 please give this one more look, I'm not sure what state it was in when you first reviewed it. I had to fix some build issues in the sample app and how the dependencies are included.

badboy:
My tests in Firefox for iOS with both build from source and using a (also locally) built binary framework were successful. Let's land this and take it to Fenix for another test round (landing on master will give us a binary framework again)

CHANGELOG.md
badboy:
```suggestion
* Android & iOS
```

PR #883, opened 2020-05-14
Enable Glean Python support outside of MinGW

Dexterp37:
Currently, Glean fails to install on Windows, outside of MinGW. This enables that support again.


Dexterp37:
This has implications for other platforms. Let's wait for @mdboom 's review as well since he's been fighting with the MinGW issues.

PR #884, opened 2020-05-14
Update the Python glean.h when regenerating it

Dexterp37:
While we were updating the one for iOS, we were failing to update the one needed by Python.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -162,6 +162,7 @@ cbindgen: ## Regenerate the FFI header file
 	RUSTUP_TOOLCHAIN=nightly \
 	cbindgen glean-core/ffi --lockfile Cargo.lock -o glean-core/ffi/glean.h
 	cp glean-core/ffi/glean.h glean-core/ios/Glean/GleanFfi.h
+	cp glean-core/ffi/glean.h glean-core/python/glean/glean.h
 .PHONY: cbindgen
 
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -162,6 +162,7 @@ cbindgen: ## Regenerate the FFI header file
 	RUSTUP_TOOLCHAIN=nightly \
 	cbindgen glean-core/ffi --lockfile Cargo.lock -o glean-core/ffi/glean.h
 	cp glean-core/ffi/glean.h glean-core/ios/Glean/GleanFfi.h
+	cp glean-core/ffi/glean.h glean-core/python/glean/glean.h
```
mdboom, line 165: This doesn't seem necessary to me.  glean-python doesn't read the header from `glean-core/python/glean/` anymore.  Instead, it reads it at build time directly from `glean-core/ffi/` at build time.  See [this](https://github.com/mozilla/glean/blob/master/glean-core/python/ffi_build.py#L31).

Maybe you somehow need to rebase on master?

mdboom: Also, maybe try deleting the venv in `glean-core/python/.venv*` to force a rebuild -- maybe something is stuck there.

Dexterp37:
Oh, thanks @mdboom, you're right. The file in my directory was a leftover. This PR is not needed.

PR #886, opened 2020-05-14
Enable the new upload mechanism in Python

Dexterp37:
This makes the Python language bindings use the new upload api and add ping gzipping on the top of that.


glean-core/python/glean/_ffi.py
```diff
@@ -96,19 +92,44 @@ def ffi_encode_vec_int32(values: List[int]) -> Any:
...
+
+    Args:
+        byte_buffer: The C byte buffer.
+        free_memory (bool): Whether or not to free the memory allocated in C.
+    """
+    try:
+        # This copies the buffer, which is slower, but makes memory management
+        # easier because we don't have to remember to free the Rust side with
+        # glean_str_free() later.
+        return bytes(ffi.buffer(byte_buffer.data, byte_buffer.len))
```
Dexterp37, line 124: @mdboom where does the copy actually happen, here? With `bytes`?

glean-core/python/tests/test_network.py
```diff
@@ -9,86 +9,50 @@
 from glean import Glean
 from glean.net import PingUploadWorker
 from glean.net.http_client import HttpClientUploader
-
-
-def test_invalid_filename():
```
Dexterp37, line 14: These tests are covered in the Rust part.

glean-core/src/lib.rs
```diff
@@ -163,7 +163,7 @@ pub struct Configuration {
 pub struct Glean {
     upload_enabled: bool,
     data_store: Option<Database>,
-    event_data_store: EventDatabase,
+    event_data_store: Option<EventDatabase>,
     core_metrics: CoreMetrics,
     internal_pings: InternalPings,
```
Dexterp37, line 168: @badboy @mdboom how do you both like this approach of creating a stub Glean object?

Creating a FFI API to only create a "PingUploader" object would require us to keep the "handle" alive like we do for the Glean object...

Note: I will need to make `core_metrics` and `internal_pings` optional as well.

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -75,65 +71,78 @@ def _test_process_sync(cls) -> bool:
...
+            upload_result = configuration.ping_uploader.do_upload(
+                url_path, body, _parse_ping_headers(headers, doc_id), configuration
+            )
+
+            # Process the response.
+            _ffi.lib.glean_process_ping_upload_response(
+                incoming_task, upload_result.to_ffi()
+            )
+        elif tag == UploadTaskTag.WAIT:
+            # TODO: what to do here?
```
Dexterp37, line 140: In Kotlin we delegate the waiting part to the `WorkManager`, here we don't have that. Since we are in a separate process, waiting that much might be an option.

badboy: This task was supposed to eventually gain a timeout, so the uploader knows how long to wait.

Right now the only way `Wait` can be returned is if the directory is still scanned.
Given that the python uploader will trigger a re-scan of the directory every time (because it starts up a manager every time it's launched) it should probably wait a couple of times (say 2-3 times if wait is returned) and otherwise just end itself.

Dexterp37: Ok, so, if I got that right, @badboy , the `time.wait` is a reasonable strategy here.

glean-core/src/lib.rs
badboy: Not a fan. :P 

More constructive: We don't need any of these fields for the upload manager.
I think we should be able to only create the uploader object. It will be created on each process launch anyway.
That will change the FFI slightly:
* We need to store it (not the worst: put it in a global)
* We either:
	* need a special API to run against that global object
	* or an additional check in the current API to either check the global Glean object or the upload manager (then again, and this is just thinking, what if we remove the upload manager from the Glean object and _only_ have the global singleton of it? Would that make things easier or more complicated?) 

Reason I dislike making more things optional:
* Having a Glean object allocated (and thus "initialized" from the Python view) but only half-filled with stuff will make it hard to detect if we call any metric recording by accident (whereas if it is not initialized it will be a no-op or crash [which would be a bug])
* `Option`s are not free. It's another few bytes on top of our struct and requires an additional branch on each access (probably not super important, as there's only one of these objects, but still overhead for consumers thatn don't require the only-upload-mode)

glean-core/python/glean/_ffi.py
mdboom: Yes.  `ffi.buffer` returns a Python `buffer` object, which is basically just a thin wrapper around the pointer and length on the Rust side.  `bytes` copies that in a new object whose memory is managed by Python.  The question to confirm is -- is the Python caller responsible for freeing the memory passed over in the BytesBuffer object?  If so, this fix (with the copy) is probably the easiest way to reliably handle that.  If we don't, we could remove the `bytes` and the `free` here.

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -75,65 +71,78 @@ def _test_process_sync(cls) -> bool:
...
-    )
-
-    if processed:
-        path.unlink()
-        log.debug("{} was deleted".format(path.name))
-
-    return processed
+        parsed_headers = json.loads(headers_as_json)
+        for header_name, header_value in parsed_headers.items():
+            headers.append((header_name, header_value))
```
mdboom, line 91: Nit: I think the above 4 lines can be replaced with `headers = list(json.loads(headers_as_json).items())`

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -75,65 +71,78 @@ def _test_process_sync(cls) -> bool:
...
+    except json.decoder.JSONDecodeError:
+        log.error("Error while parsing headers for ping " + document_id)
+    finally:
```
mdboom, line 94: Unnecessary `finally`.

glean-core/python/glean/net/ping_upload_worker.py
mdboom: Yes -- and the `time.sleep` won't block the main process or threads, so is not a problem here.  Only detail I think missing here from @badboy's comment is that we should only wait *n* times before `return False` just to prevent the case of an upload process running forever if there is some sort of bug.

glean-core/src/lib.rs
mdboom: As much as it adds additional work, I think I agree with @badboy here.  It's *almost* as if the upload stuff could/should be completely isolated in its own crate (not suggesting we do that, but we should at least think of it that way, maybe).

glean-core/python/glean/net/base_uploader.py
```diff
@@ -50,66 +48,55 @@ def _log_ping(path: str, data: str) -> None:
...
-                successfully or faced and unrecoverable error). False if there
-                was a recoverable error that callers can deal with.
+            result (UploadResult): the status code of the upload response.
         """
         if config.log_pings:
-            self._log_ping(path, data)
+            uncompressed_data = (
+                gzip.decompress(data) if self._is_gzipped(headers) else data
+            )
+            self._log_ping(path, uncompressed_data.decode("utf-8"))
```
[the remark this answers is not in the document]
Dexterp37: Oh, @badboy took that and it's almost ready, see #899 - I'll rebase on the top of that once it's merged.

glean-core/python/glean/net/base_uploader.py
```diff
@@ -50,66 +48,55 @@ def _log_ping(path: str, data: str) -> None:
...
+
+        if config.ping_tag is not None:
```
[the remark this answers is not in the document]
Dexterp37: Good idea! Filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1640575).

glean-core/python/glean/_ffi.py
Dexterp37: Nope, freeing memory is glean-core responsibility. I will fix this.

glean-core/python/glean/net/ping_upload_worker.py
Dexterp37: Oh, good idea.

glean-core/src/lib.rs
Dexterp37: All right, good feedback. I reworked this a bit!

glean-core/python/glean/glean.py
```diff
@@ -159,20 +158,17 @@ def initialize(
         # Deal with any pending events so we can start recording new ones
         @Dispatcher.launch_at_front
         def submit_pending_events():
-            if _ffi.lib.glean_on_ready_to_submit_pings():
+            if (
+                _ffi.lib.glean_on_ready_to_submit_pings()
+                or cls._upload_enabled is False
+            ):
                 PingUploadWorker.process()
```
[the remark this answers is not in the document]
Dexterp37: That's a good point. Actually I noticed that the init is fundamentally different from the other platforms. I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1640831) to investigate it and [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1640833) for the test coverage. @mdboom do we need to block this PR on them? I believe we're not making things worse than they were

glean-core/python/glean/testing/__init__.py
```diff
@@ -69,11 +71,23 @@ def __init__(self, file_path):
...
+            len([h for h in headers if h[0] == "Content-Encoding" and h[1] == "gzip"])
+            == 1
+        )
+
+        uncompressed_data = gzip.decompress(data) if is_gzipped else data
         with self.file_path.open("w") as fd:
-            fd.write(str(url_path) + "\n")
-            fd.write(serialized_ping + "\n")
+            fd.write(str(path) + "\n")
+            fd.write(uncompressed_data.decode("utf-8") + "\n")
```
badboy, line 88: Heh, just thinking here but maybe we can avoid writing files here all together. It's for tests only, we could keep stuff in memory (nothing to do here now)

glean-core/python/glean/_ffi.py
```diff
@@ -96,19 +87,36 @@ def ffi_encode_vec_int32(values: List[int]) -> Any:
...
+
+def ffi_decode_byte_buffer(byte_buffer) -> bytes:
+    """
+    Convert a ByteBuffer returned from Rust to a Python bytes object.
+    Does not free the Rust buffer.
+
+    Args:
+        byte_buffer: The C byte buffer.
+    """
+    return ffi.buffer(byte_buffer.data, byte_buffer.len)
```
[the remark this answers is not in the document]
mdboom: I think the default behavior of buffers in Python is *not* to own things passed from C (FFI).  So we should be ok here.

glean-core/python/glean/testing/__init__.py
mdboom: In memory won't work since uploading happens in a separate process in the Python bindings...  But like you say, nothing to do here for now.

CHANGELOG.md
```diff
@@ -1,5 +1,8 @@
 # Unreleased changes
 
+* Python:
+  * Refactor the ping uploader to use the new upload mechanism.
```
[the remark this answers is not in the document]
mdboom: :+1:

glean-core/python/glean/testing/__init__.py
badboy: gnah, right, even in tests. Files it is.

glean-core/python/glean/_ffi.py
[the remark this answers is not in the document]
Dexterp37: I had the same thought and watched out for this while writing code, came to the same conclusion as you folks, even though I was unable to find docs about it.

glean-core/ffi/src/lib.rs
```diff
@@ -353,6 +398,19 @@ pub extern "C" fn glean_is_first_run() -> u8 {
 // * `result`: the object the output task will be written to.
 #[no_mangle]
 pub extern "C" fn glean_get_upload_task(result: *mut FfiPingUploadTask, log_ping: u8) {
+    // If an upload manager instance is available, use that (it should only happen
+    // in processes which do not initialize Glean).
+    if glean_core::upload::global_upload_manager().is_some() {
+        with_standalone_uploader_value(|ping_uploader| {
+            let ffi_task = FfiPingUploadTask::from(ping_uploader.get_upload_task(log_ping != 0));
+            unsafe {
+                std::ptr::write(result, ffi_task);
```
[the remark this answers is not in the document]
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1641138

glean-core/ffi/src/lib.rs
```diff
@@ -389,6 +447,22 @@ pub unsafe extern "C" fn glean_process_ping_upload_response(
     // but as it controls the memory, we put something valid in place, just in case.
     let task = std::ptr::replace(task, FfiPingUploadTask::Done);
 
+    // If an upload manager instance is available, use that (it should only happen
+    // in processes which do not initialize Glean).
+    if glean_core::upload::global_upload_manager().is_some() {
+        with_standalone_uploader(|ping_uploader| {
+            if let FfiPingUploadTask::Upload { document_id, .. } = task {
```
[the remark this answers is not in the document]
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1641138

PR #890, opened 2020-05-18
Bug 1636951 - Document the requirement to configure an HTTP client

badboy:
[doc only]

### DIFF
--- a/docs/user/general-api.md
+++ b/docs/user/general-api.md
@@ -57,7 +57,7 @@ class SampleApplication : Application() {
 
         // Initialize the Glean library.
         Glean.initialize(
-            applicationContext, 
+            applicationContext,
             // Here, `settings()` is a method to get user preferences, specific to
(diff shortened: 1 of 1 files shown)


docs/user/general-api.md
```diff
@@ -78,6 +78,25 @@ Library code should never call `Glean.initialize`, since it should be called exa
 Glean.initialize(applicationContext, Configuration(channel = "beta"))
 ```
 
+> **Note**: When the Glean SDK is consumed through Android Components, it is required to configure an HTTP client to be used for upload.
```
mdboom, line 81: I was going to suggest linking to the `concept.fetch` docs, but I can't really find the relevant place.  [This link](https://mozac.org/api/mozilla.components.concept.fetch/) is too low-level for what someone would need to know here.  So probably fine without the link, but I wish there was one...

PR #891, opened 2020-05-18
Get upload tasks using an output parameter 

Dexterp37:
Due to the way we use CFFI in Python, we cannot use a `union` type as a return value. On the other
hand, this works fine with output parameters. This commit changes the API to use output parameters
to accommodate all the supported language bindings.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -16,6 +16,7 @@ import androidx.work.Worker
 import androidx.work.WorkerParameters
 import mozilla.telemetry.glean.rust.LibGleanFFI
 import mozilla.telemetry.glean.Glean
+import mozilla.telemetry.glean.net.FfiPingUploadTask
 import mozilla.telemetry.glean.utils.testFlushWorkManagerJob
 import mozilla.telemetry.glean.net.PingUploadTask
 
@@ -54,11 +55,6 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
         // NOTE: The `PINGS_DIR` must be kept in sync with the one in the Rust implementation.
         @VisibleForTesting
         internal const val PINGS_DIR = "pending_pings"
-
-        // For this error, the ping will be retried later
-        internal const val RECOVERABLE_ERROR_STATUS_CODE = 500
-        // For this error, the ping data will be deleted and no retry happens
-        internal const val UNRECOVERABLE_ERROR_STATUS_CODE = 400
 
         /**
          * Function to aid in properly enqueuing the worker in [WorkManager]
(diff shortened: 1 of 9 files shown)


Dexterp37:
Mh, I'm really not sure what clippy is trying to tell me here:

> error: this public function dereferences a raw pointer but is not marked `unsafe`
>    --> glean-core/ffi/src/lib.rs:376:38
>     |
> 376 |         let task = std::ptr::replace(task, FfiPingUploadTask::Done);
>     |                                      ^^^^
>     |
>     = note: `#[deny(clippy::not_unsafe_ptr_arg_deref)]` on by default
>     = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#not_unsafe_ptr_arg_deref
> 
> error: aborting due to previous error
> 
> error: could not compile `glean-ffi`.

The code is within an `unsafe` block already?

glean-core/ffi/src/lib.rs
```diff
@@ -342,27 +342,50 @@ pub extern "C" fn glean_is_first_run() -> u8 {
     with_glean_value(|glean| glean.is_first_run())
 }
 
+// Unfortunately, the way we use CFFI in Python ("out-of-line", "ABI mode") does not
+// allow return values to be `union`s, so we need to use an output parameter instead of
+// a return value to get the task.
```
badboy, line 347: Let's document here that we write the output into `result`.
And also call out that we take the data back in the below function to free it.

badboy:
> The code is within an `unsafe` block already?

Clippy might be a bit overeager. We can mark the function as unsafe instead though.

Dexterp37:
Thanks @badboy for the iOS fixes.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -55,11 +56,6 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
         @VisibleForTesting
         internal const val PINGS_DIR = "pending_pings"
 
-        // For this error, the ping will be retried later
-        internal const val RECOVERABLE_ERROR_STATUS_CODE = 500
```
badboy, line 59: uh, did I forgot to remove them? My bad.

glean-core/ffi/src/lib.rs
```diff
@@ -342,23 +342,58 @@ pub extern "C" fn glean_is_first_run() -> u8 {
...
 }
 
-// We need to pass the whole task instead of only the document id,
-// so that we can free the strings properly on Drop.
+/// Process and free a `FfiPingUploadTask`.
+///
+/// We need to pass the whole task instead of only the document id,
+/// so that we can free the strings properly on Drop.
+///
+/// The data behind the `task` pointer should be freed after return by the caller.
```
badboy, line 368: Reading this now again I thin we should slightly change this:

```suggestion
/// After retrun the `task` should not be used further by the caller. 
/// Any data from the `UploadRequest` has been freed.
```

(not 100% happy with that phrasing either).
I just want to call out that while the Rust side frees the contained strings (if any), the other side need to remove the variable (at best it's just on the stack, so nothing to do anyway)

glean-core/ffi/src/lib.rs
```diff
@@ -342,23 +342,58 @@ pub extern "C" fn glean_is_first_run() -> u8 {
...
 #[no_mangle]
-pub extern "C" fn glean_process_ping_upload_response(task: FfiPingUploadTask, status: u32) {
+pub unsafe extern "C" fn glean_process_ping_upload_response(
+    task: *mut FfiPingUploadTask,
+    status: u32,
+) {
+    // Safety:
+    // * We null-check the passed task before dereferencing.
+    // * We replace data behind the pointer with another valid variant.
+    // * We handle gracefully handle invalid data in strings.
```
badboy, line 381:
```suggestion
    // * We gracefully handle invalid data in strings.
```

PR #895, opened 2020-05-19
Bug 1637990 - Prevent Glean.initialize being called from iOS extensions

travis79:
This prevents `Glean.initialize` from being called within embedded extensions.  This should keep them from generating their own stores and client ids, and ultimately prevent them from sending data. This means that integrating applications will be responsible for implementing a way for extension data to be recorded by the base application.


travis79:
I added a commit that adds the documentation of the limitations of using Glean with iOS application extensions, as per [Bug 1637987](https://bugzilla.mozilla.org/show_bug.cgi?id=1637987], because it made more sense to merge these two changes together.

glean-core/ios/Glean/Glean.swift
```diff
@@ -39,6 +39,10 @@ public class Glean {
 
     private let logger = Logger(tag: Constants.logTag)
 
+    // Cache variable for checking if running in main process.  Also used to override for tests in
+    // order to simulate not running in the main process.  DO NOT SET EXCEPT IN TESTS!
```
badboy, line 43: I'm ... not super happy that this is potentially exposed to users and all that is stopping them to touch this is a comment.
Do we really need a test for this? If not this could be private and unchangeable after first setting.

If we do, couldn't we make this "internal" and then it's only available if Glean is imported `@testable`?

glean-core/ios/Glean/Glean.swift
```diff
@@ -50,6 +54,7 @@ public class Glean {
         self.initialized = false
     }
 
+    // swiftlint:disable function_body_length
```
badboy, line 57: Damn :D

glean-core/ios/Glean/Glean.swift
```diff
@@ -452,6 +467,21 @@ public class Glean {
         GleanDebugUtility.handleCustomUrl(url: url)
     }
 
+    /// Returns true if running in the base application process, otherwise returns false
+    private func checkIsMainProcess() -> Bool {
+        if isMainProcess == nil {
+            if let packageType = Bundle.main.object(forInfoDictionaryKey: "CFBundlePackageType") as? String {
+                isMainProcess = packageType != "XPC!"
```
badboy, line 474: Can we document/link to what `XPC!` means?
No one will know that this is what extensions are.
Also the way it stands if there are other package types that are not XPC but also not the main app it would report wwrong. Should this instead be an equal test (`==`) to what the package type of the main app is?

glean-core/ios/Glean/Glean.swift
travis79: Don't forget, `internal` is the default access for variables in Swift.  This would only be exposed if it were explicitly `public`

glean-core/ios/Glean/Glean.swift
travis79: Yeah, by 1 stinking line :/

glean-core/ios/Glean/Glean.swift
travis79: The only package types we are interested in now fall under XPC services (as app extensions).  [The other possibilites are "APPL" which is an app, "FMWK" for frameworks, and "BNDL", for which "XPC!" is a more specific type of "BNDL".](https://developer.apple.com/documentation/bundleresources/information_property_list/cfbundlepackagetype). [See also](https://developer.apple.com/library/archive/documentation/General/Reference/InfoPlistKeyReference/Articles/CoreFoundationKeys.html#//apple_ref/doc/uid/20001431-111321).  

I'll add some documentation in the form of a comment and I'm happy to invert the `if` statement if it makes you feel more comfortable I just didn't want to exclude potential valid package types so I followed the method of detecting running in an extension with [this code](https://searchfox.org/mozilla-mobile/rev/d5408842b4214f583bd50cdd679addb5464ce8bf/focus-ios/Shared/AppInfo.swift#14-29) taken from Focus.

glean-core/ios/Glean/Glean.swift
travis79: I originally had it as private but, as you noticed, it was easier to test this way.  I'm okay dropping the test and making this explicitly private too, but having test coverage felt useful...  (looks over shoulder for @Dexterp37)

badboy: uuuh, I constantly forget about `internal` being the default. That means under normal use one could already not read or write this field, right?

travis79: Correct, only those that import Glean as `@testable` should be able to access it.

badboy: That makes me less woried then

glean-core/ios/Glean/Glean.swift
badboy: In that case I guess I'm fine with `!= XPC`

docs/user/adding-glean-to-your-project.md
```diff
@@ -229,6 +229,8 @@ Follow these steps to automatically run the parser at build time:
    This will ignore files that are generated at build time by the `sdk_generator.sh` script.
    They don't need to be kept in version control, as they can be re-generated from your `metrics.yaml` and `pings.yaml` files.
 
+> **Important information about Glean and embedded extensions:** Applications for iOS have the ability to define [application extensions](https://developer.apple.com/library/archive/documentation/General/Conceptual/ExtensibilityPG/ExtensionOverview.html#//apple_ref/doc/uid/TP40014214-CH2-SW2) which, while they are contained within the application, do not have access to the same data or APIs, and even run in different processes. Since the extension runs in a separate sandbox from the application, Glean would run in the extension as if it were a completely separate application with different client ids and storage. This complicates the use of Glean because Glean doesn’t know or care about other processes. Because of this, Glean is purposefully prevented from running in an application extension and if metrics are to be collected from extensions, it's up to the integrating application to pass the information to the base application to record in Glean.
```
badboy, line 232: Should we start with a "Glean will not run in application extensions. Metric collection is a no-op in extensions." just to get it out of the way right away?
IMO the long explanation burries the important information a bit.

cc @Dexterp37 on this.

travis79: I'm fine with leading with that info

PR #897, opened 2020-05-19
Remove `ping_type` from `ping_info` docs

chutten:
It was removed in bug 1609149


docs/user/pings/index.md
```diff
@@ -28,7 +28,6 @@ Optional fields are marked accordingly.
 
 | Field name | Type | Description |
 |---|---|---|
-| `ping_type` | String | The name of the ping type (e.g. "baseline", "metrics") |
```
travis79, line 31: Good catch!

PR #898, opened 2020-05-20
Use modified ByteBuffer in Kotlin

badboy:


glean-core/android/src/main/java/mozilla/telemetry/glean/rust/RustBuffer.kt
```diff
@@ -0,0 +1,53 @@
...
+
+package mozilla.telemetry.glean.rust
+
+import com.sun.jna.Pointer
+import com.sun.jna.Structure
+
+/**
+ * This is a mapping for the `glean_core::byte_buffer::ByteBuffer` struct.
+ * It's a copy of the `support.native.RustBuffer` class in application-services,
+ * with the small changes that the length field is a (32-bit) Int instead.
```
Dexterp37, line 13: @badboy can you explain why this small change is even needed?

glean-core/ffi/src/byte_buffer.rs
```diff
@@ -0,0 +1,152 @@
...
+///
+/// The bytes array is allocated on the heap and must be freed on it as well. Critically, if there
+/// are multiple rust packages using being used in the same application, it *must be freed on the
+/// same heap that allocated it*, or you will corrupt both heaps.
+///
+/// Typically, this object is managed on the other side of the FFI (on the "FFI consumer"), which
+/// means you must expose a function to release the resources of `data` which can be done easily
+/// using the [`define_bytebuffer_destructor!`] macro provided by this crate.
+#[repr(C)]
+pub struct ByteBuffer {
```
Dexterp37, line 64: Would it make sense to expose a i32 and a i64 version upstream instead?

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/RustBuffer.kt
badboy: It's in the commit message!

glean-core/ffi/src/byte_buffer.rs
badboy: I can talk to Thom and see what he thinks about it.
It's simple enough code that I'm not worried about the duplication we have here.
Changing it might be problematic for a-s though.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/RustBuffer.kt
Dexterp37: I know :) I was suggesting to add it here as a comment as well, so that we don't need to go through the history to dig up the commit message if we ever stumble upon this :) But if you feel like this is overthinking, I'm fine !

glean-core/ffi/src/byte_buffer.rs
Dexterp37: Oh, I see. No problem then!

PR #899, opened 2020-05-20
Bug 1637914 - Log ping in rust

badboy:


badboy:
> I guess glean_core can't just pull this from the configuration object because it may have changed since `initialize` was called and it was passed across the FFI boundary?

It's actually not in the configuration object on the Rust side to begin with.
Mostly because it was never needed and it might change between invocations due to the debug view activity.

PR #900, opened 2020-05-20
Bug 1635865: Fix ConcurrentModificationException

mdboom:
Cc: @liuche for visibility

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -79,7 +79,7 @@ open class GleanInternalAPI internal constructor () {
     internal lateinit var metricsPingScheduler: MetricsPingScheduler
 
     // Keep track of ping types that have been registered before Glean is initialized.
-    private val pingTypeQueue: MutableSet<PingTypeBase> = mutableSetOf()
+    internal val pingTypeQueue: MutableSet<PingTypeBase> = mutableSetOf()
 
     // This is used to cache the process state and is used by the function `isMainProcess()`
     @VisibleForTesting(otherwise = VisibleForTesting.PRIVATE)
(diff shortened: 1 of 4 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -166,7 +166,13 @@ open class GleanInternalAPI internal constructor () {
 
             // If any pings were registered before initializing, do so now.
             // We're not clearing this queue in case Glean is reset by tests.
-            pingTypeQueue.forEach { registerPingType(it) }
+            synchronized(this@GleanInternalAPI) {
```
Dexterp37, line 169: Gee, I now realize that `@synchronized` is at the top of the `initialize` function but we don't hold it here.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -166,7 +166,13 @@ open class GleanInternalAPI internal constructor () {
...
+                pingTypeQueue.forEach {
+                    LibGleanFFI.INSTANCE.glean_register_ping_type(
```
Dexterp37, line 171: Why can't we call `registerPingType` directly here? We should already hold the mutex at this point so it should be fine.

I'm fine with keeping the code as it is btw, But if there's a specific reason why we can't do that, please let's add a comment about it here

mdboom: Good catch -- I was playing with that to avoid the issue and it ultimately didn't matter.  I agree "using the real API" is better.  Changing it back...

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: Exactly.  It means we're only synchronized for the small bit of things we do on the caller thread, not on the coroutine.

PR #908, opened 2020-05-25
iOS: Remove all stubs on teardown of tests

badboy:
This is recommended to ensure they don't linger around across tests:
https://github.com/AliSoftware/OHHTTPStubs#using-ohhttpstubs-in-your-unit-tests

We didn't have issues with this yet, but ... I let you decide if we pull this in :P


travis79:
Seems like a legitimate thing to do considering the author recommends it and we know what sort of "fun" sticky state can be between tests.

PR #909, opened 2020-05-25
Fix the docs for `RUST_LOG` verbose logging [doc only]

Dexterp37:
The documented value doesn't work for me, locally. This works and matches what we advertise for the android bindings.


Dexterp37:
@badboy what was your specific concern about this? I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1641148): can you please add it there?

PR #913, opened 2020-05-27
Bug 1640241 - Explicitly call out 0 as invalid 

badboy:


badboy:
@chutten Agree, what about that additional sentence?

PR #914, opened 2020-05-27
Bug 1640831: Python: Do Glean.initialize on the worker thread

mdboom:


glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
 
         cls._upload_enabled = upload_enabled
```
Dexterp37, line 133: I just noticed this: should we call `set_upload_enabled()` instead? This would make it consistent with other platforms (can be a follow-up).

glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
...
-            cls._data_dir,
-            application_id,
-            cls._upload_enabled,
-            configuration.max_events,
-        )
+        # Use `Glean._execute_task` rather than `Glean.launch` here, since we
+        # never want to put this work on the `Dispatcher._preinit_queue`.
+        @Dispatcher._execute_task
+        def initialize():
+            cfg = _ffi.make_config(
```
Dexterp37, line 139: I just noticed we're not calling `register_pings` as we do for other platforms. I believe that's ok, since we are not currently sending any ping, but this might change with [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1639882). Nothing to do here?

glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
...
+            )
+
+            cls._initialized = _ffi.lib.glean_initialize(cfg) != 0
+
+            # If initialization of Glean fails, we bail out and don't initialize
+            # further
+            if not cls._initialized:
+                return
+
+            for ping in cls._ping_type_queue:
```
Dexterp37, line 153: Is this threadsafe in Python? I'm worried about hitting the same trouble we were hitting in Kotlin. If it's thread-safe, can we add a comment about it and mentioned that we don't need to worry as we do for Kotlin?

glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
...
+            is_first_run = _ffi.lib.glean_is_first_run() != 0
+            if is_first_run:
+                cls._initialize_core_metrics()
+
+            # Glean Android sets up the metrics ping scheduler here, but we don't
+            # have one.
+
+            # Deal with any pending events so we can start recording new ones
+            @Dispatcher.launch_at_front
+            def submit_pending_events():
```
Dexterp37, line 168: This is no longer required and may break the init sequence, since it would make ping submission happen _after_ what we're doing right now.

Let's make this consistent with other platforms and [call it directly](https://searchfox.org/glean/rev/6f31bba9617f83f32ba347c6796d816fa3a64cb4/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#182-189)

glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
...
+                if (
+                    _ffi.lib.glean_on_ready_to_submit_pings()
+                    or cls._upload_enabled is False
+                ):
+                    PingUploadWorker.process()
 
-        cls._initialized = _ffi.lib.glean_initialize(cfg) != 0
+            # Glean Android checks for the "dirty bit" and sends the `baseline` ping
```
Dexterp37, line 175:
```suggestion
            # Other platforms initialize the `metrics` ping scheduler at this point.

            # Glean Android checks for the "dirty bit" and sends the `baseline` ping
```

Dexterp37: Since we're here, let's add this note..

glean-core/python/glean/glean.py
```diff
@@ -132,56 +132,60 @@ def initialize(
...
 
-        cfg = _ffi.make_config(
-            cls._data_dir,
-            application_id,
-            cls._upload_enabled,
-            configuration.max_events,
-        )
+        # Use `Glean._execute_task` rather than `Glean.launch` here, since we
+        # never want to put this work on the `Dispatcher._preinit_queue`.
+        @Dispatcher._execute_task
```
Dexterp37, line 137: Please note that everything before this point should be thread-safe (for consistency with other platforms), but I don't see this being enforced in any way. Should we be worried? If not, can we document why we believe we shouldn't (e.g. some GIL detail I do not know about).

glean-core/python/glean/glean.py
mdboom: There's no need for the register_pings static workaround in Python, since pings are always explicitly registered when they are loaded with `load_pings`.  Maybe worth a comment here, though.

glean-core/python/glean/glean.py
mdboom: With the GIL, pretty much everything is threadsafe -- that doesn't eliminate race conditions, but there's no data structures that aren't threadsafe.  Given my earlier comment, though, I wonder if we need this at all...

glean-core/python/glean/glean.py
mdboom: Makes sense.

glean-core/python/glean/glean.py
Dexterp37: Yes, having a comment here would be really helpful!

glean-core/python/glean/glean.py
Dexterp37: Will you handle this in a followup?

glean-core/python/glean/glean.py
Dexterp37: Will you investigate this in a follow-up?

mdboom: https://bugzilla.mozilla.org/show_bug.cgi?id=1641558

glean-core/python/glean/glean.py
mdboom: Ok.  Adding one in this PR.

glean-core/python/glean/glean.py
mdboom: I filed this bug to handle this as a follow-up: https://bugzilla.mozilla.org/show_bug.cgi?id=1641564

glean-core/python/glean/glean.py
mdboom: I just made this change right here.  Should be functionally equivalent, so low risk.

PR #918, opened 2020-05-28
Bug 1637675 - Single generated swift file

badboy:


CHANGELOG.md
```diff
@@ -3,10 +3,15 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v30.1.0...master)
 
 * General:
-  * The version of glean_parser has been upgraded to v1.21.0
+  * The version of `glean_parser` has been upgraded to v1.22.0
+    * A maximum of 10 `extra_keys` is now enforced for event metric types.
```
Dexterp37, line 7: Mh, are we now adding changes from the glean_parser here as well? Cool, I didn't know that

badboy: I added it because I felt it might be relevant to Glean consumers and they are more likely to check here first.

PR #921, opened 2020-05-28
Bug 1640927 - Flate2 is default oxidized

badboy:
"miniz_oxide" is the default for flate2, but if Glean is a dependency of
a bigger project that transitively also uses flate2 somewhere else, the
backend in use might change.
We don't really care about the backened in use.

This is fixing an issue when integrating back into mozilla-central,
where flate2 is already in tree and transitively used _somehwere_ with
different features.

--- 

not sure how far along the release is, the dependency summary changes might be in that already, I can drop them from here (and at the latest they would get in with the next version bump)

### DIFF
--- a/DEPENDENCIES.md
+++ b/DEPENDENCIES.md
@@ -8,6 +8,7 @@ the details of which are reproduced below.
 * [MIT License](#MIT)
 * [BSD 2-Clause "Simplified" License](#BSD-2-Clause)
 * [Mozilla Public License 2.0](#MPL-2.0)
+* [zlib License](#Zlib)
 
 -------------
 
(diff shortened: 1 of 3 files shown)


DEPENDENCIES.md
```diff
@@ -2788,8 +2823,8 @@ THE SOFTWARE.
 The following text applies to code linked from these dependencies:
 
 * [benchmark 0.1.0]( https://crates.io/crates/benchmark )
-* [glean-core 30.0.0]( https://github.com/mozilla/glean )
-* [glean-ffi 30.0.0]( https://github.com/mozilla/glean )
+* [glean-core 30.1.0]( https://github.com/mozilla/glean )
+* [glean-ffi 30.1.0]( https://github.com/mozilla/glean )
```
Dexterp37, line 2827: I believe this applies to 31.0.0, since we're cutting this?

badboy: This file is auto-generated (`bin/dependency-summary.sh`), so we just need to run this again against whatever is in master (and it now runs as part of `prepare-release.sh`, so gets updated in that)

badboy:
Ignoring Android error, @Dexterp37 can cherry-pick that commit into the release if necessary

PR #923, opened 2020-05-28
CI: Find, gather and upload core files and logs if possible

badboy:
Instead of a fixed path let's just search for it.


badboy:
Currently testing locally. We might not need this if I figure it out before that.

badboy:
Let's still pull this in in a simpler version, in case we're facing more coredumps in the future. Might help us.

PR #925, opened 2020-05-28
Bug 1640692: Python: Use Rust as canonical source for enum values

mdboom:
This was prompted by the recent bug in the Python bindings that the `Lifetime` enum in glean_parser and glean-core didn't match.  I think that rather than treating glean_parser as the canonical source of values, we should treat the Rust glean-core as the canonical source of values and get them from there where possible.

This exposes the C-like enums (Rust empty enums) through cbindgen so we can use them directly from Python and not have to make sure that the values are in sync.

Then to make sure glean_parser is correct (since glean_parser doesn't / can't have a dependency on glean-core), this just adds a test to make sure its enums are in sync with glean-core's.  It does mean we won't catch those issues until upgrading the version of glean_parser in glean, but that should be more than good enough for most things.

There might be a way to do a similar thing for Swift (I just don't know one way or the other), but I haven't been able to find a way for Kotlin/JNA (and I doubt there is one).

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -12,6 +12,7 @@ use crate::Glean;
 ///
 /// A metric's lifetime determines when its stored data gets reset.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(i32)] // Use i32 to be compatible with our JNA definition
 pub enum Lifetime {
     /// The metric is reset with each sent ping
     Ping,
--- a/glean-core/ffi/cbindgen.toml
+++ b/glean-core/ffi/cbindgen.toml
@@ -19,3 +19,6 @@ crates = ["glean-ffi"]
 [parse]
 parse_deps = true
 include = ["glean-core", "ffi-support"]
+
+[enum]
+prefix_with_name = true
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -34,6 +34,87 @@
 #define UPLOAD_RESULT_UNRECOVERABLE 2
 
 /**
+ * The supported metrics' lifetimes.
+ *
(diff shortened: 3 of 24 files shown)


glean-core/src/common_metric_data.rs
```diff
@@ -12,6 +12,7 @@ use crate::Glean;
 ///
 /// A metric's lifetime determines when its stored data gets reset.
 #[derive(Copy, Clone, Debug, PartialEq, Eq)]
+#[repr(i32)] // Use i32 to be compatible with our JNA definition
```
mdboom, line 15: We should confirm that this change doesn't break the database representation.

mdboom:
> At first I wondered about using the enum name to prefix the values, but I see that it's to prevent clashes. This LGTM!

Yeah -- it's kind of to work around the fact that C enums aren't namespaced.

glean-core/src/common_metric_data.rs
badboy: Afaik we don't store lifetimes as such in the database.
Apart from that:

* Rust optimizes size to be as small as possible (`Lifetime` was a u8 before)
* bincode encodes enums as u32, see https://github.com/servo/bincode#details

mdboom: > Afaik we don't store lifetimes as such in the database.

Yeah -- I had to confirm my memory.  `Lifetime` isn't a key in the database -- we have separate store objects for each.

> Apart from that:
> 
>     * Rust optimizes size to be as small as possible (`Lifetime` was a u8 before)
> 
>     * bincode encodes enums as u32, see https://github.com/servo/bincode#details

Given that we don't actually store the enums in the database, I think this is ok.  Do you agree?

badboy: Yup.

PR #928, opened 2020-05-28
Bug 1641631 - Update Kotlin and Coroutine dependencies for Glean-Android

travis79:
Updates to our Kotlin and Coroutines dependencies to match android-components versions.


badboy:
Might be a good idea to have the reason why we do this in the commit message.

build.gradle
```diff
@@ -14,13 +14,13 @@ buildscript {
...
         android_maven_publish_plugin: '3.6.2',
-        coroutines: '1.3.0',
+        coroutines: '1.3.7',
         dokka: '0.9.17',
         jna: '5.2.0',
         junit: '4.12',
         mockito: '2.24.5',
         mockwebserver: '3.10.0',
-        kotlin: '1.3.50',
+        kotlin: '1.3.72',
```
Dexterp37, line 23: In general, I don't think we should target Fenix requirements for our versions. These requirements, for example, are different from the one A-C uses and often bleeding edge.

A safer path would be to target what A-C does (e.g. [here](https://github.com/mozilla-mobile/android-components/blob/21816c76c7b6c6fa697338b61198fe0b509afcd0/buildSrc/src/main/java/Dependencies.kt#L7)).

travis79: Not a problem, as I mentioned in the Bugzilla comment, I've updated this PR to match A-C rather than Fenix.

build.gradle
```diff
@@ -14,13 +14,13 @@ buildscript {
         android_components: '15.0.0',
```
Dexterp37, line 14: Woha, that's very old. Why do we even need this? @travis79 any chance you could file a follow-up bug for getting rid of it?

build.gradle
```diff
@@ -14,13 +14,13 @@ buildscript {
...
+        coroutines: '1.3.5',
         dokka: '0.9.17',
         jna: '5.2.0',
         junit: '4.12',
         mockito: '2.24.5',
         mockwebserver: '3.10.0',
-        kotlin: '1.3.50',
+        kotlin: '1.3.71',
         robolectric: '4.2.1', // This is different than a-c, but we're fine, it's only tests.
         rust_android_plugin: '0.8.3',
```
Dexterp37, line 25: Any chance you could also double check if A-C is using any newer version for `androidx_*` stuff below? They are, at least, for `androidx_core`

build.gradle
travis79: Since you asked to check for other mismatches below, I'll go ahead and get rid of it in this PR (unless you see a problem with that).

build.gradle
travis79: Yep, let me go through and update everything I can find...   commits incoming.

build.gradle
travis79: Oops, doesn't look like we can get rid of it [because we are using it for some tests](https://github.com/mozilla/glean/blob/e65bf5e7be9404ea3a387eb85fd0130a1da15c08/glean-core/android/build.gradle#L210).  So, would you rather I update it or file a bug to go back and refactor our code to remove the dependency?

Dexterp37: Let's leave it be for now and file a bug to  update or refactor :)

travis79: [Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1641941)

PR #931, opened 2020-05-29
Use absolute paths in `include` list for packaging

badboy:
The list contains "globs", meaning it previously looked for `README.md`
or `LICENSE` in all subfolders, leading to the inclusion of far too
many files, e.g. from a virtualenv.

### DIFF
--- a/glean-core/Cargo.toml
+++ b/glean-core/Cargo.toml
@@ -9,12 +9,12 @@ license = "MPL-2.0"
 edition = "2018"
 keywords = ["telemetry"]
 include = [
-  "README.md",
-  "LICENSE",
-  "src/**/*",
-  "examples/**/*",
-  "tests/**/*",
-  "Cargo.toml"
+  "/README.md",
+  "/LICENSE",
+  "/src",
+  "/examples",
+  "/tests",
+  "/Cargo.toml"
 ]
 
 [badges]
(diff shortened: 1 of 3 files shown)


badboy:
> oh, wow. Does this mean [we were including](https://doc.rust-lang.org/cargo/reference/manifest.html#the-exclude-and-include-fields) useless files?

If you had them around, then yes.
E.g. the 31.0.1 release on crates.io now contains a `A  third_party/rust/glean-core/ios/.venv/lib/python3.7/site-packages/attrs-19.3.0.dist-info/LICENSE` file and I didn't catch that before I pushed it.
(I'm considering to just yank that release)

mdboom:
Thanks for catching this.  Python's packaging stuff can be configured to use the `.gitignore` file -- I wonder if cargo can do the same...?

badboy:
That's exactly what the `include` or an `exclude` list gives you.
In general cargo will _also_ respect `gitignore`, but because of our layout we _can't_ put all the non-Rust stuff into `gitignore` (e.g. all of `android`, `ios`, `python`, ...)

so using the `include` list is the right thing: we specify exactly what should land in the package.
The only problem was that we* (* pretty much I) assumed that list is from the current directory, whereas it were globs all along and requires a leading `/` to be absolute paths.

PR #934, opened 2020-06-02
Bug 1639186 - TC: Build Android build for all targets

badboy:
cc @MihaiTabara 

Skipping circleci once more. This is testing right now, we don't depend on results just yet.


.buildconfig.yml
```diff
@@ -20,5 +20,5 @@ projects:
     artifactId: glean-gradle-plugin
     publications:
       - name: glean-gradle-plugin
-        type: aar
+        type: jar
```
badboy, line 23: This was wrong to begin with.
a-c also has it as jar: https://github.com/mozilla-mobile/android-components/blob/master/.buildconfig.yml#L435-L439

gradle-plugin/build.gradle
```diff
@@ -59,4 +59,10 @@ if (project.name == 'glean-gradle-plugin') {
             }
         }
     }
+
+    // Task alias to match the main glean module's tasks.
+    // For simpler use on CI.
+    task assembleRelease {
```
badboy, line 65: as comment says.

publish.gradle
```diff
@@ -177,6 +177,4 @@ ext.configurePublish = { jnaForTestConfiguration = null ->
             }
         }
     }
-
-    task checkMavenArtifacts
```
badboy, line 181: Empty task, we don't need it (yet)

taskcluster/ci/android-build/kind.yml
```diff
@@ -25,7 +27,7 @@ jobs:
         # to be able to access `rustup` and `rustc` from within the cc script
         # and the gradle command. Another options could be to set those env vars
         # here like: [export, 'PATH=$HOME/.cargo/bin:$PATH'
-        - [source, taskcluster/scripts/rust/rustup-setup.sh]
+        - [source, taskcluster/scripts/rustup-setup.sh]
```
MihaiTabara, line 30: Maybe at some point I should have more granular folder too in app-services, good call 👍

taskcluster/ci/rust/kind.yml
```diff
@@ -21,13 +21,13 @@ jobs:
     description: 'Rust: Clippy'
     run:
       commands:
-        - ['.', './taskcluster/scripts/rust/rustup-setup.sh'] # this should be a pre-command, but those don't run in the same shell (yet)
+        - ['.', './taskcluster/scripts/rustup-setup.sh'] # this should be a pre-command, but those don't run in the same shell (yet)
```
MihaiTabara, line 24: If you're `source`-ing as opposed to running the script itself, it should work, even if it's running in the `pre-command`. That's because https://github.com/mozilla/glean/blob/master/taskcluster/glean_taskgraph/job.py#L62 is basically a chain of commands.

taskcluster/glean_taskgraph/__init__.py
```diff
@@ -25,4 +25,11 @@ def _import_modules(modules):
 
 
 def get_decision_parameters(graph_config, parameters):
-    pass
+    if parameters["tasks_for"] == "github-pull-request":
```
MihaiTabara, line 28: This reminds me I need to backport this `ci-skip/normal/full` logic in the rest of mobile repos too. We've lost it during the taskgraph migration in all the projects except this one and app-services.

taskcluster/ci/rust/kind.yml
badboy: Oh, good point. These commands were from the previous integration and we didn't change it yet.

PR #936, opened 2020-06-03
Add minimal language bindings for C#

Dexterp37:
This PR sets up the language binding structure and provides a minimal API surface to use them.

It additionally commits a sample application to test the bindings.

The plan:

- attempt to merge this PR, keeping followin PRs small and focused;
- mockup most of the system;
- scale up with other devs after the first metric type is merged.

Things it does:

- Picks the `glean_ffi.dll` and allows to call glean-core functions;
- Defines the project structure;
- Initializes Glean and wires up Rust logging.
- Adds a sample c# application that uses nuGet to consume the library.

This PR will **NOT** cover:

- adding a proper logging system;
- adding a new metric type;
- initialize core metrics;
- properly initialize Glean;
- add test coverage for the basic functionalities;
- make sure it picks the right dll on macos

### DIFF
--- a/glean-core/csharp/csharp/Dispatchers.cs
+++ b/glean-core/csharp/csharp/Dispatchers.cs
@@ -1,0 +1,26 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using System;
+
+namespace Mozilla.Glean
+{
+    internal class Dispatchers
+    {
+        internal static void AssertInTestingMode()
+        {
+
+        }
+
+        internal static void LaunchAPI(Action action)
+        {
+            action();
+        }
+
+        internal static void FlushQueuedInitialTasks()
+        {
+
+        }
+    }
+}
--- a/glean-core/csharp/csharp/Glean.cs
+++ b/glean-core/csharp/csharp/Glean.cs
@@ -1,0 +1,176 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using Mozilla.Glean.FFI;
+using System;
+
+namespace Mozilla.Glean
+{
+    /// <summary>
+    /// The Glean Gneral API.
+    /// </summary>
+    public sealed class Glean
+    {
+        // Initialize the singleton using the `Lazy` facilities.
+        private static readonly Lazy<Glean>
+          lazy = new Lazy<Glean>(() => new Glean());
+        public static Glean GleanInstance => lazy.Value;
+
+        private Glean()
+        {
+            // Private constructor to disallow instantiation since
+            // this is meant to be a singleton. It only wires up the
+            // glean-core logging on the Rust side.
+            LibGleanFFI.glean_enable_logging();
+        }
+
+        /// <summary>
+        /// Initialize the Glean SDK.
+        /// 
+        /// This should only be initialized once by the application, and not by
+        /// libraries using the Glean SDK. A message is logged to error and no
+        /// changes are made to the state if initialize is called a more than
+        /// once.
+        /// 
+        /// This method must be called from the main thread.
+        /// </summary>
+        /// <param name="applicationId">The application id to use when sending pings.</param>
+        /// <param name="applicationVersion">The version of the application sending
+        /// Glean data.</param>
+        /// <param name="uploadEnabled">A `bool` that determines whether telemetry is enabled.
+        /// If disabled, all persisted metrics, events and queued pings (except first_run_date)
+        /// are cleared.</param>
+        /// <param name="configuration">A Glean `Configuration` object with global settings</param>
+        /// <param name="dataDir">The path to the Glean data directory.If not provided, uses
+        /// a temporary directory.</param>
+        public void Initialize(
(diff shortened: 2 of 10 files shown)


glean-core/csharp/csharp/Glean.cs
```diff
@@ -0,0 +1,177 @@
...
+using Mozilla.Glean.FFI;
+using System;
+using System.IO;
+
+namespace Mozilla.Glean
+{
+    /// <summary>
+    /// The Glean Gneral API.
+    /// </summary>
+    public class Glean
```
travis79, line 14: Let's make this `sealed` also to prevent it from being inherited from.

glean-core/csharp/csharp/Glean.cs
```diff
@@ -0,0 +1,177 @@
...
+        /// <summary>
+        /// Initialize the Glean SDK.
+        /// 
+        /// This should only be initialized once by the application, and not by
+        /// libraries using the Glean SDK. A message is logged to error and no
+        /// changes are made to the state if initialize is called a more than
+        /// once.
+        /// 
+        /// A LifecycleObserver will be added to send pings when the application goes
+        /// into foreground and background.
```
travis79, line 38: I realize this is a rough framework, but since we don't know this yet, maybe it's a good idea to remove this now before it gets left behind, or maybe update it with a TODO or something?

glean-core/csharp/csharp/LibGleanFFI.cs
```diff
@@ -0,0 +1,43 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using System;
+using System.Runtime.InteropServices;
```
travis79, line 6: Oh the "fun" to be had with this :D

glean-core/csharp/csharp/Glean.cs
Dexterp37: Good point!

glean-core/csharp/csharp/LibGleanFFI.cs
```diff
@@ -0,0 +1,65 @@
...
+using System.Text;
+
+namespace Mozilla.Glean.FFI
+{
+    static class LibGleanFFI
```
daoshengmu, line 11: We should add `sealed ` here.

glean-core/csharp/csharp/Dispatchers.cs
```diff
@@ -0,0 +1,21 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using System;
+using System.Collections.Generic;
+using System.Text;
```
daoshengmu, line 7: You don't need using `System.Text`  and `System.Collections.Generic`

glean-core/csharp/csharp/Glean.csproj
```diff
@@ -0,0 +1,23 @@
...
+    <RepositoryUrl>https://github.com/mozilla/glean</RepositoryUrl>
+    <Description>The Glean SDK is a modern approach for a telemetry library by Mozilla.</Description>
+    <Version>0.0.1</Version>
+    <RootNamespace>Mozilla.Glean</RootNamespace>
+  </PropertyGroup>
+
+  <ItemGroup>
+    <!-- Make sure to pack the glean-core DLL in the final package -->
+    <!-- TODO: pick the appropriate DLL depending on the target -->
+    <Content Include="../../../target/debug/glean_ffi.dll">
```
daoshengmu, line 16: My `glean_ffi.dll` can't be added via glean/Dependencies/Add Reference. I was thinking they only allow to import the dll is written by C#.

glean-core/csharp/csharp/LibGleanFFI.cs
Dexterp37: Unfortunately, we can't have both `static` and `sealed` on the same class. However, this was clearly lacking the `internal`, so I added it! Good catch!

glean-core/csharp/csharp/Dispatchers.cs
Dexterp37: Good point. Finalizing this will be part of [a separate bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1643011).

glean-core/csharp/csharp/Glean.csproj
Dexterp37: What do you mean?

daoshengmu: I guess <Pack> and <PackagePath> are hand-writing by yourself because I can't find a way to add them in MSVC. I think adding them looks right to me because we will pack them into NuGet package for sure.

Besides that, using `$(ProjectDir)` predefined macros of VS here would be more standard when including external files.

daoshengmu:
One thing makes me interested is I saw your source code are put at `glean-core/csharp/csharp/xxx.cs`. Is this path with two `csharp` is what you want?

glean-core/csharp/csharp/Glean.csproj
Dexterp37: Yes, this will be solved in an upcoming PR.

Dexterp37: I see. I will take care of this in [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1643564).

PR #937, opened 2020-06-04
Verify if *only* deletion request was sent on android sample app tests

brizental:
Fixes [Bug 1620673](https://bugzilla.mozilla.org/show_bug.cgi?id=1620673)

As an extra, I noticed the `validateBaselinePing` integration test was also failing and fixed it. It was a known issue of getting the `foreground` ping instead of the `background` one, which doesn't have a `duration` field and fails the test. This has been fixed in a-c and fenix already.

### DIFF
--- a/samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
+++ b/samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
@@ -43,13 +43,17 @@ class DeletionRequestPingTest {
         val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
         device.waitForIdle()
 
-        // Disable upload by toggline the switch
+        // Wait for any ping to make sure there are no pending requests before going forward
+        waitForPingContent("", server)
+
+        // Disable upload by toggling the switch
         onView(withId(R.id.uploadSwitch))
             .perform(closeSoftKeyboard())
             .perform(click())
 
-        // We might receive previous baseline or events ping, let's ignore that
-        val deletionPing = waitForPingContent("deletion-request", server)!!
+        // We must get the deletion request on the first attempt,
+        // no other ping should be sent after disabling upload
+        val deletionPing = waitForPingContent("deletion-request", server, 1)!!
 
         // Validate the received data.
 
--- a/samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
+++ b/samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
@@ -44,7 +44,7 @@ class DeletionRequestPingTest {
         device.waitForIdle()
 
         // Wait for any ping to make sure there are no pending requests before going forward
-        waitForPingContent("", server)
+        waitForPingContent("", null, server)
 
         // Disable upload by toggling the switch
         onView(withId(R.id.uploadSwitch))
@@ -53,7 +53,7 @@ class DeletionRequestPingTest {
 
         // We must get the deletion request on the first attempt,
         // no other ping should be sent after disabling upload
-        val deletionPing = waitForPingContent("deletion-request", server, 1)!!
+        val deletionPing = waitForPingContent("deletion-request", null, server, 1)!!
 
         // Validate the received data.
 
@@ -68,7 +68,7 @@ class DeletionRequestPingTest {
         device.pressHome()
 
         // Validate the received data.
-        val baselinePing = waitForPingContent("baseline", server)!!
+        val baselinePing = waitForPingContent("baseline", null, server)!!
 
         clientInfo = baselinePing.getJSONObject("client_info")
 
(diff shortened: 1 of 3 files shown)


samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/SharedTestUtils.kt
```diff
@@ -36,24 +64,30 @@ internal fun createMockWebServer(): MockWebServer {
...
         val docType = request.path.split("/")[3]
         if (pingName == docType) {
-            return if (request.getHeader("Content-Encoding") == "gzip") {
-                val bodyInBytes = ByteArrayInputStream(request.body.readByteArray()).readBytes()
-                JSONObject(decompressGZIP(bodyInBytes))
-            } else {
-                JSONObject(request.body.readUtf8())
+            parsedPayload = JSONObject(request.getPlainBody())
+            if (pingReason == null) {
+                break
```
brizental, line 81: I changed this because ktlint complains about the three returns in the same function.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
```diff
@@ -43,13 +43,17 @@ class DeletionRequestPingTest {
         val device = UiDevice.getInstance(InstrumentationRegistry.getInstrumentation())
         device.waitForIdle()
 
-        // Disable upload by toggline the switch
+        // Wait for any ping to make sure there are no pending requests before going forward
+        waitForPingContent("", server)
```
badboy: Do we expect to receive any ping here? If not will this cause a 20s delay because it waits for the timeout?

badboy: Also why can we use `""`? Doesn't that make the waitForPingcontent` spin 3 times, each with a 20s timeout?

brizental: > Do we expect to receive any ping here? If not will this cause a 20s delay because it waits for the timeout?

The comment we had previously in this stated that there could be events or metrics ping.

> Also why can we use ""? Doesn't that make the waitForPingcontent` spin 3 times, each with a 20s timeout?

I can remove the timeout time from this `takeRequest` function call. Also, I did this by mistake initially and then undid it, but I think it makes sense to just break out of the while loop in case `takeRequest` returns `null`.

brizental: This timeout parameter is optional actually. I can just remove it.

badboy: Ah, so if there is something, we get it, drop it, and then wait again for a potential next ping.
So if there are multiple pings getting sent we should be quick through the loop as we don't need to wait, otherwise we most likely wait once to not receive anything, then should drop out after the first timeout.

brizental: Yes, still 20s is a long wait time. Even if we only wait once.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/SharedTestUtils.kt
Dexterp37: You can also just suppress the warning if the code looks reasonably clearer the other way.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
Dexterp37: If 20s is a concern, we should test how low we can get. CI can be very slow.

brizental: Well, I changed that `continue` to a `break`. 

MockWebServer is just a LinkedBlokingQueue of requests. We wait in case there is some background task running that still hasn't finished and might send a ping request, correct?

There is a way to run WorkManager tasks synchronously for integrations tests (see: https://developer.android.com/topic/libraries/architecture/workmanager/how-to/integration-testing), which could be a solution. I am not sure if it is worth it though.

badboy:
Test failures are unrelated and can be ignored.

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/DeletionRequestPingTest.kt
Dexterp37: > There is a way to run WorkManager tasks synchronously for integrations tests (see: https://developer.android.com/topic/libraries/architecture/workmanager/how-to/integration-testing), which could be a solution. I am not sure if it is worth it though.

That's definitely helpful, but last time we tried it was riddled of bugs and basically not very useful at all :) Things have probably changed!

PR #939, opened 2020-06-04
Add a StringMetricType for C# [ci skip]

Dexterp37:
This PR adds basic support for the StringMetricType.

Note: [The Rust FFI Omnibus](http://jakegoulding.com/rust-ffi-omnibus/) was instrumental to get the C# part moving, along with the great prior work by @daoshengmu !

**TODO**

- [x] Test coverage for the metric type;
- [x] Remove the example code from `InitializeCoreMetrics`


glean-core/csharp/csharp/LibGleanFFI.cs
```diff
@@ -25,6 +25,57 @@ internal class FfiConfiguration
...
+
+            public override bool IsInvalid
+            {
+                get { return this.handle == IntPtr.Zero; }
+            }
+
+            public string AsString()
+            {
+                int len = 0;
+                while (Marshal.ReadByte(handle, len) != 0) { ++len; }
```
daoshengmu, line 61: it is not clear to me how you get this `handle`, and it is the same with `this.handle` from other functions?

glean-core/csharp/csharp/ManualMetrics.cs
```diff
@@ -0,0 +1,28 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using Mozilla.Glean.Private;
+using System;
+
+namespace Mozilla.Glean.GleanMetrics
+{
+    internal class GleanInternalMetricsOuter
```
daoshengmu, line 10: Here might be able to add a `sealed `

Dexterp37: Yes, you're right. This can be sealed.

glean-core/csharp/csharp/LibGleanFFI.cs
Dexterp37: The `SafeHandle` integrates with the C# P/Invoke, so can be safely used as a input or output (or return!) value. The framework will take care of passing things properly, additionally providing a GC-friendly way to clean up memory. See the the [official docs](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.safehandle?view=netcore-3.1), [this nice post](https://www.meziantou.net/stop-using-intptr-for-dealing-with-system-handles.htm) and how the [FFI Omnibus](http://jakegoulding.com/rust-ffi-omnibus/string_return/) talks about it.

PR #940, opened 2020-06-05
Move the c# code to the same dir as the solution file

Dexterp37:


Dexterp37:
Test failures are unrelated and I filed a bug for them.

PR #943, opened 2020-06-05
Update TWiG index

travis79:


travis79:
Darn, forgot the [doc only] tag :|

badboy:
I assume spellcheck doesn't like `gzipped`

travis79:
> I assume spellcheck doesn't like `gzipped`

Nope, I just added it to the dictionary

PR #948, opened 2020-06-08
 Create metric to record upload failures 

brizental:
Related to [Bug 1589124](https://bugzilla.mozilla.org/show_bug.cgi?id=1589124)

I am not sure adding all this logic to the `UploadResult` is the best idea. I saw the `error_recording` module, but that is very specific to the **metric** errors we may get. I also thought about adding this to each binding, but that would mean a lot of repeated code and also if a user decides to implement the uploader themselves they would have to know about this metric.

I also didn't add a label for timeout errors in this PR and I think we should discuss whether it should be added at all, since it would mean yet another different result variant for the `UploadResult` enum. If we decide that yes we should have a new Timeout variant I can add it in a different PR.


glean-core/src/upload/result.rs
```diff
@@ -64,3 +70,122 @@ impl From<u32> for UploadResult {
...
+            UploadResult::HttpStatus(400..=499) => Some("status_code_4xx"),
+            UploadResult::HttpStatus(500..=599) => Some("status_code_5xx"),
+            UploadResult::HttpStatus(_) => Some("status_code_unknown"),
+            UploadResult::UnrecoverableFailure => Some("unrecoverable"),
+            UploadResult::RecoverableFailure => Some("recoverable"),
+        }
+    }
+
+    fn get_metric(&self) -> Option<CounterMetric> {
+        if let Some(label) = self.get_metric_label() {
```
mdboom, line 87: It probably doesn't matter in a practical sense, but to make this work more like how labeled metrics work in the language bindings, I don't think we should create the metric "from scratch" like this each time.  Instead, lets add an object for `glean.upload.ping_upload_failure` to `internal_metrics.rs` and then call `new_metric_with_name` on it to get the specific labeled metric that we want.

glean-core/src/lib.rs
```diff
@@ -485,6 +485,7 @@ impl Glean {
     /// * `uuid` - The UUID of the ping in question.
     /// * `status` - The upload result.
     pub fn process_ping_upload_response(&self, uuid: &str, status: UploadResult) {
+        status.record_error(self);
```
badboy, line 488: Note-worthy (before I give this a full look): The Python bindings won't initialize a full Glean anymore in the upload process.
Therefore Python will be unable to report any of those errors
(As this code here is in `src/lib.rs` only, it will all continue to work and not crash or anything)

brizental: Yes I saw that and it is the reason why I put it in the Glean object. We don't want to record anything in case there is no Glean object, correct?

mdboom: Good catch, @badboy.  We could create a bug for that and not block this -- I still wonder if a good way around that would be to have a pure-Rust uploading binary (which would actually do a full init so it could record these things).

glean-core/src/upload/result.rs
brizental: Well, I made this changes and there is much less code to this PR now :D

glean-core/src/lib.rs
mdboom: Let's file a bug to add support for upload error recording to Python.

brizental: https://bugzilla.mozilla.org/show_bug.cgi?id=1644359

glean-core/metrics.yaml
```diff
@@ -320,3 +320,25 @@ glean.error:
     expires: never
     no_lint:
       - COMMON_PREFIX
+
+glean.upload:
+  ping_upload_failure:
+    type: labeled_counter
+    description:
+      Counts the number of times pings fail to upload
+      and the reason why they failed.
```
badboy, line 329: Maybe it's worth mentioning that (obviously) these numbers are reported after the fact?
That is: if a metrics ping upload fails and gets retried it still won't have this in there, only the next one will.

glean-core/src/lib.rs
```diff
@@ -485,6 +485,11 @@ impl Glean {
     /// * `uuid` - The UUID of the ping in question.
     /// * `status` - The upload result.
     pub fn process_ping_upload_response(&self, uuid: &str, status: UploadResult) {
+        if let Some(label) = status.get_label() {
+            let metric = self.core_metrics.ping_upload_failure.get(label);
```
badboy, line 489: So clean! :)

glean-core/src/metrics/labeled.rs
```diff
@@ -127,7 +127,7 @@ where
     ///
     /// If the requested label is in the list of allowed labels, it is returned.
     /// Otherwise the `OTHER_LABEL` is returned.
-    fn static_label<'a>(&mut self, label: &'a str) -> &'a str {
+    fn static_label<'a>(&self, label: &'a str) -> &'a str {
```
badboy, line 130: Oh, did this come up through a lint now that these don't have to me mutable anymore?

glean-core/src/upload/result.rs
```diff
@@ -64,3 +64,18 @@ impl From<u32> for UploadResult {
         }
     }
 }
+
+impl UploadResult {
+    /// Failures are recorded in the `ping_upload_failure` metric.
```
badboy, line 69: The first sentence should describe what this function does.

E.g.: 

```suggestion
    /// Get the label to be used in recording error counts for upload.
    ///
    /// Returns `None` if the upload finished succesfully.
    /// Failures are recorded in the `ping_upload_failure` metric.
```

glean-core/src/metrics/labeled.rs
brizental: Not through a lint, before the changes suggested by @mdboom I noticed it.

glean-core/metrics.yaml
brizental: Maybe it could be confusing mention this, because this metric will count the failures for all ping uploads, not only the metrics ping. It could be misleading to mention this situation
> if a metrics ping upload fails and gets retried it still won't have this in there, only the next one will.

because it only applies for the metrics ping.

badboy: Hm, @mdboom, got an idea how to phrase this?
I'd like us to start being a bit more explicit of when and where to expect this metric

mdboom: It's not great -- this is the best I could come up with.  What do you think?

```
Counts the number of ping upload failures, by type of failure.
This includes failures for all ping types, though the counts appear in the next successfully sent `metrics` ping.
```

brizental:
This android failure on CI is probably unrelated since it failed on a commit where I only changed description text in the `metrics.yaml`.

badboy:
> This android failure on CI is probably unrelated since it failed on a commit where I only changed description text in the `metrics.yaml`.

It's also only on TaskCluster, can be ignored for now.

PR #949, opened 2020-06-08
Bug 1637609: Hand-roll uses of regex library

mdboom:


glean-core/src/metrics/labeled.rs
```diff
@@ -32,8 +39,36 @@ const MAX_LABEL_LENGTH: usize = 61;
...
+                None => return true,
+                // Valid characters.
+                Some('_') | Some('-') | Some('a'..='z') | Some('0'..='9') => (),
+                // We ended a word, so iterate over the outer loop again.
+                Some('.') => break,
+                // An invalid character
+                _ => return false,
+            }
+            count += 1;
+            if count == 29 {
```
travis79, line 66: Should this be if count == 30 since we start counting at 0 and allow for 30 characters?

mdboom: The first character (since it's handled differently) is outside of the loop, so there's only 29 of these characters allowed.  Also see regex itself in the comment above.  Worth mentioning in a comment, for sure.

travis79: Nevermind, you already peeled off the first character.

travis79: jinx :)

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v31.0.2...master)
 
+* General:
+    * The `regex` crate is no longer required, making the Glean release binary
+      about 0.5MB smaller.
```
badboy, line 7: I would be a bit more careful of the phrasing here. We deliver to so many different platforms that a blank size statement might be misleading.

mdboom: Good point.  I changed this to just "smaller"

PR #953, opened 2020-06-09
Add maximum number of retries to PingUploadWorker

brizental:
Fixes [Bug 1644290](https://bugzilla.mozilla.org/show_bug.cgi?id=1644290)

This is a temporary fix so we can stop [this issue](https://github.com/mozilla-mobile/fenix/issues/11335). 

I opened a bug to move this retrying logic to the Rust side: https://bugzilla.mozilla.org/show_bug.cgi?id=1644364

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -123,10 +123,9 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                         Glean.configuration
                     ).toFfi()
 
-                    if (result == Constants.UPLOAD_RESULT_RECOVERABLE ||
-                        result == Constants.UPLOAD_RESULT_UNRECOVERABLE) {
-                            uploadFailures++
-                        }
+                    if (result == Constants.UPLOAD_RESULT_RECOVERABLE) {
+                        uploadFailures++
+                    }
 
                     // Process the upload response
                     LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -20,6 +20,7 @@ import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.net.FfiPingUploadTask
 import mozilla.telemetry.glean.utils.testFlushWorkManagerJob
 import mozilla.telemetry.glean.net.PingUploadTask
+import mozilla.telemetry.glean.rust.Constants
 
 /**
  * Build the constraints around which the worker can be run, such as whether network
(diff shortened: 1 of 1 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -112,14 +121,20 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
...
                     // Process the upload response
-                    LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result.toFfi())
+                    LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
                 }
                 PingUploadTask.Wait -> return Result.retry()
                 PingUploadTask.Done -> return Result.success()
             }
-        } while (true)
+        } while (uploadFailures < MAX_RETRIES)
+        return Result.failure()
```
brizental, line 138: Unsure if this should really be a `failure` or a `retry`.

badboy: We use a `OneTimeWorkRequest` which defaults to exponential backoff on `retry`.
I think `failure` is the right thing here. If we fail too often, we don't want to reschedule it at all. When a ping gets submitted again (e.g. next time the app is backgrounded) any pending pings are picked up again anyway.

badboy: cc @Dexterp37 just for a quick look

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -112,14 +121,20 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                         action.request.body,
                         action.request.headers,
                         Glean.configuration
-                    )
+                    ).toFfi()
+
+                    if (result == Constants.UPLOAD_RESULT_RECOVERABLE ||
+                        result == Constants.UPLOAD_RESULT_UNRECOVERABLE) {
```
badboy, line 127: A thing I only notice now:

* We know that `unrecoverable` failures will lead to the file being deleted, we would only need to count the recoverable errors here.
  * HTTP status 400-499 are also unrecoverable, at least internally to the Rust part. Again, we don't need to care, the ping is deleted and not retried.
* android-components never uses `UPLOAD_RESULT_UNRECOVERABLE` anyway (there we catch all I/O errors as `recoverable`

If we do fail on I/O in Fenix we correctly catch that here with the `recoverable` part.

mdboom:
Should we file a follow up bug for Swift and Python?

brizental:
> Should we file a follow up bug for Swift and Python?

https://bugzilla.mozilla.org/show_bug.cgi?id=1644419

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
Dexterp37: Works for me, assuming failure doesn't mean deleting the pings

badboy: Nope, it's only marking the worker as failed. The pings are handled inside glean-core and will be retried by the next worker.

PR #954, opened 2020-06-09
Bug 1643011 - Implement a dispatcher for C#

travis79:


glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,24 +3,213 @@
...
+        /// because the only tasks that should be launched by this are the ping
+        /// upload schedulers and those should run regardless of the initialized
+        /// state.
+        ///
+        /// If `TestingMode` is enabled, then `LaunchConcurrent` will just
+        /// execute the task rather than adding it to the concurrent queue to
+        /// avoid asynchronous issues while testing.
+        /// </summary>
+        /// <param name="action">The Action to perform</param>
+        internal static void LaunchConcurrent(Action action)
```
Dexterp37, line 111: I find the `LaunchConcurrent` naming a bit confusing (I know that's how it works in Swift): isn't `LaunchAPI` concurrent as well?

The difference we have on the Kotlin side is this: `launch` adds the task at the end of the queue, while `execute` runs the task _as soon as possible_, right after the current executing task (if any). The executor is the same for both.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,24 +3,213 @@
...
 namespace Mozilla.Glean
 {
     internal class Dispatchers
     {
-        internal static void AssertInTestingMode()
-        {
+        private const string LogTag = "glean/Dispatchers";
+        internal const int MaxQueueSize = 100;
+
+        internal static bool TestingMode { get; set; } = false;
```
Dexterp37, line 19: Can we retain the docs from the other bindings?

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,24 +3,213 @@
...
+        private static int _queueInitialTasks = 1;
+        internal static bool QueueInitialTasks {
+            get => _queueInitialTasks == 1;
+            set => Interlocked.Exchange(ref _queueInitialTasks, value ? 1 : 0);
         }
 
+        // Create a ConcurrentExclusiveSchedulerPair object.
+        // All API tasks will be executed on the exclusive part of the scheduler
+        // in order to ensure single threaded behavior
+        // The writer will run on the exclusive part of the scheduler pair.
```
Dexterp37, line 30: What's the writer?

glean-core/csharp/Glean/Dispatchers.cs
Dexterp37: (also port over all the comments that apply to other entries, too)

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,24 +3,213 @@
...
+            {
+                new Task(action, cancellationToken)
+                    .Start(taskSchedulerPair.ConcurrentScheduler);
+            }
+        }
+
+        /// <summary>
+        /// Cancels any pending background tasks.
+        /// </summary>
+        internal static void CancelBackgroundTasks()
```
Dexterp37, line 127: What's this for?

glean-core/csharp/Glean/Glean.csproj
```diff
@@ -120,4 +120,11 @@
       <CopyToOutputDirectory>Always</CopyToOutputDirectory>
     </Content>
   </ItemGroup>
+
+  <!-- Expose internal access objects to unit tests -->
+  <ItemGroup>
```
Dexterp37, line 125: Since you added this, can you remove what I added at the top of `Glean.cs`?

glean-core/csharp/GleanTests/DispatchersTests.cs
```diff
@@ -0,0 +1,38 @@
...
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using Xunit;
+
+namespace Mozilla.Glean.Tests
+{
+    public class DispatchersTests
+    {
+        [Fact]
+        public async void TestTaskQueueing()
```
Dexterp37, line 12: Are you planning on porting over all the tests from Kotlin?

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,24 +3,213 @@
...
 using System;
+using System.Threading;
+using System.Threading.Tasks;
+using System.Diagnostics;
+using System.Collections.Generic;
+using System.Runtime.CompilerServices;
 
 namespace Mozilla.Glean
 {
     internal class Dispatchers
```
Dexterp37, line 14: I see you're making this `static` heavy. No concern on this, just a different paradigm with respect to the Lazy patterns used elsewhere :)

glean-core/csharp/Glean/Dispatchers.cs
travis79: Oops, copy-pasta from a MS example

glean-core/csharp/Glean/Dispatchers.cs
travis79: In this case `LaunchAPI` uses a serially executed `ThreadPool` (single threaded, executes sequentially), while `LaunchConcurrent` uses a concurrently executed `ThreadPool` (multi-threaded, tasks run concurrently).  I was a little torn on implementing this, I originally started patterning it after the Kotlin implementation when I discovered the `ConcurrentExclusiveSchedulerPair` provided both types of execution in a single object, so I ended up going more towards the Swift implementation

glean-core/csharp/Glean/Dispatchers.cs
travis79: We cancel background tasks when we toggle upload enabled to false, this is just a way to cancel all of the tasks running in the Glean background queues.  This is similar to what we are doing when we [cancel the `PingUploadWorker` in Kotlin](https://github.com/mozilla/glean/blob/5ecdd24b1dc92539903bf2246cdc0aa3a56954d9/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L274) and [cancelling the background tasks in Swift](https://github.com/mozilla/glean/blob/5ecdd24b1dc92539903bf2246cdc0aa3a56954d9/glean-core/ios/Glean/Glean.swift#L222).  In Kotlin, there just wasn't a good way to cancel a coroutine, if I remember correctly.

glean-core/csharp/GleanTests/DispatchersTests.cs
travis79: Yes, I only had one test complete when I asked you to take a look, but the plan is to have the exact same coverage and tests.

glean-core/csharp/Glean/Dispatchers.cs
travis79: I was also a little worried about this, but the option would be to use a Singleton pattern like the base Glean object.  If static ends up becoming a problem for us, I'm okay with changing it out.

glean-core/csharp/Glean/Dispatchers.cs
Dexterp37: My concern is that this would end up working weirdly around startup: in kotlin, since we use the same pool of threads, init blocks the pool until after it finishes.

travis79: These are separate ThreadPool's so I think it should be okay.  Let's see how the tests work out and go from there.

glean-core/csharp/GleanTests/DispatchersTests.cs
```diff
@@ -0,0 +1,123 @@
...
+using System.Collections.Generic;
+using System.Threading;
+using System.Threading.Tasks;
+using Xunit;
+
+namespace Mozilla.Glean.Tests
+{
+    public class DispatchersTests
+    {
+        //public DispatchersTests()
```
Dexterp37, line 15: Do we need this?

glean-core/csharp/GleanTests/DispatchersTests.cs
```diff
@@ -0,0 +1,123 @@
...
+            while (counter < 99) { Thread.Sleep(1); }
+
+            for (int num = 0; num < 99; num++)
+            {
+                Assert.Equal(num, orderedList[num]);
+            }
+        }
+
+        [Fact]
+        public void TestCancellingBackgroundTasksClearsQueue()
```
Dexterp37, line 96: There's a bunch of tests that are in Kotlin but still missing here. Are you planning on implementing them?

travis79: Yes, it was my EoD and I just pushed what I had...  I'm close enough now that I'll just flag you once I promote this from a Draft so you aren't looking at WIP stuff :)

glean-core/csharp/GleanTests/DispatchersTests.cs
travis79: WIP stuff, sorry :)

PR #964, opened 2020-06-10
Run C# tests on CI

Dexterp37:


.circleci/config.yml
```diff
@@ -984,6 +984,33 @@ jobs:
             # Upload to GitHub
             ./ghr -replace ${CIRCLE_TAG} glean-core/python/dist
 
+  ###########################################################################
+  # C#
```
badboy, line 988: C########################

.circleci/config.yml
```diff
@@ -984,6 +984,33 @@ jobs:
...
+
+  CSharp tests:
+    docker:
+      - image: mcr.microsoft.com/dotnet/core/sdk:3.1-buster
```
badboy, line 992: Just to understand: this is an actual Linux image, just with all the dotnet tools in there?

.circleci/config.yml
```diff
@@ -984,6 +984,33 @@ jobs:
...
+      - setup-rust-toolchain:
+          rust-version: "1.43.0"
+      - run:
+          name: Install the dependencies
+          command: |
+            apt update
+            apt install -y gcc
+      - run:
+          name: Build glean_ffi
+          command: cargo build
```
badboy, line 1006: It's not automatically build as part of the dotnet buildstep? :(

.circleci/config.yml
```diff
@@ -984,6 +984,33 @@ jobs:
...
+
+  CSharp tests:
+    docker:
+      - image: mcr.microsoft.com/dotnet/core/sdk:3.1-buster
+    steps:
+      - checkout
+      - skip-if-doc-only
+      - install-rustup
+      - setup-rust-toolchain:
+          rust-version: "1.43.0"
```
badboy, line 998: does it also fail with 1.44.0 as the other windows target?

.circleci/config.yml
Dexterp37: Yes! This is a Debian image to be precise. The list of images is [available here](https://hub.docker.com/_/microsoft-dotnet-core-sdk/).

.circleci/config.yml
Dexterp37: No, not yet :( We'll cross that bridge after we get out of the "MOVEEEEEE FASTTTTTTTTTT" 🐎 🐎  phase :)

.circleci/config.yml
Dexterp37: I didn't try, I cargo culted that part. Should this be `stable`?

Dexterp37:
> 
> 
> Looks good to me, surprisingly simple these days.
> 
>     1. We should probably make that task a blocking CI task here on GitHub

How can this be done?

>     2. Does it generate any artifacts we should expose? (e.g. test report, generated package) Of course that can be done in a follow-up.

We don't generate anything useful at the moment, as we didn't enable useful logging yet. Do you mean something like [this](https://circleci.com/docs/2.0/collect-test-data/#trx2junit-for-visual-studio--net-core-tests) ?

PR #960, opened 2020-06-10
Adding C# CustomPing.

daoshengmu:
It looks very close to be done and is blocked by `Glean.SubmitPings` and `Glean.RegisterPingType`.  I  also need to figure out how to verify success/failed results in unit tests.

PR #965, opened 2020-06-10
Adding C# CustomPing.

badboy:
It looks very close to be done and is blocked by `Glean.SubmitPings` and `Glean.RegisterPingType`.  I  also need to figure out how to verify success/failed results in unit tests.

--

re-open of #960 -- that was closed by accident. Original by @daoshengmu. The other PR already got some review comment. Unfortunately I can't restore them.


glean-core/csharp/GleanTests/Metrics/PingTypeTest.cs
```diff
@@ -0,0 +1,91 @@
...
+        {
+            // TODO:
+        }
+
+        [Fact]
+        public void RegistryShouldContainBuildInPings()
+        {
+            // Blow lines return false always, how to I register
+            // build-in pings?
+            //Assert.True(GleanInstance.TestHasPingType("metrics"));
```
daoshengmu, line 86: If I register pings and their names are `metrics`, `baseline`, `event` before calling below lines. These tests will return true. However, I feel these three types pings are build-in pings. We should register them internally, but I don't know where. @Dexterp37

glean-core/csharp/Glean/Glean.cs
```diff
@@ -5,7 +5,11 @@
 using Mozilla.Glean.FFI;
 using static Mozilla.Glean.GleanMetrics.GleanInternalMetricsOuter;
 using System;
+using System.Runtime.CompilerServices;
+using Mozilla.Glean.Private;
+using System.Collections.Generic;
 
+[assembly: InternalsVisibleTo("GleanTests")]
```
Dexterp37, line 12: You can remove this: we now have a project level statement that enables it (rebase on latest `main`):

glean-core/csharp/GleanTests/Metrics/PingTypeTest.cs
Dexterp37: Good point. The pings code is usually generated automatically by the build system, using the `glean_parser`, when building the Glean SDK. Unfortunately, we don't have that part yet, so we need to generate a file manually, as we do for our [core metrics](https://searchfox.org/glean/source/glean-core/csharp/Glean/ManualMetrics.cs). You need to create a `ManualPings.cs` that defines the pings.

For example, [here's the content](https://dpaste.org/RP2o) of the `Pings.kt` file generated by the Glean build system when parsing the SDK.

Ping registration can then happen [here](https://searchfox.org/glean/rev/bf1825a7523576911c46eb770912630693228a0e/glean-core/csharp/Glean/Glean.cs#92), by passing in the manual pings class.

daoshengmu: I have made a `ManualPings.cs` and was trying to lazy create these `PingTypes` as what you did in `ManualMetrics.cs`. However, when executing running all tests, we will call `Glean.TestDestroyGleanHandle()` to destroy Glean between tests. That makes me have to re-create these build-in ping types after every time we initialize Glean. For the case of running a single test, we can just lazy create them.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -104,6 +105,9 @@ string dataDir
                     return;
                 }
 
+                // Registering pings needs to be after Glean initialization.
+                RegisterPings();
+
                 /* TODO:
                 // If any pings were registered before initializing, do so now.
                 // We're not clearing this queue in case Glean is reset by tests.
```
Dexterp37, line 113: This is the part that is required to make the ping type work in the tests.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -104,6 +105,9 @@ string dataDir
                     return;
                 }
 
+                // Registering pings needs to be after Glean initialization.
+                RegisterPings();
```
Dexterp37, line 109: It shouldn't: the way we use the dispatchers should guarantee the proper execution model. Can you kindly try to move it back where it was? :) Maybe this is not even needed in this language.

glean-core/csharp/Glean/ManualPings.cs
```diff
@@ -0,0 +1,82 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using Mozilla.Glean.Private;
+using System;
+
+namespace Mozilla.Glean.GleanPings
+{
+    internal sealed class GleanInternalPingsOuter
```
Dexterp37, line 10: By looking at the [Koltin generated code](https://dpaste.org/RP2o), we don't need/want ping definitions to be lazily initialized.

Dexterp37, line 10:
```suggestion
    internal sealed class GleanInternalPings
```

glean-core/csharp/Glean/ManualPings.cs
```diff
@@ -0,0 +1,82 @@
...
+    {
+        // Initialize the singleton using the `Lazy` facilities.
+        private static readonly Lazy<GleanInternalPingsOuter>
+          lazy = new Lazy<GleanInternalPingsOuter>(() => new GleanInternalPingsOuter());
+        public static GleanInternalPingsOuter GleanInternalPings => lazy.Value;
```
Dexterp37, line 15:
```suggestion
```

glean-core/csharp/Glean/ManualPings.cs
```diff
@@ -0,0 +1,82 @@
...
+        }
+
+        internal enum EventsReasonCodes : int
+        {
+            background,
+            maxCapacity,
+            startup
+        }
+
+        private PingType<BaselineReasonCodes> baseline = null;
```
Dexterp37, line 43: You can make these types static and initialize them inline.

glean-core/csharp/Glean/ManualPings.cs
```diff
@@ -0,0 +1,82 @@
...
+        private PingType<MetricsReasonCodes> metrics = null;
+        private PingType<EventsReasonCodes> events = null;
+        private PingType<NoReasonCodes> deletionRequest = null;
+
+        public void RegisterPings()
```
Dexterp37, line 48: You don't need a `RegisterPings` here.

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,57 @@
...
+    /// <summary>
+    ///  An enum with no values for convenient use as the default set of reason codes.
+    /// </summary>
+    public enum NoReasonCodes : int
+    {
+    }
+
+    public class PingTypeBase
+    {
+        internal string name;
```
Dexterp37, line 20: Can you add a private constructor to make this class not instantiable on its own?

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,57 @@
...
+        internal LibGleanFFI.PingTypeHandle handle;
+    }
+
+    public class PingType<NoReasonCodes> : PingTypeBase
```
Dexterp37, line 24: Can you port over all the docs from the Kotlin implementation?

glean-core/csharp/GleanTests/Metrics/PingTypeTest.cs
```diff
@@ -0,0 +1,88 @@
+﻿using System.IO;
```
Dexterp37, line 1: This is missing the license header.

glean-core/csharp/GleanTests/Metrics/PingTypeTest.cs
```diff
@@ -0,0 +1,88 @@
...
+                sendInPings: new string[] { "custom" }
+            );
+
+            // Record two strings of the same type, with a little delay.
+            booleanMetric.Set(true);
+
+            // Check that data was properly recorded.
+            Assert.True(booleanMetric.TestHasValue());
+            customPing.Submit();
+        }
```
Dexterp37, line 54: Can you add a comment and a link to the rest of the code of this test (from Koltin) , with a TODO? So that we know what to implement when the time comes.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -311,5 +320,35 @@ private void InitializeCoreMetrics()
...
+                );
+            }
+
+            // We need to keep track of pings, so they get re-registered after a reset.
+            // This state is kept across Glean resets, which should only ever happen in test mode.
+            // Or by the instrumentation tests (`connectedAndroidTest`), which relaunches the application activity,
+            // but not the whole process, meaning globals, such as the ping types, still exist from the old run.
+            // It's a set and keeping them around forever should not have much of an impact.
+
+            //pingTypeQueue.Add(pingType);
```
Dexterp37, line 342: That's part of the solution to the problem with the type not surviving the tests :)

glean-core/csharp/Glean/Metrics/PingType.cs
daoshengmu: We can make it as an `internal` constructor at least because it is inherited by other public classes.

glean-core/csharp/Glean/Glean.cs
daoshengmu: Yep.

glean-core/csharp/Glean/Glean.cs
daoshengmu: I think I get the sense how to make it work. It looks like I can't just make `GleanInternalPings` to be a static class and its pings are static inline.

Because I need a place for `GleanInternalPings` to instance its attributes. Therefore, I still choose to make 'GleanInternalPings' as a lazy instantiation singleton. Currently, I only make its pings are inline. Let me know if you feel lazy instance them is better.

glean-core/csharp/Glean/Metrics/PingType.cs
Dexterp37: Oh, apparently we can further restrict this by making it `protected internal` (https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/keywords/protected-internal)

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        internal PingTypeBase() { }
+    }
+
+    /// <summary>
+    ///  This implements the developer facing API for custom pings.
+    ///  Instances of this class type are automatically generated by the parsers at build time.
+    ///  The Ping API only exposes the [send] method, which schedules a ping for sending.
+    /// </summary>
+    public class PingType<NoReasonCodes> : PingTypeBase
```
Dexterp37, line 36:
```suggestion
    public class PingType<ReasonCodesEnum> : PingTypeBase where ReasonCodesEnum: System.Enum, struct
```

The intent here is for `PingType` to declare the type for the Generic class and also add a constraint to it: we only want enums to be valid generics arguments.

**Note**: `struct` is needed in order to allow for passing a nullable enum to `Submit`

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,78 @@
...
+        /// <summary>
+        ///  Collect and submit the ping for eventual upload.
+        ///  
+        ///  While the collection of metrics into pings happens synchronously, the
+        ///  ping queuing and ping uploading happens asyncronously.
+        ///  There are no guarantees that this will happen immediately.
+        ///  
+        ///  If the ping currently contains no content, it will not be queued.
+        /// </summary>
+        public void Submit(int reasonCode)
```
Dexterp37, line 72:
```suggestion
        public void Submit(ReasonCodesEnum? reasonCode = null)
```

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,78 @@
...
+        }
+
+        /// <summary>
+        ///  Collect and submit the ping for eventual upload.
+        ///  
+        ///  While the collection of metrics into pings happens synchronously, the
+        ///  ping queuing and ping uploading happens asyncronously.
+        ///  There are no guarantees that this will happen immediately.
+        ///  
+        ///  If the ping currently contains no content, it will not be queued.
```
Dexterp37, line 70: Please document the parameters.

glean-core/csharp/GleanTests/GleanTests.cs
```diff
@@ -44,5 +44,11 @@ public void GettingUploadEnabledBeforeInitShouldNotCrash()
                 dataDir: tempDataDir
                 );
         }
+
+        [Fact]
+        public void SendAPing()
```
Dexterp37, line 49: Let's remove this test please, since this is incomplete, or add a TODO comment stating that this is incomplete and the whole test needs to be ported over.

glean-core/csharp/GleanTests/Metrics/PingTypeTest.cs
```diff
@@ -0,0 +1,95 @@
...
+                uploadEnabled: true,
+                configuration: new Configuration(),
+                dataDir: tempDataDir
+                );
+        }
+
+        [Fact]
+        public void SendCustomPings()
+        {
+            Private.PingType<Private.NoReasonCodes> customPing =
```
Dexterp37, line 32: Please make this test empty, as the other tests, since this is actually not testing the ping as the original one.

glean-core/csharp/Glean/ManualPings.cs
Dexterp37: Let's not forget to rename this

glean-core/csharp/Glean/ManualPings.cs
Dexterp37: And to remove this

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -240,6 +240,45 @@ bool disabled
...
+            byte send_if_empty,
+            string[] reason,
+            Int32 reason_codes_len
+        );
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void glean_destroy_ping_type(IntPtr handle);
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern StringAsReturnValue glean_ping_collect(PingTypeHandle ping_type_handle, string reason);
```
Dexterp37, line 274: Sorry to be nitpicky on this, but would you kindly move this function to the same position it has in the Kotlin implementation [here](https://searchfox.org/glean/rev/cfabeff6ba9c3562332dbe8640f16f26e99f0a46/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt#112)?

This belongs to the Glean General API, not to the Ping Metric type specific API.

Also, please note that `glean_submit_ping_by_name` is missing :)

glean-core/csharp/Glean/ManualPings.cs
Dexterp37: Sorry, I was wrong. Let's ignore this.

glean-core/csharp/Glean/ManualPings.cs
Dexterp37: Sorry, I was wrong. Let's ignore this.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -23,6 +27,8 @@ public sealed class Glean
         // Keep track of this flag before Glean is initialized
         private bool uploadEnabled = true;
 
+        private HashSet<PingTypeBase> pingTypeQueue = new HashSet<PingTypeBase>();
```
Dexterp37, line 30:
```suggestion
        // Keep track of ping types that have been registered before Glean is initialized.
        private HashSet<PingTypeBase> pingTypeQueue = new HashSet<PingTypeBase>();
```

glean-core/csharp/Glean/Glean.cs
```diff
@@ -104,13 +110,15 @@ string dataDir
                     return;
                 }
 
-                /* TODO:
-                // If any pings were registered before initializing, do so now.
-                // We're not clearing this queue in case Glean is reset by tests.
-                synchronized(this@GleanInternalAPI) {
-                    pingTypeQueue.forEach { registerPingType(it) }
+                //// If any pings were registered before initializing, do so now.
+                //// We're not clearing this queue in case Glean is reset by tests.
```
Dexterp37, line 114:
```suggestion
                // If any pings were registered before initializing, do so now.
                // We're not clearing this queue in case Glean is reset by tests.
```

glean-core/csharp/Glean/Glean.cs
```diff
@@ -104,13 +110,15 @@ string dataDir
...
+                lock (pingTypeQueue)
```
Dexterp37, line 115:
```suggestion
                lock (this)
```

This should lock using the same object as the `[MethodImpl(MethodImplOptions.Synchronized)]` in the methods, otherwise we'll have races.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -311,5 +319,57 @@ private void InitializeCoreMetrics()
...
+        /// <summary>
+        /// Register the pings generated from `ManualPings` with the Glean SDK.
+        /// </summary>
+        /// <param name="pings"> The `Pings` object generated for your library or application
+        /// by the Glean SDK.</param>
+        private void RegisterPings(object pings)
+        {
+            // Instantiating the Pings object to send this function is enough to
+            // call the constructor and have it registered through [Glean.registerPingType].
+            Console.WriteLine("Registering pings");
```
Dexterp37, line 332: Did you check if, when getting in the body of this function, the `GleanInternalPings` get lazily instantiated? (that's the whole point of this function)

glean-core/csharp/Glean/Glean.cs
```diff
@@ -85,7 +91,7 @@ string dataDir
             SetUploadEnabled(uploadEnabled);
```
Dexterp37, line 91: You should also add the `[MethodImpl(MethodImplOptions.Synchronized)]` annotation to the `Initialize` method for this to work safely.

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -158,7 +164,7 @@ bool disabled
              StringMetricTypeHandle metric_id,
              Int32 error_type,
              string storage_name
-         );
+        );
```
Dexterp37, line 167:
```suggestion
        );
```

glean-core/csharp/Glean/Metrics/PingType.cs
```diff
@@ -0,0 +1,79 @@
...
+        /// </summary>
+        /// <param name="reasonCode">The reason code enum for ping submit.</param>
+        public void Submit(ReasonCodesEnum? reasonCode = null)
+        {
+            // TODO: submit pings
+            //GleanInstance.SubmitPings()
```
Dexterp37, line 76: I can implement this in my PR, since it kind of need the uploader anyway. Ok?

daoshengmu: yep.

glean-core/csharp/Glean/Glean.cs
daoshengmu: yes. I checked.

PR #969, opened 2020-06-10
Bug 1643571 - Add C# Logger

travis79:
- Add Logger wrapper around Serilog
- Update pertinent TODO's to use C# Logger

### DIFF
--- a/glean-core/csharp/Glean/LibGleanFFI.cs
+++ b/glean-core/csharp/Glean/LibGleanFFI.cs
@@ -124,7 +124,6 @@ namespace Mozilla.Glean.FFI
             {
                 if (!this.IsInvalid)
                 {
-                    Console.WriteLine("Freeing string metric type handle");
                     glean_destroy_string_metric(handle);
                 }
 
@@ -160,6 +159,46 @@ namespace Mozilla.Glean.FFI
              Int32 error_type,
              string storage_name
          );
+
+        // Boolean
+
+        /// <summary>
+        /// A handle for the boolean metric type, which performs cleanup.
+        /// </summary>
+        internal sealed class BooleanMetricTypeHandle : BaseGleanHandle
+        {
+            protected override bool ReleaseHandle()
+            {
+                if (!this.IsInvalid)
+                {
+                    glean_destroy_boolean_metric(handle);
+                }
+
+                return true;
+            }
+        }
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern BooleanMetricTypeHandle glean_new_boolean_metric(
+            string category,
+            string name,
+            string[] send_in_pings,
+            Int32 send_in_pings_len,
+            Int32 lifetime,
+            bool disabled
+        );
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void glean_destroy_boolean_metric(IntPtr handle);
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern void glean_boolean_set(BooleanMetricTypeHandle metric_id, byte value);
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern byte glean_boolean_test_get_value(BooleanMetricTypeHandle metric_id, string storage_name);
+
+        [DllImport(SharedGleanLibrary, ExactSpelling = true, CallingConvention = CallingConvention.Cdecl)]
+        internal static extern byte glean_boolean_test_has_value(BooleanMetricTypeHandle metric_id, string storage_name);
 
         // Misc
 
--- a/glean-core/csharp/Glean/Metrics/BooleanMetricType.cs
+++ b/glean-core/csharp/Glean/Metrics/BooleanMetricType.cs
@@ -1,0 +1,108 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using Mozilla.Glean.FFI;
+using System;
+
+namespace Mozilla.Glean.Private
+{
+    public sealed class BooleanMetricType
+    {
+        private bool disabled;
+        private string[] sendInPings;
+        private LibGleanFFI.BooleanMetricTypeHandle handle;
+
+        public BooleanMetricType(
+            bool disabled,
+            string category,
+            Lifetime lifetime,
+            string name,
+            string[] sendInPings
+            ) : this(new LibGleanFFI.BooleanMetricTypeHandle(), disabled, sendInPings)
+        {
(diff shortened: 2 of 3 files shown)


glean-core/csharp/Glean/Glean.cs
```diff
@@ -70,18 +76,21 @@ string dataDir
                 Log.e(LOG_TAG, "Attempted to initialize Glean on a process other than the main process")
                 return
             }
+            */
 
-            if (isInitialized()) {
-                Log.e(LOG_TAG, "Glean should not be initialized multiple times")
-                return
+            if (IsInitialized()) {
+
```
Dexterp37, line 82: nit: bogus blank line

glean-core/csharp/Glean/Glean.cs
```diff
@@ -204,13 +213,9 @@ public void SetUploadEnabled(bool enabled)
...
                 {
-                    // Cancel any pending workers here so that we don't accidentally upload or
+                    // Cancel any pending tasks here so that we don't accidentally upload or
                     // collect data after the upload has been disabled.
-                    // TODO: metricsPingScheduler.cancel()
-
-                    // Cancel any pending workers here so that we don't accidentally upload
-                    // data after the upload has been disabled.
-                    // TODO: PingUploadWorker.cancel(applicationContext)
+                    Dispatchers.CancelBackgroundTasks();
```
Dexterp37, line 218: Let's leave this as it is for now. This will be taken care in the future, when the metrics ping scheduler is there.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -204,13 +213,9 @@ public void SetUploadEnabled(bool enabled)
 
                 if (!enabled)
                 {
-                    // Cancel any pending workers here so that we don't accidentally upload or
+                    // Cancel any pending tasks here so that we don't accidentally upload or
```
Dexterp37, line 216: Let's revert this for now

glean-core/csharp/Glean/Utils/Logger.cs
```diff
@@ -0,0 +1,66 @@
...
+            var format =
+                string.Format(
+                    "[{{Timestamp:HH:mm:ss}} {{Level:u3}}] {0}:{{Message:lj}}{{NewLine}}{{Exception}}",
+                    logTag);
+
+            // Setup Serilog logging
+            // **Note**: we add `Trace` sink in order to see log messages when
+            // running unit tests more conveniently.
+            log = new LoggerConfiguration()
+                .MinimumLevel.Debug()
```
Dexterp37, line 30: Is this something we could do just in tests, maybe by changing the project file for Tests?
Moreover, how can we enable debugging logging in release builds when we have to investigate problems?

glean-core/csharp/Glean/Glean.csproj
```diff
@@ -127,4 +127,12 @@
       <_Parameter1>GleanTests</_Parameter1>
     </AssemblyAttribute>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Serilog" Version="2.9.0" />
+    <PackageReference Include="Serilog.Sinks.Console" Version="3.1.1" />
+    <PackageReference Include="Serilog.Sinks.Trace" Version="2.1.0" />
+  </ItemGroup>
+  <ItemGroup>
+    <Folder Include="Utils\" />
```
Dexterp37, line 136: What is this for?

glean-core/csharp/Glean/Glean.csproj
```diff
@@ -127,4 +127,12 @@
       <_Parameter1>GleanTests</_Parameter1>
     </AssemblyAttribute>
   </ItemGroup>
+  <ItemGroup>
+    <PackageReference Include="Serilog" Version="2.9.0" />
```
Dexterp37, line 131: Would you kindly check if this dependency is transitively added to the sample application?

glean-core/csharp/Glean/Utils/Logger.cs
travis79: This just sets the minimum severity level that will be output by Serilog  but it will still output for all builds.  Think of this more like the Android logging where this means that `Log.d`, `Log.i`, `Log.e` would be output but `Log.v` would not.  So, if we change the minimum level to `Information`, then it would only log messages with severity `Information` or greater.  This is enabled in Release.

glean-core/csharp/Glean/Glean.csproj
travis79: It's a folder...  where the `Logger` class lives.  I don't guess I understand the question...

glean-core/csharp/Glean/Glean.csproj
travis79: Sure!  I was planning on trying out the suggestions from Serilog on resolving the issue I ran into with the crash so I will take it for a test drive in the sample application too.

glean-core/csharp/Glean/Glean.csproj
Dexterp37: Sorry if I was unclear: why do we need this? We don't need it for other directories (e.g, `Metrics`, `Configuration`)

travis79: Oh, this is probably what you get when you add the folder via VS instead of VSCode...

glean-core/csharp/Glean/Utils/Logger.cs
Dexterp37: I would expect this to be `Debug` only in tests, while being configurable at runtime in release (defaulting at Information).

In a release build, where do logs get printed by default? stdout? Are they turned on by default?

glean-core/csharp/Glean/Glean.cs
travis79: FWIW, just calling this shouldn't hurt anything since the metrics ping scheduler will likely be using the ConcurrentScheduler to run the upload task on, it would be canceled by this function anyway.

glean-core/csharp/Glean/Glean.csproj
travis79: Removed it and nothing broke. :)

glean-core/csharp/Glean/Glean.cs
Dexterp37: Probably, but let's take this to when we'll implement these pieces ;)

glean-core/csharp/Glean/Utils/Logger.cs
travis79: We don't allow log level configuration for any other binding, why here?  For release or debug builds this will log to the provided sinks, in this case the Console (i.e. stdout), and to Trace (which makes it show up in the application output window in VS).

Dexterp37: For python, for example, no log gets shown unless we enable them. You also pick the minimum logging level there. The same goes for Rust bindings. Before digging more into this, let me understand it a bit better:

- how can you provide a sink?
- can you turn on/off the logging on a released application?

travis79: The sinks are provided in the two lines following the one you tagged for this comment.  And no, as this is currently implemented, there isn't a way to turn it on and off from a consuming application.  Just like the Kotlin bindings doesn't turn off the logging to LogCat or provide a way to set a 'sink' for the logging.

travis79:
Wait, what just happened..   I didn't mean to merge this nor did I hit the button :O

PR #974, opened 2020-06-12
Limit the number of ping upload requests per minute

brizental:
Related to [Bug 1543612](https://bugzilla.mozilla.org/show_bug.cgi?id=1543612)

I wanted to get some feedback on my approach to this problem before I continued. My goal here was to try and replicate [what Firefox does to rate limit ping uploads](https://searchfox.org/mozilla-central/rev/4bb2401ecbfce89af06fb2b4d0ea3557682bd8ff/toolkit/components/telemetry/app/TelemetrySend.jsm#84), specifically:
> Only allow up to 10 ping uploads per minute.

This turned out to be tricky. The challenges on the current implementation and how I decided to solve them are:
- It doesn't really know exactly when the ping is uploaded.
   - My solution here was to ignore this fact and just try to limit the number of times `get_upload_task` can be called per minute.
- It sends pings to upload one at a time, not in batches.
   - I created a counter to know how many pings have been sent out in the past minute.

NOTE: I am aware this breaks the ffi code and all the bindings, I'll deal with that after.

### DIFF
--- a/docs/dev/core/internal/upload.md
+++ b/docs/dev/core/internal/upload.md
@@ -19,7 +19,7 @@ sequenceDiagram
     participant Server
 
     Glean wrapper->>Glean core: get_upload_task()
-    Glean core->>Glean wrapper: Task::Upload(task)
+    Glean core->>Glean wrapper: Task::Upload(PingRequest)
     Glean wrapper-->>Server: POST /submit/{task.id}
     Server-->>Glean wrapper: 200 OK
     Glean wrapper->>Glean core: upload_response(200)
@@ -27,9 +27,9 @@ sequenceDiagram
     Glean core->>Glean wrapper: Task::Done
 ```
 
-Glean core will take care of file management, cleanup, rescheduling and throttling[^1].
+Glean core will take care of file management, cleanup, rescheduling and rate limiting[^1].
 
-> [^1] Note: At this point throttling is not implemented. Follow [Bug 1543612](https://bugzilla.mozilla.org/show_bug.cgi?id=1543612) for updates.
+> [^1] Rate limiting is achieved by limiting the amount of times a language binding is allowed to get a `Task::Upload(PingRequest)` from `get_upload_task` in a given time interval. Currently, the default limit is for a maximum of 10 upload tasks every 60 seconds and there are no exposed methods that allow changing this default (follow [Bug 1647630](https://bugzilla.mozilla.org/show_bug.cgi?id=1647630) for updates). If the caller has reached the maximum tasks for the current interval, they will get a `Task::Wait` regardless if there are other `Task::Upload(PingRequest)`s queued.
 
 ## Available APIs
 
--- a/docs/user/pings/index.md
+++ b/docs/user/pings/index.md
@@ -79,6 +79,9 @@ See [Using the Experiments API](../experiments-api.md) on how to record experime
 ## Ping submission
 
 The pings that the Glean SDK generates are submitted to the Mozilla servers at specific paths, in order to provide additional metadata without the need to unpack the ping payload.
+
+> **Note**: To keep resource usage in check, the Glean SDK allows only up to 10 ping submissions every 60 seconds. There are no exposed methods to change these rate limiting defaults yet, follow [Bug 1647630](https://bugzilla.mozilla.org/show_bug.cgi?id=1647630) for updates.
+
 A typical submission URL looks like
 
   `"<server-address>/submit/<application-id>/<doc-type>/<glean-schema-version>/<document-id>"`
@@ -101,6 +104,7 @@ A pre-defined set of headers is additionally sent along with the submitted ping:
 | `Date` | e.g. `Mon, 23 Jan 2019 10:10:10 GMT+00:00` | Submission date/time in GMT/UTC+0 offset |
 | `X-Client-Type` | `Glean` | Custom header to support handling of Glean pings in the legacy pipeline |
 | `X-Client-Version` | e.g. `0.40.0` | The Glean SDK version, sent as a custom header to support handling of Glean pings in the legacy pipeline |
+
 
 ## Defining foreground and background state
 
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -190,13 +190,19 @@ impl Glean {
         let data_store = Some(Database::new(&cfg.data_path, cfg.delay_ping_lifetime_io)?);
         let event_data_store = EventDatabase::new(&cfg.data_path)?;
 
+        // Create an upload manager with rate limiting of 10 pings every 60 seconds.
+        let mut upload_manager = PingUploadManager::new(&cfg.data_path, false);
+        upload_manager.set_rate_limiter(
+            /* seconds per interval */ 60, /* max tasks per interval */ 10,
+        );
+
         let mut glean = Self {
             upload_enabled: cfg.upload_enabled,
             data_store,
             event_data_store,
             core_metrics: CoreMetrics::new(),
             internal_pings: InternalPings::new(),
-            upload_manager: PingUploadManager::new(&cfg.data_path, false),
+            upload_manager,
             data_path: PathBuf::from(cfg.data_path),
             application_id,
             ping_registry: HashMap::new(),
--- a/glean-core/src/upload/mod.rs
+++ b/glean-core/src/upload/mod.rs
@@ -104,7 +104,7 @@ impl RateLimiter {
     /// The counter should reset if
     ///
     /// 1. It has never started;
-    /// 2. It has been started more than one minute ago;
+    /// 2. It has been started more than the interval time ago;
     /// 3. Something goes wrong while trying to calculate the elapsed time since the last reset.
     fn should_reset(&self) -> bool {
         if self.started.is_none() {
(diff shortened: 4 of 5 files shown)


glean-core/src/upload/mod.rs
```diff
@@ -69,6 +77,74 @@ pub fn setup_upload_manager(upload_manager: PingUploadManager) -> Result<()> {
     Ok(())
 }
 
+/// A counter to help limit the number of processing pings per minute.
+///
+/// NOTE: A "processing" ping is a ping that has been dequeued through `get_upload_task`.
+#[derive(Debug)]
+struct PingCounter {
```
brizental, line 84: This doesn't really need to be a different struct, it was easier for me just to isolate the logic I was working on. I can merge it in the `PingUploadManager` if necessary.

glean-core/src/upload/mod.rs
```diff
@@ -440,11 +529,50 @@ mod test {
...
+            }
+        }
+
+        // Enqueue just one more ping...
+        upload_manager.enqueue_ping(DOCUMENT_ID, PATH, json!({}));
+
+        // Verify that we are indeed told to wait because we are at capacity
+        assert_eq!(PingUploadTask::Wait, upload_manager.get_upload_task(false));
+
+        thread::sleep(Duration::from_millis((1000 * SECONDS_IN_A_MINUTE) as u64));
```
brizental, line 563: I know this is way too much time for a test to wait. I'll think of another way to test it before finishing this.

glean-core/src/upload/mod.rs
```diff
@@ -195,8 +274,18 @@ impl PingUploadManager {
...
+        match queue.front() {
             Some(request) => {
+                // If unable to increment the counter,
+                // we have reached the rate limit of ping uploads for the time being.
+                if self.counter.increment().is_none() {
+                    log::info!(
+                        "Tried getting an upload task, but we are already at maximum processing capacity.
+                        Will come back later."
+                    );
+                    return PingUploadTask::Wait;
```
brizental, line 286: Should we actually provide the amount of time to wait? Like `PingUploadTask::Wait(SECONDS_IN_A_MINUTE)`. I remember this being discussed back when `PingUploadTask` was first created. Could be useful in the future for other features we want to implement such as jitter delay.

glean-core/src/upload/mod.rs
```diff
@@ -32,6 +34,12 @@ mod directory;
 mod request;
 mod result;
 
+static SECONDS_IN_A_MINUTE: u32 = 60;
```
Dexterp37, line 37: nit: don't we have this defined in chrono somewhere already?

glean-core/src/upload/mod.rs
Dexterp37: Yes, for the final thing, I'd rather see this as part of the upload manager, but I have no strong opinion. If it looks cleaner this way, let's stick to it.

glean-core/src/upload/mod.rs
Dexterp37: I believe we should, but maybe we don't need it just yet for this one.

glean-core/src/upload/mod.rs
Dexterp37: Maybe you can make the wait time customizable? We'll need that (in the future) so that products could customize it.

glean-core/src/upload/mod.rs
```diff
@@ -69,6 +77,74 @@ pub fn setup_upload_manager(upload_manager: PingUploadManager) -> Result<()> {
...
+impl PingCounter {
+    pub fn new() -> Self {
+        Self {
+            started: None,
+            count: 0,
+        }
+    }
+
+    fn reset(&mut self) {
+        self.started = Some(SystemTime::now());
```
Dexterp37, line 98: Is this clock monotonic? We should use a monotonic clock  :) We have "chrono" for that, too.

glean-core/src/upload/mod.rs
```diff
@@ -69,6 +77,74 @@ pub fn setup_upload_manager(upload_manager: PingUploadManager) -> Result<()> {
...
+            }
+        }
+
+        false
+    }
+
+    /// Increments the counter by 1 and returns the current count.
+    ///
+    /// This will return `None` if the counter has reached the maximum.
+    pub fn increment(&mut self) -> Option<u32> {
```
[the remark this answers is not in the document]
brizental: No need to return the count, in hindsight I don't even know why I put that there. I didn't use a `Result` here just because I wanted to avoid creating a new error type for this. The enum approach is nicer, I'll do it and we can bikeshed on the names after I send updates.

glean-core/src/upload/mod.rs
```diff
@@ -139,13 +211,29 @@ impl PingUploadManager {
...
         }
     }
 
     fn has_processed_pings_dir(&self) -> bool {
         self.processed_pending_pings.load(Ordering::SeqCst)
     }
 
+    /// Adds rate limiting capability to this upload manager.
+    ///
+    /// The rate limiter will limit the amount of calls to `get_upload_task` per interval.
```
Dexterp37, line 224: Let's mention explicitly that this resets the state of the rate limiter: e.g. if you sent the maximum amount of pings in the timeframe, then you reset, we start counting again.

glean-core/src/upload/mod.rs
```diff
@@ -9,16 +9,12 @@
 //! * Exposes `process_ping_upload_response` API to check the HTTP response from the ping upload
 //!   and either delete the corresponding ping from disk or re-enqueue it for sending.
 
-// !IMPORTANT!
-// Remove the next line when this module's functionality is in the Glean object.
-// This is here just to not have lint error for now.
-#![allow(dead_code)]
```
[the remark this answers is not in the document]
badboy: :D Hah, oops.

glean-core/src/upload/mod.rs
```diff
@@ -195,8 +283,21 @@ impl PingUploadManager {
...
-        match queue.pop_front() {
+        match queue.front() {
             Some(request) => {
+                if let Some(rate_limiter) = &self.rate_limiter {
+                    let mut rate_limiter = rate_limiter
+                        .write()
+                        .expect("Can't write to the rate limiter.");
+                    if rate_limiter.get_state() == RateLimiterState::Throttled {
+                        log::info!(
+                            "Tried getting an upload task, but we are throttled at the moment.
```
badboy, line 294: This will log with a newline and a lot of preceding whitespace. The second line should probably be inline here (or just be removed?)

docs/dev/core/internal/upload.md
```diff
@@ -19,17 +19,17 @@ sequenceDiagram
...
     Glean wrapper->>Glean core: upload_response(200)
     Glean wrapper->>Glean core: get_upload_task()
     Glean core->>Glean wrapper: Task::Done
 ```
 
-Glean core will take care of file management, cleanup, rescheduling and throttling[^1].
+Glean core will take care of file management, cleanup, rescheduling and rate limiting[^1].
 
-> [^1] Note: At this point throttling is not implemented. Follow [Bug 1543612](https://bugzilla.mozilla.org/show_bug.cgi?id=1543612) for updates.
+> [^1] Rate limiting is achieved by limiting the amount of times a language binding is allowed to get a `Task::Upload(PingRequest)` from `get_upload_task` in a given time interval. Currently, the default limit is for a maximum of 10 upload tasks every 60 seconds and there are no exposed methods that allow changing this default. If the caller has reached the maximum tasks for the current interval, they will get a `Task::Wait` regardless if there are other `Task::Upload(PingRequest)`s queued.
```
brizental, line 32: Not sure this is very clear, let me know your thoughts.

mdboom: I think this is clear for the internal docs (as it pertains to people on the Glean team writing bindings). However, I think we should *also* document this in the user docs -- maybe [on this page](https://mozilla.github.io/glean/book/user/pings/index.html)?

glean-core/src/lib.rs
```diff
@@ -190,13 +190,19 @@ impl Glean {
         let data_store = Some(Database::new(&cfg.data_path, cfg.delay_ping_lifetime_io)?);
         let event_data_store = EventDatabase::new(&cfg.data_path)?;
 
+        // Create an upload manager with rate limiting of 10 pings every 60 seconds.
+        let mut upload_manager = PingUploadManager::new(&cfg.data_path, false);
+        upload_manager.set_rate_limiter(
```
brizental, line 195: Should we also add a rate limiter to the upload manager created in [`glean_initialize_standalone_uploader`](https://github.com/mozilla/glean/blob/50cf80311315caf05f4870397e83812a5eda5731/glean-core/ffi/src/lib.rs#L480). I have just noticed this, but I am not sure exactly from looking at the code if it is necessary. If so I can open a follow up bug for it. @badboy ?

docs/dev/core/internal/upload.md
brizental: Alright, it's done in ba82a94

docs/user/pings/index.md
```diff
@@ -79,6 +79,9 @@ See [Using the Experiments API](../experiments-api.md) on how to record experime
 ## Ping submission
 
 The pings that the Glean SDK generates are submitted to the Mozilla servers at specific paths, in order to provide additional metadata without the need to unpack the ping payload.
+
+> **Note**: To keep resource usage in check, the Glean SDK allows only up to 10 ping submissions every 60 seconds. There are no exposed methods to change these rate limiting defaults.
```
brizental, line 83: > There are no exposed methods to change these rate limiting defaults.

Do you believe methods to change this will be immediately necessary? I think maybe for testing purposes they might be? If so I can file a bug and link to it in the internal and user facing documentations.

Dexterp37: The long(er) term plan is to allow products to configure their own limits. Fenix might want tighter control compared to Firefox Desktop, for example. Let's file a bug about allowing products to customize the policy and link it here.

glean-core/src/lib.rs
Dexterp37: @brizental yes, we should. Please kindly file a follow-up :)

PR #976, opened 2020-06-12
Run cbindgen on stable rust (but pretend it's nightly)

badboy:
`RUSTC_BOOTSTRAP` turns a stable compiler into pretending it is nightly,
unlocking nightly-only features.
It shouldn't be used in a normal build, ever.

But here we _only_ need it for some special nightly-only output flags
for rustc, as used by cbindgen, to determine some metadata about the
crates it looks at.

A recent cargo nightly regression caused that to fail, so this will try
out if we can depend on (pseudo-)stable instead.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -309,8 +309,7 @@ jobs:
     steps:
       - checkout
       - skip-if-doc-only
-      - setup-rust-toolchain:
-          rust-version: "nightly"
+      - setup-rust-toolchain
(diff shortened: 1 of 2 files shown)


badboy:
Now this failed because the test ran on _my_ CircleCI account, which didn't have access to the medium+ resource class.

badboy:
Nightly cargo should be fixed. Dropping this for now.

PR #977, opened 2020-06-15
Enable the upload manager in C#

Dexterp37:
This PR introduces the C# language bindings bits to communicate with the Rust upload APIs.

It DOES the following:

- add the ping submission APIs to the `Glean` object;
- implement the `BaseUploader` classe;
- provide test coverage for the `PingType` (which required an uploader) and the `BaseUploader`.

It doesn't:

- Implement an HTTP uploader (will come as a separate PR to ease review);
- Implement an off-the-main thread upload (will come as yet another PR);


glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -0,0 +1,141 @@
...
+                            // Delegate the actual upload and get its return value.
+                            UploadResult result = Upload(path, body, headers, config);
+
+                            if (result is RecoverableFailure)
+                            {
+                                uploadFailures += 1;
+                            }
+
+                            // Process the upload response in the core.
+                            LibGleanFFI.glean_process_ping_upload_response(ref incomingTask, result.ToFfi());
```
Dexterp37, line 118: All data is read correctly without the last commit, but it crashes after executing this function, because Rust frees `incomingTask`

glean-core/csharp/Glean/Glean.cs
```diff
@@ -76,16 +90,16 @@ string dataDir
                 return
             }
 
-            this.applicationContext = applicationContext
+            this.applicationContext = applicationContext*/
+
+            this.configuration = configuration;
+            httpClient = new BaseUploader(configuration.httpClient);
```
daoshengmu, line 96: IIRC, `configuration.httpClient` here is null?

daoshengmu:
I will start to review once #965 merged and rebase from it.

glean-core/csharp/Glean/Glean.cs
Dexterp37: Yes, that's fine. This PR will not provide any HTTP uploader, it will only create the upload manager.

glean-core/csharp/Glean/Configuration/Configuration.cs
```diff
@@ -16,6 +16,10 @@ public sealed class Configuration
         public string serverEndpoint;
         public string channel;
         public int? maxEvents;
+        public bool logPings;
+        public string pingTag;
+
```
daoshengmu, line 21: It looks like you miss adding `userAgent`.

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -8,6 +8,63 @@
...
+    }
+
+    [StructLayout(LayoutKind.Sequential)]
+    internal struct FfiUploadTaskBody
+    {
+        public byte tag;
+        public IntPtr documentId;
+        public IntPtr path;
+        public int bodyLen;
+        public IntPtr body;
```
daoshengmu, line 51: nit: It looks ok to me. But if you wanna match to `UploadBody.kt`, there is supposed to have a RustBuffer struct to contain these.

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -8,6 +8,63 @@
...
+    /// <summary>
+    /// Represent an upload task by simulating the union passed through
+    /// the FFI layer.
+    /// </summary>
+    [StructLayout(LayoutKind.Explicit)]
+    internal struct FfiUploadTask
+    {
+        [FieldOffset(0)]
+        public byte tag;
+        [FieldOffset(0), MarshalAs(UnmanagedType.Struct)]
```
daoshengmu, line 64: Q: I am confused with `FieldOffset(0)` here. It supposes to be `1` for `body` because it shifts one byte from `tag`. But, it looks like you're right. If giving `1`, some tests are skipped.

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -0,0 +1,154 @@
...
+        /// </summary>
+        internal void TriggerUpload(Configuration config)
+        {
+            // TODO: must not work like this, it should work off the main thread.
+            // FOR TESTING Implement the upload worker here and call this from Glean.cs
+
+            int waitAttempts = 0;
+            int uploadFailures = 0;
+
+            while (uploadFailures < 3)
```
daoshengmu, line 89: You might want to make `3` as a constant and comment it a bit.
```
// Maximum number of recoverable errors allowed before aborting the ping uploader
        const maxRetries = 3;
```

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -0,0 +1,154 @@
...
+                            // Process the upload response in the core.
+                            LibGleanFFI.glean_process_ping_upload_response(ptrCopy, result.ToFfi());
+
+                            // Free the allocated.
+                            Marshal.FreeHGlobal(ptrCopy);
+                        }
+                        break;
+                    case UploadTaskTag.Wait:
+                        {
+                            if (waitAttempts < 3)
```
daoshengmu, line 129: is it `maxRetries` ?

glean-core/csharp/Glean/Configuration/Configuration.cs
Dexterp37: Oh, good point. I'll add it, but we just filed a bug to consider [removing it from all the bindings](https://bugzilla.mozilla.org/show_bug.cgi?id=1646441).

glean-core/csharp/Glean/Glean.cs
```diff
@@ -343,6 +350,83 @@ internal void HandleBackgroundEvent()
...
+        /// policies.
+        /// 
+        /// If the ping currently contains no content, it will not be assembled and
+        /// queued for sending.
+        /// </summary>
+        /// <param name="ping">Ping to submit.</param>
+        /// <param name="reason">The reason the ping is being submitted.</param>
+        internal void SubmitPing(PingTypeBase ping, string reason = null)
+        {
+            SubmitPingByName(ping.name, reason);
```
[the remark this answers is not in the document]
Dexterp37: Oh, the fact that we return `Job` on Android is vestigial: we only needed it for testing, we don't even need it anymore and it's bound to be removed.

glean-core/csharp/Glean/LibGleanFFI.cs
Dexterp37: In Kotlin, we copied over that implementation from Application Services. In this case, we don't have any reference implementation available, so we're fine to leave this as it is. There it was more for convenience, than for need.

glean-core/csharp/Glean/LibGleanFFI.cs
Dexterp37: Yes, as you found, the offset is required to be `0` in both places. That's because we're mapping to a [C `union`](https://searchfox.org/glean/rev/73522afaaa64a2c904ec28e209e53c92fb94967b/glean-core/ffi/glean.h#282,289). The `tag` is present both in `FfiUploadTask` and `FfiUploadTaskBody`, at the same location. If tag is "Upload", then we consider this to have a body a proceed with further parsing.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -343,6 +350,83 @@ internal void HandleBackgroundEvent()
...
+        /// 
+        /// If the ping currently contains no content, it will not be assembled and
+        /// queued for sending, unless explicitly specified otherwise in the registry
+        /// file.
+        /// </summary>
+        /// <param name="name">Name of the ping to submit.</param>
+        /// <param name="reason">The reason the ping is being submitted.</param>
+        internal void SubmitPingByName(string name, string reason = null)
+        {
+            Dispatchers.LaunchAPI(() => {
```
travis79, line 388: Format nit, what does StyleCop think of this curly brace on the same line?

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -0,0 +1,154 @@
...
+        }
+
+        /// <summary>
+        /// This function triggers the actual upload.
+        /// 
+        /// It calls the implementation-specific upload function.
+        /// </summary>
+        /// <param name="url">the URL path to upload the data to</param>
+        /// <param name="data">the serialized text data to send</param>
+        /// <param name="headers">a [HeadersList] containing the headers to add</param>
```
travis79, line 36: `HeadersList` doesn't match the type passed in of an array of tuples.

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -0,0 +1,154 @@
...
+                }
+            }
+        }
+
+        /// <summary>
+        /// Cancel any outstanding upload.
+        /// </summary>
+        internal void CancelUploads()
+        {
+
```
travis79, line 151: Missing TODO or bug to complete this?

glean-core/csharp/Glean/Net/PingUploader.cs
```diff
@@ -0,0 +1,84 @@
...
+    /// The interface defining how to send pings.
+    /// </summary>
+    public interface IPingUploader
+    {
+        /// <summary>
+        /// Synchronously upload a ping to a server.
+        /// </summary>
+        /// <param name="url">the URL path to upload the data to</param>
+        /// <param name="data">the serialized text data to send</param>
+        /// <param name="headers">a list containing the headers to add</param>
```
travis79, line 79: nit, "an array of tuples", not a "list"

glean-core/csharp/Glean/Configuration/Configuration.cs
Dexterp37: This actually led me to find that Glean has a bug currently: it's not setting the right user agent: https://bugzilla.mozilla.org/show_bug.cgi?id=1646626

glean-core/csharp/Glean/Glean.cs
```diff
@@ -140,8 +148,7 @@ string dataDir
                 // 2. Upload is disabled, to upload a possible deletion-request ping.
```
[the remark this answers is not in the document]
Dexterp37: I changed this to reference `BaseUploader` now. The two points remains largely correct.

glean-core/csharp/Glean/Net/BaseUploader.cs
Dexterp37: Nope, this is really wait attempts: https://searchfox.org/glean/rev/73522afaaa64a2c904ec28e209e53c92fb94967b/glean-core/python/glean/net/ping_upload_worker.py#154 - we don't want to get into any infinite loop if there's no connectivity!

glean-core/csharp/Glean/Glean.cs
Dexterp37: Mh, it was, but really not that visible in the IDE. I'll file a bug to add linting to CI :)

PR #978, opened 2020-06-15
Fix a race condition in C# Dispatchers

travis79:
This uses a custom TaskScheduler + Factory from Microsoft example: https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler?view=netcore-3.1

### DIFF
--- a/glean-core/csharp/Glean/Dispatchers.cs
+++ b/glean-core/csharp/Glean/Dispatchers.cs
@@ -4,11 +4,11 @@
 
 using System;
 using System.Collections.Concurrent;
-using System.Collections.Generic;
 using System.Diagnostics;
 using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
+using System.Threading.Tasks.Schedulers;
 
 namespace Mozilla.Glean
 {
@@ -23,7 +23,7 @@ namespace Mozilla.Glean
         /// This is the number of milliseconds that are allowed for the initial
         /// tasks queue to process all of the queued tasks.
         /// </summary>
-        private const int QueueProcessingTimeout = 5000;
+        private const int QueueProcessingTimeoutMs = 5000;
 
         /// <summary>
         /// This is the maximum number of tasks that will be queued before
@@ -119,7 +119,7 @@ namespace Mozilla.Glean
                 {
                     // If we are in testing mode, then go ahead and await
                     // the task to ensure synchronous execution.
-                    factory.StartNew(action).Wait();
+                    action.Invoke();
                 }
             }
 
@@ -161,9 +161,9 @@ namespace Mozilla.Glean
             }
             else
             {
-                // This is not invoked on the apiScheduler to ensure that it
-                // gets executed now, ahead of anything on the queue. The task
-                // is returned so that it can be awaited if needed.
+                // **NOTE** This does not ensure that this task is executed at
+                // the front of the apiScheduler. This will be addressed with
+                // https://bugzilla.mozilla.org/show_bug.cgi?id=1646750
                 return factory.StartNew(action);
             }
         }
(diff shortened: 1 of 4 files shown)


glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -275,8 +213,8 @@ private static void AddActionToQueue(Action action)
             {
                 //TODO Log.i(LOG_TAG, "Task queued for execution and delayed until flushed")
             }
-
-            taskQueue.Enqueue(action);
+
```
Dexterp37, line 216:
```suggestion

```

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -39,61 +39,44 @@ internal class Dispatchers
...
 
         /// <summary>
-        /// Provides the token source for cancellation tokens used to cancel
-        /// running background tasks.
+        /// This Queue holds the initial Actions that are launched before Glean
+        /// is initialized.
         /// </summary>
-        private static readonly CancellationTokenSource tokenSource =
-            new CancellationTokenSource();
+        internal static List<Action> preInitActionQueue = new List<Action>();
```
Dexterp37, line 64: HA! Now I see what the problem was (now that you fixed it somewhere else in the code :) ). This should really be a [`ConcurrentQueue`](https://docs.microsoft.com/en-us/dotnet/api/system.collections.concurrent.concurrentqueue-1?view=netcore-3.1)! We use a `ConcurrentLinkQueue` in Kotlin (lock-free), but it's fine to use any other thing here as log as it is thread safe (and `ConcurrentQueue` is).

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -39,61 +39,44 @@ internal class Dispatchers
...
-            ConcurrentExclusiveSchedulerPair taskSchedulerPair =
-            new ConcurrentExclusiveSchedulerPair();
-
-        /// <summary>
-        /// This Queue holds the initial tasks that are launched before Glean is
-        /// initialized.
-        /// </summary>
-        internal static ConcurrentQueue<Action> taskQueue =
-            new ConcurrentQueue<Action>();
+        // Create a scheduler that uses a single thread
```
Dexterp37, line 52:
```suggestion
        // Create a scheduler that uses a single thread.
```

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -39,61 +39,44 @@ internal class Dispatchers
...
+        private static readonly LimitedConcurrencyLevelTaskScheduler lcts =
```
Dexterp37, line 53:
```suggestion
        private static readonly LimitedConcurrencyLevelTaskScheduler apiScheduler =
```

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -39,61 +39,44 @@ internal class Dispatchers
...
+            new LimitedConcurrencyLevelTaskScheduler(1);
 
-        /// <summary>
-        /// The number of tasks added to the queue beyond the MaxQueueSize.
-        /// </summary>
-        private static int OverflowCount { get; set; } = 0;
+        // Create a new TaskFactory and pass it the scheduler and cancellation
```
Dexterp37, line 56: We're not passing a cancellation token anymore!

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -104,21 +87,22 @@ internal static Task LaunchAPI(Action action)
         {
             Task task = null;
 
-            if (QueueInitialTasks)
+            lock (preInitActionQueue)
```
Dexterp37, line 90: If we go for the `ConcurrentQueue`, we don't need this anymore.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -130,140 +114,94 @@ internal static Task LaunchAPI(Action action)
...
         /// </summary>
-        /// <param name="action">The Action to perform</param>
-        /// <returns>A Task or null if run synchronously</returns>
-        internal static Task LaunchConcurrent(Action action)
+        internal static void FlushQueuedInitialTasks()
         {
-            Task task = null;
+            // Create a list to store the created tasks in order to await
+            // their execution
+            List<Task> tasks = new List<Task>();
```
Dexterp37, line 138: Let's stick to the implementation on the Kotlin side: let's create a new action, that runs on the API scheduler, which is responsible for switching.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -130,140 +114,94 @@ internal static Task LaunchAPI(Action action)
...
+                // If we are in testing mode, then await the task and return
+                // null
+                factory.StartNew(action).Wait();
+                return null;
+            }
+            else
             {
-                //TODO GleanError.preinitTasksOverflow
-                //         .addSync(MaxQueueSize + OverflowCount)
+                return factory.StartNew(action);
```
Dexterp37, line 184: Does this guarantee that `action` will be executed *right after* the current dispatched action, instead of being put at the end of the queue?

If not, this might require slightly changing `LimitedConcurrencyLevelTaskScheduler`.

glean-core/csharp/Glean/Dispatchers.cs
travis79: Yeah, that was part of the problem, I think.  If you are okay with it I'll file a follow-up bug to implement that part back (if we even need it).

Dexterp37: No, I believe it's fine to not have that.

glean-core/csharp/Glean/Dispatchers.cs
travis79: I can change it back now that I have fixed the real issue :)

glean-core/csharp/Glean/Dispatchers.cs
travis79: It appears to work this way from the only context in which it is called, which is inside of initialize, before queueing is turned off, so it should always be called when there are no other tasks queued.

travis79: I removed any other call to it and replaced them with just invoking the action directly.

travis79: I went ahead an changed this to not use the custom factory and instead use the default one and just throw this on the ThreadPool to ensure it gets executed off of the calling thread, and not on the queue.

travis79: Actually, the more I think about this...  this should be launched on the apiScheduler but it's good to call out that Initialize is the only thing using this just to sidestep the pre-init queuing.

travis79: Because that way it guarantees that initialize (and FlushQueuedTasks) get's called before anything else gets into the apiScheduler.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -3,26 +3,27 @@
...
         /// </summary>
         private const string LogTag = "glean/Dispatchers";
 
         /// <summary>
-        /// This is the number of seconds that are allowed for the initial
+        /// This is the number of milliseconds that are allowed for the initial
         /// tasks queue to process all of the queued tasks.
         /// </summary>
-        private const int QueueProcessingTimeout = 5;
+        private const int QueueProcessingTimeout = 5000;
```
Dexterp37, line 26:
```suggestion
        private const int QueueProcessingTimeoutMs = 5000;
```

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -106,164 +90,101 @@ internal static Task LaunchAPI(Action action)
...
+                    });
                 }
                 else
                 {
-                    // If we are in testing mode, then go ahead and invoke the
-                    // action to ensure synchronous execution.
-                    action.Invoke();
+                    // If we are in testing mode, then go ahead and await
+                    // the task to ensure synchronous execution.
+                    factory.StartNew(action).Wait();
```
Dexterp37, line 122: Why does this have to change? Can't this be just `action.Invoke()`?

glean-core/csharp/Glean/Dispatchers.cs
Dexterp37: Please note that this **has** to get executed _before_ any task in queue (except the one that's currently being executed). Is this checked for by any of our tests already? Did you check this manually?

> Actually, the more I think about this... this should be launched on the apiScheduler but it's good to call out that Initialize is the only thing using this just to sidestep the pre-init queuing.

This will change once we'll add the metrics ping scheduler. That's why we need to get this right. I'm fine with not keeping this PR waiting, we should file a bug and deal with this specific aspect later.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -289,5 +210,135 @@ public static void AssertInTestingMode()
         {
             Debug.Assert(TestingMode);
         }
+
+        // Provides a task scheduler that ensures a maximum concurrency level while
+        // running on top of the thread pool.  Copied from example provided here:
+        // https://docs.microsoft.com/en-us/dotnet/api/system.threading.tasks.taskscheduler?view=netcore-3.1
+        private class LimitedConcurrencyLevelTaskScheduler : TaskScheduler
```
Dexterp37, line 217: Would you kindly move this to a separate file?  It's also a bit unclear if we're allowed to use this and what's the license for this.

glean-core/csharp/Glean/Dispatchers.cs
travis79: It can be just `Invoke()` here, good call.

glean-core/csharp/Glean/Dispatchers.cs
travis79: Sure, that makes it a little more open within the project, but I can put it under the Dispatchers namespace to help with that.  [The original code sample](https://docs.microsoft.com/en-us/samples/dotnet/samples/parallel-programming-extensions-extras-cs/) uses the MIT license, which I think is compatible for us to use?

travis79: > This will change once we'll add the metrics ping scheduler. That's why we need to get this right. I'm fine with not keeping this PR waiting, we should file a bug and deal with this specific aspect later.

[Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1646750)

glean-core/csharp/GleanTests/DispatchersTests.cs
```diff
@@ -66,23 +43,25 @@ public void TestTaskQueueing()
...
             for (int i = 0; i < 3; ++i)
             {
-                Dispatchers.LaunchAPI(() =>
+                tasks.Add(Dispatchers.LaunchAPI(() =>
                 {
                     threadCanary += 1;
-                });
-            }
+                }));
+            }
```
Dexterp37, line 56:
```suggestion
            }
```

I don't really like the linter not complaining about this :\

travis79: I don't really like VS putting it there either, they aren't exactly visible in the editor.

PR #979, opened 2020-06-15
Add brief overview section on book's main page

badboy:
I remember we talked about this briefly, don't remember if we filed a bug.

But I wanted to get _something_ done today.
I felt that there's only 3 big chapters worth listing, but then also mentioning the TWiG felt like the right thing to do.

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -12,6 +12,22 @@ To contact us you can:
 
 The source code is available [on GitHub](https://github.com/mozilla/glean/).
 
-## License
+## Using this book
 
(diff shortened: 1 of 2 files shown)


docs/README.md
```diff
@@ -12,6 +12,27 @@ To contact us you can:
 
 The source code is available [on GitHub](https://github.com/mozilla/glean/).
 
-## License
+## Using this book
 
-This Source Code Form is subject to the terms of the Mozilla Public License, v. 2.0. If a copy of the MPL was not distributed with this file, You can obtain one at <https://mozilla.org/MPL/2.0/>.
```
mdboom, line 17: Shouldn't we retain the license to Glean somewhere on the README?  (I know it doesn't really pertain to the docs, but I think it's probably still an important detail about the project)

mdboom: (But totally fine with moving it to the bottom / de-emphasizing it etc.)

PR #982, opened 2020-06-17
Make android debug activity usage documentation clearer

brizental:
Related to [Bug 1646002](https://bugzilla.mozilla.org/show_bug.cgi?id=1646002)

[doc only]

### DIFF
--- a/docs/appendix/glossary.md
+++ b/docs/appendix/glossary.md
@@ -54,7 +54,7 @@ The Glean SDK provides default pings and allows for custom ping, see [Glean Ping
 
 ## Submission
 
-"To submit" means to collect & to en-queue a ping for uploading.
+"To submit" means to collect & to enqueue a ping for uploading.
 
 The Glean SDK stores locally all the metrics set by it or by its clients.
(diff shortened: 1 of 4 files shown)


docs/appendix/glossary.md
```diff
@@ -54,7 +54,7 @@ The Glean SDK provides default pings and allows for custom ping, see [Glean Ping
 
 ## Submission
 
-"To submit" means to collect & to upload a ping.
+"To submit" means to collect & to en-queue a ping for uploading.
```
travis79, line 57: I don't think `en-queue` should have a hyphen, [according to the dictionary](https://www.lexico.com/definition/enqueue)

brizental: Me either, but our syntax checker complained about that. There is no discussing with the dictionary though, I'll see what I can do about the syntax checker.

travis79: You can add the word to the [dictionary file](https://github.com/mozilla/glean/blob/main/.dictionary) :)

PR #983, opened 2020-06-17
Bug 1646173: Python: Fix database not found on app shutdown

mdboom:
Cc: @hackebrot

### DIFF
--- a/glean-core/python/glean/_dispatcher.py
+++ b/glean-core/python/glean/_dispatcher.py
@@ -63,7 +63,10 @@ class _ThreadWorker:
         # queue, just run them now. This is required for synchronous testing
         # mode, and also to run the tasks in the expected order.
         if threading.get_ident() == self._ident:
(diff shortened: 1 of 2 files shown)


mdboom:
> Was this creating problems because an exception was thrown in a dispatched task?

No, not really, but getting more information out was helpful to resolve the bug, so I decided to leave it in there.

The main issue is that Burnham shuts down (and kills the Glean singleton) before the worker thread had a chance to complete all of its work (which gets worse when there is a networking issue that slows down uploading).

Dexterp37:
> 
> 
> > Was this creating problems because an exception was thrown in a dispatched task?
> 
> No, not really, but getting more information out was helpful to resolve the bug, so I decided to leave it in there.

Ok, interesting. Because exception eating from the dispatched tasks is the default behaviour we have in the other bindings, but we forgot to have in Pytho n:(

PR #985, opened 2020-06-18
Add a default ping uploader in C#

Dexterp37:
This adds an HTTP stack to enable ping sending. This is modelled after the Kotlin implementation.

### DIFF
--- a/glean-core/csharp/Glean/Net/HttpClientUploader.cs
+++ b/glean-core/csharp/Glean/Net/HttpClientUploader.cs
@@ -1,0 +1,91 @@
+﻿// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+using System;
+using System.Net.Http;
+
+namespace Mozilla.Glean.Net
+{
+    /// <summary>
+    /// A simple ping uploader that uses the `HttpClient` C# object.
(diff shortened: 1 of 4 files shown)


glean-core/csharp/Glean/Net/HttpClientUploader.cs
```diff
@@ -0,0 +1,91 @@
...
+
+namespace Mozilla.Glean.Net
+{
+    /// <summary>
+    /// A simple ping uploader that uses the `HttpClient` C# object.
+    /// </summary>
+    public class HttpClientUploader : IPingUploader
+    {
+        // The timeout, in milliseconds, to use when connecting to the server.
+        internal const int DEFAULT_CONNECTION_TIMEOUT = 10000;
```
travis79, line 16:
```suggestion
        internal const int DEFAULT_CONNECTION_TIMEOUT_MS = 10000;
```

Dexterp37:
> 
> 
> One name change requested, along with a request to add `samples/csharp/glean_data/` to the .gitignore file.

This shouldn't be needed: the sample app should create its data under `samples/csharp/bin`, which is already under .gitignore -

travis79:
> > One name change requested, along with a request to add `samples/csharp/glean_data/` to the .gitignore file.
> 
> This shouldn't be needed: the sample app should create its data under `samples/csharp/bin`, which is already under .gitignore -

Maybe it does something different on Mac, because it created this directory when I took this for a test run last week.

PR #986, opened 2020-06-18
Bug 1646402: Python: Don't use threading in multiprocessing subprocesses

mdboom:
Cc: @wlach -- I'd appreciate your review as well, even though permissions don't seem to let me request it...

### DIFF
--- a/glean-core/python/tests/test_dispatcher.py
+++ b/glean-core/python/tests/test_dispatcher.py
@@ -211,6 +211,7 @@ def test_single_threaded_in_multiprocessing_subprocess():
     p.start()
     p.join()
 
+    # This must match the definition in _subprocess() above.
     string_metric = metrics.StringMetricType(
         disabled=False,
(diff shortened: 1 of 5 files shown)


glean-core/python/tests/test_dispatcher.py
```diff
@@ -179,3 +183,40 @@ def slow_task():
...
+    not sys.platform.startswith("linux"), reason="Test only works on Linux"
+)
+def test_single_threaded_in_multiprocessing_subprocess():
+    import multiprocessing
+
+    p = multiprocessing.Process(target=_subprocess)
+    p.start()
+    p.join()
+
+    string_metric = metrics.StringMetricType(
```
Dexterp37, line 214: nit: can you add a comment stating that this definition must match the one in `_subprocess`?

mdboom: :+1:

PR #988, opened 2020-06-18
Bug 1643571 - Add C# Logger

travis79:
- Adds a GleanLogger wrapper around the Serilog Logger to make configuration simple.
- Update any commented out logging or Console.WriteLine to use the GleanLogger

### DIFF
--- a/glean-core/csharp/Glean/Dispatchers.cs
+++ b/glean-core/csharp/Glean/Dispatchers.cs
@@ -9,6 +9,8 @@ using System.Runtime.CompilerServices;
 using System.Threading;
 using System.Threading.Tasks;
 using System.Threading.Tasks.Schedulers;
+using Serilog;
+using static Mozilla.Glean.Utils.GleanLogger;
 
 namespace Mozilla.Glean
 {
@@ -70,6 +72,11 @@ namespace Mozilla.Glean
         private static int overflowCount = 0;
 
         /// <summary>
+        /// A logger configured for this class.
+        /// </summary>
+        private static readonly ILogger Log = GetLogger(LogTag);
+
+        /// <summary>
         /// Launch a block of work asynchronously.
         ///
         /// Takes an Action and launches it using the TaskFactory ensuring the
@@ -103,16 +110,17 @@ namespace Mozilla.Glean
                     // task asynchronously                    
                     task = factory.StartNew(() =>
                     {
-                            // In order to prevent tasks from causing exceptions
-                            // we wrap the action invocation in try/catch
-                            try
+                        // In order to prevent tasks from causing exceptions
+                        // we wrap the action invocation in try/catch
+                        try
                         {
                             action.Invoke();
                         }
                         catch (Exception)
                         {
-                                //TODO Exception eaten by Glean
-                            }
+                            Log.Error("Exception thrown by queued initial task and swallowed.");
+                            
+                        }
                     });
                 }
                 else
@@ -178,7 +186,7 @@ namespace Mozilla.Glean
         {
             if (preInitActionQueue.Count >= MaxQueueSize)
             {
-                //TODO Log.e(LOG_TAG, "Exceeded maximum queue size, discarding task")
+                Log.Error("Exceeded maximum queue size, discarding task");
 
                 // This value ends up in the `preinit_tasks_overflow` metric,
                 // but we can't record directly there, because that would only
@@ -190,11 +198,11 @@ namespace Mozilla.Glean
 
             if (TestingMode)
             {
-                //TODO Log.i(LOG_TAG, "Task queued for execution in test mode")
+                Log.Information("Task queued for execution in test mode");
             }
             else
             {
-                //TODO Log.i(LOG_TAG, "Task queued for execution and delayed until flushed")
+                Log.Information("Task queued for execution and delayed until flushed");
             }
             
             preInitActionQueue.Enqueue(action);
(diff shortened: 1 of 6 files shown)


glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -68,7 +80,7 @@ internal static (string, string)[] GetHeadersFromJSONString(string documentId, s
             }
             catch (JsonException)
             {
-                Console.WriteLine($"Error while parsing headers for ping {documentId}");
+                Log.Error("Error while parsing headers for ping {0}", documentId);
```
travis79, line 83: [According to Serilog's docs](https://github.com/serilog/serilog/wiki/Writing-Log-Events) using "templates" is preferred over "messages". I'm not entirely sure that string interpolation gets treated as a template so I changed this to be in line with the other places we format in values.

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -103,16 +110,17 @@ internal static Task LaunchAPI(Action action)
...
-                            // In order to prevent tasks from causing exceptions
-                            // we wrap the action invocation in try/catch
-                            try
+                        // In order to prevent tasks from causing exceptions
+                        // we wrap the action invocation in try/catch
+                        try
                         {
                             action.Invoke();
                         }
                         catch (Exception)
```
Dexterp37, line 119:
```suggestion
                        catch (Exception e)
```

Dexterp37: If we declare a variable for holding the exception...

glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -103,16 +110,17 @@ internal static Task LaunchAPI(Action action)
...
                         {
-                                //TODO Exception eaten by Glean
-                            }
+                            Log.Error("Exception thrown by queued initial task and swallowed.");
```
Dexterp37, line 121:
```suggestion
                            Log.Error(e, "Exception thrown by queued initial task and swallowed.");
```

... then we can use it here to print useful information :)

glean-core/csharp/Glean/Net/BaseUploader.cs
Dexterp37:
```suggestion
                Log.Error("Error while parsing headers for ping {documentId}");
```

Can this form be used instead here and in other places?

glean-core/csharp/Glean/Glean.csproj
```diff
@@ -1,4 +1,4 @@
-﻿<Project Sdk="Microsoft.NET.Sdk">
+<Project Sdk="Microsoft.NET.Sdk">
```
Dexterp37, line 1: What changed on this line?

glean-core/csharp/Glean/Net/BaseUploader.cs
travis79: Looks like [c# string interpoloation](https://docs.microsoft.com/en-us/dotnet/csharp/tutorials/string-interpolation) is built on top of composite formatting, so this may not be a performance issue with the way Serilog caches templates after all, so yes, we can use the string interpoloation formatting here and other places.  This is cool because I much prefer the readability of that over composite.

glean-core/csharp/Glean/Glean.csproj
travis79: Um... I don't know?  Line ending maybe?  This is weird.

travis79: That doesn't show up as a change on my local diff, just the two Package Refs below. Maybe something weird with GitHub?

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -130,7 +142,7 @@ protected override bool ReleaseHandle()
             {
                 if (!this.IsInvalid)
                 {
-                    Console.WriteLine("Freeing string handle");
+                    Log.Information("Freeing string handle");
```
Dexterp37, line 145: This was a stray print. Would you kindly remove this?

glean-core/csharp/Glean/LibGleanFFI.cs
```diff
@@ -114,6 +116,16 @@ public static string GetFromRustString(IntPtr pointer)
...
+            /// <summary>
+            /// This is the tag used for logging from this class.
+            /// </summary>
+            private const string LogTag = "glean/StringAsReturnValue";
+
+            /// <summary>
+            /// A logger configured for this class
+            /// </summary>
+            private static readonly ILogger Log = GetLogger(LogTag);
+
```
Dexterp37, line 128: No need for these, here, see below

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -68,7 +80,7 @@ internal static (string, string)[] GetHeadersFromJSONString(string documentId, s
             }
             catch (JsonException)
```
Dexterp37, line 81:
```suggestion
            catch (JsonException e)
```

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -68,7 +80,7 @@ internal static (string, string)[] GetHeadersFromJSONString(string documentId, s
...
             {
-                Console.WriteLine($"Error while parsing headers for ping {documentId}");
+                Log.Error($"Error while parsing headers for ping {documentId}");
```
Dexterp37, line 83:
```suggestion
                Log.Error(e, $"Error while parsing headers for ping {documentId}");
```

travis79:
Using Admin rights to override the unrelated failing Android tests.  Note, the intermittent failures are being tracked in [Bug 1646253](https://bugzilla.mozilla.org/show_bug.cgi?id=1646253)

PR #989, opened 2020-06-19
Adding C# DatetimeMetricType

daoshengmu:
- [x] Adding DatetimeMetricType class
- [x] Adding DatetimeMetricTypeTest

There is an uncertain thing regarding to why the month of date has a month difference [1]. I know we added a month at [2]. But, in that way, it will make the current timestamp dismatch to the timestamp from `GetValue()` [3].

@Dexterp37 do you know the reason?

[1]  https://searchfox.org/glean/rev/01d30f83427d2a727694ae5dae03221d6fee72dc/glean-core/android/src/test/java/mozilla/telemetry/glean/private/DatetimeMetricTypeTest.kt#47-51
[2] https://searchfox.org/glean/rev/01d30f83427d2a727694ae5dae03221d6fee72dc/glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt#81
[3] https://searchfox.org/glean/rev/01d30f83427d2a727694ae5dae03221d6fee72dc/glean-core/android/src/test/java/mozilla/telemetry/glean/private/DatetimeMetricTypeTest.kt#112-118


daoshengmu:
In C#, `FindSystemTimeZoneById` methods looking for a time zone identifier on the local system. Otherwise, it will throw a `TimeZoneNotFoundException ` exception.

Because the registry serves as the repository of time zone information in Windows XP and Windows Vista, this exception indicates that the registry contains no information about a particular time zone. Time zone information is stored in the subkeys of HKEY_LOCAL_MACHINE\Software\Microsoft\Windows NT\CurrentVersion\Time Zones. [1]

Therefore, the best way to have a timezone offset is by creating `CustomTimeZone` if we need a timezone offset.

[1] https://docs.microsoft.com/en-us/dotnet/api/system.timezonenotfoundexception?view=netcore-3.1

PR #993, opened 2020-06-22
Fix the `Date` header generated by glean-core

Dexterp37:
Drop the sub-second precision and report GMT
as timezone identifier, as required by https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date

### DIFF
--- a/glean-core/src/upload/request.rs
+++ b/glean-core/src/upload/request.rs
@@ -88,11 +88,25 @@ impl PingRequest {
         gzipper.finish().ok()
     }
 
+    /// Creates a formatted date string that can be used with Date headers.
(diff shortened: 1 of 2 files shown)


glean-core/src/upload/request.rs
```diff
@@ -88,11 +88,25 @@ impl PingRequest {
         gzipper.finish().ok()
     }
 
+    /// Creates a formatted date string that can be used with Date headers.
+    fn create_date_header_value(current_time: DateTime<Utc>) -> String {
```
badboy, line 92: Not wrong or anything, but this could just be a free-standing function in this module.

Dexterp37: Next time @badboy , next time! :-)

PR #994, opened 2020-06-22
Provide a better sample app for C#

Dexterp37:
This moves the sample under the same solution as the Glean SDK, in order to allow building the sample without confusing Visual Studio.

This additionally manually adds a ping and its metrics.

Ping tagging works already, as shown [here](https://debug-ping-preview.firebaseapp.com/pings/csharp-test).


Dexterp37:
> 
> 
> I'm glad solution files allow for projects in different places in the directories!

Me too! That was a nice discovery this morning :)

PR #998, opened 2020-06-23
Bug 1640575 - Move "debug view tagging" to Rust

brizental:
Fixes [Bug 1640575](https://bugzilla.mozilla.org/show_bug.cgi?id=1640575).

~As a bonus also fixes (or at least partially fixes) [Bug 1605097](https://bugzilla.mozilla.org/show_bug.cgi?id=1605097), because persisting the tag was the best way I found to move it to Rust.~ Not anymore, this was defered to a later PR. Because the tag was not being persisted any longer, I had to drop the commit where I removed the debug view tag management from python in favour of the rust implementation. Since python instantiates a standalone upload manager to do the upload, without persisting the tag we don't pck up on it during upload. When I work on the PR to persist the headers I can bring the python commit back.

I realized after the fact that this could be broken down into smaller PRs, which can stil be done if you prefer since this is already broken down into coherent commits. Please let me know and I'll close this if necessary.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -586,6 +586,21 @@ open class GleanInternalAPI internal constructor () {
     }
 
     /**
+     * Set a tag to be applied to headers when uploading pings for debug view.
+     * This is only meant to be used internally by the `GleanDebugActivity`.
+     *
+     * @param value The value of the tag, which must be a valid HTTP header value.
+     */
+    fun setDebugViewTag(value: String): Boolean {
+        if (isInitialized()) {
+            return LibGleanFFI.INSTANCE.glean_set_debug_view_tag(value).toBoolean()
+        } else {
+            Log.e(LOG_TAG, "Glean must be initialized before setting a debug view tag.")
+            return false
+        }
+    }
+
+    /**
      * TEST ONLY FUNCTION.
      * This is called by the GleanTestRule, to enable test mode.
      *
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
@@ -37,9 +37,6 @@ class GleanDebugActivity : Activity() {
          * The value must match the pattern `[a-zA-Z0-9-]{1,20}`.
          */
         const val TAG_DEBUG_VIEW_EXTRA_KEY = "tagPings"
-
-        // Regular expression filter for debugId
-        internal val pingTagPattern = "[a-zA-Z0-9-]{1,20}".toRegex()
     }
 
     // IMPORTANT: These activities are unsecured, and may be triggered by
@@ -83,19 +80,15 @@ class GleanDebugActivity : Activity() {
 
             // Check for ping debug view tag to apply to the X-Debug-ID header when uploading the
             // ping to the endpoint
-            var pingTag: String? = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
+            var debugViewTag: String? = intent.getStringExtra(TAG_DEBUG_VIEW_EXTRA_KEY)
 
-            // Validate the ping tag against the regex pattern
-            pingTag?.let {
-                if (!pingTagPattern.matches(it)) {
-                    Log.e(LOG_TAG, "tagPings value $it does not match accepted pattern $pingTagPattern")
-                    pingTag = null
-                }
+            // Set the debug view tag, if the tag is invalid it won't be set
+            debugViewTag?.let {
+                Glean.setDebugViewTag(debugViewTag)
             }
 
             val debugConfig = Glean.configuration.copy(
-                logPings = intent.getBooleanExtra(LOG_PINGS_EXTRA_KEY, Glean.configuration.logPings),
-                pingTag = pingTag
+                logPings = intent.getBooleanExtra(LOG_PINGS_EXTRA_KEY, Glean.configuration.logPings)
             )
 
             // Finally set the default configuration before starting
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
@@ -104,14 +104,14 @@ class GleanTest {
     }
 
     @Test
-    fun `X-Debug-ID header is correctly added when pingTag is not null`() {
+    fun `X-Debug-ID header is correctly added when debug view tag is set`() {
         val server = getMockWebServer()
         resetGlean(context, Glean.configuration.copy(
             serverEndpoint = "http://" + server.hostName + ":" + server.port,
-            logPings = true,
-            pingTag = "this-ping-is-tagged"
+            logPings = true
         ))
 
+        Glean.setDebugViewTag("this-ping-is-tagged")
         Glean.handleBackgroundEvent()
 
         // Now trigger it to upload
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -642,6 +642,8 @@ uint8_t glean_quantity_test_has_value(uint64_t metric_id, FfiStr storage_name);
 
 void glean_register_ping_type(uint64_t ping_type_handle);
 
+uint8_t glean_set_debug_view_tag(FfiStr tag);
+
 void glean_set_dirty_flag(uint8_t flag);
 
 void glean_set_experiment_active(FfiStr experiment_id,
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -492,4 +492,12 @@ pub extern "C" fn glean_initialize_standalone_uploader(data_dir: FfiStr) -> u8 {
     })
 }
 
+#[no_mangle]
+pub extern "C" fn glean_set_debug_view_tag(tag: FfiStr) -> u8 {
+    with_glean_mut(|glean| {
+        let tag = tag.to_string_fallible()?;
+        Ok(glean.set_debug_view_tag(&tag))
+    })
+}
+
 define_string_destructor!(glean_str_free);
--- a/glean-core/ios/Glean/Config/Configuration.swift
+++ b/glean-core/ios/Glean/Config/Configuration.swift
@@ -8,7 +8,6 @@ public struct Configuration {
     let serverEndpoint: String
     var logPings: Bool
     let maxEvents: Int32?
-    var pingTag: String?
     let channel: String?
 
     struct Constants {
@@ -24,21 +23,17 @@ public struct Configuration {
     ///   * logPings: whether to log ping contents to the console.
     ///     This is only meant to be used internally by the `GleanDebugActivity`.
     ///   * maxEvents: the number of events to store before the events ping is sent.
-    ///   * pingTag: String tag to be applied to headers when uploading pings for debug view.
-    ///     Used internally by the `GleanDebugActivity`.
     ///   * channel: the release channel the application is on, if known.
     ///     This will be sent along with all the pings, in the `client_info` section.
     internal init(
         serverEndpoint: String = Constants.defaultTelemetryEndpoint,
         logPings: Bool = Constants.defaultLogPings,
         maxEvents: Int32? = nil,
-        pingTag: String? = nil,
         channel: String? = nil
     ) {
         self.serverEndpoint = serverEndpoint
         self.logPings = logPings
         self.maxEvents = maxEvents
-        self.pingTag = pingTag
         self.channel = channel
     }
 
@@ -56,7 +51,6 @@ public struct Configuration {
         self.serverEndpoint = serverEndpoint ?? Constants.defaultTelemetryEndpoint
         self.logPings = Constants.defaultLogPings
         self.maxEvents = maxEvents
-        self.pingTag = nil
         self.channel = channel
     }
 
--- a/glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
+++ b/glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
@@ -19,25 +19,6 @@ class GleanDebugUtilityTests: XCTestCase {
         OHHTTPStubs.removeAllStubs()
     }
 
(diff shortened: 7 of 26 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -104,14 +104,14 @@ class GleanTest {
     }
 
     @Test
-    fun `X-Debug-ID header is correctly added when pingTag is not null`() {
+    fun `X-Debug-ID header is correctly added when debug view tag is set`() {
```
brizental, line 107: Is it weird that I left these tests for checking the `X-Debug-Id` header value in the bindings tests? Should I remove them and just add a test to [`request.rs`](https://github.com/mozilla/glean/blob/main/glean-core/src/upload/request.rs)  about it?

I'm commenting about this in the android test, but there are tests for these in python and swift also.

glean-core/preview/src/lib.rs
```diff
@@ -207,5 +207,23 @@ pub fn submit_ping_by_name(ping: &str, reason: Option<&str>) -> bool {
...
+///
+/// Returns true if the tag was set, false otherwise.
+pub fn set_debug_view_tag(value: &str) -> bool {
+    with_glean_mut(|glean| glean.set_debug_view_tag(value))
+}
+
+/// Remove the debug view tag.
+///
+/// See `glean_core::Glean.clear_debug_view_tag`.
+pub fn clear_debug_view_tag() {
```
brizental, line 224: Do we really need a function to clear the tag? I supposed that since we have one to set the tag we must need a way to clear it, but turns out none of the other bindings actually have a way to clear the tag once it's set.

glean-core/src/lib.rs
```diff
@@ -689,6 +696,34 @@ impl Glean {
         self.is_first_run
     }
 
+    /// Set a debug view tag.
```
Dexterp37, line 699: Since this is a public API, can you expand a bit on what a "debug view tag" is and what happens to pings tagged this way? You can also link to the docs, if needed.

glean-core/src/util.rs
```diff
@@ -271,3 +271,39 @@ mod test {
...
 }
+
+/// The debug view tag is the value for the `X-Debug-Id` header of tagged ping requests,
+/// thus is it must be a valid header value.
+///
+/// In other words, it must match the regex: "[a-zA-Z0-9-]{1,20}"
+///
+/// The regex crate isn't used here because it adds to the binary size,
+/// and the Glean SDK doesn't use regular expressions anywhere else.
+pub fn validate_debug_view_tag(value: &str) -> Option<String> {
```
Dexterp37, line 282: Can you add unit testing for this?

glean-core/src/upload/request.rs
```diff
@@ -32,7 +32,12 @@ impl PingRequest {
     ///
     /// Automatically creates the default request headers.
     /// Clients may add more headers such as `userAgent` to this list.
-    pub fn new(document_id: &str, path: &str, body: JsonValue) -> Self {
+    pub fn new(
+        document_id: &str,
+        path: &str,
+        body: JsonValue,
+        debug_view_tag: Option<&String>,
```
Dexterp37, line 39: Can you document the parameters?

glean-core/src/lib.rs
```diff
@@ -689,6 +696,34 @@ impl Glean {
...
+    ///
+    /// This will return `false` in case `value` is not a valid tag.
```
Dexterp37, line 701: Can you document `value`, its limits and its expected behaviour?

glean-core/src/ping/mod.rs
```diff
@@ -282,6 +283,10 @@ impl PingMaker {
             file.write_all(url_path.as_bytes())?;
             file.write_all(b"\n")?;
             file.write_all(::serde_json::to_string(ping_content)?.as_bytes())?;
+            if let Some(tag) = debug_view_tag {
+                file.write_all(b"\n")?;
```
Dexterp37, line 287: Can we split this out to a separate PR? I'm not sure it makes sense to handle this here and now: since we're going to add other tags that we'll need to persist, we might need to provide a flexible solution here from day 0, to not need to care about migrating pings in a different format.

glean-core/preview/src/lib.rs
Dexterp37: This might be needed for fog. @badboy ?

In general, though, I'm not sure it makes sense to keep "preview" up to date. What's your take on it, @badboy ?

glean-core/preview/src/lib.rs
```diff
@@ -207,5 +207,23 @@ pub fn submit_ping_by_name(ping: &str, reason: Option<&str>) -> bool {
...
 }
 
+/// Add a debug tag to pings.
+///
+/// See `glean_core::Glean.set_debug_view_tag`.
+///
+/// ## Return value
+///
+/// Returns true if the tag was set, false otherwise.
+pub fn set_debug_view_tag(value: &str) -> bool {
```
Dexterp37, line 217: Do we need a public facing API for these? On Android and Swift, we expose other APIs to tag pings. On non-mobile platforms, we'll add environment variable detection within the SDK itself. I'd rather not provide an ad-interim public API

brizental: That's I good question. I am not sure really, I added this to the preview like  added it to other bindings and just didn't think about it very deeply. If we are going to set debug options through enviroment variables only, it makes more sense to just remove the commit for the preview changes from this PR altogether and wait for when I work on getting this options from the environment.

glean-core/preview/src/lib.rs
brizental: I will remove this commit for the `preview` changes for now. It's easy enough to bring it back later for a future PR and it is very decoupled from this PR anyways.

glean-core/src/ping/mod.rs
brizental: I removed this changes from this PR, we can discuss how to better persist this header value when I deal with that specific bug, as suggested. 

Now, this made python stop working, which makes sense since the tag is now on the Glean object and python actually instantiates a standalone instance of the Upload Manager for doing the uploads. I will have to drop the python changes from this PR too, so that nothing is broken. When we persist the headers then we can propagate this changes there.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
mdboom: I think it's fine to leave the test in the binding, even though it no longer happens there, just to confirm that it continues to work.

glean-core/ffi/glean.h
```diff
@@ -359,6 +361,8 @@ char *glean_datetime_test_get_value_as_string(uint64_t metric_id, FfiStr storage
 
 uint8_t glean_datetime_test_has_value(uint64_t metric_id, FfiStr storage_name);
 
+char *glean_debug_view_tag(void);
```
mdboom, line 364: Maybe we should remove this (and the Swift test function that uses it)?  I don't think there's a real use case for *getting* it outside of Rust, and we can test that it's working by just looking at the HTTP headers.

glean-core/ffi/src/lib.rs
```diff
@@ -489,4 +489,27 @@ pub extern "C" fn glean_initialize_standalone_uploader(data_dir: FfiStr) -> u8 {
...
+    })
+}
+
+#[no_mangle]
+pub extern "C" fn glean_clear_debug_view_tag() {
+    with_glean_value_mut(|glean| glean.clear_debug_view_tag())
+}
+
+#[no_mangle]
+pub extern "C" fn glean_debug_view_tag() -> *mut c_char {
```
mdboom, line 506: As above, maybe we remove this?

glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
```diff
@@ -25,17 +25,17 @@ class GleanDebugUtilityTests: XCTestCase {
...
         Glean.shared.handleCustomUrl(url: url!)
-        XCTAssertNil(Glean.shared.configuration?.pingTag)
+        XCTAssertEqual(Glean.shared.testDebugViewTag(), "")
 
         // Check valid tag: This should update the pingTag value of the
         // configuration object.
         url = URL(string: "test://glean?tagPings=Glean-Ping")
         Glean.shared.handleCustomUrl(url: url!)
-        XCTAssertEqual("Glean-Ping", Glean.shared.configuration?.pingTag)
+        XCTAssertEqual("Glean-Ping", Glean.shared.testDebugViewTag())
```
mdboom, line 38: Maybe there's a way to test this by looking at the HTTP headers -- or, honestly, I'd be ok with just removing this test case.

glean-core/src/lib.rs
```diff
@@ -689,6 +695,41 @@ impl Glean {
...
+    }
+
+    /// Remove the debug view tag.
+    pub fn clear_debug_view_tag(&mut self) {
+        self.debug_view_tag = None;
+    }
+
+    /// Return the value for the debug view tag or `None` if it hasn't been set.
+    pub fn debug_view_tag(&self) -> Option<&String> {
+        self.debug_view_tag.as_ref()
```
mdboom, line 730: I find it a bit confusing to have a member with the same name as a method here.  But maybe this is idiomatic Rust?  @badboy?

glean-core/src/upload/mod.rs
```diff
@@ -147,14 +147,20 @@ impl PingUploadManager {
     }
 
     /// Creates a `PingRequest` and adds it to the queue.
-    pub fn enqueue_ping(&self, document_id: &str, path: &str, body: JsonValue) {
+    pub fn enqueue_ping(
+        &self,
+        document_id: &str,
+        path: &str,
+        body: JsonValue,
+        debug_ping_value: Option<&String>,
```
mdboom, line 155: Nit: I think `debug_ping_tag` might be more descriptive for this variable.

brizental: That's a mistake. During development I interchangeably used `debug_view_tag` and `debug_ping_tag`, but the latter was used by mistake. I thought I had removed all these cases, but clearly not :) Thansk for pointing that out!

glean-core/src/lib.rs
brizental: I just followed the convention I saw for other getters in the Glean rust struct, such as [`start_time`](https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L629).

glean-core/preview/src/lib.rs
Dexterp37: We'll allow both: on Android and iOS, we'll support both environment variable and command-line based workflows

glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
brizental: I removed this and added a test to the rust upload module about the X-Debug-ID header just to be sure.

glean-core/ffi/glean.h
brizental: Removed this one and made it onl;y visible to the rust core crate. I also removed the`clear_debug_view_tag` function, which is not used anywhere. As much as it is wierd to me that add a `set` without a `remove`/`clear` there is no use case for it. If ever there is it is a simple enough function to just re-add.

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
```diff
@@ -83,19 +80,15 @@ class GleanDebugActivity : Activity() {
...
 
-            // Validate the ping tag against the regex pattern
-            pingTag?.let {
-                if (!pingTagPattern.matches(it)) {
-                    Log.e(LOG_TAG, "tagPings value $it does not match accepted pattern $pingTagPattern")
-                    pingTag = null
-                }
+            // Set the debug view tag, if the tag is invalid it won't be set
+            debugViewTag?.let {
+                Glean.setDebugViewTag(debugViewTag)
```
Dexterp37, line 87: We might not always be already initialized at this point, we're actually racing against `Glean.initialize` here. This will be especially true in Fenix.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -585,6 +585,21 @@ open class GleanInternalAPI internal constructor () {
...
     }
 
+    /**
+     * Set a tag to be applied to headers when uploading pings for debug view.
+     * This is only meant to be used internally by the `GleanDebugActivity`.
+     *
+     * @param value The value of the tag, which must be a valid HTTP header value.
+     */
+    fun setDebugViewTag(value: String): Boolean {
+        if (isInitialized()) {
```
Dexterp37, line 595: This might be intermittently failing when QA is testing, which is bad. Can we do as we do in [setUploadEnabled](https://searchfox.org/glean/rev/c6e1349fdb05baec59aaa5f465ba65e6cdd3b7d3/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#289) and cache the tag internally, then apply it when glean inits?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -9,7 +9,6 @@ public struct Configuration {
     let userAgent: String
     var logPings: Bool
     let maxEvents: Int32?
-    var pingTag: String?
```
Dexterp37, line 12: Would you kindly file a bug for removing this from the C# bindings too _after_ we have environment variables ready?=

glean-core/src/lib.rs
badboy: Don't see a problem with that. It's idiomatic to provide "getters" that way
(Making individual fields of a struct "pub" makes them both readable and writable if you own the instance)

glean-core/ios/Glean/Config/Configuration.swift
brizental: Sorry, didn't think to check the C# bindings here. Filed the bug https://bugzilla.mozilla.org/show_bug.cgi?id=1648007

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
brizental: Filed a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1648009

PR #1000, opened 2020-06-23
Bug 1646441 - Remove options to set custom user agent header

brizental:
Fixes [Bug 1646441](https://bugzilla.mozilla.org/show_bug.cgi?id=1646441) and relates to [Bug 1646626](https://bugzilla.mozilla.org/show_bug.cgi?id=1646626)

Although this just removes user agent related code from all bindings it doesn't really change anything since we are currently not setting the user agent header at all. I will follow with a PR to do that right after this one.


glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -110,10 +106,6 @@ data class Configuration internal constructor(
          * The default server pings are sent to.
          */
         const val DEFAULT_TELEMETRY_ENDPOINT = "https://incoming.telemetry.mozilla.org"
-        /**
-         * The default user agent used when sending pings.
-         */
-        const val DEFAULT_USER_AGENT = "Glean/${BuildConfig.LIBRARY_VERSION} (Android)"
```
Dexterp37, line 116: Can you double check that the `BuildConfig.LIBRARY_VERSION` value is the same we'd see in glean-core? It should be :)

glean-core/android/src/test/java/mozilla/telemetry/glean/net/BaseUploaderTest.kt
```diff
@@ -18,9 +18,7 @@ class BaseUploaderTest {
     private val testPath: String = "/some/random/path/not/important"
     private val testPing: String = "{ 'ping': 'test' }"
     private val testHeaders: HeadersList = mutableListOf(Pair("X-Test-Glean", "nothing-to-see-here"))
-    private val testDefaultConfig = Configuration().copy(
-        userAgent = "Glean/Test 25.0.2"
-    )
+    private val testDefaultConfig = Configuration().copy()
```
Dexterp37, line 21:
```suggestion
    private val testDefaultConfig = Configuration()
```

No need to copy: copy is useful when overriding specific fields :)

glean-core/android/src/test/java/mozilla/telemetry/glean/net/HttpURLConnectionUploaderTest.kt
```diff
@@ -35,9 +35,7 @@ import java.net.URI
 class HttpURLConnectionUploaderTest {
     private val testPath: String = "/some/random/path/not/important"
     private val testPing: String = "{ 'ping': 'test' }"
-    private val testDefaultConfig = Configuration().copy(
-        userAgent = "Glean/Test 25.0.2"
-    )
+    private val testDefaultConfig = Configuration().copy()
```
Dexterp37, line 38:
```suggestion
    private val testDefaultConfig = Configuration()
```

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
brizental: Yes it is.

brizental:
Android failure is unrelated. See: https://bugzilla.mozilla.org/show_bug.cgi?id=1648397

PR #1001, opened 2020-06-23
TC: Limit CI runs to default branch only 

badboy:
[skip ci]

---

cc @MihaiTabara


badboy:
Android intermittent. Merging anyway.

PR #1003, opened 2020-06-23
Enable sending well-formed pings from the C# bindings

Dexterp37:
This enables and populates all the core metrics required to produce well-formed Glean pings.

Some sample pings generated with the C# sample app are available [on the Glean debug view](https://debug-ping-preview.firebaseapp.com/pings/csharp-valid-ping). The one that does not validate is old :)


Dexterp37:
@daoshengmu would you kindly review as well?

glean-core/csharp/Glean/Glean.cs
```diff
@@ -334,13 +336,55 @@ internal void TestDestroyGleanHandle()
...
+            }
+
+            if (configuration.channel != null)
+            {
+                GleanInternalMetrics.appChannel.SetSync(configuration.channel);
+            }
+
+            // Try to get the version of the product using the Glean SDK. Unfortunately,
+            // this uses reflection.
+            var mainAssembly = System.Reflection.Assembly.GetEntryAssembly();
```
travis79, line 377: Ugh, I was hoping we could avoid refection but I don't expect there is much choice here.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -334,13 +336,55 @@ internal void TestDestroyGleanHandle()
...
+            if (mainAssembly != null)
+            {
+                GleanInternalMetrics.appDisplayVersion.SetSync(mainAssembly.GetName().Version.ToString());
+            }
+            else
+            {
+                GleanInternalMetrics.appDisplayVersion.SetSync("Unknown");
+            }
+
+            GleanInternalMetrics.appBuild.SetSync(configuration.buildId ?? "Unknown");
```
daoshengmu, line 387: In Python, we were using `_application_id` and `_application_version`
```
 metrics.glean.internal.metrics.app_build.set(cls._application_id)
 if cls._application_version is not None:
      metrics.glean.internal.metrics.app_display_version.set(
         cls._application_version)
```
I suppose we might be able to consider to adopt {`applicationId, applicationVersion`} from `Glean Initialize()`

glean-core/csharp/Glean/Glean.cs
```diff
@@ -334,13 +336,55 @@ internal void TestDestroyGleanHandle()
...
+
+            // Possible values for `RuntimeInformation.ProcessArchitecture` are documented here:
+            // https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.architecture?view=netstandard-2.0
+            GleanInternalMetrics.architecture.SetSync(RuntimeInformation.ProcessArchitecture.ToString());
+
+            try
+            {
+                // CurrentUiCulture is used for representing the locale of the UI, coming from the OS,
+                // while CurrentCulture is the general locale used for other things (e.g. currency).
+                GleanInternalMetrics.locale.SetSync(CultureInfo.CurrentUICulture.Name);
```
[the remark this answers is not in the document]
mdboom: We definitely want `en-US`.  The `English-United States` thing in Python is a bug (see #1004).

daoshengmu:
The difference among bindings might not affect too much because every application has its own table. I suppose an application should not be possible to use two even more different Glean bindings at the same time.

glean-core/csharp/Glean/Glean.cs
Dexterp37: I know :( I was very disappointed about this :(

glean-core/csharp/Glean/Glean.cs
```diff
@@ -334,13 +336,55 @@ internal void TestDestroyGleanHandle()
...
+            // The `Environment.OSVersion` object will return a version that represents an approximate
+            // version of the OS. As the MSDN docs state, this is unreliable:
+            // https://docs.microsoft.com/en-us/dotnet/api/system.environment.osversion?redirectedfrom=MSDN&view=netstandard-2.0
+            // However, there's really nothing we could do about it. Unless the product using the
+            // Glean SDK correctly marks their manifest as Windows 10 compatible, this API will report
+            // Windows 8 version (6.2). See the remarks section here:
+            // https://docs.microsoft.com/en-gb/windows/win32/api/winnt/ns-winnt-osversioninfoexa?redirectedfrom=MSDN#remarks
+            try
+            {
+                GleanInternalMetrics.osVersion.SetSync(Environment.OSVersion.Version.ToString());
```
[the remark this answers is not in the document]
Dexterp37: Nope, we don't provide any guarantee of consistency on the values there. If this ever becomes a future requirement, we'll deal with that.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -334,13 +336,55 @@ internal void TestDestroyGleanHandle()
...
+            }
+            catch (InvalidOperationException)
+            {
+                GleanInternalMetrics.osVersion.SetSync("Unknown");
+            }
+
+            // Possible values for `RuntimeInformation.ProcessArchitecture` are documented here:
+            // https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.architecture?view=netstandard-2.0
+            GleanInternalMetrics.architecture.SetSync(RuntimeInformation.ProcessArchitecture.ToString());
```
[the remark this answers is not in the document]
Dexterp37: Oh, good catch, we want os architecture here.

glean-core/csharp/Glean/Glean.cs
Dexterp37: In Python we're doing the wrong thing: it's a bug, we're setting the wrong value for `app_build` (see [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1647782)). Moreover, in Python I'm not sure we can determine the application version easily, while this is possible in C#. We're currently providing the `buildId` from the configuration here, which is something that we won't be able to infer automatically.

Dexterp37:
> 
> 
> The difference of environment info among bindings might not affect too much because every application has its own table. I suppose an application should not be possible to use two even more different Glean bindings at the same time.

Truth to be told, we will be able to use multiple bindings in a single product (in Firefox, for example). However, only on language binding will be allowed to call `initialize()`. The values that are being populated will be coming from that binding and they will be consistent.

PR #1005, opened 2020-06-23
Making C# Configuration.DefaultTelemetryEndpoint be able to be accessed in public.

daoshengmu:
I think it makes sense to do this adjustment when we need to give `channel` name after `endpoint` for initializing Glean in C#.

```
 GleanInstance.Initialize(
                applicationId: "org.mozilla.csharp.tests",
                applicationVersion: "1.0-test",
                uploadEnabled: true,
                configuration: new Configuration(Configuration.DefaultTelemetryEndpoint, "fxr"),
                dataDir: tempDataDir
 );
```

### DIFF
--- a/glean-core/csharp/Glean/Configuration/Configuration.cs
+++ b/glean-core/csharp/Glean/Configuration/Configuration.cs
@@ -11,7 +11,7 @@ namespace Mozilla.Glean
         /// <summary>
         /// The default server pings are sent to.
         /// </summary>
-        public const string DefaultTelemetryEndpoint = "https://incoming.telemetry.mozilla.org";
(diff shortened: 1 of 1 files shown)


glean-core/csharp/Glean/Configuration/Configuration.cs
```diff
@@ -11,7 +11,7 @@ public sealed class Configuration
         /// <summary>
         /// The default server pings are sent to.
         /// </summary>
-        public const string DefaultTelemetryEndpoint = "https://incoming.telemetry.mozilla.org";
+        public static readonly string DefaultTelemetryEndpoint = "https://incoming.telemetry.mozilla.org";
```
Dexterp37, line 14: Why was this changed?

daoshengmu:
I confirmed. We don't need to make this change :)

PR #1010, opened 2020-06-24
CI: Remove Android build tasks from CircleCI

badboy:
They take a total of about 20 Minutes on CI.
We don't need the artifacts.
We got TaskCluster now to generate the artifacts AND publish straight to Maven.


badboy:
Just plain CI removal. Going ahead without review.

PR #1012, opened 2020-06-24
Bug 1648140: Python: Make sure core metrics are in every ping

mdboom:
With the recent changes to make Python threading more like the other platforms, this workaround for setting the core metrics that already exists in the other bindings was missed.

@hackebrot: Tested inside of Burnham and known to resolve the issue there.


badboy:
Test failure reported in https://bugzilla.mozilla.org/show_bug.cgi?id=1648397.

Ready for merge.

PR #1016, opened 2020-06-25
Bug 1609968 - Remove locale from baseline ping

brizental:
Fixes [Bug 1609968](https://bugzilla.mozilla.org/show_bug.cgi?id=1609968)

Opening this as a draft because I believe the `send_if_empty` addition to the baseline ping should be discussed. 

Removing the locale means that when `foreground` or `dirty_startup` the baseline ping will have no metrics, so I believe adding this flag makes sense. If everyone agrees on that, still this would need to wait for a fix in the glean_parser, because the `send_if_empty` breaks the baseline ping documentation.

### DIFF
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
@@ -232,19 +232,16 @@ class GleanTest {
                     assertEquals(1, json.getJSONArray("events").length())
                 } else if (docType == "baseline") {
                     val seq = json.getJSONObject("ping_info").getInt("seq")
-                    val baselineMetricsObject = json.getJSONObject("metrics")
-                    val baselineStringMetrics = baselineMetricsObject.getJSONObject("string")
-                    assertEquals(1, baselineStringMetrics.length())
-                    assertNotNull(baselineStringMetrics.get("glean.baseline.locale"))
 
                     // There are three baseline pings:
                     //   - seq: 0, reason: foreground, duration: null
                     //   - seq: 1, reason: background, duration: non-null
                     //   - seq: 2, reason: foreground, duration: null
                     if (seq == 0 || seq == 2) {
-                        assertFalse(baselineMetricsObject.has("timespan"))
+                        assertFalse(json.has("metrics"))
                         assertEquals("foreground", json.getJSONObject("ping_info").getString("reason"))
                     } else if (seq == 1) {
+                        val baselineMetricsObject = json.getJSONObject("metrics")
                         assertEquals("background", json.getJSONObject("ping_info").getString("reason"))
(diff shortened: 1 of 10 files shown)


glean-core/pings.yaml
```diff
@@ -17,6 +17,7 @@ baseline:
     The `baseline` ping is automatically sent when the application is moved to
     the background.
   include_client_id: true
+  send_if_empty: true
```
brizental, line 20: Would we need data review for this addition?

Dexterp37: That's a good question for @chutten

chutten: To my understanding the original data reviews cover sending the "baseline" ping whether it is empty or not, at rates up to the stated frequency. No data review needed in my mind.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -232,19 +232,16 @@ class GleanTest {
...
-                    assertEquals(1, baselineStringMetrics.length())
-                    assertNotNull(baselineStringMetrics.get("glean.baseline.locale"))
 
                     // There are three baseline pings:
                     //   - seq: 0, reason: foreground, duration: null
                     //   - seq: 1, reason: background, duration: non-null
                     //   - seq: 2, reason: foreground, duration: null
                     if (seq == 0 || seq == 2) {
-                        assertFalse(baselineMetricsObject.has("timespan"))
+                        assertFalse(json.has("metrics"))
```
Dexterp37, line 241: I think we can just drop this.

glean-core/metrics.yaml
```diff
@@ -25,24 +25,6 @@ glean.baseline:
     notification_emails:
       - glean-team@mozilla.com
     expires: never
-  locale:
```
Dexterp37, line 28: You should check with @fbertsch if it's fine for us to remove this. Or maybe @mdboom knows as well!

I believe it should be fine.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
Dexterp37: Oh, no, wait, we don't. Ignore this :)

glean-core/metrics.yaml
mdboom: Yeah -- it should be fine.  The column will remain in the table (since it exists in historical revisions) so old versions will continue to successfully record it, but new ones will now.

brizental:
> Is this intentionally a draft?

Yeah, was not sure about the `send_if_empty` and even if you all are okay with that this brask the baseline pings documentation due to a bug in the glean_parser, would need to fix that before merge.

PR #1017, opened 2020-06-25
Bug  1648009 - Let debug view tag be set before init on Android and iOS

brizental:
Fixes [Bug 1648009](https://bugzilla.mozilla.org/show_bug.cgi?id=1648009)

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -27,6 +27,7 @@ public class Glean {
 
     var initialized: Bool = false
     private var uploadEnabled: Bool = true
+    private var debugViewTag: String?
     var configuration: Configuration?
     private var observer: GleanLifecycleObserver?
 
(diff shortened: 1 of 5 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -110,6 +111,10 @@ public class Glean {
                 return
             }
 
+            if self.debugViewTag != nil {
```
travis79, line 114: You can fix the cyclomatic complexity lint by either refactoring this function into more functions, or you can just disable it for init by putting the comment `// swiftlint:disable cyclomatic_complexity` before the function and `// swiftlint:enable cyclomatic_complexity` after the function.  This will only disable it for the one function.

brizental: I will disable the lint, because right now this function looks very similar to Android and that simplifies development (at least for me).

PR #1020, opened 2020-06-26
TC: Ensure we're compiling dependencies in non-debug mode. [ci full]

badboy:
[ci full]


badboy:
I used the artifacts from https://firefox-ci-tc.services.mozilla.com/tasks/LzNrrrJ9TOO7snJK6PsN3g in a local build of samples-glean in the AC repository.
They now pass.

A local build without `NDEBUG` fails the tests with:

```
11:00:32 V/InstrumentationResultParser: dlopen failed: cannot locate symbol "stderr" referenced by "libglean_ffi.so"...
```

Test device is the x86 emulator running Android API 21.

PR #1021, opened 2020-06-26
Bug 1648695 - Fix failing rust upload tests on CI

brizental:
The tests on the upload module either used a standalone upload manager or the upload manager that is created when you create a new Glean instance. When creating a standalone one, the tests "wait" for upload manager to be done scanning the pending pings directory. When using the one inside Glean the don't.

This is most probably not causing the CI failures, if it is that would mean I understand those failures even less than I thought. Still, it is a good thing to add to the tests.

Also, opening this PR will allow me to send new commits to test fixes to the CI problem.

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -499,7 +499,7 @@ impl Glean {
         }
 
         self.upload_manager
-            .process_ping_upload_response(uuid, status);
+            .process_ping_upload_response(uuid, status, self.debug_view_tag());
     }
 
     /// Take a snapshot for the given store and optionally clear it.
--- a/glean-core/src/upload/directory.rs
+++ b/glean-core/src/upload/directory.rs
@@ -12,8 +12,10 @@ use std::path::{Path, PathBuf};
 use serde_json::Value as JsonValue;
 use uuid::Uuid;
 
-use super::PingRequest;
 use crate::{DELETION_REQUEST_PINGS_DIRECTORY, PENDING_PINGS_DIRECTORY};
+
+/// A representation of the PingData extracted from a ping file.
+type PingData = (String, String, JsonValue);
 
 /// Get the file name from a path as a &str.
 ///
@@ -82,14 +84,14 @@ impl PingDirectoryManager {
         };
     }
 
-    /// Reads a ping file and returns a `PingRequest` from it.
+    /// Reads a ping file and returns the data from it.
     ///
     /// If the file is not properly formatted, it will be deleted and `None` will be returned.
     ///
     /// ## Arguments
     ///
     /// * `document_id` - The UUID of the ping file to be processed
-    pub fn process_file(&self, document_id: &str) -> Option<PingRequest> {
+    pub fn process_file(&self, document_id: &str) -> Option<PingData> {
         let path = match self.get_file_path(document_id) {
             Some(path) => path,
             None => {
@@ -113,7 +115,7 @@ impl PingDirectoryManager {
         let mut lines = BufReader::new(file).lines();
         if let (Some(Ok(path)), Some(Ok(body))) = (lines.next(), lines.next()) {
             if let Ok(parsed_body) = serde_json::from_str::<JsonValue>(&body) {
-                return Some(PingRequest::new(document_id, &path, parsed_body, None));
+                return Some((document_id.to_string(), path, parsed_body));
             } else {
                 log::warn!(
                     "Error processing ping file: {}. Can't parse ping contents as JSON.",
@@ -130,7 +132,7 @@ impl PingDirectoryManager {
         None
     }
 
-    /// Process the pings directory and return a vector of `PingRequest`s
+    /// Process the pings directory and return a vector with the ping data
     /// corresponding to each valid ping file in the directory.
     /// This vector will be ordered by file `modified_date`.
     ///
@@ -141,8 +143,9 @@ impl PingDirectoryManager {
     ///
     /// # Return value
     ///
-    /// `Vec<PingRequest>` - see [`PingRequest`](struct.PingRequest.html) for more information.
-    pub fn process_dir(&self) -> Vec<PingRequest> {
+    /// `Vec<(String, String, JsonValue)>` -
+    ///     a vector of tuples containing the document_id, path and body of each request.
+    pub fn process_dir(&self) -> Vec<PingData> {
         log::info!("Processing persisted pings.");
 
         // Walk the pings directory and process each file in it,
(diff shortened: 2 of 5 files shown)


brizental:
Okay I think I figured out what is happening and why `a596734` seems to "fix" it. I was not able to reproduce the bug with this commit ever, but removing it made things work.

I think we have a bug in the upload manager, which is: we shouldn't enqueue a ping if the pending pings directory is being scanned, because if a ping is submitted before that is over we may enqueue a ping by submitting and the scanning enqueues that ping again! This would explain the intermittent failure here.

Two possible fixes I can think of are:
1. We don't enqueue a ping if scanning is in process. That would mean maybe creating a temporary queue of pings that we then unload after scanning is done, skipping duplicates or something of the sort.
2. We just don't enqueue if a ping with the same document id is already in the queue. This is simpler.

I will think of a way to prove that this is the case, but even if that is not the what is causing the intermittents we still should fix it. Do you agree @badboy ?

badboy:
Hm, so both scanning and enqueueing lock the queue, so at least they won't override each other.
But what does happen is this:

* We start scanning, it locks the queue and reads the directory
* A ping is submitted and thus written to the disk, then blocks waiting for the queue to be writable.
* The scanning thread picks up that file and enqueues it. It then finishes and releases the lock. 
* The enqueuer unblocks, gets writable access to the queue, then adds the ping to the queue
    * Now the same ping (by document id and file) is in the queue twice.

I think solution 2 sounds better to me. That way we can be sure that ping is properly enqueued (we don't know whether the scanner has seen the file or not).

We should not have _a lot_ of pings enqueued, so we can do a quick search through what is there.
If that's becoming a problem we can switch to a more efficient way.

glean-core/src/upload/directory.rs
```diff
@@ -237,7 +237,7 @@ mod test {
     }
 
     #[test]
-    fn test_creates_requests_correctly_from_valid_ping_file() {
+    fn test_gets_correct_data_from_valid_ping_file() {
```
badboy, line 240: Side-note: (I only noticed that recently again) the `test_` prefix for test functions is not necessary. They are already in a module called `test`, so in the output they show up as `test::test_...` which is kinda double info (no need to change that here now, we have multiple of methods named like that)

glean-core/src/upload/mod.rs
```diff
@@ -219,6 +225,16 @@ impl PingUploadManager {
         self.processed_pending_pings.load(Ordering::SeqCst)
     }
 
+    /// Gets the ping queue and checks in a ping with a certain document_id is already enqueued.
+    fn has_enqueued(
```
badboy, line 229: I think `is_enqueued` is a better name.

glean-core/src/upload/mod.rs
```diff
@@ -219,6 +225,16 @@ impl PingUploadManager {
         self.processed_pending_pings.load(Ordering::SeqCst)
     }
 
+    /// Gets the ping queue and checks in a ping with a certain document_id is already enqueued.
```
badboy, line 228:
```suggestion
    /// Checks if a ping with a certain `document_id` is already enqueued.
```

glean-core/src/upload/mod.rs
```diff
@@ -219,6 +225,16 @@ impl PingUploadManager {
...
+    fn has_enqueued(
+        queue: &RwLockWriteGuard<'_, VecDeque<PingRequest>>,
```
badboy, line 230: We should be able to take just a slice here, the function shouldn't care about the fact it's behind a lock.
 
```suggestion
        queue: &VecDeque<PingRequest>,
```

glean-core/src/upload/mod.rs
```diff
@@ -366,7 +394,12 @@ impl PingUploadManager {
     ///
     /// `document_id` - The UUID of the ping in question.
     /// `status` - The HTTP status of the response.
-    pub fn process_ping_upload_response(&self, document_id: &str, status: UploadResult) {
+    pub fn process_ping_upload_response(
+        &self,
+        document_id: &str,
+        status: UploadResult,
+        debug_view_tag: Option<&String>,
```
badboy, line 401: This parameter is undocumented above.
Will this eventually be removed here again, when the debug view tag is also persisted?

glean-core/src/lib.rs
```diff
@@ -499,7 +499,7 @@ impl Glean {
         }
 
         self.upload_manager
-            .process_ping_upload_response(uuid, status);
+            .process_ping_upload_response(uuid, status, self.debug_view_tag());
```
badboy, line 502: Why was it added in this PR?

glean-core/src/upload/directory.rs
```diff
@@ -12,9 +12,11 @@ use std::path::{Path, PathBuf};
 use serde_json::Value as JsonValue;
 use uuid::Uuid;
 
-use super::PingRequest;
 use crate::{DELETION_REQUEST_PINGS_DIRECTORY, PENDING_PINGS_DIRECTORY};
 
+/// A representation of the PingData extracted from a ping file.
+type PingData = (String, String, JsonValue);
```
badboy, line 18: *puts on a @Dexterp37-hat*
Naming something [`Data`](https://pharr.org/matt/blog/2020/02/28/naming-maxims.html) is not saying much.
Maybe `PingPayload`?

This being a tuple it's also not clear what is what.
At the very least the doc string needs to tell us that the first part is the document ID, the second the path and the JsonValue the actual JSON-encoded payload.

glean-core/src/upload/mod.rs
badboy: or maybe `queue_contains`. 
Nah, I think I like the first better.

brizental: Hum, funny. `is_enqueued` is the name I initally gave to this function and then changed it.

Anyways, I have no strong opinions here, I'll change it.

glean-core/src/upload/mod.rs
brizental: Yes, it wil be removed. I'll still document it.

glean-core/src/lib.rs
brizental: Because when I changed the `process_ping_upload_response` function to use `enqueue_ping` instead of just adding a request to the queue I realized it was not taking the debug tag in consideration.

glean-core/src/upload/directory.rs
brizental: I documented. I agree with a tuple not being very expressive though, it doesn't look too bad when I actually call `process_file`:
```rust
if let Some((document_id, path, body)) =
    self.directory_manager.process_file(document_id)
{
    self.enqueue_ping(&document_id, &path, body, debug_view_tag);
}
```
Do you think it would be better if I make this into a struct?

badboy: Nah, I think for now the tuple is fine. The documentation on the type alias should help.
And in the code it's more obvious.

glean-core/src/upload/mod.rs
badboy: I think you missed that. Your old phrasing has a typo in there and the function doesn't so much "gets the ping queue".
It just checks that something is in the queue it gets passed.

The shorter description captures that.

PR #1023, opened 2020-06-26
Add extra key to event breaks key-value matching of extras.

travis79:


mdboom:
See https://github.com/mozilla/glean_parser/pull/196

Dexterp37:
This is being fixed in #1024.

PR #1024, opened 2020-06-29
Use glean_parser 1.23.0 to fix event's key/value mismatch

Dexterp37:
When 2 or more extra_key are present in event definition, there could be a mismatch between the reported keys and their
value. This PR fixes the potential mismatch.


Dexterp37:
This was manually merged, I had to do this due to a GitHub operational issue. The commit on main is 01644dc0b3cb92d0c87e02dbfa6ec09a44b6228b

PR #1028, opened 2020-06-30
Update glean_parser verion to 1.24.0

brizental:


gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -34,7 +34,7 @@ class GleanMetricsYamlTransform extends ArtifactTransform {
 @SuppressWarnings("GrPackage")
 class GleanPlugin implements Plugin<Project> {
     // The version of glean_parser to install from PyPI.
-    private String GLEAN_PARSER_VERSION = "1.23.0"
+    private String GLEAN_PARSER_VERSION = "1.24.0"
```
Dexterp37, line 37: Did you try to build locally and see if docs were updated? Because they are not included in this PR.

brizental: Yes, I built locally and nothing changed. That is why no changes here.

PR #1030, opened 2020-06-30
Bug 1646626 - Add User-Agent headers to ping requests

brizental:
Fixes [Bug 1646626](https://bugzilla.mozilla.org/show_bug.cgi?id=1646626)

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -52,7 +52,7 @@ data class GleanTimerId internal constructor(internal val id: Long)
 open class GleanInternalAPI internal constructor () {
     companion object {
         private const val LOG_TAG: String = "glean/Glean"
-        private const val BINDING_LANGUAGE_NAME: String = "Kotlin"
+        private const val LANGUAGE_BINDING_NAME: String = "Kotlin"
         internal const val GLEAN_DATA_DIR: String = "glean_data"
     }
 
@@ -156,7 +156,7 @@ open class GleanInternalAPI internal constructor () {
             val cfg = FfiConfiguration(
                 dataDir = gleanDataDir.path,
                 packageName = applicationContext.packageName,
-                bindingLanguageName = BINDING_LANGUAGE_NAME,
+                languageBindingName = LANGUAGE_BINDING_NAME,
                 uploadEnabled = uploadEnabled,
                 maxEvents = configuration.maxEvents,
                 delayPingLifetimeIO = false
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -52,7 +52,7 @@ data class GleanTimerId internal constructor(internal val id: Long)
 open class GleanInternalAPI internal constructor () {
     companion object {
         private const val LOG_TAG: String = "glean/Glean"
-        private const val PLATFORM_NAME: String = "Android"
+        private const val BINDING_LANGUAGE_NAME: String = "Kotlin"
         internal const val GLEAN_DATA_DIR: String = "glean_data"
     }
 
@@ -156,7 +156,7 @@ open class GleanInternalAPI internal constructor () {
             val cfg = FfiConfiguration(
                 dataDir = gleanDataDir.path,
                 packageName = applicationContext.packageName,
-                platform = PLATFORM_NAME,
+                bindingLanguageName = BINDING_LANGUAGE_NAME,
                 uploadEnabled = uploadEnabled,
                 maxEvents = configuration.maxEvents,
                 delayPingLifetimeIO = false
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -52,6 +52,7 @@ data class GleanTimerId internal constructor(internal val id: Long)
 open class GleanInternalAPI internal constructor () {
     companion object {
         private const val LOG_TAG: String = "glean/Glean"
+        private const val PLATFORM_NAME: String = "Android"
         internal const val GLEAN_DATA_DIR: String = "glean_data"
     }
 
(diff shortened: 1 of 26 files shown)


glean-core/preview/src/lib.rs
```diff
@@ -108,6 +109,7 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) -> Result<
         upload_enabled: cfg.upload_enabled,
         data_path: cfg.data_path.clone(),
         application_id: cfg.application_id.clone(),
+        platform: cfg.platform.clone(),
```
brizental, line 112: Should preview have a default value like the other bindings? Something like `Rust on <OS>`.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -53,6 +53,7 @@ data class GleanTimerId internal constructor(internal val id: Long)
 open class GleanInternalAPI internal constructor () {
     companion object {
         private const val LOG_TAG: String = "glean/Glean"
+        private const val PLATFORM_NAME: String = "Android"
```
Dexterp37, line 56: Instead of calling this `PLATFORM` here and in the other places, can we be a bit more specific about what this is about? This is really a `USER_AGENT_PLATFORM`, since we specifically want to use it for the user agent part.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -47,6 +47,11 @@ public sealed class Glean
         /// </summary>
         private const string LogTag = "glean/Glean";
 
+        /// <summary>
+        /// This is the name of the platform this instance is running on.
+        /// </summary>
+        private readonly static string PlatformName = $"C# on {Environment.OSVersion}";
```
Dexterp37, line 53: I think this should really be `Environment.OSVersion.Platform`, otherwise you'll get a version number in this (e.g. "Microsoft Windows NT 6.2.9200.0" vs "Win32NT").

glean-core/ffi/examples/glean_app.c
```diff
@@ -11,6 +11,7 @@ int main(void)
   FfiConfiguration cfg = {
     "/tmp/glean_data",
     "c-app",
+    "C App",
```
Dexterp37, line 14:
```suggestion
    "Sample C App",
```

glean-core/preview/src/lib.rs
Dexterp37: Yes! And you can use [this](https://searchfox.org/glean/rev/a259debf8250d74732d1442ffb0ddd3068234a9b/glean-core/src/lib.rs#683) to get the OS part... actually, can this be used for all the bindings, so that everything can be within glean-core? Maybe we don't need the FFI changes at all.

brizental: >  Maybe we don't need the FFI changes at all.

Oh yes, that would be so much better!

brizental: Only problem is if we use this we don't get tha language name only the OS. Instead of `Python on Windows` we would get `Windows`. I think we need the ffi part, we can definitely remove the OS name responsability from the bindings, but would still need the language binding name. If we are fine with not having the language name, then ok.

Dexterp37: Oh, you're correct. To make things consistent, let's just allow to set the "language bindings name" from the FFI configuration. Let's take the OS from Rust.

glean-core/ffi/examples/glean_app.c
```diff
@@ -11,6 +11,7 @@ int main(void)
   FfiConfiguration cfg = {
     "/tmp/glean_data",
     "c-app",
+    "C",
```
Dexterp37, line 14:
```suggestion
    "C Sample",
```

glean-core/src/upload/request.rs
```diff
@@ -11,6 +11,32 @@ use flate2::{read::GzDecoder, write::GzEncoder, Compression};
...
+    // as documented here:
+    // https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Date
+    // Unfortunately we can't use `current_time.to_rfc2822()` as it
+    // formats as "Mon, 22 Jun 2020 10:40:34 +0000", with an ending
+    // "+0000" instead of "GMT". That's why we need to go with manual
+    // formatting.
+    current_time.format("%a, %d %b %Y %T GMT").to_string()
+}
+
+fn create_user_agent_header_value(language_binding_name: &str) -> String {
```
Dexterp37, line 31: Can you add a unit test for this?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -53,6 +53,7 @@ data class GleanTimerId internal constructor(internal val id: Long)
 open class GleanInternalAPI internal constructor () {
     companion object {
         private const val LOG_TAG: String = "glean/Glean"
+        private const val LANGUAGE_BINDING_NAME: String = "Kotlin"
```
Dexterp37, line 56: This will result in the User-Agent being `Glean/31.x.x (Kotlin on Android)`, right?

glean-core/ffi/examples/glean_app.c
brizental: Ah, I thought this wouldn't apply after the changes. This raises the question if we should be more specific for the preview, benchmark, example. They are all "Rust". Should each be "Rust Preview", "Rust Benchmark" etc. 

Personally I think this adds subjectivity and goes against what we expect this "language_binding_name" to be. On the other hand could be useful. I vote for not doing it, what do you think? That would also mean this wouldn't be "C Sample", but "C".

Dexterp37: Mh, you're correct. Let's go with `C`.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
brizental: Yes. That means it would be different from before, but also means now this header will look the same for all platform + binding combinations.

Dexterp37: Cool, let's just update the docs then!

PR #1032, opened 2020-07-01
Using application version as display id instead of using reflection.

daoshengmu:
`applicationVersion` is not optional when calling `GleanInstance.Initialize` in C#, and Python binding is also doing this way.

Besides, using Reflection approach sounds like not a good idea, some applications are made by Unity ` callling System.Reflection.Assembly.GetEntryAssembly()` will return null.


daoshengmu:
Per [the debug ping](https://debug-ping-preview.firebaseapp.com/pings/fxr-pc-debug1), its `app_display_version` is `inaccessible`.

PR #1034, opened 2020-07-02
Bug 1650012 - Always wrap `Action` in try/catch when dispatching in C#

Dexterp37:
See the individual commit messages for a detailed explanation.

### DIFF
--- a/glean-core/csharp/Glean/Dispatchers.cs
+++ b/glean-core/csharp/Glean/Dispatchers.cs
@@ -95,12 +95,26 @@ namespace Mozilla.Glean
         {
             Task task = null;
 
+            // Wrap the provided action in a try/catch block: we don't want to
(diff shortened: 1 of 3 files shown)


glean-core/csharp/Glean/Dispatchers.cs
```diff
@@ -95,39 +95,42 @@ internal static Task LaunchAPI(Action action)
         {
             Task task = null;
 
+            // Wrap the provided action in a try/catch block: we don't want to
+            // break execution if something throws.
+            Action safeAction = () => {
+                // In order to prevent tasks from causing exceptions
+                // we wrap the action invocation in try/catch
```
brizental, line 102: A little comment redundancy here, I think you should pick one and delete the other.

PR #1037, opened 2020-07-06
Introduce a docker image for building the C# packages

Dexterp37:
[skip ci]

This also enables building the C# package on CI.

Notes:
- I had to switch from `full-ci` to `normal-ci` in the `kind.yml`. thanks to @MihaiTabara suggestion, otherwise TC was not showing all my tasks. This makes sure that the tasks run as part of the PR.


taskcluster/ci/csharp-build/kind.yml
```diff
@@ -0,0 +1,37 @@
...
+    max-run-time: 1800
+  run:
+    pre-dotnet:
+      # XXX: scripts subshell at runtime so we need to source this here
+      # to be able to access `rustup` and `rustc` from within the cc script
+      # and the gradle command. Another options could be to set those env vars
+      # here like: [export, 'PATH=$HOME/.cargo/bin:$PATH'
+      - [source, taskcluster/scripts/rustup-setup.sh]
+      - [source, taskcluster/scripts/cross-compile-setup.sh]
+      - [bash, '-c', 'echo "rust.targets=x86_64,x86,darwin,linux-x86-64,win32-x86-64-gnu\n" > local.properties']
```
Dexterp37, line 33: @badboy this only makes sense when running gradle, right? I guess I should add something similar to the C# project or manually run `cargo build --target ...` here, right?

taskcluster/ci/csharp-build/kind.yml
```diff
@@ -0,0 +1,37 @@
...
+transforms:
+  - glean_taskgraph.transforms.module_build:transforms
+  - taskgraph.transforms.job:transforms
+  - taskgraph.transforms.task:transforms
+
+job-defaults:
+  attributes:
+    run-on-pr-type: full-ci
+  run-on-tasks-for: [github-pull-request, github-push]
+  description: "{module_name} - Build and test"
```
Dexterp37, line 17: Is there any documentation for the TC syntax? What's `module_name` here?

taskcluster/docker/csharp/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+        # Required to build sqlcipher.
+        tclsh \
+        # Required in libs/ by some scripts patching the source they download.
+        patch \
+        # For windows cross-compilation.
+        mingw-w64 \
+        ## NSS build dependencies
+        gyp \
+        ninja-build \
+        zlib1g-dev \
```
Dexterp37, line 75: Do we need to build NSS on this image?

taskcluster/docker/csharp/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+ENV \
+    # Some APT packages like 'tzdata' wait for user input on install by default.
+    # https://stackoverflow.com/questions/44331836/apt-get-install-tzdata-noninteractive
+    DEBIAN_FRONTEND=noninteractive
+
+# System.
+RUN apt-get update -qq \
+    && apt-get install -qy --no-install-recommends \
+        # To fetch Glean.
+        git \
```
Dexterp37, line 44: Where does the clone happen? Is it automatically performed by TC? Does it mean I can get rid of this dependency?

taskcluster/ci/csharp-build/kind.yml
```diff
@@ -0,0 +1,37 @@
...
+  scopes:
+    - project:releng:services/tooltool/api/download/internal
+  worker-type: b-linux
+  worker:
+    chain-of-trust: true
+    docker-image: { in-tree: csharp }
+    max-run-time: 1800
+  run:
+    pre-dotnet:
```
badboy, line 26: There's no `pre-dotnet`.
This can just be `pre-commands` though.
It is defined here: https://github.com/mozilla/glean/blob/main/taskcluster/glean_taskgraph/job.py#L54

taskcluster/ci/csharp-build/kind.yml
badboy: Yup, local.properties is only used by Gradle, and the rust-android-gradle plugin uses that `rust.targets` parameter to determine what it needs to pass to `cargo build` internally).

As the dotnet build system doesn't handle building the rust part yet, we need to manually invoke the individual `cargo build` commands with the right `--target` parameters.

badboy: And I _think_ we might actually need to set up the compilers/linkers and flags in `.cargo/config` for this, e.g. the extra flags requires for a macOS build: https://github.com/mozilla/glean/blob/main/taskcluster/scripts/cross-compile-setup.sh#L8

taskcluster/ci/csharp-build/kind.yml
```diff
@@ -0,0 +1,37 @@
...
+      # XXX: scripts subshell at runtime so we need to source this here
+      # to be able to access `rustup` and `rustc` from within the cc script
+      # and the gradle command. Another options could be to set those env vars
+      # here like: [export, 'PATH=$HOME/.cargo/bin:$PATH'
+      - [source, taskcluster/scripts/rustup-setup.sh]
+      - [source, taskcluster/scripts/cross-compile-setup.sh]
+      - [bash, '-c', 'echo "rust.targets=x86_64,x86,darwin,linux-x86-64,win32-x86-64-gnu\n" > local.properties']
+    dotnet:
```
badboy, line 34: This should be `commands:`

taskcluster/ci/csharp-build/kind.yml
badboy: `{...}` is just python string interpolation.
`module_name` is only defined for `module_build` transforms here: https://github.com/mozilla/glean/blob/main/taskcluster/glean_taskgraph/transforms/module_build.py#L24

`module_builds` are the different modules for Gradle though, so it doesn't apply here.

* Remove the `transform` line above
* Set a static name here.

taskcluster/docker/csharp/Dockerfile
```diff
@@ -0,0 +1,117 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+FROM mcr.microsoft.com/dotnet/core/sdk:3.1-buster
+
+LABEL maintainer="aplacitelli@mozilla.com"
+
+RUN echo "trigger rebuild"
```
badboy, line 9: Let's actually get rid of this :D 
That's a leftover when I copied over the other Dockerfile. It was used sometime ago in a-s repo to trigger a rebuild of the image. As this is the first time it's build we don't need to have it.

taskcluster/docker/csharp/Dockerfile
badboy: afaik TC inserts that automatically into jobs, meaning we need to keep `git` around.

taskcluster/docker/csharp/Dockerfile
badboy: Actually no, I never cleaned up the full dependency list here.

taskcluster/docker/csharp/Dockerfile
```diff
@@ -0,0 +1,117 @@
...
+        # Required by tooltool to extract tar.xz archives.
+        xz-utils \
+        # Required to build libs/.
+        make \
+        # Required to build sqlcipher.
+        tclsh \
+        # Required in libs/ by some scripts patching the source they download.
+        patch \
+        # For windows cross-compilation.
+        mingw-w64 \
```
badboy, line 71: If we need to build 32bit we might need mingw-w32 (if such a thing exists)

taskcluster/glean_taskgraph/transforms/module_build.py
```diff
@@ -19,8 +19,9 @@ def build_task(config, tasks):
         name = module_info["name"]
         version = get_version()
 
-        for i,item in enumerate(task["run"]["gradlew"]):
-            task["run"]["gradlew"][i] = task["run"]["gradlew"][i].format(module_name=name)
+        if "gradlew" in task["run"]:
```
badboy, line 22: Once we remove the above `transform` in kind.yaml we won't require these changes. 
The module_build transform doesn't apply here and is Gradle-specific.

taskcluster/docker/csharp/Dockerfile
Dexterp37: Looks like `mingw-w64` can do both by providing the appropriate options. Let's take this to a follow-up!

badboy: even better then

Dexterp37:
Given the change in our priorities, I'm closing this.

PR #1039, opened 2020-07-06
Bug 1644359: Support error reporting in Python bindings

mdboom:
This works by initializing a full Glean in the subprocess that
*doesn't* do any startup work such as setting metrics or submitting
pings. This makes it safe to initialize Glean in the subprocess,
and thus errors can be reported there.

### DIFF
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -12,7 +12,6 @@ use ffi_support::{define_string_destructor, ConcurrentHandleMap, FfiStr, IntoFfi
 pub use glean_core::metrics::MemoryUnit;
 pub use glean_core::metrics::TimeUnit;
 pub use glean_core::upload::ffi_upload_result::*;
-use glean_core::upload::PingUploadManager;
 use glean_core::Glean;
 pub use glean_core::Lifetime;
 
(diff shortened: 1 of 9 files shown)


glean-core/src/lib.rs
```diff
@@ -201,21 +200,29 @@ impl Glean {
...
-        let mut glean = Self {
+        Ok(Self {
             upload_enabled: cfg.upload_enabled,
             data_store,
             event_data_store,
             core_metrics: CoreMetrics::new(),
             internal_pings: InternalPings::new(),
             upload_manager,
-            data_path: PathBuf::from(cfg.data_path),
+            data_path: PathBuf::from(cfg.data_path.to_string()),
```
brizental, line 210: You can borrow here instead: `&cfg.data_path`.

glean-core/ffi/src/lib.rs
```diff
@@ -479,25 +407,22 @@ pub unsafe extern "C" fn glean_process_ping_upload_response(
...
 
+/// # Safety
+///
+/// A valid and non-null configuration object is required for this function.
 #[no_mangle]
-pub extern "C" fn glean_initialize_standalone_uploader(
-    data_dir: FfiStr,
-    language_binding_name: FfiStr,
-) -> u8 {
+pub unsafe extern "C" fn glean_initialize_for_subprocess(cfg: *const FfiConfiguration) -> u8 {
```
Dexterp37, line 414: What happens if two processes write/open the database at the same time? Thin of the 'pingsender' case on Windows: FOG might be still running while the pingsender is being used.

mdboom: My understanding is that LMDB handles this for us.  See the [Threads and Processes section here](http://www.lmdb.tech/doc/starting.html).

What *isn't* supported is opening the DB multiple times from the same process.

Dexterp37: That's very interesting. Would you mind adding a small comment about this guarantee here or in some other place? It would also be great to have test coverage for this, but I guess it's fine to do this as a follow-up (or at least just file a bug)

mdboom: The new Python tests in this PR are already sort of testing this case.  Even in test mode, it fires off a subprocess (the difference is that it "joins" and waits for it to finish).  To be extra sure, we could write a test that repeatedly changes metrics while the ping uploader is running and reporting network errors...  Should be easy enough, so I might as well do that now.

mdboom: I was wrong about testing mode still using real subprocesses.  So good catch -- I updated the tests to actually do that so we are testing things running on separate processes, and also added one that sets metrics from both processes at the same time.

PR #1041, opened 2020-07-07
Bug 1650752 - Reapply async upload toggle after initialize

badboy:
From the commit message:

The task running the intialization runs asynchronously.
It therefore can happen that we externally modify the uploadEnabled
flag.
Up to now we forgot to recheck it.
By doing this as the last thing before flushing the queue of tasks it
should be in the expected order for users:

1. First Glean fully initializes, with the uploadEnabled flag as set in
   the `initialize` call.
2. Potentially queued up tasks (such as metric recordings) are replayed
3. Finally toggling the flag is respected.

---

This fixes the one encountered issue where changes to the flag are simply not reflected at all.

### DIFF
--- a/glean-core/python/glean/glean.py
+++ b/glean-core/python/glean/glean.py
@@ -147,7 +147,7 @@ class Glean:
                 cls._application_build_id = application_build_id
 
             # We know we're not initialized,
-            # so we can skip the check inside `setUploadEnabled`
+            # so we can skip the check inside `set_upload_enabled`
             # by setting the variable directly.
             cls._upload_enabled = upload_enabled
 
--- a/glean-core/python/glean/glean.py
+++ b/glean-core/python/glean/glean.py
@@ -146,7 +146,10 @@ class Glean:
             else:
                 cls._application_build_id = application_build_id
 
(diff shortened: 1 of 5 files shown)


badboy:
I think this is some previous discussion around that `setUploadEnabled` part that Dexter mentioned: https://github.com/mozilla/glean/pull/786/files#r399284200

It's not directly related though, it only talks about which way the task is launched

badboy:
> Needs Python changes, I think...

I ... don't even know why I didn't check that -_-
Coming in next.

glean-core/python/glean/glean.py
```diff
@@ -146,7 +146,10 @@ def initialize(
             else:
                 cls._application_build_id = application_build_id
 
-            cls.set_upload_enabled(upload_enabled)
+            # We know we're not initialized,
+            # so we can skip the check inside `setUploadEnabled`
```
mdboom, line 150: Supernit:

```suggestion
            # so we can skip the check inside `set_upload_enabled`
```

badboy: coooopy pasta.

PR #1044, opened 2020-07-08
Fix link ordering in the changelog (and others)

badboy:


badboy:
> Should we also add the Bugzilla link to maintain the pattern from the other entries?

If there is one the PR should have it's bug number anyway

PR #1046, opened 2020-07-09
Bug 1649926 - Always enqueue an async task to change upload and deprecate `getUploadEnabled`

badboy:
Replicating the commit messages for information:

    Always enqueue an async task to change upload

    That way it follows what a user expect when calling it in between other calls:
    It executes in the right order.

    Because the dispatch queue is halted until Glean is fully initialized
    we can safely enqueue here and it will execute after initialization.

and

    Deprecate `getUploadEnabled()` across all implementations

    Due to Glean's asynchronous initialization the return value can be incorrect.
    Applications should not rely on Glean's internal state.
    Upload enabled status should be tracked by the application and communicated to Glean if it changes.
    Note: The method was removed from the C# implementation.

--- 

This probably deserves some extra eyes on it. Putting you 3 on it, but maybe one or 2 should be enough.

I'm now reasonably sure it does what we expect it to do and the tests capture that.
However the tests are a bit brittle and (at least in some cases) might pass just fine on the old code base too.

Getting rid of `getUploadEnabled` means we also don't need to additionally track that upload enabled flag anymore.

Unfortunately I can't write a C# test because we don't have all the necessary utilities in place yet.
I did verify that adding `GleanInstance.SetUploadEnabled(false);` in the sample app just after `Initialize` does NOT send the custom ping, but DOES send the deletion-request ping.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -293,7 +293,7 @@ open class GleanInternalAPI internal constructor () {
         // we can safely enqueue here and it will execute after initialization.
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.launch {
-            val originalEnabled = getUploadEnabled()
+            val originalEnabled = internalGetUploadEnabled()
             LibGleanFFI.INSTANCE.glean_set_upload_enabled(enabled.toByte())
 
             if (!enabled) {
@@ -322,8 +322,25 @@ open class GleanInternalAPI internal constructor () {
      * Get whether or not Glean is allowed to record and upload data.
      *
      * Caution: the result is only correct if Glean is already initialized.
+     *
+     * **THIS METHOD IS DEPRECATED.**
+     * Applications should not rely on Glean's internal state.
+     * Upload enabled status should be tracked by the application and communicated to Glean if it changes.
      */
+    @Deprecated("Upload enabled should be tracked by the application and communicated to Glean if it changes")
     fun getUploadEnabled(): Boolean {
+        return internalGetUploadEnabled()
+    }
+
+    /**
+     * Get whether or not Glean is allowed to record and upload data.
+     *
+     * Caution: the result is only correct if Glean is already initialized.
+     *
+     * Note: due to the deprecation notice and because warnings break the build,
+     * we pull out the implementation into an internal method.
+     */
+    internal fun internalGetUploadEnabled(): Boolean {
         if (isInitialized()) {
             return LibGleanFFI.INSTANCE.glean_is_upload_enabled().toBoolean()
         } else {
@@ -593,7 +610,7 @@ open class GleanInternalAPI internal constructor () {
             return
         }
 
-        if (!getUploadEnabled()) {
+        if (!internalGetUploadEnabled()) {
             Log.e(LOG_TAG, "Glean disabled: not submitting any pings.")
             return
         }
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -344,7 +344,7 @@ open class GleanInternalAPI internal constructor () {
         if (isInitialized()) {
             return LibGleanFFI.INSTANCE.glean_is_upload_enabled().toBoolean()
         } else {
-            return true
+            return false
         }
     }
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -57,6 +57,10 @@ open class GleanInternalAPI internal constructor () {
     }
 
     private var initialized: Boolean = false
+    // Set when `initialize()` returns.
+    // This allows to detect calls that happen before `Glean.initialize()` was called.
+    // Note: The initialization might still be in progress, as it runs in a separate thread.
+    private var initFinished: Boolean = false
 
     internal lateinit var configuration: Configuration
 
@@ -249,6 +253,7 @@ open class GleanInternalAPI internal constructor () {
                 ProcessLifecycleOwner.get().lifecycle.addObserver(gleanLifecycleObserver)
             }
         }
+        this.initFinished = true
     }
 
     /**
@@ -285,6 +290,15 @@ open class GleanInternalAPI internal constructor () {
      * @param enabled When true, enable metric collection.
      */
     fun setUploadEnabled(enabled: Boolean) {
+        if (!this.initFinished) {
+            val msg = """
+            Changing upload enabled before Glean is initialized is not supported.
+            Pass the correct state into `Glean.initialize()`.
+            See documentation at https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk
+            """.trimIndent()
+            Log.w(LOG_TAG, msg)
+            return
+        }
         // Changing upload enabled always happens asynchronous.
         // That way it follows what a user expect when calling it inbetween other calls:
         // It executes in the right order.
@@ -763,6 +777,7 @@ open class GleanInternalAPI internal constructor () {
 
         LibGleanFFI.INSTANCE.glean_destroy_glean()
         initialized = false
+        initFinished = false
     }
 
     /**
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -776,8 +776,12 @@ open class GleanInternalAPI internal constructor () {
         }
 
         LibGleanFFI.INSTANCE.glean_destroy_glean()
-        initialized = false
+
+        // Reset all state.
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.setTaskQueueing(true)
         initFinished = false
+        initialized = false
     }
 
     /**
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -72,9 +72,6 @@ open class GleanInternalAPI internal constructor () {
 
     private lateinit var gleanDataDir: File
 
-    // Keep track of this flag before Glean is initialized
-    private var uploadEnabled: Boolean = true
-
     // Keep track of this value before Glean is initialized
     private var debugViewTag: String? = null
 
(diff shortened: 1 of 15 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -284,46 +269,66 @@ open class GleanInternalAPI internal constructor () {
...
+     *
+     * Note: due to the deprecation notice and because warnings break the build,
+     * we pull out the implementation into an internal method.
+     */
+    internal fun internalGetUploadEnabled(): Boolean {
         if (isInitialized()) {
             return LibGleanFFI.INSTANCE.glean_is_upload_enabled().toBoolean()
         } else {
-            return uploadEnabled
+            return true
```
mdboom, line 331: Just looking at where this is called, this should never happen before Glean is initialized.  However, maybe it's safer to return `false` here so we err on the side of not recording things if there is a bug?  (I was also going to suggest throwing an exception -- maybe that's fine if we really consider that undefined, but having that exception accidentally "bubble out" of the API would be bad...)

glean-core/python/glean/glean.py
```diff
@@ -315,32 +297,35 @@ def set_upload_enabled(cls, enabled: bool) -> None:
...
             enabled (bool): When True, enable metric collection.
         """
-        if cls.is_initialized():
+        # Changing upload enabled always happens asynchronous.
+        # That way it follows what a user expect when calling it inbetween other calls:
+        # It executes in the right order.
+        #
+        # Because the dispatch queue is halted until Glean is fully initialized
+        # we can safely enqueue here and it will execute after initialization.
+        @Dispatcher.launch
```
mdboom, line 306: This is a nit, but since there is literally nothing in this method that needs to happen on the main thread, you could just put the `@Dispatcher.task` decorator on the method itself and not have a nested function.

glean-core/ios/Glean/Glean.swift
```diff
@@ -223,45 +210,61 @@ public class Glean {
...
+
+    /// Get whether or not Glean is allowed to record and upload data.
+    ///
+    /// Caution: the result is only correct if Glean is already initialized.
+    func internalGetUploadEnabled() -> Bool {
         if isInitialized() {
             return glean_is_upload_enabled() != 0
         } else {
-            return uploadEnabled
+            return true
```
Dexterp37, line 267: Do not forget to update this based on Mike's input.

glean-core/csharp/Glean/Glean.cs
```diff
@@ -321,7 +306,7 @@ public bool GetUploadEnabled()
             }
             else
             {
-                return uploadEnabled;
+                return true;
```
Dexterp37, line 309: Do not forget to update this based on Mike's input.

glean-core/python/glean/glean.py
```diff
@@ -315,32 +297,35 @@ def set_upload_enabled(cls, enabled: bool) -> None:
...
+                cls._initialize_core_metrics()
 
-        else:
-            cls._upload_enabled = enabled
+            if original_enabled is True and cls.get_upload_enabled() is False:
+                # If uploading is disabled, we need to send the deletion-request ping
+                PingUploadWorker.process()
 
     @classmethod
     def get_upload_enabled(cls) -> bool:
```
Dexterp37, line 319: Should we simply remove this/change this to `_get_upload_enabled` so that the intent is clearer? I don't think anybody is using this.

glean-core/python/glean/glean.py
```diff
@@ -315,32 +297,35 @@ def set_upload_enabled(cls, enabled: bool) -> None:
...
         """
         Get whether or not Glean is allowed to record and upload data.
+
+        Caution: the result is only correct if Glean is already initialized.
         """
         if cls.is_initialized():
             return bool(_ffi.lib.glean_is_upload_enabled())
         else:
-            return cls._upload_enabled
+            return True
```
Dexterp37, line 328: Do not forget to update this based on Mike's input.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Good point.

glean-core/python/glean/glean.py
badboy: Thanks, that helps improve my python foo

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -284,46 +269,66 @@ open class GleanInternalAPI internal constructor () {
...
-
-                if (!enabled) {
-                    // Cancel any pending workers here so that we don't accidentally upload or
-                    // collect data after the upload has been disabled.
-                    metricsPingScheduler.cancel()
-                    // Cancel any pending workers here so that we don't accidentally upload
-                    // data after the upload has been disabled.
-                    PingUploadWorker.cancel(applicationContext)
-                }
+        // Changing upload enabled always happens asynchronous.
```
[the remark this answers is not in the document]
badboy: Yes, that's the behavior I _would_ expect. From a user perspective `uploadEnabled` was set to false only _after_ the ping was submitted.

However ... (gosh this is complicated) ... because of the delay of the uploader picking it up it could still get cancelled :/

glean-core/csharp/Glean/Glean.cs
```diff
@@ -301,16 +289,13 @@ public void SetUploadEnabled(bool enabled)
...
+            });
         }
 
         /// <summary>
         /// Get whether or not Glean is allowed to record and upload data.
+        ///
+        /// Caution: the result is only correct if Glean is already initialized.
         /// </summary>
         /// <returns>`true` if Glean is allowed to record and upload data.</returns>
         public bool GetUploadEnabled()
```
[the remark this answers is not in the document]
badboy: Failed re-ordering of my commits and then lost that change. Indeed this should have become private.

glean-core/python/glean/glean.py
badboy: 👍

badboy:
> What happens in the following order of events?
> 
> ```
> Glean.setUploadEnabled(false)
> Glean.initialize(uploadEnabled=true)
> ```
> 
> I would expect the state to be upload is enabled, but since the first call would be deferred until after the second, that might not happen (we should confirm with a test -- I could be wrong).

Damn, you are probably right. This is the _one_ case why we had to preserve the state in an instance variable.
I need to think about it.

badboy:
Ok, following up on what Mike said.

The positive first: It's easy to write a test!
Now the rest:
I think the only sensible solution is to actually ignore `setUploadEnabled` completely before Glean is initialized.
If that happens the order is (mostly) clear and the initialize should be the last operation, so is what should take precedence.
We already expect the `uploadEnabled` passed to `initialize` be correct. The whole reason we introduce that parameter was because people kept doing it wrong with `setUploadEnabled`.

I propose we add this to the top of `setUploadEnabled`:

```kotlin
        if (!isInitialized()) {
            Log.w(LOG_TAG, "Changing upload enabled before Glean is initialized is not supported.")
            return
        }
```

cc @Dexterp37 @mdboom

badboy:
I spoke too soon.

We can't differentiate between

```
Glean.setUploadEnabled(false)
Glean.initialize(...)
```

and

```
Glean.initialize(...)
Glean.setUploadEnabled(false)
```

right now.

Do we really need _more_ state? -_-

badboy:
So yeah, if we track that `initialize` was called at all we might be able to prevent this properly.

wdyt?

Dexterp37:
> I propose we add this to the top of `setUploadEnabled`:
> 
> ```kotlin
>         if (!isInitialized()) {
>             Log.w(LOG_TAG, "Changing upload enabled before Glean is initialized is not supported.")
>             return
>         }
> ```
> 
> cc @Dexterp37 @mdboom

I support this approach and like it. My only recommendation would be to have a more comprehensive message or add a link to our docs that explain the initial state should be provided with `Glean.Initialize`.

badboy:
As I won't get this done today here's what's left:

* [ ] Incorporate the comments from above 
	* `true -> false` in `getUploadEnabled` across all implementations
	* In Python make `getUploadEnabled` private
* [ ] Add a new variable `initWasCalled` (name bikesheddable), set to true before returning from `initialize`
* [ ] Check that new variable in `setUploadEnabled`
    * Print warning with extended message if `initWasCalled == false`

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -300,46 +290,75 @@ open class GleanInternalAPI internal constructor () {
...
-                    // data after the upload has been disabled.
-                    PingUploadWorker.cancel(applicationContext)
-                }
+        if (!this.initFinished) {
+            val msg = """
+            Changing upload enabled before Glean is initialized is not supported.
+            Pass the correct state into `Glean.initialize()`.
+            See documentation at https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk
+            """.trimIndent()
+            Log.w(LOG_TAG, msg)
```
Dexterp37, line 299:
```suggestion
            Log.e(LOG_TAG, msg)
```

Should this be an error?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -757,6 +776,11 @@ open class GleanInternalAPI internal constructor () {
         }
 
         LibGleanFFI.INSTANCE.glean_destroy_glean()
+
+        // Reset all state.
```
Dexterp37, line 780: Great catch!

glean-core/csharp/Glean/Glean.cs
```diff
@@ -271,11 +262,22 @@ internal bool IsInitialized()
...
         public void SetUploadEnabled(bool enabled)
         {
-            if (IsInitialized())
-            {
-                bool originalEnabled = GetUploadEnabled();
+            if (!this.initFinished) {
+                string msg = "Changing upload enabled before Glean is initialized is not supported.\n" +
+                    "Pass the correct state into `Glean.initialize()`.\n" +
+                    "See documentation at https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk";
+                Log.Warning(msg);
```
Dexterp37, line 269:
```suggestion
                Log.Error(msg);
```

glean-core/ios/Glean/Glean.swift
```diff
@@ -228,45 +223,71 @@ public class Glean {
...
     public func setUploadEnabled(_ enabled: Bool) {
-        if isInitialized() {
-            let originalEnabled = getUploadEnabled()
+        if !self.initFinished {
+            let msg = """
+            Changing upload enabled before Glean is initialized is not supported.
+            Pass the correct state into `Glean.initialize()`.
+            See documentation at https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk
+            """
+            self.logger.warn(msg)
```
Dexterp37, line 232: Error here?

glean-core/python/glean/glean.py
```diff
@@ -315,32 +305,45 @@ def set_upload_enabled(cls, enabled: bool) -> None:
...
-            def set_upload_enabled():
-                _ffi.lib.glean_set_upload_enabled(enabled)
+        if not cls._init_finished:
+            msg = """
+                  Changing upload enabled before Glean is initialized is not supported.
+                  Pass the correct state into `Glean.initialize()`.
+                  See documentation at
+                  https://mozilla.github.io/glean/book/user/general-api.html#initializing-the-glean-sdk
+                  """
+            log.warning(inspect.cleandoc(msg))
```
Dexterp37, line 315:
```suggestion
            log.error(inspect.cleandoc(msg))
```

badboy:
Waiting with the rebase, given that there's a release incoming and thus I need to change the changelog anyway.

PR #1047, opened 2020-07-10
Bug 1646706 - Manage debug options in the core and get them from the environment

brizental:
This is still a work in progress. I wanted to get some feedback on my approcah before moving forward.

For now I only have the logic abstracting away the debug options into their own struct and getting the values from the env. I still need to deal with: better testing coverage, move log_pings management to the core, updating documentation.

I tried to make the logic a bit generic in case we add more debugging options in the future, which I know we will: https://bugzilla.mozilla.org/show_bug.cgi?id=1634064

Now, still it might be a little overkill, so please let me know your thoughts. 

**NOTE** Thanks to @badboy I was able to make this code look a little nicer, sorry for nerdsniping you, hehe.

[UPDATE]

This is now ready for review. I will make this part 1 of 2.

This introduces the `DebugOptions` abstraction. Next PR will add the `set_log_pings` function to the Glean object and to the FFI and make the necessary changes to the bindings.

### DIFF
--- a/glean-core/src/debug.rs
+++ b/glean-core/src/debug.rs
@@ -1,0 +1,216 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! # Debug options
+//!
+//! The debug options for Glean may be set by calling one of the `set_*` functions
+//! or by setting specific environment variables.
+//!
+//! The environment variables will be read only once when the options are initialized.
+//!
+//! The possible debugging features available out of the box are:
+//!
+//! * **Ping logging** - logging the contents of ping requests that are correctly assembled;
+//!         This may be set by calling glean.set_log_pings(value: bool)
+//!         or by setting the environment variable GLEAN_LOG_PINGS="true";
+//! * **Debug tagging** - Adding the X-Debug-ID header to every ping request,
+//!         allowing these tagged pings to be sent to the ["Ping Debug Viewer"](https://mozilla.github.io/glean/book/dev/core/internal/debug-pings.html).
+//!         This may be set by calling glean.set_debug_view_tag(value: &str)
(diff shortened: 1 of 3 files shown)


badboy:
To surface the things I posted privately earlier, and so that @Dexterp37 sees them:

I think this is quite clever code (especially now that we got rid of the allocation).
However I think it might be _too clever_ of a piece of code for 2 (maybe in the future 3) possible debug options.

The alternative would be to have the `DebugOptions` struct, let it read from the environment on creation and expose methods `set_log_pings` and `set_debug_tag` on it, where we individually handle the validation.

Dexterp37:
> 
> 
> To surface the things I posted privately earlier, and so that @Dexterp37 sees them:
> 
> I think this is quite clever code (especially now that we got rid of the allocation).
> However I think it might be _too clever_ of a piece of code for 2 (maybe in the future 3) possible debug options.

I think I'm fine with the current approach, as long as we provide a bit more documentation and testing for the `DebugOption` part.

glean-core/src/debug.rs
```diff
@@ -0,0 +1,216 @@
...
+//!
+//! The debug options for Glean may be set by calling one of the `set_*` functions
+//! or by setting specific environment variables.
+//!
+//! The environment variables will be read only once when the options are initialized.
+//!
+//! The possible debugging features available out of the box are:
+//!
+//! * **Ping logging** - logging the contents of ping requests that are correctly assembled;
+//!         This may be set by calling glean.set_log_pings(value: bool)
```
brizental, line 15: This function will come in the next PR.

brizental: Do you mean add it to the code as a comment or add it here? I haven't opened a new bug for it, I was just going to do it as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1646706

Dexterp37:
```suggestion
//!         This may be set by calling glean.set_log_pings(value: bool) (will be added by bug 1646706).
```

brizental: I will open a PR adding this function immediatelly after I merge this one so I think this addition is not necessary.

PR #995, opened 2020-06-22
Python: Upgrade to Python 3.8 as the "default" Python version

Harshitha1234:
Signed-off-by: Harshitha1234 <harshithachowdary.t17@iiits.in>

**Changes**
Upgrade to Python 3.8 [Bug 1624680](https://bugzilla.mozilla.org/show_bug.cgi?id=1624680)

PR #1048, opened 2020-07-10
Fix doc-only builds

mdboom:
A full Python build needs to run in order for a Python documentation build to work.

As of #995, the Python 3.8 test job needs to run even for doc-only builds,
rather than the Python 3.7 test job.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -701,6 +701,7 @@ jobs:
       - image: circleci/python:3.7.6
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


badboy:
ah, haha. Let's see what's faster: 1049

mdboom:
Sure -- this is probably faster in the common case of non-doc-only PRs.  But slower for doc-only PRs (but that's how we've done it forever -- the bug is just changing the Python version for doc building yesterday without doing this...)

PR #1049, opened 2020-07-10
Setup the Python environment for generating Python documentation

badboy:
[doc only]

We skip the tests if a patch set is doc-only, but we still need the
environment around to generate docs.
The test and doc tasks don't _need_ to depend on each other.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -781,8 +781,9 @@ jobs:
       - image: circleci/python:3.8.2
     steps:
       - checkout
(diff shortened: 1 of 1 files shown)


badboy:
This fails because we actually do need a fully built glean-sdk installed in the virtual env.
Let's take mike's approach.

mdboom:
Yeah -- Python documentation tools generally have the dark pattern that they "import" (rather than just parse) the code they document.

PR #1050, opened 2020-07-10
Bug 1649925 - Make sure Glean C# doesn't crash on x86

Dexterp37:
This replaces the use of `SafeHandle`-derived classes when passing Glean handles back and forth from the FFI layer. It was crashing on x86 because, while Glean correctly uses uint64 as handles, `SafeHandle` was using `IntPtr` under the hood, which is 4 bytes on 32 bit architectures.

This made glean-core panic because it was seeing an unknown handle when it was passed back from C# to glean-core over the FFI.

All in all, we don't really need `SafeHandle`(s) for everything: we just need it for Win32 handles or pointers coming from the FFI. Moreover, we don't really care yet about cleaning up, so no need to get desperate about re-implementing `SafeHandle`-like mechanism for uint64: that would also be complicated, because the .NET marshaler knows about `SafeHandle` and behaves in a special way with that, when doing P/Invoke on native libraries.

### DIFF
--- a/glean-core/csharp/Glean/Glean.csproj
+++ b/glean-core/csharp/Glean/Glean.csproj
@@ -25,6 +25,7 @@
     <IsWindows Condition="'$(OS)' == 'Windows_NT'">true</IsWindows>
     <IsOSX Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::OSX)))' == 'true'">true</IsOSX>
     <IsLinux Condition="'$([System.Runtime.InteropServices.RuntimeInformation]::IsOSPlatform($([System.Runtime.InteropServices.OSPlatform]::Linux)))' == 'true'">true</IsLinux>
+    <Platforms>AnyCPU;x86</Platforms>
   </PropertyGroup>
 
   <!--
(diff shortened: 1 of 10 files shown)


mdboom:
Ok -- it makes sense to me that `SafeHandle` isn't really needed here.

This is in the category of "only if it's easy / I don't feel that strongly":  One thing we seem to be losing in the process is a little bit of type checking help, though. For instance, this would seem to catch accidentally passing a string handle into a boolean FFI method.  Could we retain those specific handle types but just make them all subclasses or typedefs of u64?

If C# makes that hard / impossible, it wouldn't be worth the effort.  This is all internal stuff anyway, and I'm not sure it's super helpful at catching mistakes -- since this is all hand-written code anyway, we could just as easily make a mistake to make something work that shouldn't...

Dexterp37:
> 
> 
> Ok -- it makes sense to me that `SafeHandle` isn't really needed here.
> 
> This is in the category of "only if it's easy / I don't feel that strongly": One thing we seem to be losing in the process is a little bit of type checking help, though. For instance, this would seem to catch accidentally passing a string handle into a boolean FFI method. Could we retain those specific handle types but just make them all subclasses or typedefs of u64?

I believe this could be possible: however, we're not doing it for Kotlin nor other bindings. Maybe it's worth filing a bug and address this holistically?

Dexterp37:
> 
> 
> I agree with @mdboom that the type checking is a nice-to-have, but I don't see a strong need to have it as no-one outside of the Glean internals should care about it. So it's basically just an issue for us and not consumers.

Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1652052

PR #1052, opened 2020-07-10
1641148: Integrate Python and Rust logging

mdboom:
This works by setting up a named pipe between Python and Rust logging.

This requires adding another thread to the Python side to read from the log.
The overhead of that *could* be problematic.

I have no idea if this approach will work on Windows.

This at least *does* work -- any callback-based approach has been ruled
out due to limitations of CFFI or the need to go to a compiling-against-
a-specific-version-of-Python approach (see details in bug).

### DIFF
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -402,6 +402,17 @@ void glean_destroy_uuid_metric(uint64_t v);
  */
 void glean_enable_logging(void);
 
+/**
+ * Initialize the logging system to send JSON messages to a pipe.
+ *
+ * `fd` is a writable file descriptor (on Unix) or file handle (on Windows).
+ *
+ * # Safety
+ * Unsafe because the fd u64 passed in will be interpreted as either a file
+ * descriptor (Unix) or file hangle (Windows) without any checking.
+ */
+void glean_enable_logging_to_pipe(uint64_t fd);
+
 void glean_event_record(uint64_t metric_id,
                         uint64_t timestamp,
                         RawIntArray extra_keys,
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -409,7 +409,7 @@ void glean_enable_logging(void);
  *
  * # Safety
  * Unsafe because the fd u64 passed in will be interpreted as either a file
- * descriptor (Unix) or file hangle (Windows) without any checking.
+ * descriptor (Unix) or file handle (Windows) without any checking.
  */
 void glean_enable_logging_to_pipe(uint64_t fd);
 
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -403,7 +403,10 @@ void glean_destroy_uuid_metric(uint64_t v);
 void glean_enable_logging(void);
 
 /**
- * Initialize the logging system to send JSON messages to a pipe.
+ * Initialize the logging system to send JSON messages to a file descriptor
+ * (Unix) or file handle (Windows).
+ *
+ * Not available on Android and iOS.
  *
  * `fd` is a writable file descriptor (on Unix) or file handle (on Windows).
  *
@@ -411,7 +414,7 @@ void glean_enable_logging(void);
  * Unsafe because the fd u64 passed in will be interpreted as either a file
  * descriptor (Unix) or file handle (Windows) without any checking.
  */
-void glean_enable_logging_to_pipe(uint64_t fd);
+void glean_enable_logging_to_fd(uint64_t fd);
 
 void glean_event_record(uint64_t metric_id,
                         uint64_t timestamp,
--- a/glean-core/ffi/src/fd_logger.rs
+++ b/glean-core/ffi/src/fd_logger.rs
@@ -1,0 +1,83 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fs::File;
+use std::io::Write;
+use std::sync::RwLock;
+
+#[cfg(target_os = "windows")]
+use std::os::windows::io::FromRawHandle;
+
+#[cfg(target_os = "windows")]
+use std::ffi::c_void;
+
+#[cfg(not(target_os = "windows"))]
+use std::os::unix::io::FromRawFd;
+
+use serde::Serialize;
+
+/// An implementation of log::Log that writes log messages in JSON format to a
+/// file descriptor/handle. The logging level is ignored in this implementation:
+/// it is up to the receiver of these log messages (on the language binding
+/// side) to filter the log messages based on their level.
+/// The JSON payload of each message in an object with the following keys:
+///    - `level` (string): One of the logging levels defined here:
+///      https://docs.rs/log/0.4.11/log/enum.Level.html
+///    - `message` (string): The logging message.
+pub struct FdLogger {
+    pub file: RwLock<File>,
+}
+
+#[derive(Serialize)]
+struct FdLoggingRecord {
+    level: String,
+    message: String,
+}
+
+#[cfg(target_os = "windows")]
+unsafe fn get_file_from_fd(fd: u64) -> File {
+    File::from_raw_handle(fd as *mut c_void)
+}
+
+#[cfg(not(target_os = "windows"))]
+unsafe fn get_file_from_fd(fd: u64) -> File {
+    File::from_raw_fd(fd as i32)
+}
+
+impl FdLogger {
+    pub unsafe fn new(fd: u64) -> Self {
+        FdLogger {
+            file: RwLock::new(get_file_from_fd(fd)),
+        }
+    }
+}
+
+impl log::Log for FdLogger {
+    fn enabled(&self, _metadata: &log::Metadata) -> bool {
+        // This logger always emits logging messages of any level, and the
+        // language binding consuming these messages is responsible for
+        // filtering and routing them.
+        true
+    }
+
+    fn log(&self, record: &log::Record) {
+        // Normally, classes implementing the Log trait would filter based on
+        // the log level here. But in this case, we want to emit all log
+        // messages and let the logging system in the language binding filter
+        // and route them.
+        let payload = FdLoggingRecord {
+            level: record.level().to_string(),
+            message: record.args().to_string(),
+        };
+        let _ = writeln!(
+            self.file.write().unwrap(),
+            "{}",
+            serde_json::to_string(&payload).unwrap()
+        );
+    }
+
+    fn flush(&self) {
+        let _ = self.file.write().unwrap().flush();
+    }
+}
--- a/glean-core/ffi/src/lib.rs
+++ b/glean-core/ffi/src/lib.rs
@@ -189,7 +189,10 @@ pub extern "C" fn glean_enable_logging() {
 #[cfg(all(not(target_os = "android"), not(target_os = "ios")))]
 static FD_LOGGER: OnceCell<fd_logger::FdLogger> = OnceCell::new();
 
-/// Initialize the logging system to send JSON messages to a pipe.
+/// Initialize the logging system to send JSON messages to a file descriptor
+/// (Unix) or file handle (Windows).
+///
+/// Not available on Android and iOS.
 ///
 /// `fd` is a writable file descriptor (on Unix) or file handle (on Windows).
 ///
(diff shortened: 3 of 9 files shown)


mdboom:
And... this doesn't work on Windows as written, but it looks like the [filedescriptor](https://docs.rs/filedescriptor/0.7.1/filedescriptor/) crate could help here.  But that's for next week...

mdboom:
@badboy : This isn't ready for a final review around code organization / documentation or any niceties like that. But I am curious to get your thoughts on the overall approach. Seems to be working on Linux, Mac and Windows now.

glean-core/ffi/src/lib.rs
```diff
@@ -128,10 +141,50 @@ where
...
+#[derive(Serialize)]
+struct FdLoggingRecord {
+    level: String,
+    message: String,
+}
+
+static mut FD_LOGGER: FdLogger = FdLogger { f: None };
+
+impl log::Log for FdLogger {
+    fn enabled(&self, _metadata: &log::Metadata) -> bool {
```
badboy, line 157: We should document why this is always enabled.

glean-core/ffi/src/lib.rs
```diff
@@ -128,10 +141,50 @@ where
...
+        true
+    }
+
+    fn log(&self, record: &log::Record) {
+        if self.enabled(record.metadata()) {
```
badboy, line 162: Given that we implemented `enabled` to always return true, we can just skip the check here (and maybe add a note why it's unnecessary?)

glean-core/ffi/src/lib.rs
```diff
@@ -128,10 +141,50 @@ where
     with_glean_mut(|glean| Ok(callback(glean)))
 }
 
+struct FdLogger {
```
badboy, line 144: Can we move the logger to its own module?

badboy: And we should document what an FdLogger is:

* unfiltered (filtering needs to happen on the other side)
* writes JSON-formatted logs to a file descriptor
* expectation: file descriptor is valid and points to either a file or a file descriptor controlled by the other side of FFI (so that it can read from it)

glean-core/ffi/src/lib.rs
```diff
@@ -128,10 +141,50 @@ where
...
+            let _ = file.write().unwrap().flush();
+        }
+    }
+}
+
 /// Initialize the logging system based on the target platform. This ensures
 /// that logging is shown when executing the Glean SDK unit tests.
 #[no_mangle]
-pub extern "C" fn glean_enable_logging() {
+pub extern "C" fn glean_enable_logging(fd: usize) {
```
badboy, line 187: Maybe this should be `c_int` or `c_uint`? `usize`'s size depends on the platform (32 vs 64bit), using `c_int` it's almost always a 32bit integer and thus easier to match the types on the other side.

glean-core/python/glean/_ffi.py
```diff
@@ -31,7 +35,40 @@ def get_shared_object_filename() -> str:  # pragma: no cover
...
+        "DEBUG": logging.DEBUG,
+    }
+
+    def log_handler():
+        while True:
+            data = reader.readline().rstrip()
+            json_content = json.loads(data)
+            log.log(level_map.get(json_content["level"], 0), json_content["message"])
+
+    logging_thread = threading.Thread(target=log_handler)
```
badboy, line 66: I have a slight dislike that this requires a thread, but I guess with this design it's the only way we can ensure to continously getting all data.

mdboom:
> I also understand that the approach that a-s takes with providing a "log callback" won't work for Python because of the other multi-threading/multi-processing things we do (or won't it? we're now initializing Glean in all processes, right?)
> So maybe for now this is the right way for Python.
> 
> The only other way for this too work for Python is building our own `python_logger` crate that actually calls back into Python, e.g. through something like [PyO3](https://github.com/PyO3/pyo3), so we can interact with Python objects. That's a bit more involved though.

The issue with callbacks in Python is well-explained [here](https://cffi.readthedocs.io/en/latest/using.html#callbacks).  It's basically because we communicate with the Rust shared object "dynamically" by reading the API and putting things on the call stack at runtime.  Callbacks in this context require running dynamically generated code, which most OSes make rather difficult these days.  The alternative is to build a shared object that links directly to libpython (which is what the PyO3 solution would amount to).  But since the Python ABI isn't stable across versions of Python, we would need to produce {3.6|3.7|3.8|3.9}x{linux|mac|win32|win64} different binaries rather than just the 4 we do now.  So we'd trade the *runtime* complexity of this pipe and the extra thread etc. for extra CI / distribution complexity (in particular, I'm concerned about that with mach where we would have to vendor all of those builds and pick the right one in so many different contexts...)

badboy:
> The issue with callbacks in Python is well-explained [here](https://cffi.readthedocs.io/en/latest/using.html#callbacks). [...] But since the Python ABI isn't stable across versions of Python, we would need to produce {3.6|3.7|3.8|3.9}x{linux|mac|win32|win64} different binaries rather than just the 4 we do now. [...]

Ah, this explains a lot. And now I see you commented that exact same thing a month ago on the bug, sorry for not noticing.
That makes the pipe/file descriptor approach much more reasonable and I'd say we go ahead with it (doesn't preclude us from fixing the other platforms eventually)

glean-core/ffi/src/lib.rs
mdboom: It looks like a Windows file handle is a pointer, which is why I went with `usize`.  I guess this matters for JNA where you have to write something that will work everywhere (but not for Python cffi where the pointer size is known and used at runtime).  I guess the best choice is probably a `u64` (which will only contain 32-bits of meaningful stuff on i686).

badboy: Ah, I see. 
At least in the _definition_ on the Kotlin side we could use [`IntegerType`](https://java-native-access.github.io/jna/4.2.0/com/sun/jna/IntegerType.html), which seems to be the equivalent?

Given that we don't pass in any value other than `0` from Kotlin it's less of an issue.
We should document properly what the `fd` should be if set and that `0` is treated as disabling fd logging.

badboy: I guess `u64` is also fine. At least it's defined and consistent across platforms.

glean-core/ffi/src/fd_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+}
+
+#[derive(Serialize)]
+struct FdLoggingRecord {
+    level: String,
+    message: String,
+}
+
+#[cfg(target_os = "windows")]
+fn get_file_from_fd(fd: u64) -> File {
```
badboy, line 39: Technically this should be `unsafe fn get_file_from_fd` as it doesn't make any further guarantees before it calls something unsafe.

glean-core/ffi/src/fd_logger.rs
```diff
@@ -0,0 +1,83 @@
...
+    unsafe { File::from_raw_handle(fd as *mut c_void) }
+}
+
+#[cfg(not(target_os = "windows"))]
+fn get_file_from_fd(fd: u64) -> File {
+    unsafe { File::from_raw_fd(fd as i32) }
+}
+
+impl FdLogger {
+    pub fn new(fd: u64) -> Self {
```
badboy, line 49: And then this also becomes `pub unsafe fn new`.

And yes, that means this unsafe propagates upwards, but there's not much we can do.
From the docs:

> This function is also unsafe as the primitives currently returned have the contract that they are the sole owner of the file descriptor they are wrapping. Usage of this function could accidentally allow violating this contract which can cause memory unsafety in code that relies on it being true.

and there is no way for us to check this precondition, so we have to mark it unsafe.
The good thing is: this is already called from an `extern "C"` function, and C land is unsafe, so it's unsafe all the way ~~down~~ up!

glean-core/ffi/src/lib.rs
```diff
@@ -165,11 +174,27 @@ pub extern "C" fn glean_enable_logging() {
...
+        // Set up logging to a file descriptor/handle. For this usage, the
+        // language binding should setup a pipe and pass in the descriptor to
+        // the writing side of the pipe as the `fd` parameter. Log messages are
+        // written as JSON to the file descriptor.
+        if fd != 0 && FD_LOGGER.set(fd_logger::FdLogger::new(fd)).is_ok() {
+            // Set the level so everything goes through to the language
+            // binding side where it will be filtered by the language
+            // binding's logging system.
+            if log::set_logger(FD_LOGGER.get().unwrap())
+                .map(|()| log::set_max_level(log::LevelFilter::Info))
```
badboy, line 188: Wouldn't this miss `debug` level?

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -69,7 +69,7 @@ internal interface LibGleanFFI : Library {
 
         internal var INSTANCE: LibGleanFFI = try {
             val lib = Native.load(JNA_LIBRARY_NAME, LibGleanFFI::class.java) as LibGleanFFI
-            lib.glean_enable_logging()
+            lib.glean_enable_logging(0)
```
Dexterp37, line 72: nit: can we add a comment explaining why we're passing 0 here? (I know it's explained elsewhere, too)

glean-core/csharp/Glean/Glean.cs
```diff
@@ -66,7 +66,7 @@ private Glean()
             // Private constructor to disallow instantiation since
             // this is meant to be a singleton. It only wires up the
             // glean-core logging on the Rust side.
-            LibGleanFFI.glean_enable_logging();
+            LibGleanFFI.glean_enable_logging(0);
```
Dexterp37, line 69: nit: can we add a short comment explaining why we're passing 0, here?

glean-core/ffi/examples/glean_app.c
```diff
@@ -7,7 +7,7 @@
 
 int main(void)
 {
-  glean_enable_logging();
+  glean_enable_logging(0);
```
Dexterp37, line 10: nit: can we add a short comment explaining why we're passing 0, here?

glean-core/ios/Glean/Glean.swift
```diff
@@ -54,7 +54,7 @@ public class Glean {
         // intentionally left private, no external user can instantiate a new global object.
 
         // Enable logging in the Rust library
-        glean_enable_logging()
+        glean_enable_logging(0)
```
Dexterp37, line 57: nit: can we add a short comment explaining why we're passing 0, here?

mdboom:
@Dexterp37: Given your comments about the `0` argument, I wonder if it's better to just add a new FFI API here called `glean_enable_logging_over_pipe(fd)`.  Most language bindings would just continue to use the old `glean_enable_logging()` and only Python would change to the other one.   It has the added advantage that only the new API would need to be explicitly declared `unsafe`.  Does that make sense?

Dexterp37:
> 
> 
> @Dexterp37: Given your comments about the `0` argument, I wonder if it's better to just add a new FFI API here called `glean_enable_logging_over_pipe(fd)`. Most language bindings would just continue to use the old `glean_enable_logging()` and only Python would change to the other one. It has the added advantage that only the new API would need to be explicitly declared `unsafe`. Does that make sense?

I think it makes sense!

mdboom:
Ah -- the other advantage I just realized -- we could completely skip it on Android and iOS builds so the new `fd` logger wouldn't add to binary size.  It would be harder to not do it on both Python from Desktop (FOG) builds, but binary size is less of an issue on those platforms.

mdboom:
Resubmitting for review -- changed from adding a parameter to an existing API to the (IMHO) cleaner adding a new API.

glean-core/ffi/glean.h
```diff
@@ -402,6 +402,17 @@ void glean_destroy_uuid_metric(uint64_t v);
  */
 void glean_enable_logging(void);
 
+/**
+ * Initialize the logging system to send JSON messages to a pipe.
```
badboy, line 406: While that is how we _use_ it we don't really care if that's a pipe, a file descriptor to an actual file or even a tcp socket.
Do you think this is still the right docs & name?

(Given that Python is the only user of this  and there we DO use a pipe this is a weakly held opinion on naming)

glean-core/ffi/src/lib.rs
```diff
@@ -180,6 +186,36 @@ pub extern "C" fn glean_enable_logging() {
...
+/// Initialize the logging system to send JSON messages to a pipe.
+///
+/// `fd` is a writable file descriptor (on Unix) or file handle (on Windows).
+///
+/// # Safety
+/// Unsafe because the fd u64 passed in will be interpreted as either a file
+/// descriptor (Unix) or file handle (Windows) without any checking.
+#[cfg(all(not(target_os = "android"), not(target_os = "ios")))]
+#[no_mangle]
+pub unsafe extern "C" fn glean_enable_logging_to_pipe(fd: u64) {
```
badboy, line 201: Given that this function _does_ land in the generated header file, should we have an empty implementation for Android/iOS?
On the other hand if we ever end up using it (by accident) on these platforms it will fail at load time with an unresolved symbol.

At the very least we should document it that it's not available everywhere.

glean-core/ffi/glean.h
mdboom: That's good point.  Can rename.

glean-core/ffi/src/lib.rs
mdboom: +1 on documenting.

I think there's some feature in cbindgen to wrap things in C `#define`s -- cbindgen warns about that, but I couldn't find the docs on that feature anywhere.  Ultimately, I gave up (but should have mentioned that here).

PR #1056, opened 2020-07-13
Bug 1640201 - Add the high level design docs for the upload module [doc only]

Dexterp37:
This is a braindump of the desired architecture for the upload modules implemented in the language bindings, that directly interact with the lower level upload module in glean-core.

A preview of the mermaid plot:

image

### DIFF
--- a/docs/dev/core/internal/upload.md
+++ b/docs/dev/core/internal/upload.md
@@ -7,6 +7,67 @@ The Glean SDK leverages the available platform capabilities to implement any net
 Glean core controls all upload and coordinates the platform side with its own internals.
 All language bindings implement ping uploading around a common API and protocol.
 
+## The upload module in the language bindings
+
+```mermaid
+classDiagram
+    class UploadResult {
+        ~ToFFI()* int
+    }
+
+    class HttpResponse {
+        int statusCode
+        ~ToFFI() int
+    }
+
+    class UnrecoverableFailure {
+        ~ToFFI() int
+    }
+
+    class RecoverableFailure {
+        ~ToFFI() int
+    }
+
+    class PingUploader {
+        <<interface>>
+        +Upload() UploadResult
+    }
+
+    class BaseUploader {
+        +BaseUploader(PingUploader)
+        ~TriggerUploads()
+        ~CancelUploads()
+    }
+
+    class HttpUploader {
+        +Upload() UploadResult
+    }
+
+    UploadResult <|-- HttpResponse
+    UploadResult <|-- UnrecoverableFailure
+    UploadResult <|-- RecoverableFailure
+    PingUploader <|-- HttpUploader
+    PingUploader o-- BaseUploader
+
+```
+
+### `PingUploader`
+The `PingUploader` interface describes the contract between the `BaseUploader` and the SDK or user-provided upload modules.
+
+### `BaseUploader`
+The `BaseUploader` component is responsible for interfacing with the lower level `get_upload_task` calls and dealing with the logic in a platform-coherent way.
+
+  * Every Glean instance will always have a single `BaseUploader` instance.
+  * The `BaseUploader` is fed, at Glean initialization, with an instance of an implementation of the `PingUploader` interface.
+  * Whenever `BaseUploader` thinks it should perform an upload, it will call the provided instance of the `PingUploader` interface and call `upload` with the data it's getting from the glean-core/FFI.
+  * Any throttling happens at this layer: the core will orchestrate the throttling, while this layer will be responsible to abide to what the core is telling it to do.
+  * Any logic for aborting uploads or triggering uploads is provided by this object.
+
+### `HttpClientUploader`
+The `HttpClientUploader` is the default SDK-provided HTTP uploader. It acts as an adapter between the platform-specific upload library and the Glean upload APIs.
+
+> Note that most of the languages have now diverged, due to the many iterations, from this design. For example, in Kotlin, the [`BaseUploader` is mostly empty](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt) and its functionalities are spread in the [`PingUploadWorker`](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt).
+
 ## Upload task API
 
 The following diagram visualizes the communication between Glean core (the Rust crate),
--- a/docs/dev/docs.md
+++ b/docs/dev/docs.md
@@ -14,7 +14,7 @@ API docs are also generated from docstrings for Rust, Kotlin, Swift and Python.
 The `mdbook` crate is required in order to build the narrative documentation:
 
 ```sh
-cargo install mdbook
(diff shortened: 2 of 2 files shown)


docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,62 @@ The Glean SDK leverages the available platform capabilities to implement any net
...
+  * Every Glean instance will always have a single `BaseUploader` instance.
+  * The `BaseUploader` is fed, at Glean initialization, with an instance of an implementation of the `PingUploader` interface.
+  * Whenever `BaseUploader` thinks it should perform an upload, it will call the provided instance of the `PingUploader` interface and call `upload` with the data it's getting from the glean-core/FFI.
+  * Any throttling happens at this layer!
+  * Any logic for aborting uploads or triggering uploads is provided by this object.
+* The `HttpClientUploader` is the default SDK-provided HTTP uploader. It acts as an adapter between the platform-specific upload library and the Glean upload APIs.
+
+> Note that most of the language have now diverged, due to the many iterations that got us here, from this design. For example, in Kotlin, the [`BaseUploader` is mostly empty](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt) and its functionalities are spread in the [`PingUploadWorker`](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt).
```
brizental, line 64:
```suggestion
> Note that most of the languages have now diverged, due to the many iterations, from this design. For example, in Kotlin, the [`BaseUploader` is mostly empty](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/net/BaseUploader.kt) and its functionalities are spread in the [`PingUploadWorker`](https://searchfox.org/glean/source/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt).
```

docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,62 @@ The Glean SDK leverages the available platform capabilities to implement any net
...
+    UploadResult <|-- HttpResponse
+    UploadResult <|-- UnrecoverableFailure
+    UploadResult <|-- RecoverableFailure
+    PingUploader <|-- HttpUploader
+    PingUploader o-- BaseUploader
+
+```
+
+* The `PingUploader` interface describes the contract between the `BaseUploader` and the SDK or user-provided upload modules.
+* There is a `BaseUploader` component that is responsible for interfacing with the lower level FFI `get_upload_*` calls and dealing with the logic in a platform-coherent way.
```
brizental, line 56:
```suggestion
* There is a `BaseUploader` component that is responsible for interfacing with the lower level FFI `get_upload_task` calls and dealing with the logic in a platform-coherent way.
```

There is just the `get_upload_task` function and no other "getter"s. No need for asterisk I think.

docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,62 @@ The Glean SDK leverages the available platform capabilities to implement any net
...
+  * Every Glean instance will always have a single `BaseUploader` instance.
+  * The `BaseUploader` is fed, at Glean initialization, with an instance of an implementation of the `PingUploader` interface.
+  * Whenever `BaseUploader` thinks it should perform an upload, it will call the provided instance of the `PingUploader` interface and call `upload` with the data it's getting from the glean-core/FFI.
+  * Any throttling happens at this layer!
```
brizental, line 60:
```suggestion
  * Any throttling happens at this layer.
```
Any specific reason for the exclamation here? One more thing, do you think it is necessary to point out that the core is what should orchestrate the throttling and the binding will just abide to what the core is telling it to do?

docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,62 @@ The Glean SDK leverages the available platform capabilities to implement any net
 Glean core controls all upload and coordinates the platform side with its own internals.
 All language bindings implement ping uploading around a common API and protocol.
 
+## The upload module in the language bindings
+This paragraph illustrates the recommended structure of the upload module implementation in the language bindings (regardless of the target language), in order to allow greater flexibility with respect to potential future requirements and features.
```
chutten, line 11: I don't think this "intent sentence" is necessary.

docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,62 @@ The Glean SDK leverages the available platform capabilities to implement any net
...
+        +Upload() UploadResult
+    }
+
+    UploadResult <|-- HttpResponse
+    UploadResult <|-- UnrecoverableFailure
+    UploadResult <|-- RecoverableFailure
+    PingUploader <|-- HttpUploader
+    PingUploader o-- BaseUploader
+
+```
```
chutten, line 53: Is there a way to provide a text caption to this figure?

docs/dev/core/internal/upload.md
chutten: Also, not all language bindings use FFI calls, so I think we can skip the word `FFI` here.

docs/dev/core/internal/upload.md
Dexterp37: Not that I know of, I'm afraid :(

docs/dev/core/internal/upload.md
Dexterp37: Oh, I tend to get overexcited with punctuation! Sorry about that!

Yes, it makes sense to clarify the throttling part. Let me add that.

docs/dev/core/internal/upload.md
```diff
@@ -7,6 +7,67 @@ The Glean SDK leverages the available platform capabilities to implement any net
...
+
+### `PingUploader`
+The `PingUploader` interface describes the contract between the `BaseUploader` and the SDK or user-provided upload modules.
+
+### `BaseUploader`
+The `BaseUploader` component that is responsible for interfacing with the lower level `get_upload_task` calls and dealing with the logic in a platform-coherent way.
```
chutten, line 58:
```suggestion
The `BaseUploader` component is responsible for interfacing with the lower level `get_upload_task` calls and dealing with the logic in a platform-coherent way.
```

docs/dev/core/internal/upload.md
brizental: Since markdown accepts any valid HTML, you could put this inside a [`figure`](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/figure) tag and provide a `figcaption` right next to it. Apparently mermaid creates an `img` tag where all this code is, so that would work perfectly. I could be wrong as to how mermaid works though.

Dexterp37: This doesn't seem to work, at least locally, for me :\ The mermaid doesn't seem to get parsed?

brizental: Hum, oh well. Maybe we can open a follow up to look into this with more time. I see the other upload graph we have also doesn't have a caption so it could be a more broad thing where we add captios to all images in the book.

Dexterp37: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1652705

PR #1057, opened 2020-07-13
Bug 1651669: Fix warnings about outdated pip

mdboom:
This just simply upgrades miniconda to the latest version available.

We had held it back previously due to breakage on Windows, so we need
to confirm that has been fixed by:

1) Deleting the `~/.gradle/glean` directory
2) Building a clean build of Glean inside AndroidStudio

That should be sufficient to confirm this change works.

This is the *easy* solution that doesn't require forcibly updating pip etc.

### DIFF
--- a/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
+++ b/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
@@ -37,7 +37,7 @@ class GleanPlugin implements Plugin<Project> {
     private String GLEAN_PARSER_VERSION = "1.24.0"
     // The version of Miniconda is explicitly specified.
     // Miniconda3-4.5.12 is known to not work on Windows.
-    private String MINICONDA_VERSION = "4.5.11"
+    private String MINICONDA_VERSION = "py38_4.8.3"
 
     private String TASK_NAME_PREFIX = "gleanGenerateMetrics"
 


Dexterp37:
Unfortunately this still leads to `pip is configured with locations that require TLS/SSL, however the ssl module in Python is not available.` on my machine.

mdboom:
This is the same issue we ran into the last time we tried to upgrade miniconda?

It might be https://github.com/conda/conda/issues/6064, which seems to suggest that other things on your PATH break this (but also that recent versions of Python no longer make it an issue).  But it seems like something pretty full of terrible interactions between various things...

Dexterp37:
> 
> 
> This is the same issue we ran into the last time we tried to upgrade miniconda?

Yes, that's the same exact issue :(

> It might be [conda/conda#6064](https://github.com/conda/conda/issues/6064), which seems to suggest that other things on your PATH break this (but also that recent versions of Python no longer make it an issue). But it seems like something pretty full of terrible interactions between various things...

Yes, that's the issue. Last time I followed some of the proposed solutions there, without any luck. I had to revert back :\ Unfortunately, even removing the potentially interacting things from the PATH did not help.

@mdboom maybe we can schedule some time to investigate this a bit further? I wonder if we could get some other Windows dev to test this as well...

mdboom:
Interesting. The linked bug seems to suggest it should either be resolved or WONTFIX (it's a little hard to tell).

To fix bug 1651669, we could also update pip immediately after installing miniconda using conda -- that seemed kind of hackish so I didn't want to try that first.

In the long run, we'll still need to be able to do this -- eventually there will be bug or security fixes to Python etc. we need that we can only get by upgrading miniconda...

Yesterday, I also looked into using PyInstaller to make a "standalone" executable of glean_parser that we run during the build (instead of using miniconda), and that's full of foot guns, too...  Probably could be made to work, but that's just very brittle tech still...

mdboom:
I'm closing this as WONTFIX.  The warnings haven't really ever been noticed by anyone prior to the requirement for offline builds.  The currently proposed solution for offline builds is for users to "bring their own Python", and they can therefore "bring their own pip", so I don't think we need to address it at this level.

PR #1058, opened 2020-07-14
Bug 1646706 - Manage log_pings from core and add docs about debugging through env vars

brizental:
I also fix an intermittent test failure introduced by the previous PR, I didn't realize this was a problem until I ran some final tests before pushing this branch.

This changes will  also fix https://bugzilla.mozilla.org/show_bug.cgi?id=1648007 and https://bugzilla.mozilla.org/show_bug.cgi?id=1648012 for free, in case we only want C# and Python to manage debug optins through environment variables.

Finally I also add docs about debugging through environment variables here.


docs/user/debugging/ios.md
```diff
@@ -2,6 +2,14 @@
 
 For debugging and validation purposes on iOS, Glean makes use of a custom URL scheme which is implemented _within the application_ that is consuming Glean.  Glean provides some convenience functions to facilitate this, but it's up to the consuming application to enable this functionality.  Applications that enable this Glean SDK feature will be able to launch the application from a URL with the Glean debug commands embedded in the URL itself.
 
+Debugging features in iOS can also be enabled using environment variables.
+For more information on the available features accessible through this method
+and how to enable them, see [Enabling debugging features through environment variables](./index.md).
+
+These environment variables must be set on the device that is running the application.
+
+> **Note** To set environment variables to the process running your app in an iOS device or emulator you need to edit the scheme for your app. In the Xcode IDE, you can use the shortcut `Cmd + <` to open the scheme editor popup. The environment variables editor is under the `Arguments` tab on this popup.
```
brizental, line 11: Hey @travis79 , do you know any other ways to do this and do you think this description is clear?

glean-core/ffi/examples/glean_app.c
```diff
@@ -45,14 +45,15 @@ int main(void)
...
+  glean_set_log_pings(1);
+  glean_get_upload_task(&task);
 
   while (task.tag != FfiPingUploadTask_Done) {
       printf("tag: %d\n", task.tag);
 
       if (task.tag == FfiPingUploadTask_Upload) {
           printf("path: %s\n", task.upload.path);
-          printf("body length: %lld\n", task.upload.body.len);
+          printf("body length: %d\n", task.upload.body.len);
```
brizental, line 56: I was getting a warning about this everytime I built the C app. Was there a readon for the previous choice?

docs/user/debugging/android.md
```diff
@@ -54,6 +54,12 @@ persist until the application is closed or manually reset.
 >   --es tagPings test-metrics-ping
 > ```
 
+Debugging features in Android can also be enabled using environment variables.
+For more information on the available features accessible through this method
+and how to enable them, see [Enabling debugging features through environment variables](./index.md).
+
+These environment variables must be set on the device that is running the application.
```
brizental, line 61: Should we add a note here saying we actually don't know how to do this? Or should we just add not add this to the Android docs yet?

docs/user/debugging/ios.md
travis79: One thing I would ask is can we put this under a separate heading?  I think it may lead the user to believe that they must set the environment variables to use the other iOS debugging methods, so anything we can do to indicate that these are separate and exclusive debugging methods.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -62,59 +62,23 @@ internal class FfiConfiguration(
...
  * @property maxEvents the number of events to store before the events ping is sent
- * @property logPings whether to log ping contents to the console. This is only meant to be used
- *           internally by the `GleanDebugActivity`.
  * @property httpClient The HTTP client implementation to use for uploading pings.
  * @property channel the release channel the application is on, if known. This will be
  *           sent along with all the pings, in the `client_info` section.
  */
-data class Configuration internal constructor(
-    val serverEndpoint: String,
+data class Configuration @JvmOverloads internal constructor(
```
travis79, line 69: Does the `internal` here (since the public one is being removed) still allow access for the consuming app to create a Configuration object?

glean-core/ios/Glean/Glean.swift
```diff
@@ -28,6 +28,7 @@ public class Glean {
     var initialized: Bool = false
     private var uploadEnabled: Bool = true
     private var debugViewTag: String?
+    var logPings: Bool?
```
travis79, line 31: This probably could have been left in the configuration and made internal, but I'm okay with this being a member variable of Glean.

brizental: I was just following the same method we used for the `uploadEnabled` and the `debugViewTag`. The only reason it is not also private is so that we can look at this property in the tests.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
brizental: This is a great question and shows my lack of experience with Kotlin. I actually don't know, but makes sense that making it internal would not allow access to consuming apps. I will verify, thanks for noticing.

docs/user/debugging/android.md
Dexterp37: Let's remove these docs, if we don't support it. Or clearly state that debugging using environment variables is not supported on Android.

docs/user/debugging/index.md
```diff
@@ -12,12 +12,25 @@
...
+Some of the debugging features described above may also be enabled using environment variables:
+
+- `logPings`: May be set by the `GLEAN_LOG_PINGS` environment variable. The accepted values are
+`true` or `false`. Any other value will be ignored.
+- `tagPings`: May be set by the `GLEAN_DEBUG_VIEW_TAG` environment variable. Any valid HTTP header value maybe set here
+(e.g. any value that matches the regex `[a-zA-Z0-9-]{1,20}`). Invalid values will be ignored.
+
+These variables must be set at runtime, not at compile time. They will be checked upon Glean initialization.
+
+Enabling debugging features using environment variables is available for all supported platforms.
```
Dexterp37, line 32: Except Android :)

glean-core/ffi/examples/glean_app.c
Dexterp37: Which warning?

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -6,35 +6,11 @@
 /// property for calculating the `FfiConfiguration`
 public struct Configuration {
```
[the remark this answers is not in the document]
travis79: I must have been doing that while you were reviewing the other parts 😄.  I'll give it one more look now.

glean-core/ios/Glean/Glean.swift
```diff
@@ -465,6 +470,19 @@ public class Glean {
         }
     }
 
+    /// Set the log_pings debug option,
+    /// when this option is `true` the pings that are successfully submitted get logged.
+    ///
+    /// - parameters:
+    ///     * value: The value of the option.
+    public func setLogPings(_ value: Bool) {
```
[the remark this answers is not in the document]
travis79: This would default to internal if `public` is removed.

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -63,7 +63,7 @@ class GleanDebugUtility {
             }
 
             if let logPings = parsedCommands.logPings {
-                Glean.shared.configuration?.logPings = logPings
+                Glean.shared.setLogPings(logPings)
```
travis79, line 66: This should still work if `setLogPings` is `internal` access.

glean-core/ios/Glean/Glean.swift
brizental: If you have any ideas on how to test without making this public, that might be best.

travis79: The default of `internal` here is fine, it allows us to access it in tests and it doesn't expose it to the public API.

glean-core/ios/Glean/Glean.swift
```diff
@@ -119,6 +120,10 @@ public class Glean {
                 _ = self.setDebugViewTag(self.debugViewTag!)
             }
 
+            if self.logPings != nil {
+                _ = self.setLogPings(self.logPings!)
```
travis79, line 124: Since `setLogPings` doesn't return anything, you should be able to get rid of the `_ =` part.  This pattern is just used to satisfy the compiler that we _know_ that the function is returning a value and we are explicitly ignoring it.

travis79: Another more swift-like way of writing this might be:
```Swift
if let logPings = self.logPings {
    self.setLogPings(logPings)
}
```

This is a form of optional unwrapping that only falls into the condition if the value is not nil.  Assignment of an optional value in swift returns true if there is a value or false if it is nil.

glean-core/ios/Glean/Glean.swift
```diff
@@ -465,6 +470,19 @@ public class Glean {
...
     }
 
+    /// Set the log_pings debug option,
+    /// when this option is `true` the pings that are successfully submitted get logged.
+    ///
+    /// - parameters:
+    ///     * value: The value of the option.
+    public func setLogPings(_ value: Bool) {
+        if self.isInitialized() {
+            glean_set_log_pings(value.toByte())
```
travis79, line 480: Should we also update the cached `logPings` member variable here?  I think it should also be updated to reflect the state of ping logging in the core, in this case if we are initialized then it sets it in the core but the local variable doesn't get set.

glean-core/ios/Glean/Glean.swift
```diff
@@ -570,5 +587,7 @@ public class Glean {
         // Init Glean.
         testDestroyGleanHandle()
         initialize(uploadEnabled: uploadEnabled, configuration: configuration)
+        // Enable ping logging for all tests
+        setLogPings(true)
```
travis79, line 591: if you set this after `Glean.initialize` then we might miss logging the pings generated in `initialize`

glean-core/ios/Glean/Glean.swift
travis79, line 31: This could also be given a default value of false rather than being an Optional.

glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
```diff
@@ -20,38 +20,37 @@ class GleanDebugUtilityTests: XCTestCase {
...
-        XCTAssertFalse(Glean.shared.configuration!.logPings)
+        // We destroy the Glean handle so that Glean in in an unitialized state,
+        // this way it will save the value of `logPings` in the `logPings` prop.
+        Glean.shared.testDestroyGleanHandle()
 
         // Test toggle true
         var url = URL(string: "test://glean?logPings=true")
         Glean.shared.handleCustomUrl(url: url!)
-        XCTAssertTrue(Glean.shared.configuration!.logPings)
+        XCTAssertTrue(Glean.shared.logPings!)
```
travis79, line 30: If `logPings` weren't Optional, you wouldn't need the `!` (force-unwrap) operator here, which could cause a runtime exception in tests if `logPings` were `nil`

glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
```diff
@@ -20,38 +20,37 @@ class GleanDebugUtilityTests: XCTestCase {
...
         Glean.shared.handleCustomUrl(url: url!)
-        XCTAssertEqual(previousValue, Glean.shared.configuration!.logPings)
-    }
+        XCTAssertEqual(previousValue, Glean.shared.logPings)
 
-    func testHandleCustomUrlWrongHost() {
         // This should NOT set the logPings to true or false because it doesn't
         // match the required host "glean".
-        let url = URL(string: "test://not-glean?logPings=true")
+        url = URL(string: "test://not-glean?logPings=true")
```
travis79, line 51: I think you need to leave the `let` alone here, it is part of the variable definition of the `url`

glean-core/ios/Glean/Config/Configuration.swift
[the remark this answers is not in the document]
Dexterp37: Ha! Nice :-D Reviewing this took me a bit longer than expected :-D Thanks for taking care of this!

docs/user/debugging/index.md
brizental: Hm, but it is available for Android. We just haven't figured out how to set environment variables for Android apps without changing the binary, if the user doesn't care about changing the binary they can totally use this method.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -62,59 +62,23 @@ internal class FfiConfiguration(
...
+    val serverEndpoint: String = DEFAULT_TELEMETRY_ENDPOINT,
     val channel: String? = null,
     val maxEvents: Int? = null,
-    val logPings: Boolean = DEFAULT_LOG_PINGS,
```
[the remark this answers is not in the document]
brizental: Good reminder that I actually need to add a changelog entry to this PR.

glean-core/ffi/examples/glean_app.c
brizental:
```
cc  -I.. -o glean_app glean_app.c ../../../target/debug/libglean_ffi.dylib
glean_app.c:55:41: warning: format specifies type 'long long' but the argument has type
      'int32_t' (aka 'int') [-Wformat]
          printf("body length: %lld\n", task.upload.body.len);
                               ~~~~     ^~~~~~~~~~~~~~~~~~~~
                               %d
```

glean-core/ios/Glean/Glean.swift
brizental: >  This pattern is just used to satisfy the compiler that we know that the function is returning a value and we are explicitly ignoring it.

Oops, copy pasted from the `debugViewTag` part and didn't realize this was not necessary. Good catch.

glean-core/ios/GleanTests/Debug/GleanDebugUtilityTests.swift
brizental: I can't, I made it all into one test (see https://github.com/mozilla/glean/pull/1058/files#diff-4f1b6d3f583d696ae3f4d66b64ae89d2L49). Leaving `let` gives me an error for trying to re-instantiate a variable with the same identifier as a previously created one.

travis79: Oh, sorry, I overlooked that :)  Carry on!

docs/user/debugging/index.md
Dexterp37: Mh, ok, you have a point hee.

docs/user/debugging/android.md
brizental: Like I said in the other comment it is supported. My thinking when I added this docs was to add it and let the user try to figure out how to set environment variables in an Android device. When I saw it in the PR I started doubting this and thought maybe we could just remove this docs from Android for now, with the intention of not bringing attention to something we might not know how to answer in case a user asks us.

Dexterp37: I don't think "it's supported": users can't actually use it, by design. If they ask us how to do this, we'd just shrug 🤷 - Let's clearly state that while the Glean SDK can read environment variables, we don't know how to set them on the device, so other methods should be used.

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
Dexterp37: No, it doesn't - this should be removed as suggested in my review :)

glean-core/ios/Glean/Glean.swift
brizental: That would go against what we are doing for the `debugViewTag` and `setUploadEnabled`. I am not necessarily opposed to it, I just think if we do it for `logPings` we should do it for the others.

docs/user/debugging/android.md
brizental: I agree, I will remove this docs and add a note on `index.md`.

glean-core/ios/Glean/Glean.swift
Dexterp37: I'm not sure we need to do this, since we should only hit the "uninitialized" part before... well, being initialized

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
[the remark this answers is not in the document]
brizental: Now that I look at it, I'm wondering: is it really a breaking change if it only changes the `internal` constructor?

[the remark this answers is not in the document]
Dexterp37: You are correct. Not a breaking change, but still requires an entry in the changelog :)

glean-core/ios/Glean/Glean.swift
travis79: Works for me then!

glean-core/ios/Glean/Glean.swift
```diff
@@ -115,8 +116,12 @@ public class Glean {
                 return
             }
 
-            if self.debugViewTag != nil {
-                _ = self.setDebugViewTag(self.debugViewTag!)
+            if let debugViewTag = self.debugViewTag {
+                self.setDebugViewTag(debugViewTag)
```
travis79, line 120: Since this _does_ return a value that we are ignoring, you probably need to leave the `_ =` to satisfy the compiler/lint warning.

travis79: But it looks like it passed so nevermind...  /shrug

PR #1062, opened 2020-07-16
Bug 1650787 - Implement the JWE metric type on the core and FFI

brizental:
I decided to separate this into two parts, this one just with the core and ffi changes, the next with the addition of this metric type to the bindings.

I found a bug on https://github.com/mozilla/glean/blob/main/glean-core/src/error_recording.rs#L61, which the bindings PR will depend on too. If you notice, we jump on the count from 2 directly to 4. I'll send a separate PR to fix that.

This implementation was based on [this document](https://docs.google.com/document/d/1nntNIiE6braTGzoKf-lx21OVDd8ssyeIeJu3jnQQfEE/edit?usp=sharing).

### DIFF
--- a/glean-core/src/lib_unit_tests.rs
+++ b/glean-core/src/lib_unit_tests.rs
@@ -372,6 +372,7 @@ fn correct_order() {
         Timespan(Duration::new(5, 0), TimeUnit::Second),
         TimingDistribution(Histogram::functional(2.0, 8.0)),
         MemoryDistribution(Histogram::functional(2.0, 8.0)),
+        Jwe("a...a.".into()),
     ];
 
     for metric in all_metrics {
@@ -396,6 +397,7 @@ fn correct_order() {
             Timespan(..)                      => assert_eq!(10, disc),
             TimingDistribution(..)            => assert_eq!(11, disc),
             MemoryDistribution(..)            => assert_eq!(12, disc),
+            Jwe(..)                           => assert_eq!(13, disc),
         }
     }
 }
--- a/glean-core/src/lib_unit_tests.rs
+++ b/glean-core/src/lib_unit_tests.rs
@@ -372,7 +372,7 @@ fn correct_order() {
         Timespan(Duration::new(5, 0), TimeUnit::Second),
         TimingDistribution(Histogram::functional(2.0, 8.0)),
         MemoryDistribution(Histogram::functional(2.0, 8.0)),
-        Jwe("a...a.".into()),
+        Jwe("eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkEyNTZHQ00ifQ.OKOawDo13gRp2ojaHV7LFpZcgV7T6DVZKTyKOMTYUmKoTCVJRgckCL9kiMT03JGeipsEdY3mx_etLbbWSrFr05kLzcSr4qKAq7YN7e9jwQRb23nfa6c9d-StnImGyFDbSv04uVuxIp5Zms1gNxKKK2Da14B8S4rzVRltdYwam_lDp5XnZAYpQdb76FdIKLaVmqgfwX7XWRxv2322i-vDxRfqNzo_tETKzpVLzfiwQyeyPGLBIO56YJ7eObdv0je81860ppamavo35UgoRdbYaBcoh9QcfylQr66oc6vFWXRcZ_ZT2LawVCWTIy3brGPi6UklfCpIMfIjf7iGdXKHzg.48V1_ALb6US04U3b.5eym8TW_c8SuK0ltJ3rpYIzOeDQz7TALvtu6UG9oMo4vpzs9tX_EFShS8iB7j6jiSdiwkIr3ajwQzaBtQD_A.XFBoMYUZodetZdvTiFvSkQ".into()),
     ];
 
     for metric in all_metrics {
--- a/glean-core/src/metrics/jwe.rs
+++ b/glean-core/src/metrics/jwe.rs
@@ -1,0 +1,374 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fmt;
+use std::str::FromStr;
+
+use serde::Serialize;
+
+use crate::error_recording::{record_error, ErrorType};
+use crate::metrics::{Metric, MetricType};
+use crate::storage::StorageManager;
+use crate::CommonMetricData;
+use crate::Glean;
+
+const DEFAULT_MAX_CHARS_PER_ELEMENT: usize = 1024;
+
+/// Verifies if a string is [`BASE64URL`](https://tools.ietf.org/html/rfc4648#section-5) compliant.
+///
+/// As such, the string must match the regex: `[a-zA-Z0-9\-\_]*`.
+///
+/// > **Note** As described in the [JWS specification](https://tools.ietf.org/html/rfc7516#section-2),
+/// > the BASE64URL encoding used by JWE discards any padding,
+/// > that is why we can ignore that for this validation.
+///
+/// The regex crate isn't used here because it adds to the binary size,
+/// and the Glean SDK doesn't use regular expressions anywhere else.
+fn validate_base64url_encoding(value: &str) -> bool {
+    let mut iter = value.chars();
+
+    loop {
+        match iter.next() {
+            // We are done, so the whole expression is valid.
+            None => return true,
+            // Valid characters.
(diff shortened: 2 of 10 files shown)


glean-core/tests/jwe.rs
```diff
@@ -0,0 +1,65 @@
...
+
+    metric.set_with_compact_repr(&glean, SAMPLE_JWE_VALUE);
+    let snapshot = StorageManager
+        .snapshot_as_json(glean.storage(), "core", true)
+        .unwrap();
+
+    assert_eq!(
+        json!({"jwe": {"local.jwe_metric": SAMPLE_JWE_VALUE }}),
+        snapshot
+    );
```
Dexterp37, line 35: Please, let's use `metric.test_get_value` instead of this :)

glean-core/tests/jwe.rs
```diff
@@ -0,0 +1,65 @@
...
+    // Check that the data was correctly set in each store.
+    for store_name in store_names {
+        let snapshot = StorageManager
+            .snapshot_as_json(glean.storage(), &store_name, true)
+            .unwrap();
+
+        assert_eq!(
+            json!({"jwe": {"local.jwe_metric": SAMPLE_JWE_VALUE}}),
+            snapshot
+        );
```
Dexterp37, line 63: Let's use `test_get_value`

glean-core/tests/jwe.rs
```diff
@@ -0,0 +1,65 @@
...
+mod common;
+use crate::common::*;
+
+use serde_json::json;
+
+use glean_core::metrics::*;
+use glean_core::storage::StorageManager;
+use glean_core::{CommonMetricData, Lifetime};
+
+const SAMPLE_JWE_VALUE: &str = "eyJhbGciOiJFQ0RILUVTIiwia2lkIjoiMFZFRTdmT0txbFdHVGZrY0taRUJ2WWl3dkpMYTRUUGlJVGxXMGJOcDdqVSIsImVwayI6eyJrdHkiOiJFQyIsImNydiI6IlAtMjU2IiwieCI6InY3Q1FlRWtVQjMwUGwxV0tPMUZUZ25OQlNQdlFyNlh0UnZxT2kzSWdzNHciLCJ5IjoiNDBKVEpaQlMwOXpWNHpxb0hHZDI5NGFDeHRqcGU5a09reGhELVctUEZsSSJ9LCJlbmMiOiJBMjU2R0NNIn0..A_wzJya943vlHKFH.yq0JhkGZiZd6UiZK6goTcEf6i4gbbBeXxvq8QV5_nC4.Knl_sYSBrrP-aa54z6B6gA";
```
Dexterp37, line 14: Can you kindly also add a test to cover `test_get_value_as_json_string`?

glean-core/src/metrics/jwe.rs
```diff
@@ -0,0 +1,374 @@
...
+use crate::CommonMetricData;
+use crate::Glean;
+
+const DEFAULT_MAX_CHARS_PER_ELEMENT: usize = 1024;
+
+/// Verifies if a string is [`BASE64URL`](https://tools.ietf.org/html/rfc4648#section-5) compliant.
+///
+/// As such, the string must match the regex: `[a-zA-Z0-9\-\_]*`.
+///
+/// > **Note** As described in the [JWS specification](https://tools.ietf.org/html/rfc7516#section-2),
```
Dexterp37, line 22: Is this URL correct? Where is the discarding part specified?

glean-core/src/metrics/jwe.rs
```diff
@@ -0,0 +1,374 @@
...
+        let value = value.into();
+        match Jwe::from_str(&value) {
+            Ok(_) => glean
+                .storage()
+                .record(glean, &self.meta, &Metric::Jwe(value)),
+            Err((error_type, msg)) => record_error(glean, &self.meta, error_type, msg, None),
+        };
+    }
+
+    /// Build a JWE value from it's elements and set to it.
```
Dexterp37, line 201:
```suggestion
    /// Build a JWE value from its elements and set to them.
```

glean-core/src/metrics/jwe.rs
```diff
@@ -0,0 +1,374 @@
...
+        for element in [&header, &cipher_text].iter() {
+            if element.is_empty() {
+                return Err((
+                    ErrorType::InvalidValue,
+                    "Elements `header` and `cipher_text` must not be empty.",
+                ));
+            }
+        }
+
+        for element in [&header, &key, &init_vector, &cipher_text, &auth_tag].iter() {
```
Dexterp37, line 90: Note that only the variable-sized  fields should have at most 1024 characters, as [per our spec](https://docs.google.com/document/d/1nntNIiE6braTGzoKf-lx21OVDd8ssyeIeJu3jnQQfEE/edit?ts=5ee76aa1#heading=h.l4ya7hop8p4q). The others have other sizes (see the SDK API section)

glean-core/src/metrics/jwe.rs
brizental: https://tools.ietf.org/html/rfc7515#section-2, off by one. Nice catch!

brizental:
> In hindsight, I would have preferred to have at least one language binding implemented here, to have better confidence on the "language to core" mechanism.

I have kotlin, swift and c# implemented already. I can add any one of them to this PR if you prefer.

glean-core/tests/jwe.rs
brizental: Hm, I believe not getting it from `get_test_value` gives us another assurance that we are storing the value we expect to be stored.

brizental: I can always add an explicit test to `get_test_value` which might be even better.

glean-core/src/metrics/jwe.rs
Dexterp37: Mh, by looking at the referenced [sample implementation](https://tools.ietf.org/html/rfc7515#appendix-C), looks like `=` is still an allowed character.

Dexterp37:
> 
> 
> > In hindsight, I would have preferred to have at least one language binding implemented here, to have better confidence on the "language to core" mechanism.
> 
> I have kotlin, swift and c# implemented already. I can add any one of them to this PR if you prefer.

Nah, that's fine, let's not over-rotate on this for now. Let's fix the problems here and move on. Next time we can do it differently! :)

glean-core/src/metrics/jwe.rs
brizental: It is allowed, BASE64URL may have the `=`. For the purposes of JWE and JWS though, the padding is not present. From the RFC:
> Base64 encoding using the URL- and filename-safe character set defined in Section 5 of RFC 4648 [RFC4648], **with all trailing '=' characters omitted (as permitted by Section 3.2)** [...]

Notice that the sample implementation removes the trailing `=` upon encoding, and re-adds them upon decoding. 

Since we need to validate that a string is valid BASE64URL (without padding) and not encode or decode, we don't expect any `=`.

glean-core/src/lib_unit_tests.rs
```diff
@@ -372,6 +372,7 @@ fn correct_order() {
         Timespan(Duration::new(5, 0), TimeUnit::Second),
         TimingDistribution(Histogram::functional(2.0, 8.0)),
         MemoryDistribution(Histogram::functional(2.0, 8.0)),
+        Jwe("a...a.".into()),
```
brizental, line 375: Reminder for my Monday self: put a valid value here!

I know there is no validation, but just for aesthetic reasons.

glean-core/src/metrics/jwe.rs
```diff
@@ -0,0 +1,468 @@
...
+            // An invalid character.
+            Some(_) => return false,
+        }
+    }
+}
+
+/// Representation of a [JWE](https://tools.ietf.org/html/rfc7516).
+///
+/// **Note** Variable sized elements will be constrained to a length of DEFAULT_MAX_CHARS_PER_VARIABLE_SIZE_ELEMENT,
+/// this is a constraint introduced by Glean and not part of the spec.
```
Dexterp37, line 46:
```suggestion
/// this is a constraint introduced by Glean to prevent abuses and not part of the spec.
```

glean-core/src/metrics/jwe.rs
```diff
@@ -0,0 +1,468 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fmt;
+use std::str::FromStr;
+
+use serde::Serialize;
+
+use crate::error_recording::{record_error, ErrorType};
```
Dexterp37, line 10: Would you kindly add a changelog entry?

PR #1069, opened 2020-07-17
Bug 1653244 - Fix experiment info getting cleared before getting into metrics ping

travis79:
This is the fix for the Android/Kotlin bindings and a modified test to cover the case.

### DIFF
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
@@ -476,7 +476,8 @@ class MetricsPingSchedulerTest {
     }
 
     @Test
-    fun `startup ping sends old version when upgraded`() {
+    @Suppress("LongMethod")
+    fun `startup ping sends correct data when upgraded`() {
         // Start the web-server that will receive the metrics ping.
         val server = getMockWebServer()
         val configuration = Configuration(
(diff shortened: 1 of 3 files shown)


travis79:
> LGTM -- but maybe this one especially should have @Dexterp37's eye given the metrics ping scheduler involvement?
> 
> Does Swift also suffer the same issue and should we file a bug to do that work?

Swift, fortunately, does not suffer from this so nothing to change there.  This appeared to be an Android only issue, and only since init was made async.

glean-core/android/src/test/java/mozilla/telemetry/glean/scheduler/MetricsPingSchedulerTest.kt
```diff
@@ -510,10 +511,15 @@ class MetricsPingSchedulerTest {
             val expectedValue = "canary"
             expectedStringMetric.set(expectedValue)
 
+            // Set an experiment active to verify it gets sent in the pings
+            Glean.setExperimentActive("test-experiment", "a")
```
Dexterp37, line 515: This should really be an integration test for our sample app, along with most of the tests in this file. Nothing you can do here, though - this whole thing would require its own bug. @travis79 would you kindly file it?

travis79: [Bug filed](https://bugzilla.mozilla.org/show_bug.cgi?id=1653955)

PR #1073, opened 2020-07-20
Bug 1650787 - Implement JWE metric in the bindings

brizental:


glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
```diff
@@ -0,0 +1,172 @@
...
+     * @return true if metric value exists, otherwise false
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    @JvmOverloads
+    fun testHasValue(pingName: String = sendInPings.first()): Boolean {
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.assertInTestingMode()
+
+        val res = LibGleanFFI.INSTANCE.glean_jwe_test_has_value(this.handle, pingName)
+        return res.toBoolean()
```
Dexterp37, line 106:
```suggestion
        return LibGleanFFI.INSTANCE.glean_jwe_test_has_value(this.handle, pingName).toBoolean()
```

?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
```diff
@@ -0,0 +1,172 @@
...
+            disabled = disabled.toByte()
+        )
+    }
+
+    /**
+     * Set a JWE value.
+     *
+     * @param value The compact representation of a JWE value.
+     */
+    fun setWithCompactRepr(value: String) {
```
Dexterp37, line 56:
```suggestion
    fun setWithCompactRepresentation(value: String) {
```

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
```diff
@@ -0,0 +1,172 @@
...
+     * writing to the the metric's storage engine before returning a value.
+     *
+     * @param pingName represents the name of the ping to retrieve the metric for.
+     *                 Defaults to the first value in `sendInPings`.
+     * @return value of the stored metric
+     * @throws [NullPointerException] if no value is stored
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    @JvmOverloads
+    fun testGetValueAsPeriodDelimitedString(pingName: String = sendInPings.first()): String {
```
Dexterp37, line 143:
```suggestion
    fun testGetCompactRepresentation(pingName: String = sendInPings.first()): String {
```

Maybe that's clearer?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
brizental: https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringMetricType.kt#L91-L92 copy paste.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
brizental: Maybe. The name came from the [the document](https://docs.google.com/document/d/1nntNIiE6braTGzoKf-lx21OVDd8ssyeIeJu3jnQQfEE/edit?usp=sharing).

brizental: I don't have strong opinions though, if yo uthink it's best I can change.

Dexterp37: Yeah, I know the source (me :-) ) - I just wonder if this makes sense, since we already call it compact representation above. Up to you to decide!

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
Dexterp37: Yeah, not sure why the original does that. Maybe we used to do something else as well

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
brizental: Sure, let's make it consistent. I'll change.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
```diff
@@ -0,0 +1,173 @@
...
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    @JvmOverloads
+    fun testGetValue(pingName: String = sendInPings.first()): String {
+        @Suppress("EXPERIMENTAL_API_USAGE")
+        Dispatchers.API.assertInTestingMode()
+
+        if (!testHasValue(pingName)) {
+            throw NullPointerException()
+        }
+        val ptr = LibGleanFFI.INSTANCE.glean_jwe_test_get_value_as_json_string(this.handle, pingName)!!
```
mdboom, line 128: Is this is really in JSON format, that should probably be documented.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/JweMetricTypeTest.kt
```diff
@@ -0,0 +1,141 @@
...
+    @Test(expected = NullPointerException::class)
+    fun `testGetValue() throws NullPointerException if nothing is stored`() {
+        val jweMetric = JweMetricType(
+            disabled = true,
+            category = "telemetry",
+            lifetime = Lifetime.Application,
+            name = "jwe_metric",
+            sendInPings = listOf("store1")
+        )
+        jweMetric.testGetValue()
```
mdboom, line 95: We should also test the result of `testGetValue` (in another test).

glean-core/python/glean/metrics/jwe.py
```diff
@@ -0,0 +1,192 @@
...
+        Returns:
+            value (int): value of the stored metric.
+        """
+        if ping_name is None:
+            ping_name = self._send_in_pings[0]
+
+        if not self.test_has_value(ping_name):
+            raise ValueError("metric has no value")
+
+        return _ffi.ffi_decode_string(
```
mdboom, line 139: Is this is JSON, it's probaly more natural to return the parsed JSON objects rather than a string of JSON. (But I also don't fully understand the use cases...)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
brizental: Yes, really JSON format: https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/jwe.rs#L310-L312.

The original idea in the design document was to return a custom struct (e.g. `JweData`), returning a JSON string was easier on the FFI and also more in sync with what we do for example for [custom distributions](https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/custom_distribution.rs#L170).

Anyways, I'll document :)

glean-core/android/src/test/java/mozilla/telemetry/glean/private/JweMetricTypeTest.kt
brizental: We already have a test for it on the core: https://github.com/mozilla/glean/blob/main/glean-core/tests/jwe.rs#L92, do you think it's necessary to test again on each binding?

glean-core/python/glean/metrics/jwe.py
brizental: I agree that it probably makes sense to parse the JSON on the bindings, like I said in a previous response: returning this as a JSON was to follow a guideline from the [design doc](https://docs.google.com/document/d/1nntNIiE6braTGzoKf-lx21OVDd8ssyeIeJu3jnQQfEE/edit). I also am not sure exactly about the use cases, now that I think about it. @Dexterp37 what do you think?

Dexterp37: In general, we use JSON here and in other types as a convenient way to just pass a string over FFI instead of defining structures. We should probably do this properly, someday, in order to unblock the [Android test speedup as well](https://github.com/mozilla/glean/pull/837#issuecomment-617640739).

Looking back at the spec, we don't ever say that this should return JSON. We say that ` returns an object representing the different parts that make up the JWE token, i.e. header, key, initVector, cipherText, authTag`. A tuple or a struct is fine. 

Good catch @mdboom

glean-core/python/glean/metrics/jwe.py
```diff
@@ -0,0 +1,192 @@
...
+    def test_get_value(self, ping_name: Optional[str] = None) -> str:
+        """
+        Returns the stored value for testing purposes only.
+
+        Args:
+            ping_name (str): (default: first value in send_in_pings) The name
+                of the ping to retrieve the metric for.
+
+        Returns:
+            value (int): value of the stored metric.
```
Dexterp37, line 131: nit: this is returning a string, not an int

glean-core/python/glean/metrics/jwe.py
```diff
@@ -0,0 +1,192 @@
...
+        """
+        Returns the stored value as the compact representation
+        for testing purposes only.
+
+        Args:
+            ping_name (str): (default: first value in send_in_pings) The name
+                of the ping to retrieve the metric for.
+
+        Returns:
+            value (int): value of the stored metric.
```
Dexterp37, line 155: wrong docs :)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/JweMetricType.kt
mdboom: LGTM. This matches the approach we've taken a bunch of other places (like events, for example).

glean-core/ios/Glean/Metrics/JweMetric.swift
```diff
@@ -0,0 +1,179 @@
...
+            throw "Missing value"
+        }
+
+        var data: JweData?
+
+        let jsonString = String(freeingRustString: glean_jwe_test_get_value_as_json_string(self.handle, pingName))
+        if let jsonData: Data = jsonString.data(using: .utf8, allowLossyConversion: false) {
+            if let json = try? JSONSerialization.jsonObject(with: jsonData, options: []) as? [String: Any] {
+                data = JweData(
+                    header: json["header"] as! String,
```
travis79, line 123: Using the force cast (`as!`) could cause a crash here if anything went wrong here. Maybe it would be safer to provide a default value instead of crashing?   This can be done something like:
```suggestion
                    header: (json["header"] as? String) ?? "default",
```

Or you might also want to ensure that the value is available under that key before passing it into the `JweData` struct initializer.  If for some reason the value wasn't available for that key, it could also cause the force cast to fail/crash.

glean-core/ios/Glean/Metrics/JweMetric.swift
```diff
@@ -0,0 +1,179 @@
...
+                    key: json["key"] as! String,
+                    initVector: json["init_vector"] as! String,
+                    cipherText: json["cipher_text"] as! String,
+                    authTag: json["auth_tag"] as! String
+                )
+            }
+        }
+
+        return data!
```
travis79, line 132: This could also cause a crash if `data` is `nil` for some reason.  Is there any chance that it would be `nil` at this point?  If so, we should probably try and avoid the crash and record/log an error here instead.

glean-core/ios/GleanTests/Metrics/JweMetricTests.swift
```diff
@@ -0,0 +1,133 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+@testable import Glean
+import XCTest
+
+// swiftlint:disable force_cast
+// REASON: Used in a test
+class JweMetricTests: XCTestCase {
+    // swiftlint:disable line_length
+    private var header: String = "eyJhbGciOiJSU0EtT0FFUCIsImVuYyI6IkEyNTZHQ00ifQ"
```
travis79, line 12: nit: If the test doesn't change these values anywhere, then they should be defined as constant with `let`

glean-core/ios/Glean/Metrics/JweMetric.swift
travis79: If you say it's not possible to fail here, then I trust you :)

brizental: I knew that this could cause a crash upon implementing, here is my thinking:
1. This is a testing only function;
2. A crash here would mean something is terribly wrong (this should never really crash, crashing would mean we either stored an invalid JWE value on `set` or that something went wrong while passing the JSON string through the FFI);

In this case I think crashing would be fine, since it is extremely unlikely and would be a helpful debugging crash.

glean-core/ios/Glean/Metrics/JweMetric.swift
brizental: Same crash thinking as the above comment.

glean-core/ios/Glean/Metrics/JweMetric.swift
travis79: Okay, since this is test-only and it should never happen unless something is wrong, I am fine with using the force-cast here.

PR #1074, opened 2020-07-20
Bug 1634064 - Implement the 'X-Source-Tags' header

Dexterp37:
This implements the new ping tagging system in the Glean SDK. It adds a CLI option for `adb`, all the glean-core implementation and environment variable support.

Please note that I had to refactor the debug options module a bit in order to allow using non-`String` option types.

### DIFF
--- a/docs/user/debugging/android.md
+++ b/docs/user/debugging/android.md
@@ -22,6 +22,7 @@ In the above:
 | `logPings` | boolean (`--ez`)  | If set to `true`, pings are dumped to logcat; defaults to `false` |
 | `sendPing` | string (`--es`)  | Sends the ping with the given name immediately |
 | `debugViewTag` | string (`--es`)  | Tags all outgoing pings as debug pings to make them available for real-time validation, on the [Glean Debug View](./debug-ping-view.md). The value must match the pattern `[a-zA-Z0-9-]{1,20}`. **Important**: in older versions of the Glean SDK, this was named `tagPings` |
+| `sourceTags` | string array (`--esa`)  | Tags outgoing pings with a maximum of 5 comma-separated tags. The tags must match the pattern `[a-zA-Z0-9-]{1,20}`. The `automation` tag is meant for tagging pings generated on automation: such pings will be specially handled on the pipeline (i.e. discarded from [non-live views](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming)). Tags starting with `glean` are reserved for future use. Subsequent calls of this overwrite any previously stored tag |
 
 All [the options](https://developer.android.com/studio/command-line/adb#am) provided to start the activity are passed over to the main activity for the application to process.
 This is useful if SDK users wants to debug telemetry while providing additional options to the product to enable specific behaviors.  
--- a/docs/user/debugging/android.md
+++ b/docs/user/debugging/android.md
@@ -21,7 +21,7 @@ In the above:
 |---|----|-----------|
 | `logPings` | boolean (`--ez`)  | If set to `true`, pings are dumped to logcat; defaults to `false` |
 | `sendPing` | string (`--es`)  | Sends the ping with the given name immediately |
-| `tagPings` | string (`--es`)  | Tags all outgoing pings as debug pings to make them available for real-time validation, on the [Glean Debug View](./debug-ping-view.md). The value must match the pattern `[a-zA-Z0-9-]{1,20}` |
+| `debugViewTag` | string (`--es`)  | Tags all outgoing pings as debug pings to make them available for real-time validation, on the [Glean Debug View](./debug-ping-view.md). The value must match the pattern `[a-zA-Z0-9-]{1,20}`. **Important**: in older versions of the Glean SDK, this was named `tagPings` |
 
 All [the options](https://developer.android.com/studio/command-line/adb#am) provided to start the activity are passed over to the main activity for the application to process.
 This is useful if SDK users wants to debug telemetry while providing additional options to the product to enable specific behaviors.  
@@ -34,7 +34,7 @@ For example, to direct a release build of the Glean sample application to (1) du
 adb shell am start -n org.mozilla.samples.gleancore/mozilla.telemetry.glean.debug.GleanDebugActivity \
   --ez logPings true \
   --es sendPing metrics \
-  --es tagPings test-metrics-ping
+  --es debugViewTag test-metrics-ping
 ```
 
 The `logPings` command doesn't trigger ping submission and you won't see any output until a ping has been sent. You can use the `sendPing` command to force a ping to be sent, but it could be more desirable to trigger the pings submission on their normal schedule. For instance, the `baseline` and `events` pings can be triggered by moving the app out of the foreground and the `metrics` ping can be triggered normally if it is overdue for the current calendar day.
@@ -54,7 +54,7 @@ persist until the application is closed or manually reset.
 > adb shell am start -n org.mozilla.samples.gleancore/mozilla.components.service.glean.debug.GleanDebugActivity \
 >   --ez logPings true \
 >   --es sendPing metrics \
->   --es tagPings test-metrics-ping
+>   --es debugViewTag test-metrics-ping
 > ```
 
 ### Glean Log messages
--- a/docs/user/debugging/index.md
+++ b/docs/user/debugging/index.md
@@ -13,7 +13,7 @@
 There are 3 available commands that you can use with the Glean SDK debug tools
 
 - `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log.
-- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
+- `debugViewTag`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
 - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
 
 Different platforms have different ways to send these commands.
@@ -24,7 +24,7 @@ Some of the debugging features described above may also be enabled using environ
 
 - `logPings`: May be set by the `GLEAN_LOG_PINGS` environment variable. The accepted values are
 `true` or `false`. Any other value will be ignored.
-- `tagPings`: May be set by the `GLEAN_DEBUG_VIEW_TAG` environment variable. Any valid HTTP header value maybe set here
+- `debugViewTag`: May be set by the `GLEAN_DEBUG_VIEW_TAG` environment variable. Any valid HTTP header value maybe set here
 (e.g. any value that matches the regex `[a-zA-Z0-9-]{1,20}`). Invalid values will be ignored.
 
 These variables must be set at runtime, not at compile time. They will be checked upon Glean initialization.
--- a/docs/user/debugging/index.md
+++ b/docs/user/debugging/index.md
@@ -10,11 +10,12 @@
 
 ### Available commands and query format
 
-There are 3 available commands that you can use with the Glean SDK debug tools
+There are 4 available commands that you can use with the Glean SDK debug tools
 
 - `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log.
 - `debugViewTag`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
 - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
+- `sourceTags`: Tags outgoing pings with a maximum of 5, comma-separated, tags.
 
 Different platforms have different ways to send these commands.
 
@@ -25,6 +26,8 @@ Some of the debugging features described above may also be enabled using environ
 - `logPings`: May be set by the `GLEAN_LOG_PINGS` environment variable. The accepted values are
 `true` or `false`. Any other value will be ignored.
 - `debugViewTag`: May be set by the `GLEAN_DEBUG_VIEW_TAG` environment variable. Any valid HTTP header value maybe set here
+(e.g. any value that matches the regex `[a-zA-Z0-9-]{1,20}`). Invalid values will be ignored.
+- `sourceTags`: May be set by the `GLEAN_SOURCE_TAGS` environment variable. A comma-separated list of valid HTTP header values may be set here
 (e.g. any value that matches the regex `[a-zA-Z0-9-]{1,20}`). Invalid values will be ignored.
 
 These variables must be set at runtime, not at compile time. They will be checked upon Glean initialization.
--- a/docs/user/debugging/ios.md
+++ b/docs/user/debugging/ios.md
@@ -17,7 +17,7 @@ For debugging and validation purposes on iOS, Glean makes use of a custom URL sc
 There are 3 available commands that you can use with the Glean SDK debug tools
 
 - `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log
-- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
+- `debugViewTag`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters. **Important**: in older versions of the Glean SDK, this was named `tagPings`.
 - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
 
 The structure of the custom URL uses the following format:
@@ -90,7 +90,7 @@ Perhaps the simplest way to invoke the Glean SDK debug functionality is to open 
 Using the glean-sample-app as an example: to activate ping logging, tag the pings to go to the Glean Debug View, and force the `events` ping to be sent, enter the following URL in a web browser on the iOS device:
 
 ```shell
-glean-sample-app://glean?logPings=true&tagPings=My-ping-tag&sendPing=events
+glean-sample-app://glean?logPings=true&debugViewTag=My-ping-tag&sendPing=events
 ```
 
 This should cause iOS to prompt you with a dialog asking if you want to open the URL in the Glean Sample App, and if you select "Okay" then it will launch (or resume if it's already running) the application with the indicated commands and parameters and immediately force the collection and submission of the events ping.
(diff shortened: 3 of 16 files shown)


glean-core/src/debug.rs
```diff
@@ -33,63 +39,77 @@ pub struct DebugOptions {
...
+    pub sourge_tags: DebugOption<Vec<String>>,
 }
 
 impl DebugOptions {
     pub fn new() -> Self {
         Self {
-            log_pings: DebugOption::new(GLEAN_LOG_PINGS, None),
-            debug_view_tag: DebugOption::new(GLEAN_DEBUG_VIEW_TAG, Some(validate_debug_view_tag)),
+            log_pings: DebugOption::new(GLEAN_LOG_PINGS, get_bool_from_str, None),
+            debug_view_tag: DebugOption::new(GLEAN_DEBUG_VIEW_TAG, identity, Some(validate_tag)),
```
badboy, line 51: Fun trick but this should work:

```suggestion
            debug_view_tag: DebugOption::new(GLEAN_DEBUG_VIEW_TAG, Some, Some(validate_tag)),
```

`Some` is not only the variant of this enum, but also a function that constructs that variant and thus can be used as such.

glean-core/src/debug.rs
```diff
@@ -147,23 +191,49 @@ pub fn validate_debug_view_tag(value: String) -> Option<String> {
...
+        return false;
+    }
+
+    for tag in tags {
+        if !validate_tag(tag) {
+            return false;
+        }
+    }
+
+    true
```
badboy, line 234:
```suggestion
    tags.iter().all(validate_tag)
```

docs/user/debugging/android.md
```diff
@@ -21,7 +21,8 @@ In the above:
...
 | `sendPing` | string (`--es`)  | Sends the ping with the given name immediately |
-| `tagPings` | string (`--es`)  | Tags all outgoing pings as debug pings to make them available for real-time validation, on the [Glean Debug View](./debug-ping-view.md). The value must match the pattern `[a-zA-Z0-9-]{1,20}` |
+| `debugViewTag` | string (`--es`)  | Tags all outgoing pings as debug pings to make them available for real-time validation, on the [Glean Debug View](./debug-ping-view.md). The value must match the pattern `[a-zA-Z0-9-]{1,20}`. **Important**: in older versions of the Glean SDK, this was named `tagPings` |
+| `sourceTags` | string array (`--esa`)  | Tags outgoing pings with a maximum of 5, comma-separated, tags. The tags must match the pattern `[a-zA-Z0-9-]{1,20}`. The `automation` tag is meant for tagging pings generated on automation: such pings will be special-handled on the pipeline (i.e. discarded from [non-live views](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming)). Tags starting with `glean` are reserved for future use.|
```
mdboom, line 25:
```suggestion
| `sourceTags` | string array (`--esa`)  | Tags outgoing pings with a maximum of 5 comma-separated tags. The tags must match the pattern `[a-zA-Z0-9-]{1,20}`. The `automation` tag is meant for tagging pings generated on automation: such pings will be specially handled on the pipeline (i.e. discarded from [non-live views](https://docs.telemetry.mozilla.org/cookbooks/bigquery/querying.html#table-layout-and-naming)). Tags starting with `glean` are reserved for future use.|
```

glean-core/src/debug.rs
```diff
@@ -33,63 +39,77 @@ pub struct DebugOptions {
...
+            ),
         }
     }
 }
 
 /// A representation of a debug option,
 /// where the value can be set programmatically or come from an environment variable.
 #[derive(Debug)]
-pub struct DebugOption<T, F = fn(T) -> Option<T>> {
+pub struct DebugOption<T, E = fn(String) -> Option<T>, F = fn(T) -> bool> {
```
badboy, line 64: Generics don't _have_ to eb single letter. At some point we could consider making them more expressive.
`F` here definitely doesn't capture anything useful about this.

docs/user/debugging/index.md
```diff
@@ -13,7 +13,7 @@
 There are 3 available commands that you can use with the Glean SDK debug tools
 
 - `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log.
-- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
+- `debugViewTag`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
 - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
```
mdboom, line 17: Should we mention `sourceTags` here as well, to be consistent with the Android-specific instructions above.  (Or does that not make sense because it isn't available in iOS yet?)

docs/user/debugging/ios.md
```diff
@@ -17,7 +17,7 @@ For debugging and validation purposes on iOS, Glean makes use of a custom URL sc
 There are 3 available commands that you can use with the Glean SDK debug tools
 
 - `logPings`: This is either true or false and will cause pings that are submitted to also be echoed to the device's log
-- `tagPings`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
+- `debugViewTag`: This command will tag outgoing pings with the provided value, in order to identify them in the Glean Debug View. Tags need to be string with upper and lower case letters, numbers and dashes, with a max length of 20 characters.
```
mdboom, line 20: Let's add the comment that this used to be called `tagPings` here as well.

glean-core/src/lib.rs
```diff
@@ -732,6 +732,27 @@ impl Glean {
...
 
+    /// Set source tags.
+    ///
+    /// This will return `false` in case `value` contains invalid tags.
+    ///
+    /// Ping tags will show in the destination datasets, after ingestion.
+    ///
+    /// ## Arguments
+    ///
+    /// * `value` - A vector of at most 5 valid HTTP header values. Must match the regex: "[a-zA-Z0-9-]{1,20}".
```
badboy, line 743:
```suggestion
    /// * `value` - A vector of at most 5 valid HTTP header values. Individual tags must match the regex: "[a-zA-Z0-9-]{1,20}".
```

glean-core/src/lib.rs
```diff
@@ -732,6 +732,27 @@ impl Glean {
...
+    pub fn set_source_tags(&mut self, value: Vec<String>) -> bool {
+        self.debug.sourge_tags.set(value)
+    }
+
+    /// Return the value for the source tags or `None` if it hasn't been set.
+    ///
+    /// The source_tags may be set from an environment variable (GLEAN_SOURCE_TAGS)
+    /// or through the `set_source_tags` function.
+    pub(crate) fn source_tags(&self) -> Option<&Vec<String>> {
+        self.debug.sourge_tags.get()
```
badboy, line 753: `&Vec<...>` is a weird type to see. Usually one would expect a `&[...]`

Since 1.40.0 there's [`as_deref`](https://doc.rust-lang.org/std/option/enum.Option.html#method.as_deref) on `Option`, so I think this should do the trick:
```suggestion
    pub(crate) fn source_tags(&self) -> Option<&[String]> {
        self.debug.sourge_tags.get().as_deref()
```

badboy: Your code is not wrong though, so I leave it to you to decide if that additional call is worth it.

glean-core/src/debug.rs
```diff
@@ -112,10 +132,10 @@ where
     ///
     /// Validates the value in case a validation function is available.
     pub fn set(&mut self, value: T) -> bool {
-        let validated = self.validate(value);
-        if validated.is_some() {
+        let validated = self.validate(value.clone());
```
mdboom, line 135: Does passing by reference remove the need for the `clone`?  Since `validate` doesn't hold on to the value, it should at least theoretically be possible to avoid the memory copy.

glean-core/src/debug.rs
```diff
@@ -128,17 +148,41 @@ where
...
+    if trimmed.is_empty() {
+        return None;
+    }
+
+    Some(
+        trimmed
+            .split(',')
+            .map(|s| s.trim().to_string())
+            // Filter out tags starting with "glean". They are reserved.
+            .filter(|s| !s.starts_with("glean"))
```
mdboom, line 170: *Maybe* silently removing these will bite us later.  Maybe we should at least log when this happens, or better yet return an error. Since this is part of debugging API, I don't know if the "never throw an exception out to the user" rule should apply.

glean-core/src/debug.rs
```diff
@@ -147,23 +191,49 @@ pub fn validate_debug_view_tag(value: String) -> Option<String> {
...
+
+    if tags.len() > GLEAN_MAX_SOURCE_TAGS {
+        log::error!(
+            "A list of tags cannot contain more than {} elements.",
+            GLEAN_MAX_SOURCE_TAGS
+        );
+        return false;
+    }
+
+    for tag in tags {
```
mdboom, line 228: Unlike `tokenize_string`, I don't think this rejects tags starting with `glean`.  We should be consistent (and also log / return error).

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
```diff
@@ -37,7 +37,12 @@ class GleanDebugActivity : Activity() {
          * Tags all outgoing pings as debug pings to make them available for real-time validation.
          * The value must match the pattern `[a-zA-Z0-9-]{1,20}`.
          */
-        const val TAG_DEBUG_VIEW_EXTRA_KEY = "tagPings"
+        const val TAG_DEBUG_VIEW_EXTRA_KEY = "debugViewTag"
```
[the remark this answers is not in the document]
Dexterp37: Mh, yes. Let me just leave an alias to the command in the code. I'd still remove references from the docs, though.

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
```diff
@@ -93,8 +98,12 @@ class GleanDebugActivity : Activity() {
                 Glean.setLogPings(logPings)
             }
 
-            intent.getStringExtra(SEND_PING_EXTRA_KEY)?.let {
-                Glean.submitPingByName(it)
+            intent.getStringExtra(SEND_PING_EXTRA_KEY)?.let { name ->
+                Glean.submitPingByName(name)
+            }
+
+            intent.getStringArrayExtra(SOURCE_TAGS_KEY)?.let { tags ->
```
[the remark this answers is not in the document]
Dexterp37: Good catch.

docs/user/debugging/android.md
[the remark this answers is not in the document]
Dexterp37: > 
> 
> They don't get discarded automatically, right?
> We discard only based on some known tags?

@badboy yes, they will be discarded automatically from the stable views, if they have the `automation` tag. Any ping with `automation` will only be available in live-views for ~30 days. This logic will be implemented right after this is merged.

Moreover, this and the `glean` things are the only one with special behaviours. Other tags are only useful for data consumers.

glean-core/src/debug.rs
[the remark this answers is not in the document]
Dexterp37: It would then complain at line 99:

>  ^^^^^ expected type parameter `T`, found `&T`

glean-core/src/debug.rs
Dexterp37: WOHA. This is great :-)

docs/user/debugging/android.md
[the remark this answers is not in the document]
badboy: Ah! I should have read that sentence more closely, it mentions `"automation"` explicitly

glean-core/src/debug.rs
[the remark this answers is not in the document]
badboy: Did you pass in `&value`?

glean-core/src/debug.rs
Dexterp37: @badboy any suggestion or something I should tackle here? Do you want me to file a bug?

badboy: What about using `V` instead? Or `VALDN` / `Valdn` or `Validation`?

glean-core/src/lib.rs
Dexterp37: Oh, TIL! Unfortunately this also means I should change `DebugOption`, otherwise the compiler is sad with

>  expected slice, found struct `std::vec::Vec`

glean-core/src/debug.rs
Dexterp37: Makes sense. I made tokenize now be only responsible of the extraction. The error will be printed in the validation phase below.

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
```diff
@@ -68,8 +76,10 @@ class GleanDebugActivity : Activity() {
         }
 
         // Make sure that at least one of the supported commands was used.
-        val supportedCommands =
-            listOf(SEND_PING_EXTRA_KEY, LOG_PINGS_EXTRA_KEY, TAG_DEBUG_VIEW_EXTRA_KEY)
+        val supportedCommands = listOf(
+                SEND_PING_EXTRA_KEY, LOG_PINGS_EXTRA_KEY,
+            TAG_DEBUG_VIEW_EXTRA_KEY, SOURCE_TAGS_KEY, LEGACY_TAG_PINGS
```
badboy, line 81: Formatting.
Shouldn't these be indented the same?

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
```diff
@@ -86,15 +96,28 @@ class GleanDebugActivity : Activity() {
             // Set the debug view tag, if the tag is invalid it won't be set
             debugViewTag?.let {
                 Glean.setDebugViewTag(debugViewTag)
+            } ?: run {
+                // If the 'debugViewTag' was not used, try to look for the legacy
+                // way of setting debug view tags. We should leave this block of
+                // code around at most until December 2020.
```
badboy, line 102: Bug or calendar reminder?

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -100,10 +100,10 @@ class GleanDebugUtility {
             }
 
             switch param.name {
-            case "tagPings":
+            case "debugViewTag":
```
badboy, line 103: I assume we don't need to keep backward-compatibility here because ... well, there's just not enough people out there that knew the old way?

glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -22,7 +22,7 @@ class GleanDebugUtility {
     /// There are 3 available commands that you can use with the Glean SDK debug tools
     ///
     /// - `logPings`: If "true", will cause pings that are submitted successfully to also be echoed to the device's log
-    /// - `tagPings`:  This command expects a string to tag the pings with and redirects them to the Glean Debug View
+    /// - `debugViewTag`:  This command expects a string to tag the pings with and redirects them to the Debug View
     /// - `sendPing`: This command expects a string name of a ping to force immediate collection and submission of.
```
badboy, line 26: Is there a plan to support the source tags on iOS, too?

badboy: (I'm fine with not doing it right now)

glean-core/src/debug.rs
```diff
@@ -18,78 +18,107 @@
...
+            true
         }
     }
 
     fn set_from_env(&mut self) {
+        let extract = &self.extraction;
         match env::var(&self.env) {
-            Ok(env_value) => match T::from_str(&env_value) {
-                Ok(v) => {
+            Ok(env_value) => match extract(env_value.clone()) {
```
badboy, line 117: That `clone` seems unnecessary (we do own `env_value` here), but it's also not that big of a deal. This happens once and only for debugging.

glean-core/src/debug.rs
```diff
@@ -128,17 +157,30 @@ where
...
+
+/// A tag is the value used in both the `X-Debug-ID` and `X-Source-Tags` headers
+/// of tagged ping requests, thus is it must be a valid header value.
 ///
 /// In other words, it must match the regex: "[a-zA-Z0-9-]{1,20}"
 ///
 /// The regex crate isn't used here because it adds to the binary size,
 /// and the Glean SDK doesn't use regular expressions anywhere else.
-pub fn validate_debug_view_tag(value: String) -> Option<String> {
+fn validate_tag(value: &String) -> bool {
```
badboy, line 180: To not make Clippy unhappy here (we need to use `&String` if we don't want to do more generic shenanigans)

```suggestion
#[allow(clippy::ptr_arg)]
fn validate_tag(value: &String) -> bool {
```

glean-core/src/debug.rs
```diff
@@ -147,23 +189,49 @@ pub fn validate_debug_view_tag(value: String) -> Option<String> {
...
+            return false;
         }
     }
 }
 
+/// Validate the list of source tags.
+///
+/// This builds upon the existing `validate_tag` function, since all the
+/// tags should respect the same rules to make the pipeline happy.
+fn validate_source_tags(tags: &Vec<String>) -> bool {
```
badboy, line 213:
```suggestion
#[allow(clippy::ptr_arg)]
fn validate_source_tags(tags: &Vec<String>) -> bool {
```

glean-core/src/debug.rs
```diff
@@ -172,23 +240,20 @@ mod test {
...
 
     #[test]
     fn debug_option_is_correctly_validated_when_necessary() {
-        fn validate(value: String) -> Option<String> {
-            if value == "test" {
-                Some(value)
-            } else {
-                None
-            }
+        fn validate(value: &String) -> bool {
```
badboy, line 249:
```suggestion
		#[allow(clippy::ptr_arg)]
        fn validate(value: &String) -> bool {
```

glean-core/ios/Glean/Debug/GleanDebugTools.swift
Dexterp37: Yup, [filed this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1654197) for the implementation work there. I might need somebody else's help :)

glean-core/src/debug.rs
Dexterp37: You're right :) Moreover, if we change things to use a `&` here, the `Some` trick instead of the `identity` function won't work anymore. I'd rather leave it as it is and eventually change it if it becomes a problem.

glean-core/android/src/main/java/mozilla/telemetry/glean/debug/GleanDebugActivity.kt
Dexterp37: Both! https://bugzilla.mozilla.org/show_bug.cgi?id=1654207

glean-core/ios/Glean/Debug/GleanDebugTools.swift
travis79: And here I spend all this time teaching the iOS folks about this... :D  

I'll do a PSA about this to the iOS folks that know about it just to be friendly.

Dexterp37: Wait :) There will be a PSA about this aimed at all

PR #1079, opened 2020-07-20
Bug 1650781: Use overflow rather than invalid for long strings

mdboom:


mdboom:
Thanks. I'll fix the iOS failure before merging -- without access to a macOS machine right now, I'm leaning on CI to check my work there ;)

PR #1084, opened 2020-07-21
Use one more e in "queueing" to make it consistent with Kotlin

badboy:
This unifies the naming across Kotlin and Swift.
I picked `queueing` because there were more occurences already.


badboy:
Let's wait for #1046, then the rebase is easier.

PR #1088, opened 2020-07-21
Bug 1653196: Always record *some* value for app_display_version

mdboom:
We can't *make* people give us a version, but we can at least make sure the ping is valid if they don't...


glean-core/python/glean/glean.py
```diff
@@ -102,7 +102,10 @@ def initialize(
         Args:
             application_id (str): The application id to use when sending pings.
             application_version (str): The version of the application sending
-                Glean data.
+                Glean data. The meaning of this field is application-specific,
+                but it is highly recommened to set this to something meaningful
```
brizental, line 106:
```suggestion
                but it is highly recommended to set this to something meaningful
```

brizental:
I believe a rebase will fix the CI errors.

PR #1092, opened 2020-07-23
Bug 1652506 - Integrate 'glean_parser' in the C# build system

Dexterp37:
This PR does the following things:

- Add a 'GleanParser.cs` custom build task ([docs from MS](https://docs.microsoft.com/en-gb/visualstudio/msbuild/task-writing?view=vs-2015&redirectedfrom=MSDN));
- Exposes the task using "props" files ([docs from MS](https://docs.microsoft.com/en-us/visualstudio/msbuild/customize-your-build?view=vs-2019)) so that the task can be used from the DLL file in external projects.
- Use the task directly (by instantiating the task loaded from the source file itself) in the C# bindings and the sample app.


glean-core/csharp/Glean/GleanParser.cs
```diff
@@ -0,0 +1,208 @@
...
+    subprocess.check_call([
+        sys.executable,
+        '-m',
+        module_name
+    ] + sys.argv[3:])
+except:
+    # We don't need to show a traceback in this helper script.
+    # Only the output of the subprocess is interesting.
+    sys.exit(1)
+";
```
mdboom, line 58: I'm glad we're able to do the same auto-update thing here. That will save some headaches for our users.

glean-core/csharp/Glean/GleanParser.cs
```diff
@@ -0,0 +1,208 @@
...
+        /// </summary>
+        /// <returns>A string representing the path to the tool</returns>
+        protected override string GenerateFullPathToTool()
+        { 
+            return ToolExe;
+        }
+
+        private string GetVirtualEnvironmentPath(string venvDir)
+        {
+            // TODO: "Scripts" is only there on windows. On Linux/Mac is "bin"
```
travis79, line 110: Is there a bug to implement this?  I know we don't have any consumers but it might be good to track this TODO.  In the case of OS detection, I wonder if [something like this](https://docs.microsoft.com/en-us/dotnet/api/system.runtime.interopservices.runtimeinformation.isosplatform?view=netframework-4.7.2) would help.

mdboom: I bet we'll want to support Linux at least to access cheaper CI...

Dexterp37: Oh no bug, that's what I meant in the PR comment with "TODO:  Fix using the task on Windows vs Linux". I took care of it this morning.

glean-core/csharp/Glean/GleanParser.cs
Dexterp37: Yes! I'm a bit unhappy about the duplication, but I couldn't find any easy way to work around this.

Dexterp37:
> 
> 
> No real complaints from me, providing you get the failed build/tests sorted.

Yeah, I'm afraid this is now broken because of two reasons:

- #1098 was merged, and it uses the `MemoryDistributionMetricType` which is not yet available in C#.
- `GLEANPARSER : error : invalid command 'bdist_wheel' [/root/project/glean-core/csharp/Glean/Glean.csproj]`. This happens also locally in WSL. `pip` is returning an exit error code because some package needs bdist_wheel, but then the package falls back to use non wheels, so everything works. Re-running the build locally works, because the glean_parser version hasn't changed since the venv was created in the last run. I'm not sure why this is not a problem in Gradle to be honest. Unfortunately, we can either disable the "stderr as Errors" completely or not, in [ToolTask](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask?view=msbuild-16-netcore). @mdboom are you aware of any way we could make `pip` not return an error if bdist is the only problem, since packages can work anyway?

mdboom:
>     * `GLEANPARSER : error : invalid command 'bdist_wheel' [/root/project/glean-core/csharp/Glean/Glean.csproj]`. This happens also locally in WSL. `pip` is returning an exit error code because some package needs bdist_wheel, but then the package falls back to use non wheels, so everything works. Re-running the build locally works, because the glean_parser version hasn't changed since the venv was created in the last run. I'm not sure why this is not a problem in Gradle to be honest. Unfortunately, we can either disable the "stderr as Errors" completely or not, in [ToolTask](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask?view=msbuild-16-netcore). @mdboom are you aware of any way we could make `pip` not return an error if bdist is the only problem, since packages can work anyway?

Do you know what package it is that complains about the missing `bdist_wheel`?  Seems like a bug in that package, probably.  We could always install `wheel` before anything else -- it wouldn't be a bad idea as that makes packages faster to install in general.

Dexterp37:
> 
> 
> > ```
> > * `GLEANPARSER : error : invalid command 'bdist_wheel' [/root/project/glean-core/csharp/Glean/Glean.csproj]`. This happens also locally in WSL. `pip` is returning an exit error code because some package needs bdist_wheel, but then the package falls back to use non wheels, so everything works. Re-running the build locally works, because the glean_parser version hasn't changed since the venv was created in the last run. I'm not sure why this is not a problem in Gradle to be honest. Unfortunately, we can either disable the "stderr as Errors" completely or not, in [ToolTask](https://docs.microsoft.com/en-us/dotnet/api/microsoft.build.utilities.tooltask?view=msbuild-16-netcore). @mdboom are you aware of any way we could make `pip` not return an error if bdist is the only problem, since packages can work anyway?
> > ```
> 
> Do you know what package it is that complains about the missing `bdist_wheel`? Seems like a bug in that package, probably. We could always install `wheel` before anything else -- it wouldn't be a bad idea as that makes packages faster to install in general.

@mdboom would we want this for Gradle as well? Where would we install `wheel`?

Looks like the problem is with `PyYAML` and `pyrsistent`.

mdboom:
> @mdboom would we want this for Gradle as well? Where would we install `wheel`?

We could probably add `python3-wheel` to the `apt install` line.  That will help CI.

For local devs, though, we could run `pip install wheel` from the little bootstrap script before `pip install glean_parser`.
 
> Looks like the problem is with `PyYAML` and `pyrsistent`.

Makes sense, as those ones have binary components.

CHANGELOG.md
```diff
@@ -6,10 +6,14 @@
   * Limit ping request body size to 1MB. ([#1098](https://github.com/mozilla/glean/pull/1098))
 * iOS
   * Implement ping tagging (i.e. the `X-Source-Tags` header) through custom URL ([#1100](https://github.com/mozilla/glean/pull/1100)).
+  * Update `glean_parser` to `v1.28.3`
+    * BUGFIX: Generate valid C# code when using Labeled metric types.
+    * BUGFIX: Support `HashSet` and `Dictionary` in the C# generated code.
```
mdboom, line 11: Probably a merge issue -- but this is appearing under the iOS section, but relates to C#

Dexterp37: Good catch!

glean-core/csharp/Glean/GleanParser.cs
```diff
@@ -0,0 +1,270 @@
...
+namespace GleanTasks
+{
+    public class GleanParser : ToolTask
+    {
+        // The name of the directory in which the Python virtual
+        // environment must be created.
+        private const string DefaultVirtualEnvDir = ".venv";
+
+        // The glean_parser pypi package version 
+        private const string GleanParserVersion = "1.28.3";
```
travis79, line 21: Might want to consider coordinating with #1116 to ensure this gets updated by the script there.

Dexterp37: Good point! @mdboom how should we manage this? :) I'm happy to have #1116 merge first and then change it a bit, but I'm not super expert with sed tbh.

mdboom: Let's land #1116 first and then update that script in this PR. I'm not much of a sed expert either, I just copied what @badboy did elsewhere.

PR #1094, opened 2020-07-23
Fix title of JWE payload on the Glean Book [doc only]

brizental:


badboy:
> I guess the [doc only] tag isn't working? @badboy?

I think it is, though we still run the full Python stuff because we need it to build Python docs.

travis79:
> > I guess the [doc only] tag isn't working? @badboy?
> 
> I think it is, though we still run the full Python stuff because we need it to build Python docs.

Ahh, okay. Nothing to see here then :)

PR #1097, opened 2020-07-27
Bug 1648436 - Add the c# labeled boolean and string

Dexterp37:
This adds support for labeled booleans and labeled strings in the C# language binding.

The code and tests were converted from Kotlin. The one important thing to watch out for is the type casting in the c# indexer operator (aka subscript): I had to use a little trick to pacify the compiler. See the comments.


glean-core/csharp/Glean/Metrics/LabeledMetricType.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// If the type of  `T`  is currently not supported.
+        /// </exception>
+        public T this[string label]
+        {
+            get {
+                // Note the double `(T)(ILabeledSubmetricInterface)` cast before returning. This is
+                // required in order to make the compiler not complain. Since the `where` clause for
+                // this class cannot list more than one class, we need all our supported subtypes to
+                // implement a common interface and use that interface as the T type constraint. This
+                // allows us to then explicitly cast bach to T, which is otherwise impossible.
```
travis79, line 105:
```suggestion
                // allows us to then explicitly cast back to T, which is otherwise impossible.
```

glean-core/csharp/Glean/Metrics/LabeledMetricType.cs
```diff
@@ -0,0 +1,153 @@
...
+        /// is used, the metric will be recorded in the special `__other__` label.
+        /// </summary>
+        /// <param name="label">The label</param>
+        /// <exception cref="InvalidOperationException">
+        /// If the type of  `T`  is currently not supported.
+        /// </exception>
+        public T this[string label]
+        {
+            get {
+                // Note the double `(T)(ILabeledSubmetricInterface)` cast before returning. This is
```
travis79, line 101: Thanks for leaving such a detailed comment, this technique of the double cast is weird!

PR #1098, opened 2020-07-27
Bug 1597761 - Don't create PingRequest if body exceeds 1MB

brizental:
This required a bit more changes than I expected because of the need to record metrics.

Data review for the metrics recorded here is still ongoing.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,10 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v31.6.0...main)
 
+* General
+  * Limit ping request body size to 1MB. ([#1098]((https://github.com/mozilla/glean/pull/1098)))
+* iOS
+  * Implement ping tagging (i.e. the `X-Source-Tags` header) through custom URL ([#1100](https://github.com/mozilla/glean/pull/1100)).
 * C#
   * Add support for Labeled Strings and Labeled Booleans.
   * Add support for the Counter metric type and Labeled Counter.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v31.6.0...main)
 
 * General
-  * Limit ping request body size to 1MB. ([#1098]((https://github.com/mozilla/glean/pull/1098)))
+  * Limit ping request body size to 1MB. ([#1098](https://github.com/mozilla/glean/pull/1098))
 * iOS
   * Implement ping tagging (i.e. the `X-Source-Tags` header) through custom URL ([#1100](https://github.com/mozilla/glean/pull/1100)).
 * C#
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -374,3 +374,36 @@ glean.upload:
     expires: never
     no_lint:
       - COMMON_PREFIX
+
+  discarded_exceeding_ping_size:
+    type: memory_distribution
+    description: >
+      The size of pings that exceeded the maximum ping size allowed for upload.
+    memory_unit: kilobyte
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761#c4
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+    no_lint:
+      - COMMON_PREFIX
+
+  discarded_exceeding_pings:
+    type: counter
+    description: >
+      Counts the pings that get discarded for exceeding the maximum ping size allowed for upload.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761#c4
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+    no_lint:
+      - COMMON_PREFIX
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -375,28 +375,11 @@ glean.upload:
     no_lint:
       - COMMON_PREFIX
 
-  discarded_exceeding_ping_size:
+  discarded_exceeding_pings_size:
     type: memory_distribution
     description: >
       The size of pings that exceeded the maximum ping size allowed for upload.
     memory_unit: kilobyte
-    bugs:
-      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761
-    data_reviews:
-      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761#c4
-    data_sensitivity:
-      - technical
-    notification_emails:
-      - glean-team@mozilla.com
-    expires: never
-    no_lint:
-      - COMMON_PREFIX
-
-  discarded_exceeding_pings:
-    type: counter
-    description: >
-      Counts the pings that get discarded for
-      exceeding the maximum ping size allowed for upload.
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761
     data_reviews:
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -395,7 +395,8 @@ glean.upload:
   discarded_exceeding_pings:
     type: counter
     description: >
-      Counts the pings that get discarded for exceeding the maximum ping size allowed for upload.
+      Counts the pings that get discarded for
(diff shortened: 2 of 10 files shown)


glean-core/metrics.yaml
```diff
@@ -374,3 +374,37 @@ glean.upload:
...
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1597761#c4
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+    no_lint:
+      - COMMON_PREFIX
+
+  discarded_exceeding_pings:
```
badboy, line 395: Isn't this something we could get from the number of samples in the distribution above?
It might be easier to query just the counter, but is that really helpful for us?
I'm sure chutten might have a good idea about that.

glean-core/src/upload/request.rs
```diff
@@ -155,19 +158,25 @@ impl Builder {
...
     /// `document_id`, `path` and `body`.
-    pub fn build(self) -> PingRequest {
-        PingRequest {
+    pub fn build(self) -> Result<PingRequest> {
+        let body = self
+            .body
+            .expect("body must be set before attempting to build PingRequest");
+
+        if body.len() > self.body_max_size {
+            return Err(ErrorKind::PingBodyOverflow(body.len() as u64).into());
```
badboy, line 167: What's the reason we don't keep this as a `usize`?

glean-core/src/upload/request.rs
```diff
@@ -60,11 +61,12 @@ pub struct Builder {
     path: Option<String>,
     body: Option<Vec<u8>>,
     headers: HeaderMap,
+    body_max_size: usize,
 }
 
 impl Builder {
     /// Creates a new builder for a PingRequest.
-    pub fn new(language_binding_name: &str) -> Self {
+    pub fn new(language_binding_name: &str, body_max_size: usize) -> Self {
```
badboy, line 69: Let's document the arguments, especially that `body_max_size` is in bytes.

glean-core/src/upload/request.rs
```diff
@@ -194,8 +203,9 @@ impl PingRequest {
     ///
     /// * `language_binding_name` - The name of the language used by the binding that instantiated this Glean instance.
     ///                             This is used to build the User-Agent header value.
-    pub fn builder(language_binding_name: &str) -> Builder {
-        Builder::new(language_binding_name)
+    /// * `body_max_size` - The maximum size the compressed ping body may have to be eligible for upload.
```
badboy, line 206: Ah, the parameter is documented here (because that's how we use the builder).
Still should say `in bytes`

glean-core/src/upload/request.rs
```diff
@@ -262,4 +273,15 @@ mod test {
         assert!(request.headers.contains_key("X-Client-Version"));
         assert!(request.headers.contains_key("Content-Length"));
     }
+
+    #[test]
+    fn errors_when_request_body_exceeds_max_size() {
+        let request = Builder::new(/* language_binding_name */ "Rust", 0)
```
badboy, line 279: `0` here doesn't say much. Often `0` is also a special value to disable stuff.

Let's document that this sets an arbitrarily small value (and maybe use 1? Just a thought, don't care much about the exact value)

glean-core/src/upload/mod.rs
```diff
@@ -16,14 +16,19 @@ use std::sync::{Arc, RwLock, RwLockWriteGuard};
...
+use crate::{internal_metrics::UploadMetrics, Glean};
+use directory::{PingDirectoryManager, PingPayload};
 pub use request::{HeaderMap, PingRequest};
 pub use result::{ffi_upload_result, UploadResult};
 
 mod directory;
 mod request;
 mod result;
 
+// The maximum size a ping body may have to be eligible for upload.
```
badboy, line 29:
```suggestion
// The maximum size in bytes a ping body may have to be eligible for upload.
```

glean-core/src/upload/mod.rs
```diff
@@ -16,14 +16,19 @@ use std::sync::{Arc, RwLock, RwLockWriteGuard};
...
+const PING_BODY_MAX_SIZE: usize = 1024 * 1024;
```
badboy, line 30:
```suggestion
const PING_BODY_MAX_SIZE: usize = 1024 * 1024; // 1 MB
```

glean-core/src/upload/mod.rs
badboy: Let's be explicit about units.

glean-core/src/upload/mod.rs
```diff
@@ -207,25 +207,55 @@ impl PingUploadManager {
...
     ///
-    /// * `interval` - the amount of seconds in each rate limiting window.
-    /// * `max_tasks` - the maximum amount of task requests allowed per interval.
-    pub fn set_rate_limiter(&mut self, interval: u64, max_tasks: u32) {
-        self.rate_limiter = Some(RwLock::new(RateLimiter::new(
-            Duration::from_secs(interval),
-            max_tasks,
-        )));
+    /// Returns the `PingRequest` or `None` if unable to build,
+    /// in which case also deletes the ping file and records an error.
```
badboy, line 213:
```suggestion
    /// in which case it will delete the ping file and records an error.
```

glean-core/src/upload/mod.rs
```diff
@@ -288,19 +331,28 @@ impl PingUploadManager {
...
                 "Tried getting an upload task, but processing is ongoing. Will come back later."
             );
             return PingUploadTask::Wait;
         }
 
+        let mut pending_pings = self
+            .pending_pings
+            .write()
+            .expect("Can't write to pending pings cache.");
+        while let Some((document_id, path, body, headers)) = pending_pings.pop() {
```
badboy, line 352: `pop` takes the last element, wouldn't this result in the exact reversed order we'd like them in? Probably not that big of an impact.

If we want to iterate in order and drain the vector at the same time and ALSO don't want to waste processing time/allocations on shifting stuff around we can use the [`drain`](https://doc.rust-lang.org/std/vec/struct.Vec.html#method.drain) iterator:

```rust
for (document_id, path, body, headers) in pending_pings.drain(..) {
    // ...
}

Comment (brizental) [reply glean-core/src/upload/mod.rs:352]: Oh you are right! Oops.

File: glean-core/metrics.yaml (hunk 1 above)
Comment (brizental) [reply glean-core/metrics.yaml:395]: You are right. I guess I just copied the Desktop fuctionality without giving it so much thought. Still, I would like to get @chutten 's input in case there is a good reason to keep this counter.

File: glean-core/src/upload/request.rs (hunk 2 above)
Comment (brizental) [reply glean-core/src/upload/request.rs:167]: Hm, no reason. I'll change it.

**Comment by brizental** [comment]
> Do we already have something (e.g. for the rate limiting)?

For the rate limiting we have a note on the internal documentation: https://mozilla.github.io/glean/book/dev/core/internal/upload.html?highlight=rate,limiting#1 and another note on the "Pings" page: https://mozilla.github.io/glean/book/user/pings/index.html?highlight=rate#ping-submission.

We could maybe create a "Limitations" section on the "Pings" page to list both in a more structure manner.

File: glean-core/metrics.yaml (hunk 1 above)
Comment (chutten) [reply glean-core/metrics.yaml:395]: No need to have a separate counter when we can reduce that information straightforwardly in SQL (there might even be a UDF for that someplace. Oh wait, maybe there's not because I have this in my toolbox file
```sql
(SELECT
        SUM(values.value)
    FROM UNNEST(metrics.timing_distribution.gfx_checkerboard_potential_duration.values) AS values)
```
)

At any rate, no need to clutter glean's metrics.yaml and the instrumentation code with a counter in my opinion.

Dexterp37:
> 
> 
> > Do we already have something (e.g. for the rate limiting)?
> 
> For the rate limiting we have a note on the internal documentation: https://mozilla.github.io/glean/book/dev/core/internal/upload.html?highlight=rate,limiting#1 and another note on the "Pings" page: https://mozilla.github.io/glean/book/user/pings/index.html?highlight=rate#ping-submission.
> 
> We could maybe create a "Limitations" section on the "Pings" page, under the "Ping submission" section to list both in a more structured manner.

@badboy @brizental I think the above is a good idea!

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v31.6.0...main)
 
+* General
+  * Limit ping request body size to 1MB. ([#1098](https://github.com/mozilla/glean/pull/1098))
+* iOS
+  * Implement ping tagging (i.e. the `X-Source-Tags` header) through custom URL ([#1100](https://github.com/mozilla/glean/pull/1100)).
```
badboy, line 8: :D was that missed in that PR? Good to still get it in.

brizental: Yeap, I am trying to remind myself everytime, but still miss it sometimes.

PR #1100, opened 2020-07-27
Implement ping tagging for iOS

brizental:
Tested locally and it works.

Not sure if I should have added tests anywhere for this changes. I followed the `debugViewTag` implementation which also doesn't have explicit tests.

### DIFF
--- a/glean-core/ios/Glean/Debug/GleanDebugTools.swift
+++ b/glean-core/ios/Glean/Debug/GleanDebugTools.swift
@@ -19,7 +19,7 @@ class GleanDebugUtility {
     /// - parameters:
     ///     * url: A `URL` object containing the Glean debug commands as query parameters
     ///
-    /// There are 3 available commands that you can use with the Glean SDK debug tools
+    /// There are 4 available commands that you can use with the Glean SDK debug tools
     ///
     /// - `logPings`: If "true", will cause pings that are submitted successfully to also be echoed to the device's log.
     /// - `debugViewTag`:  This command expects a string to tag the pings with and redirects them to the Debug View.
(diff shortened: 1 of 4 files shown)


glean-core/ios/Glean/Debug/GleanDebugTools.swift
```diff
@@ -21,8 +21,12 @@ class GleanDebugUtility {
     ///
     /// There are 3 available commands that you can use with the Glean SDK debug tools
```
badboy, line 22:
```suggestion
    /// There are 4 available commands that you can use with the Glean SDK debug tools
```

glean-core/ios/Glean/Glean.swift
```diff
@@ -121,6 +122,10 @@ public class Glean {
                 _ = self.setDebugViewTag(debugViewTag)
             }
 
+            if let sourceTags = self.sourceTags {
```
badboy, line 125: On Android we set logPings befors sourceTags in init.
While I don't think this has semantic meaning, it would be good to keep them in the same order, so reasoning about both code bases is easier.

glean-core/ios/GleanTests/GleanTests.swift
```diff
@@ -328,6 +329,32 @@ class GleanTests: XCTestCase {
...
+        // Set the last time the "metrics" ping was sent to now. This is required for us to not
+        // send a metrics pings the first time we initialize Glean and to keep it from interfering
+        // with these tests.
+        let now = Date()
+        Glean.shared.metricsPingScheduler.updateSentDate(now)
+        // Restart glean
+        Glean.shared.resetGlean(clearStores: false)
+
+        let host = URL(string: Configuration.Constants.defaultTelemetryEndpoint)!.host!
+        stub(condition: isHost(host)) { data in
```
badboy, line 347: hmpf, eventually we should extend `stubServerReceive` to have the headers.
Fine for now here though.

PR #1102, opened 2020-07-27
Bug 1634410: Python: Don't use a temporary directory by default

mdboom:


mdboom:
Cc: @wlach, @zalun, @globau for your awareness

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * C#
   * Add support for Labeled Strings and Labeled Booleans.
+* Python
+  * Breaking change: `data_dir` must always be passed to `Glean.initialize`. Prior to this, a missing value would store Glean data in a temporary directory.
```
badboy, line 8: Previously we used `**Breaking change**` just to make it stand out a bit more when rendered.

```suggestion
  * **Breaking change**: `data_dir` must always be passed to `Glean.initialize`. Prior to this, a missing value would store Glean data in a temporary directory.
```

glean-core/python/glean/glean.py
```diff
@@ -101,7 +101,7 @@ def initialize(
         application_version: str,
         upload_enabled: bool,
         configuration: Optional[Configuration] = None,
-        data_dir: Optional[Path] = None,
+        data_dir: Path = None,
```
badboy, line 104: `Path` can be not-optioanl and still be `None`? And the type checker is okay with that? That highly confuses me :D

glean-core/python/glean/glean.py
```diff
@@ -138,10 +137,7 @@ def initialize(
                 configuration = Configuration()
 
             if data_dir is None:
-                data_dir = Path(tempfile.TemporaryDirectory().name)
-                cls._destroy_data_dir = True
-            else:
-                cls._destroy_data_dir = False
+                raise TypeError("data_dir must be provided")
```
badboy, line 140: If only we had a type system to ensure this :P

glean-core/python/glean/glean.py
```diff
@@ -213,6 +209,37 @@ def initialize():
...
+        Initialize Glean to use a temporary data directory. Use for internal
+        unit testing only.
+
+        The temporary directory will be destroyed when Glean is initialized
+        again or at process shutdown.
+        """
+        if data_dir is None:
+            actual_data_dir = Path(tempfile.TemporaryDirectory().name)
+        else:
+            actual_data_dir = data_dir
```
badboy, line 232: Is there any reason to keep this behavior and not just always use a temp directory when this function is used?

glean-core/python/glean/glean.py
zalun: Please also add the `data_dir` to the docstring so it will appear in the docs.

mdboom: We have to give it *some* value so it remains a keyword argument, since keyword arguments have to come after non-keyword arguments (otherwise we have to re-order the arguments which would be an even larger breaking change).  The type checker doesn't check this because it is a common pattern to use `None` as meaning "not provided" to keyword arguments.  The type checker *will* catch passing None in by a user of this function (if our users run a type checker).

@zalun: `data_dir` is already documented in the docstring -- or maybe I'm missing something?

glean-core/python/glean/glean.py
mdboom: Yep -- for example: https://github.com/mozilla/glean/blob/main/glean-core/python/tests/test_glean.py#L156

glean-core/python/glean/glean.py
mdboom: Yeah -- Python has ended up in a truly strange place with respect to type-checking, but here we are.

glean-core/python/glean/glean.py
badboy: The example on top is missing the data_dir parameter:

```
    >>> Glean.initialize(
    ...     application_id="my-app",
    ...     application_version="0.0.0",
    ...     upload_enabled=True
    ... )
    """
```

glean-core/python/glean/glean.py
badboy: But ... that's calling `Glean.initialize` already, which handles the data_dir.
The test-only function here (`_initialize_with_tempdir_for_testing`) shouldn't need it?

glean-core/python/glean/glean.py
mdboom: Good catch, thanks.

glean-core/python/glean/glean.py
mdboom: Ah, I see your point, now.  Makes sense.  I'll simplify this (and update the tests).

PR #1107, opened 2020-07-28
Make Glean-iOS build as a static lib

travis79:


travis79:
Since we are [making progress towards an iOS megazord with A-S](https://github.com/mozilla/application-services/pull/3554), I'm closing this in favor of that approach.

PR #1110, opened 2020-07-30
Bug 1601550 - Enforce disk quota on pending pings directory

brizental:
I am opening this as a draft because I would like to get some opinions on the approach.

Please refer to comments I'll leave in the code to see what I mean.


glean-core/src/upload/mod.rs
```diff
@@ -276,6 +279,55 @@ impl PingUploadManager {
         }
     }
 
+    /// Enqueue pings that might have been cached.
```
brizental, line 282: The bulk of the logic that enforces the quota is in this function. All the rest is to support that we can do what we do here.

I have some questions regarding this approach, though. I would like feedback before proceeding and adding the metric recording to this.

1. I wonder if it is the best idea to only check for the pending pings size when scanning the directory, or if I should check everytime a ping is stored. I remember asking this before, but I don't remember the answer. Looking at this implementation it really makes me question how effective this limitation is if we are not checking everytime a ping is stored. Doing that would require many changes, specially since a new `PingMaker` is created everytime we want to store a new ping and it holds no state. We would basically have to scan the pending pings folder everytime we store a new ping, which is suboptimal to say the least, or instantiate the PingMaker only once, accumulate the size of every stored ping and decrement the size of every deleted ping. Writing this down, I see that it is probably not that best way. I am sure there are other solutions if we decide to go this route.

2. I decided to not even check the size of the deletion request folder. I explain why in this docstring. I wonder if it would be best to sum the size of the pending_ping with the deletion_request directory instead and then enforce that the sum of both directories size doesn't exceed the quota.

Dexterp37:
I redirected this to @badboy as I won't be able to get to it today and won't be around tomorrow.

glean-core/src/upload/mod.rs
brizental: One more thing I just realized: this approch will not take into account the size of pings submitted after initialization of Glean (specifically the initialization of the upload manager inside Glean) when calculating the size of the folder.

badboy: To 1:
I had to check the code again, but Firefox Telemetry on Desktop enforces that quota only on start (where it runs asynchronously and gets canceled on shutdown if not finished)

I think that's still a reasonable choice.
We should _assume_ that we 1) generate only a reasonable amount of pings while running, 2) most of the time are able to sent them out, 3) we already check that pings are <1mb (correct me if I'm wrong, we do this on storage, not on upload, right?)
And as you figured out doing it differently is much more complex.

Desktop tries to get the size to 90% of the quota if it was larger. It does so by deleting the oldest pings first.
Is that what you intend to do here as well?

To 2:
Sounds reasonable. There shouldn't even be multiple deletion-request pings around
(the only way this can happen if some user's device is offline and they keep toggling data upload, right?)

brizental: Yes, I mostly tried to copy over what desktop does. It just looks very different because implementation in general here is very different.

> Desktop tries to get the size to 90% of the quota if it was larger. It does so by deleting the oldest pings first.
Is that what you intend to do here as well?

Yes, since the cached pings list is ordered by date, starting with the newest and ending with the oldest we will delete older pings instead of newer. I do not check if we are at 90% though, I guess that was an oversight when copying the logic over. Do you think that it makes sense to do that for glean too?

brizental: One more point I forgot to raise, for telemetry we have this 15MB quota, but that is for desktop and we havea different quota for mobile. Should we also do that for glean? (e.g. by exposing the quota as a config option for Glean).

That would probably be a follow up. In any case, do you think a quota of 15MB as a starting point is good enough? Should we file a follow up to make the quota configurable?

glean-core/src/upload/directory.rs
```diff
@@ -76,41 +95,49 @@ impl PingDirectoryManager {
...
-            ],
+            pending_pings_dir: data_path.join(PENDING_PINGS_DIRECTORY),
+            deletion_request_pings_dir: data_path.join(DELETION_REQUEST_PINGS_DIRECTORY),
         }
     }
 
     /// Attempts to delete a ping file.
     ///
-    /// ## Arguments
+    /// # Arguments
```
badboy, line 105: We're _nearly_ consistent in using a level-2 header (`##`), which seems to be the exact opposite of what the Rust libstd does.
It's also noted in the [API guidelines](https://rust-lang.github.io/api-guidelines/documentation.html) (an old thing that's not fully maintained anymore I think) and this RFC uses level-1 header: [RFC 1574](https://github.com/rust-lang/rfcs/blob/master/text/1574-more-api-documentation-conventions.md).

Where am I going with this? I initially intended to argue for keeping level-2, but I guess it's better if I file a bug and we streamline this to level-1 eventually.

Please move on, nothing to see here.

glean-core/src/upload/directory.rs
```diff
@@ -76,41 +95,49 @@ impl PingDirectoryManager {
...
     ///
     /// * `uuid` - The UUID of the ping file to be deleted
     ///
-    /// ## Panics
+    /// # Return value
+    ///
+    /// `bool` - `true` if file was deleted succesfully and `false` otherwise.
```
badboy, line 111: No need to mention the type again, that's in the function declaration.
```suggestion
    /// Returns `true` if the file was deleted succesfully and `false` otherwise.
```

glean-core/src/upload/directory.rs
```diff
@@ -151,29 +178,40 @@ impl PingDirectoryManager {
...
     ///
-    /// Files that are not correctly formatted will also be deleted.
-    ///
     /// # Return value
     ///
-    /// `Vec<(String, String, JsonValue, HeaderMap)>` -
-    ///     a vector of tuples containing the document_id, path, body and headers of each request.
-    pub fn process_dir(&self) -> Vec<PingPayload> {
+    /// `Vec<(fs::Metadata, PingPayload)>` -
+    ///     a vector of tuples with the file metadata and payload of each ping file in the directory.PathBuf
```
badboy, line 199: Heh, see, just means we always need to adjust the type in a comment. 
Let's just remove it here.
```suggestion
    ///     Returns a vector of tuples with the file metadata and payload of each ping file in the directory.
```

glean-core/src/upload/directory.rs
```diff
@@ -151,29 +178,40 @@ impl PingDirectoryManager {
...
+    fn process_dir(&self, dir: &PathBuf) -> Vec<(fs::Metadata, PingPayload)> {
```
badboy, line 200: PathBuf and Path are akin to String and str, therewore we should probably take `&Path` (the same as you wouldn't pass around (`&String`) most of the time.

glean-core/src/upload/directory.rs
```diff
@@ -151,29 +178,40 @@ impl PingDirectoryManager {
...
-            .into_iter()
+        let entries: fs::ReadDir;
+        match dir.read_dir() {
+            Ok(r) => entries = r,
+            Err(_) => {
+                // This may error simply because the directory doesn't exist,
+                // which is expected if no pings were stored yet.
+                return Vec::new();
+            }
+        }
```
badboy, line 211: `match` is an expression and returns a value, we can therefore do this:

```suggestion
        let entries = match dir.read_dir() {
            Ok(entries) => entries,
            Err(_) => {
                // This may error simply because the directory doesn't exist,
                // which is expected if no pings were stored yet.
                return Vec::new();
            }
        };
```

glean-core/src/upload/directory.rs
```diff
@@ -183,50 +221,39 @@ impl PingDirectoryManager {
...
                     }
-                    // In case we can't process the file we just ignore it.
                     if let Some(data) = self.process_file(file_name) {
-                        // Get the modified date of the file, which will later be used
-                        // for sorting the resulting vector.
-                        let modified_date = fs::metadata(&path).and_then(|data| data.modified());
-                        return Some((modified_date, data));
+                        // This may panic in case the path does not exist
+                        // or the user does not have permissions to read the file.
+                        // Both are extremely unlike so we can panic! here if that happens.
```
badboy, line 227: Uhm ... not so sure about that, I've seen to many broken things with files :D 
If this ever _does_ panic it will take down the whole app.
But in case of an error we can just ignore it, can't we?

glean-core/src/upload/directory.rs
```diff
@@ -19,6 +19,23 @@ use crate::{DELETION_REQUEST_PINGS_DIRECTORY, PENDING_PINGS_DIRECTORY};
 /// this will contain the document_id, path, JSON encoded body of a ping and the persisted headers.
 pub type PingPayload = (String, String, String, Option<HeaderMap>);
 
+/// A struct to hold the result of scanning all pings directories.
+#[derive(Clone, Debug, Default)]
+pub struct PingPayloadsByDirectory {
```
badboy, line 24: Somehow this name irks me, but I also don't have a better name.
Just `PingPayloads` is too confusable with `PingPayload`. Hmmm.
Unless you have another good idea we can keep it though.

glean-core/src/upload/mod.rs
```diff
@@ -29,6 +29,9 @@ mod result;
 // The maximum size in bytes a ping body may have to be eligible for upload.
 const PING_BODY_MAX_SIZE: usize = 1024 * 1024; // 1 MB
 
+// The maximum size in byte the pending pings directory may have on disk.
+const PENDING_PINGS_DIRECTORY_QUOTA: usize = 15 * 1024 * 1024; // 15 MB
```
badboy, line 33: @Dexterp37 On desktop we keep 15 MB in the directory, but set it to 1MB for mobile gecko (Fennec).
15 MB seems like _a lot_ here, should we adjust that?

glean-core/src/upload/mod.rs
```diff
@@ -276,6 +279,64 @@ impl PingUploadManager {
         }
     }
 
+    /// Enqueue pings that might have been cached.
+    ///
+    /// The size of the PENDING_PINGS_DIRECTORY directory will be calculated
+    /// (by accumulating each pings size in that directory)
```
badboy, line 285:
```suggestion
    /// (by accumulating each ping's size in that directory)
```

glean-core/src/upload/mod.rs
```diff
@@ -276,6 +279,64 @@ impl PingUploadManager {
...
+    /// and in case we extrapolate the quota, defined by the `quota` arg,
```
badboy, line 286:
```suggestion
    /// and in case we exceed the quota, defined by the `quota` arg,
```

glean-core/src/upload/mod.rs
```diff
@@ -276,6 +279,64 @@ impl PingUploadManager {
...
+    /// outstanding pings get deleted and are not enqueued.
+    ///
+    /// The size of the DELETION_REQUEST_PINGS_DIRECTORY will not be calculated
+    /// and no deletion-request pings will be deleted. Deletion request pings
+    /// are not very common and usually don't contain any data,
+    /// we don't expect that directory to ever reach quota.
+    /// Most importantly, we don't want to ever delete deletion-request pings.
+    ///
+    /// Arguments
```
badboy, line 295:
```suggestion
    /// # Arguments
```

glean-core/src/upload/mod.rs
```diff
@@ -276,6 +279,64 @@ impl PingUploadManager {
...
+    ///
+    /// * `glean` - The Glean object holding the database.
+    /// * `quota` - The quota, in bytes, for the size of the pending pings directory.
+    fn enqueue_cached_pings(&self, glean: &Glean, quota: usize) {
```
badboy, line 299: If I understand this logic right, we essentialy try to do this whenever someone asks for a ping file to be uploaded, but we only read "cached pings" from the directories at startup.
So we effectively clean up the directory only once on first request.
Correct?

glean-core/src/upload/directory.rs
```diff
@@ -19,6 +19,23 @@ use crate::{DELETION_REQUEST_PINGS_DIRECTORY, PENDING_PINGS_DIRECTORY};
...
+    pub pending_pings: Vec<(fs::Metadata, PingPayload)>,
```
badboy, line 25: Do we actually need the full metadata? I think we only use the size.
`fs::Metadata` [is a wooping 176 bytes](https://play.rust-lang.org/?version=stable&mode=debug&edition=2018&gist=6e574ebe39ec305cd963d67808807b8b) (on Linux at least).

glean-core/src/upload/mod.rs
```diff
@@ -1054,4 +1100,39 @@ mod test {
...
+        let n = 10;
+        for _ in 0..n {
+            glean.submit_ping(&ping_type, None).unwrap();
+        }
+
+        // Create a new upload manager pointing to the same data_path as the glean instance.
+        let upload_manager = PingUploadManager::new(tmpdir.path(), "Rust", true);
+
+        // Enqueue cached pings and set an arbitrarily low quota.
+        upload_manager.enqueue_cached_pings(&glean, 1);
```
badboy, line 1122: An empty ping should have a rather stable size. 
Would it make sense to set the quota to just slightly above that so we see that the last ping is kept and others are deleted?

badboy: And if we keep `deleted_pings_after_quota_hit` we could check that it deleted the right amount.

glean-core/metrics.yaml
```diff
@@ -391,3 +391,41 @@ glean.upload:
...
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1601550#c3
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+    no_lint:
+      - COMMON_PREFIX
+
+  deleted_pings_after_quota_hit:
```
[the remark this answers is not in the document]
brizental: The idea is to know how many pings we have lost from going over quota. Yes, it could be incremented several times, but in the end it will still report how many pings were deleted because of going over quota, so, in my opinion, still good information.

In desktop we do have a histogram for this, I do agree that a distribution could also make sense here, specially in case we wanted to know specifically the average of pings deleted per scanning of the pings directory.

From my initial understanding of the usage for this metric, a counter works just fine. But now I am not so sure anymore. @Dexterp37 what do you think?

glean-core/src/upload/directory.rs
brizental: I'll be happy to take that bug, when you file it. I did not know about this official guidelines, I just get annoyed everytime I realize this stuff is not standardized and try to do just that.

glean-core/src/upload/directory.rs
brizental: We can't really ignore, we would have to not enqueue that file all together, or I guess set a default size for it or something of the sort.

My point in this being very unlikely is that we read the file in the lines right above this one in this same closure. If the path didn't exist / user does not have permissions we wouldn't really get to this line of the function.

For good measure I can change this to just ignore any failures though.

glean-core/src/upload/directory.rs
brizental: Irks me too, but I also couldn't think of a better name.

glean-core/src/upload/mod.rs
brizental: I gave this suggestion on my initial comment for this PR, but one thing we could do is make this into a config option. Each binding could define the quota that makes sense for it.

glean-core/src/upload/mod.rs
brizental: Correct, especially since we drain the `cached_pings` vector the first time we read it and never write to it again.

glean-core/src/upload/directory.rs
brizental: Not really, we only use the size. I can change that.

glean-core/src/upload/directory.rs
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1657112

glean-core/src/upload/directory.rs
badboy: Hm, in that case if we actually panic and crash we have a better indicator than not seeing it ever.

glean-core/src/upload/mod.rs
Dexterp37: @badboy when we added that different mobile quota on Desktop/Fennec, the world was very different place, SDs had lower space, etc. etc. - From a Firefox Desktop POV, 15MB means about 45days worth of pings. In the mobile world, that's twice as much (at least!). I'd advocate for using one single quota, make things simpler. Possibly lower this to 10MB.

badboy: I agree I want a single quota (for now, maybe configurable in the future -> separate bug).
Sticking with 10 MB for now sounds reasonable.

glean-core/metrics.yaml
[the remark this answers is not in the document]
brizental: I talked with @Dexterp37 a bit about this offline. He does agree it would be nice to have a distribution so that we can tell exactly how many pings were deleted per time we reach the quota, right now though we only have memory distrbutions or timing distribuitons and for this we would need a [custom distribution](https://mozilla.github.io/glean/book/user/metrics/custom_distribution.html), which is not available for usage outside of gecko view.

The counter at least will tell us how many pings we are losing from reaching quota, it's arguable how necessary that is. Let me know what you think.

glean-core/src/upload/mod.rs
brizental: This discussion continued in: https://github.com/mozilla/glean/pull/1110#discussion_r465010205

docs/user/pings/index.md
```diff
@@ -94,10 +94,13 @@ where:
...
+To keep resource usage in check, the Glean SDK enforces some limitations on ping uploading and ping storage.
 
 - **Rate limiting**: only up to 10 ping submissions every 60 seconds are allowed. There are no exposed methods to change these rate limiting defaults yet, follow [Bug 1647630](https://bugzilla.mozilla.org/show_bug.cgi?id=1647630) for updates.
 - **Request body size limiting**: the body of a ping request may have up to 1MB. Pings that exceed this size are discarded and don't get uploaded. Size and number of discarded pings are recorded on the internal Glean metric [`glean.upload.discarded_exceeding_pings_size`](../collected-metrics/metrics.md#metrics-1).
+- **Storage quota**: pings are stored on the `pending_pings` directory. That directory is scanned every time Glean is initialized and upon scanning Glean checks its size. If this directory exceeds a quota of 10MB, pings are deleted to get the directory back to an accepted size. The deleted pings are always the oldest pings in the directory and the number of deleted pings is the necessary to get the directory back to an accepted size.
```
badboy, line 101: The name is an internal detail.

```suggestion
- **Storage quota**: Pending pings are stored on disk. The storage directory is scanned every time Glean is initialized and upon scanning Glean checks its size. If this directory exceeds a size of 10MB, pings are deleted to get the directory back to an accepted size. Pings are deleted oldest first, until the directory size is below the quota.
```

Maybe like this?

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/DeletionPingTest.kt
```diff
@@ -133,6 +134,7 @@ class DeletionPingTest {
             serverEndpoint = "http://" + server.hostName + ":" + server.port
         ), clearStores = true, uploadEnabled = false)
         triggerWorkManager(context)
+        waitForEnqueuedWorker(context, PingUploadWorker.PING_WORKER_TAG)
```
badboy, line 137: Oh was this causing test failures?

glean-core/src/upload/mod.rs
```diff
@@ -1136,4 +1182,44 @@ mod test {
...
+        //
+        // From manual testing I figured out an empty ping file is 324bytes,
+        // I am setting this a little over just so that minor changes to the ping structure
+        // don't mmediatelly break this.
+        upload_manager.enqueue_cached_pings(&glean, 500);
+
+        // Get a task once
+        // One ping should have been enqueued.
+        match upload_manager.get_upload_task(&glean, false) {
+            PingUploadTask::Upload(_) => {}
```
badboy, line 1214: Hm, we _could_ check that we're seeing the last written ping (seq_id 9), but that would require us to parse the body into JSON, then look into that.
(or do some ugly string check).

Did you by chance check that it is the latest ping, just to verify our own assumptions in which order we delete?

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/DeletionPingTest.kt
brizental: Yeap, this is a hacky fix for now. I figure this is a potential problem for all tests that use the `triggerWorkManager` function and I opened [a bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1657589) to investigate later. Could really even be a problem in other bindings.

glean-core/src/upload/mod.rs
brizental: Hm, really good that you asked for this. Actually found a bug in the implementation where we are deleting the newest pings, not the oldest. Somehow we will need to "reverse drain" the cached_pings. Something for tomorrow's Bea.

badboy: Hah, good I asked :D 
In that case let's make sure to encode this into the test after you fix it.
Guess we need 2 runs through the list then? 
One to trim it down to the maximum size, then one to insert in the right order?

brizental: Also glad you asked! And you are correct, we need two run throughs and we need to reverse the vector for the first run thorugh and then reverse it back for enqueueing... I guess this will need another pass before I am confident to merge. You can look at the changes in this commit https://github.com/mozilla/glean/pull/1110/commits/586d84e628e4c1280d9c66e23a8f95eb289e5f37.

glean-core/src/upload/mod.rs
```diff
@@ -297,6 +300,75 @@ impl PingUploadManager {
...
+    /// * `quota` - The quota, in bytes, for the size of the pending pings directory.
+    fn enqueue_cached_pings(&self, glean: &Glean, quota: usize) {
+        let mut cached_pings = self
+            .cached_pings
+            .write()
+            .expect("Can't write to pending pings cache.");
+
+        let mut pending_pings_directory_size: usize = 0;
+        let mut deleting = false;
+        // The pending pings vector is sorted by date in asceding order (oldest -> newest).
```
badboy, line 328:
```suggestion
        // The pending pings vector is sorted by date in ascending order (oldest -> newest).
```

glean-core/src/upload/mod.rs
```diff
@@ -297,6 +300,75 @@ impl PingUploadManager {
...
+        // We need to calculate the size of the pending pings directory
+        // and delete the **oldest** pings in case quota is reached.
+        // Thus, we reverse the order of the pending pings vector,
+        // so that we iterate in descending order (newest -> oldest).
+        cached_pings.pending_pings.reverse();
+        cached_pings.pending_pings.retain(|(file_size, (document_id, _, _, _))| {
+            pending_pings_directory_size += *file_size as usize;
```
badboy, line 335: Two things:
1. If we have the size as u64, we should just have the quota as that as well and then the `pending_pings_directory_size` can be u64 and we can remove the cast.
2. I think `&file_size` in the above line should work, then we don't need the dereference. Personally I find that more readable, but ymmv.

PR #1111, opened 2020-07-30
Bug 1648439 - Add the MemoryDistributionMetricType to C#

Dexterp37:
Note that this not add the `accumulateSamples` function available in Kotlin, since that's only required for project EXTRACT, which is not available to C#.


CHANGELOG.md
```diff
@@ -9,6 +9,7 @@
 * C#
   * Add support for Labeled Strings and Labeled Booleans.
   * Add support for the Counter metric type and Labeled Counter.
+  * Add support for the MemoryDistributionMetricType.
```
mdboom, line 12: To make the spellchecker happy.

```suggestion
  * Add support for the `MemoryDistributionMetricType`.
```

PR #1112, opened 2020-07-30
Bug 1654811 - Add "fake" legacy telemetry ID to deletion-request integration tests

travis79:
This adds a fake legacy telemetry id to test out the payload of the deletion-request ping to ensure that we can indeed send other ids in it.


travis79:
Using my superpowers to override the Swiftlint check. [The bug to resolve the linting issue](https://bugzilla.mozilla.org/show_bug.cgi?id=1656443) has been filed (thanks @badboy)

PR #1116, opened 2020-07-31
Bug 1655481, Bug 1655760: Add glean_parser version update script

mdboom:
Also, check the glean_parser version at runtime from the Python bindings.


bin/update-glean-parser-version.sh
```diff
@@ -0,0 +1,50 @@
+#!/usr/bin/env bash
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Update the glean_parser version
```
Dexterp37, line 7: fly-by: can you please add docs for this?

PR #1120, opened 2020-08-03
Bug 1644364 -  Move logic to limit max retries on recoverable failures to glean-core

brizental:


glean-core/src/upload/mod.rs
```diff
@@ -112,7 +117,16 @@ pub enum PingUploadTask {
...
-    /// A flag signaling that the pending pings queue is empty and requester is done.
+    /// A flag signaling that requester doesn't need to request any more upload tasks at this moment.
+    ///
+    /// There are two possibilities for this scenario:
+    /// * Pending pings queue is empty, no more pings to request;
+    /// * Requester has reported more than MAX_RECOVERABLE_FAILURES_PER_PERIOD
+    ///   recoverable upload failures on the same period[1]
+    ///   and should stop requesting at this moment.
+    ///
+    /// [1]: A "period" starts when a requester gets a new `PingUploadTask::Upload(PingRequest)`
```
brizental, line 128: Not really happy with the "period" terminology here, couldn't think of anything better at the time. Please help me in bikeshedding.

Dexterp37: Time window? Window? Time-frame?

glean-core/src/upload/mod.rs
```diff
@@ -326,17 +347,7 @@ impl PingUploadManager {
...
-    /// ## Arguments
-    ///
-    /// * `glean` - The Glean object holding the database.
-    /// * `log_ping` - Whether to log the ping before returning.
-    ///
-    /// # Return value
-    ///
-    /// `PingUploadTask` - see [`PingUploadTask`](enum.PingUploadTask.html) for more information.
-    pub fn get_upload_task(&self, glean: &Glean, log_ping: bool) -> PingUploadTask {
+    fn get_upload_task_job(&self, glean: &Glean, log_ping: bool) -> PingUploadTask {
```
Dexterp37, line 350:
```suggestion
    fn get_upload_task_internal(&self, glean: &Glean, log_ping: bool) -> PingUploadTask {
```

?

glean-core/src/upload/mod.rs
```diff
@@ -352,6 +363,14 @@ impl PingUploadManager {
             self.enqueue_ping(glean, &document_id, &path, &body, headers);
         }
 
+        if self.recoverable_failure_count() >= MAX_RECOVERABLE_FAILURES_PER_PERIOD {
```
Dexterp37, line 366: Once you adjust the terminology, don't forget to fix it here as well.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -98,14 +95,9 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
      */
     @Suppress("ReturnCount")
     override fun doWork(): Result {
-        // We counte the number of failures,
-        // if we get more than three failures in upload we will stop this worker.
-        //
-        // This is a hack before a more robust mechanism is implemented in Rust.
-        var uploadFailures = 0
+        // Create a slot of memory for the task: glean-core will write data into
```
Dexterp37, line 98: Where does the creation happen? Are you referring to the `val incomingTask = ...` below? If so, please move the comment closer.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -121,17 +113,12 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
...
-
                     // Process the upload response
                     LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
                 }
                 PingUploadTask.Wait -> return Result.retry()
                 PingUploadTask.Done -> return Result.success()
             }
-        } while (uploadFailures < MAX_RETRIES)
-        return Result.failure()
+        } while (true)
```
Dexterp37, line 122: Let's add a short comment explaining why this won't lead to an infinite loop :-) (i.e. we're taking care of this in glean-core)

glean-core/csharp/Glean/Net/BaseUploader.cs
```diff
@@ -100,9 +97,8 @@ internal void TriggerUpload(Configuration config)
             // FOR TESTING Implement the upload worker here and call this from Glean.cs
 
             int waitAttempts = 0;
-            int uploadFailures = 0;
 
-            while (uploadFailures < MAX_RETRIES)
+            while (true)
```
Dexterp37, line 101: nit: same comment as the one requested for Kotlin

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -98,18 +95,14 @@ public class HttpPingUploader {
     /// It will report back the task status to Glean, which will take care of deleting pending ping files.
     /// It will continue upload as long as it can fetch new tasks.
     func process() {
-        var uploadFailures = 0
-        while uploadFailures < Constants.maxRetries {
+        while true {
```
Dexterp37, line 98: nit: same comment as the one requested for Kotlin

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -120,9 +116,7 @@ def _process(data_dir: Path, application_id: str, configuration) -> bool:
 
     wait_attempts = 0
 
-    upload_failures = 0
-
-    while upload_failures < MAX_RETRIES:
+    while True:
```
Dexterp37, line 119: nit: same comment as the one requested for Kotlin

glean-core/python/tests/test_network.py
```diff
@@ -81,11 +81,12 @@ def test_500_error_submit(safe_httpserver, monkeypatch):
     ProcessDispatcher._wait_for_last_process()
 
     # This kind of recoverable error will be tried 10 times
-    assert 10 == len(safe_httpserver.requests)
+    # The number of retries is defined on glean-core
+    assert 3 == len(safe_httpserver.requests)
```
Dexterp37, line 85: Ah, was it retried 10 times before in Python?

glean-core/src/upload/mod.rs
```diff
@@ -26,6 +26,11 @@ mod directory;
 mod request;
 mod result;
 
+/// The maximum recoverable failures allowed per period.
```
Dexterp37, line 29: Terminolgy :)

glean-core/src/upload/mod.rs
brizental: Much nicer, clearly yesterday was not a good naming day for me.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
brizental: Hm, this only shows as an added line bcause I moved the lines around it.

I see I did move the lines around it incorrectly and it should be right before `val incomingTask = ...`. Good catch!

glean-core/python/tests/test_network.py
brizental: Uhum, actually this was a bug in the `MAX_RETRIES` python implementation. And subsequently a bug in the tests, actually it wouldn't only be retried 10 times, there was no limitation.

I was not taking into account the the recoverable failures related to 5xx status codes. 

+1 for centralizing code and removing duplication I guess.

brizental: Remider for myself to update the comment in line 83 here.

Dexterp37: I see. If this was a bug in Python, would you kindly add a related BUGFIX entry in the changelog?

CHANGELOG.md
```diff
@@ -9,6 +9,9 @@
 * C#
   * Add support for the String List metric type.
   * Enable generating the C# APIs using the glean_parser.
+* Python
```
Dexterp37, line 12: Can you also add a "General" changelog entry for this PR?

brizental: Sure. I was not sure that was necessary since I am not introducing any new functionality.

PR #1121, opened 2020-08-03
WIP: 1655480: Address startup pings being sent with incomplete core metrics

mdboom:
We are getting an incomplete set of core metrics on some pings sent at
startup: "startup" (events), "upgrade" (metrics), and "overdue" (metrics)

1) This tries to address that by only sending these startup pings if
all the expected metrics are there -- resulting in data loss.

2) Alternatively, we could fill in the missing values with values from the
current run (which would result in incorrect data).

3) Alternatively, we could fill in the missing values with "MISSING", which
would create noisy data, but at least something that DS could filter on.

I'm leaning toward (3) but putting up this WIP PR in any event to solicit
comments...

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringMetricType.kt
@@ -74,6 +74,15 @@ class StringMetricType internal constructor(
     }
 
     /**
+     * Internal only. Check whether a field has a value. Used for determining if
+     * all core metrics have been set.
+     */
+    internal fun hasValue(): Boolean {
+        val res = LibGleanFFI.INSTANCE.glean_string_test_has_value(this.handle, sendInPings.first())
+        return res.toBoolean()
+    }
+
+    /**
      * Tests whether a value is stored for the metric for testing purposes only. This function will
      * attempt to await the last task (if any) writing to the the metric's storage engine before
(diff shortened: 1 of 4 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/StringMetricType.kt
```diff
@@ -73,6 +73,15 @@ class StringMetricType internal constructor(
         LibGleanFFI.INSTANCE.glean_string_set(this.handle, value)
     }
 
+    /**
+     * Internal only. Check whether a field has a value. Used for determining if
+     * all core metrics have been set.
+     */
+    internal fun hasValue(): Boolean {
```
Dexterp37, line 80: Brrrrrr please let's not use this. If crashes are really the culprit here, we can also just rely on the dirty flag.

mdboom: Unfortunately, the dirty flag isn't sufficient here.  We are still "ok" with these pings if the app follows:

"foreground (dirty) -> background -> successfully sends ping (not dirty) -> app shutdown"

I thought about adding another flag, but thought that would be even worse.

In any event, I think I'm preferring the approach in #1123 at this point anyway...

mdboom:
Closing this.  This is *definitely* not the direction we want to go in.

PR #1123, opened 2020-08-03
WIP: Bug 1655480: Fill in missing client_info fields

mdboom:
This is an alternative to #1121.  This just makes sure that all required client_info fields have *something* in them.

It is probably impossible to prevent Android from killing the app while the core metrics are being written to the database, so it will always be the case that the database has some missing fields for some clients some of the time.  As discussed elsewhere, we could fill them in with values for the *current run* of the app, but this data is supposed to come from the *previous run* of the app, so that would be incorrect.  We could also not send pings when this happens, but that would lose some data (and would be logically the same as what we have now where those pings are dropped at ingestion).  This is the middle ground solution -- to just put come placeholder in there, so the data is collected, but data analysis could filter it out if needed.

### DIFF
--- a/glean-core/src/ping/mod.rs
+++ b/glean-core/src/ping/mod.rs
@@ -9,6 +9,7 @@ use std::io::Write;
 use std::path::{Path, PathBuf};
 
 use log::info;
+use once_cell::sync::Lazy;
 use serde_json::{json, Value as JsonValue};
 
 use crate::common_metric_data::{CommonMetricData, Lifetime};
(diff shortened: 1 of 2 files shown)


glean-core/src/ping/mod.rs
```diff
@@ -41,6 +42,21 @@ impl Default for PingMaker {
...
+static REQUIRED_CLIENT_INFO_FIELDS: Lazy<Vec<String>> = Lazy::new(|| {
+    vec![
+        "app_build".into(),
+        "app_display_version".into(),
+        "architecture".into(),
+        "first_run_date".into(),
+        "os".into(),
+        "os_version".into(),
+    ]
+});
```
badboy, line 58: No need to allocate (we will need to allocate later on insert though):

```suggestion
static REQUIRED_CLIENT_INFO_FIELDS: &[&str] = &[
    "app_build",
    "app_display_version",
    "architecture",
    "first_run_date",
    "os",
    "os_version",
];
```

Dexterp37:
I'd favour this approach more than #1121: the other one relies on testing APIs in production code :)

While I have no strong objection to this PR, I'd love to understand a bit more about what's causing this, since we're curing the symptoms here. Are we confident this is exclusively due to crashes? Did we try to correlate malformed pings (maybe trying to recover the client ids from the error stream) with clients who sent dirty startup baselines and/or crash data?

mdboom:
Closing as WONTFIX -- it is below the "we should care" threshold.  See [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1655480) for details.

PR #1126, opened 2020-08-04
Bug 1624626: Python: Add measure context manager

mdboom:


badboy:
Looks like the wrong bug is mentioned here.

badboy:
Fixed the PR title

glean-core/python/glean/metrics/timespan.py
```diff
@@ -100,6 +100,37 @@ def cancel(self) -> None:
...
+        A context manager for recording timings. Used by the `measure` method.
+        """
+
+        def __init__(self, timespan: "TimespanMetricType"):
+            self._timespan = timespan
+
+        def __enter__(self) -> None:
+            self._timespan.start()
+
+        def __exit__(self, type, value, tb) -> None:
```
badboy, line 114: I read up on context managers and it seems all 3 values are referring to a potential exception.
Any reason to check tb specifically?
(According to https://docs.python.org/2/library/sys.html#sys.exc_info if there's no exception than all 3 are none, but it doesn't say that `tb` can't be `None` when there _is_ an exception. Then again I don't see how that would happen, so I defer that to you if this is correct)

glean-core/python/tests/metrics/test_timing_distribution.py
```diff
@@ -150,3 +150,57 @@ def override_time_ns():
...
+        disabled=False,
+        category="telemetry",
+        lifetime=Lifetime.APPLICATION,
+        name=f"timing_distribution",
+        send_in_pings=["baseline"],
+        time_unit=TimeUnit.NANOSECOND,
+    )
+
+    with metric.measure():
+        override_time = 1000
```
badboy, line 176: First reading that I was slightly confused. Maybe add a comment taht we actively manipulate the timing function?

glean-core/python/glean/metrics/timing_distribution.py
```diff
@@ -142,6 +142,37 @@ def cancel():
...
+            else:
+                self._timing_distribution.cancel(self._timer_id)
+
+    def measure(self) -> "_TimingDistributionContextManager":
+        """
+        Provides a context manager for measuring the time it takes to execute
+        snippets of code in a `with` statement.
+
+        If the contents of the `with` statement raise an exception, the timing
+        is not recorded.
```
badboy, line 168: Is the exception re-raised from that block?

glean-core/python/glean/metrics/timespan.py
mdboom: I was just copying the code example in the PEP about context managers: https://docs.python.org/2.5/whatsnew/pep-343.html

But, yeah, maybe you're right that we need to be more careful.  This stuff isn't very well documented -- let me see if I can find a more canonical example.

glean-core/python/glean/metrics/timing_distribution.py
mdboom: Yes -- the `with` block doesn't catch the exception, so it bubbles out.

glean-core/python/glean/metrics/timespan.py
badboy: That's good enough for me.

PR #1128, opened 2020-08-04
Fix iOS tests

travis79:


glean-core/ios/GleanTests/GleanTests.swift
```diff
@@ -309,24 +309,42 @@ class GleanTests: XCTestCase {
...
             )
         }
+
+        expectation = expectation(description: "Completed upload")
+
+        // Resetting Glean doesn't trigger pings in tests so we must call the method
+        // directly to invoke a ping to be created
+        Glean.shared.submitPingByName(pingName: "baseline")
+
+        waitForExpectations(timeout: 5.0) { error in
```
brizental, line 345: TIL: I should **always** wait for the expectations in these tests to fulfill. In retrospect, it does make sense.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -110,8 +110,11 @@ class MetricsPingSchedulerTests: XCTestCase {
         let mps = MetricsPingScheduler()
         let now = Date()
 
+        Glean.shared.resetGlean(configuration: Configuration(), clearStores: true, uploadEnabled: false)
```
brizental, line 113: Do you add `uploadEnabled: false` here so that no pings are uploaded?

brizental:
Tested it all out and it works, no more pings are sent. I tested by adding the debugViewTag to *every* request sent out by the tests.

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
travis79: Yes, it's an added measure since the other tests in this test class handle initializing Glean in the way they need.  Adding comment now 👍

glean-core/ios/GleanTests/GleanTests.swift
travis79: Yeah, that is how we "await" the asynchronous nature of the HTTP request.  I should have caught this earlier in reviews but we found it now :)

travis79:
> As I mentioned in a private message I wonder how that deletion-request we were sending before stopped being sent. Maybe because of the fact that we were not waiting for expectation to fulfill in the debug view tag tests?

That could be it, there could also have been another test that was affected by the weird state of Glean in the MPS tests that toggled upload enabled and triggered it.  We should keep our eyes out for more "valid-tag" pings in the Debug View just in case.

PR #1130, opened 2020-08-05
Use same glean_parser version in Python part as everywhere else

badboy:


badboy:
Fixes CI failure I saw in #1110

mdboom:
This was added very recently. I think it was a merge ordering snafu that caused this.

PR #1132, opened 2020-08-05
Bug 1657292 - Ping registration off main thread

badboy:
~~I don't like the second commit that disables the metrics ping by setting the last sent date, awkardly.
Not sure why that is necessary, at a minimum we should have a nicer way to deal with it.~~

Otherwise this seems to work in our tests.
We would need to take this to a test drive in Fenix to figure out if that fixes the startup delay for them.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -156,8 +156,6 @@ open class GleanInternalAPI internal constructor () {
         // Execute startup off the main thread.
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.executeTask {
-            registerPings(Pings)
-
             val cfg = FfiConfiguration(
                 dataDir = gleanDataDir.path,
                 packageName = applicationContext.packageName,
@@ -198,15 +196,21 @@ open class GleanInternalAPI internal constructor () {
             val isDirtyFlagSet = LibGleanFFI.INSTANCE.glean_is_dirty_flag_set().toBoolean()
             LibGleanFFI.INSTANCE.glean_set_dirty_flag(false.toByte())
 
+            // Register builtin pings.
+            // Unfortunately we need to manually list them here to guarantee they are registered synchronously
+            // before we need them.
+            // We don't need to handle the deletion-request ping. It's never touched from the language implementation.
+            LibGleanFFI.INSTANCE.glean_register_ping_type(Pings.baseline.handle)
+            LibGleanFFI.INSTANCE.glean_register_ping_type(Pings.metrics.handle)
+            LibGleanFFI.INSTANCE.glean_register_ping_type(Pings.events.handle)
+
             // If any pings were registered before initializing, do so now.
             // We're not clearing this queue in case Glean is reset by tests.
             synchronized(this@GleanInternalAPI) {
                 pingTypeQueue.forEach {
                     // We're registering pings synchronously here,
                     // as this whole `initialize` block already runs off the main thread.
-                    LibGleanFFI.INSTANCE.glean_register_ping_type(
-                        it.handle
-                    )
+                    LibGleanFFI.INSTANCE.glean_register_ping_type(it.handle)
                 }
             }
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -201,7 +201,13 @@ open class GleanInternalAPI internal constructor () {
             // If any pings were registered before initializing, do so now.
             // We're not clearing this queue in case Glean is reset by tests.
             synchronized(this@GleanInternalAPI) {
-                pingTypeQueue.forEach { registerPingType(it) }
+                pingTypeQueue.forEach {
+                    // We're registering pings synchronously here,
+                    // as this whole `initialize` block already runs off the main thread.
+                    LibGleanFFI.INSTANCE.glean_register_ping_type(
+                        it.handle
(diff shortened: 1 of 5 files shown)


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -670,6 +674,15 @@ class GleanTest {
 
     @Test
     fun `test sending of startup baseline ping with application lifetime metric`() {
+        // Let's tell the MPS that we just sent a metrics ping.
+        // This way we won't trigger it again here.
+        val fakeNowDoNotSend = Calendar.getInstance()
```
Dexterp37, line 679: @badboy why do we need this?

badboy: From my commit message:

> We don't expect to receive a metrics ping in these tests.
> An "easy" way to prevent it is to mark a metrics ping as recently sent.

For some reason this test (and the other one) suddenly saw a metrics ping before the ping it expected.
I could have also gone and ignore that while waiting for the baseline ping.
I'm not sure _why_ it started to happen now though.

This is the part where I'm unsure about this solution. I just tried to make it work.
The fact that I needed to change other tests could point out that something else is wrong now, or that metrics ping scheduling is very weird there.

badboy: I _think_ it's worth taking this to a Fenix build and test it out. If it doesn't fix the underlying issue we don't need to waste time to clean up the tests.

If it does, we'll find out what's going on in the tests and find a proper solution.

badboy:
> I think this is fine, since we only send built-in pings (defined in the Rust core) at startup.
>
> This does indicate that we have a bug: custom pings that include events should also be flushed at startup because they have the same timing issues as the events. But that's a separate bug and not a live issue right now.

We already do that.
We [do check for pending events on startup](https://github.com/mozilla/glean/blob/fa2810fdeeea08f3f827c71ed5f37f0aef240c1e/glean-core/src/lib.rs#L329-L331) and send those out in their corresponding pings. For that we need to have that ping registered (because we need to know if it should include the client ID).
For most user (and for all that I know) events always only land in the `events` ping, so it's no problem (we register that ping synchronously before we need it), but in theory we could have a custom ping that also has events.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
Dexterp37: Cool, we can also ask @mcomella for support to double check the fix solves the problem, if that helps. Thanks for tackling this.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -789,10 +795,17 @@ open class GleanInternalAPI internal constructor () {
...
     internal fun registerPingType(pingType: PingTypeBase) {
-        if (this.isInitialized()) {
-            LibGleanFFI.INSTANCE.glean_register_ping_type(
-                pingType.handle
-            )
+        // If this happens after Glean.initialize is called (and returns),
+        // we dispatch ping registration on the thread pool.
+        // Registering a ping should not block the application.
+        // Submission itself is also dispatched, so it will always come after the registration.
+        if (this.initFinished) {
```
Dexterp37, line 802: Mh, this is interesting. [initFinished](https://github.com/mozilla/glean/blob/fa2810fdeeea08f3f827c71ed5f37f0aef240c1e/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L262) is set outside of the dispatched init, which means this is automatically set to `true` whenever `Glean.Initialize` is called, regardless of the fact that the dispatched init has executed or not. This might be another bug, but leads to two questions:

- was that intentional?
- is the `initFinished` really a misnamed variable that should be `initDispatched`?
- If that was intentional, then this check is not really helping too much. as we really want to make sure this is exclusively happening [after this](https://github.com/mozilla/glean/blob/fa2810fdeeea08f3f827c71ed5f37f0aef240c1e/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L203-L211) , otherwise we might loose some ping registration.

badboy: >     * was that intentional?

Yes, we only need to track that `initialize` was called.

>     * is the `initFinished` really a misnamed variable that should be `initDispatched`?

Happy to bikeshed on the name, yes.

>     * If that was intentional, then this check is not really helping too much. as we really want to make sure this is exclusively happening [after this](https://github.com/mozilla/glean/blob/fa2810fdeeea08f3f827c71ed5f37f0aef240c1e/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L203-L211) , otherwise we might loose some ping registration.

It is helping. We put it on the dispatch queue, which is not worked on until `Dispatchers.API.flushQueuedInitialTasks()` is called late in the dispatched initialize part.

Dexterp37: Got it. I'm now convinced of this :) Thanks for bearing with me.

badboy:
I asked for help with profiling. Until then we shouldn't merge it.

badboy:
I modified it slightly and changed tests.
Please give it a thorough review

badboy:
> This looks good. I agree the removed test doesn't really test anything correctly anymore.
> 
> Are there C# changes also required? Since we don't have any C# users, happy to just file a bug and handle that later.

I file one.

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=1660815

PR #1133, opened 2020-08-05
Bug 1655100 - Capture device logs to diagnose intermittent iOS integration test failure

travis79:
This adds a 100ms delay after launching the app to the `testDeletionRequestPing` test in order to prevent a race condition that is happening when the test is triggering `setUploadEnabled(false)` very quickly after app launch, which is intermittently causing the deletion-request ping to not be sent.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -526,6 +526,10 @@ jobs:
             - target
           key: v1-cargo-cache-{{arch}}-{{checksum "buildtype.txt"}}-{{checksum "Cargo.lock"}}
       - run:
+          name: Configure device logging
+          command: |
+            xcrun simctl spawn booted log config --mode "level:debug" --subsystem org.mozilla.glean-sample-app
+      - run:
           name: Run iOS tests
           command: |
             if git log -1 "$CIRCLE_SHA1" | grep -q '\[doc only\]'; then
@@ -542,6 +546,14 @@ jobs:
       - store_artifacts:
           path: raw_xcodetest.log
           destination: raw_xcodetest.log
+      - run:
+          name: Collect device logs
+          command: |
+            xcrun simctl spawn booted log collect --output $(pwd)/ios-tests.logarchive
+            zip -r $(pwd)/device_logs.zip $(pwd)/ios-tests.logarchive
+      - store_artifacts:
+          path: device_logs.zip
+          destination: device_logs.zip
       - persist_to_workspace:
           root: build/
           paths: docs/swift
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -633,12 +633,24 @@ jobs:
           path: raw_sample_xcodebuild.log
           destination: raw_sample_xcodebuild.log
       - run:
+          name: Configure device logging
+          command: |
+            xcrun simctl spawn booted log config --mode "level:debug" --subsystem org.mozilla.glean-sample-app
+      - run:
           name: Run sample app tests
           command: |
             bash bin/run-ios-sample-app-test.sh
       - store_artifacts:
           path: raw_sample_xcodetest.log
           destination: raw_sample_xcodetest.log
+      - run:
+          name: Collect device logs
+          command: |
+            xcrun simctl spawn booted log collect --output $(pwd)/ios-integration-test.logarchive
+            zip ~/project/device_logs.zip ~/project/ios-integration-test.logarchive
+      - store_artifacts:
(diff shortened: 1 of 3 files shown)


samples/ios/app/glean-sample-appUITests/DeletionRequestPingTest.swift
```diff
@@ -43,6 +43,11 @@ class DeletionRequestPingTest: XCTestCase {
         app.launchArguments = ["USE_MOCK_SERVER", "\(port)"]
         app.launch()
 
+        // Sleep for a 100ms to prevent an unlikely race condition that, while
+        // not practically acheiveable in real-world situations, does appear to
```
Dexterp37, line 47: What is this race condition? Can you describe what's racing against what and why that's unlikely to happen?

badboy: Yeah, I was about to ask the same.
If we _are_ running into a race condition here we should try to plug that as well (even if that happens in a followup and we do this hack for now)

travis79: The 'race' was more hypothesis than anything concrete.  I could never get this to reproduce locally so the only thing I could think of that might affect this is that the test toggles the telemetry enabled switch as soon as the app is launched which could potentially still be during Glean's initialization.  The race condition is I believe [here](https://searchfox.org/glean/source/glean-core/ios/Glean/Dispatchers.swift#205) where we are switching between queuing and not queuing when flushing tasks.  That's a pretty narrow windows, I admit, but the only other thing I could think of was that the `setUploadEnabled` was getting called _before_ init was finished and thus ran into [this](https://searchfox.org/glean/source/glean-core/ios/Glean/Glean.swift#231).

Dexterp37: > 
> 
> The 'race' was more hypothesis than anything concrete. I could never get this to reproduce locally so the only thing I could think of that might affect this is that the test toggles the telemetry enabled switch as soon as the app is launched which could potentially still be during Glean's initialization. 

Shouldn't this be covered by #1046?

>The race condition is I believe [here](https://searchfox.org/glean/source/glean-core/ios/Glean/Glean.swift#231), if the app beats Glean to here with toggling the switch before init is complete, then no deletion-request ping.

I believe we should probably let this race here then, and add more instrumentation (= log) to test the theories instead of suppressing it

badboy: As said in a private chat looking at this again (and oh, I wrote that code), that should probably be an atomic to prevent a possible race I can think of.

Otherwise I agree: Can we add logging and see if we trigger it again on CI?

travis79: I have no issues with that, it looks like we already log this as an error inside of glean but I don't see any indication of this in the test logs. The test logs don't appear to capture device logging.

travis79:
Okay, I've added a step that should capture the simulator device logs when we run the iOS integration tests. Running the iOS tests to ensure it works, so not ready for review yet.

.circleci/config.yml
```diff
@@ -636,9 +636,15 @@ jobs:
           name: Run sample app tests
           command: |
             bash bin/run-ios-sample-app-test.sh
+            SIMULATOR_UDID=$(xcrun instruments -s devices | grep 'iPhone 11 (13' | awk -F '[][]' '{print $2}')
+            xcrun simctl spawn $SIMULATOR_UDID log collect --output ~/project/ios-integration-test.logarchive 2>/dev/null
```
badboy, line 640: TIL if there's only one simulator running, this works:

```suggestion
            xcrun simctl spawn booted log collect --output ~/project/ios-integration-test.logarchive 2>/dev/null
```

Then we don't even need the UUID

travis79: Worth a shot, I was just getting ready to try `fastlane scan`

travis79:
Okay, with @badboy's help I think we have this where we are actually saving the logs in CI in such a way that we can see the Glean logs that we are interested in.

Dexterp37:
I defer to @badboy for the review of this, seems it seems very iOS specific.

travis79:
Okay, changelog entry added and I'll leave the bug open for now (I'll untake it and make it a P2 for now)

PR #1135, opened 2020-08-06
Bug 1655613 - Work around a crash on Android 8 buggy devices

Dexterp37:
This works around a bug in the Java compiler on some Android 8/8.1 devices (SDK=25). See [here](https://issuetracker.google.com/issues/110848122#comment17) for more details. On buggy devices, an `AssertionError` is raised because the timezone name `enum` switch is compiled into racy code. Since we only use the timezone name in our logs, we can safely swap that out.

We don't need this defensive programming in `DateUtis.kt` as we're not using timezone names there.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
@@ -63,6 +63,26 @@ internal class MetricsPingScheduler(
     }
 
     /**
+     * Safely transforms a `Date` to a `String`.
+     *
+     * This is only useful for working around a crash bug on Android 8 devices.
+     *
+     * @return the `Date` as a `String` or `<buggy Android 8>` if running on a buggy OS.
+     */
+    internal fun safeDateToString(date: Date): String
+    {
+        // Work around a bug in Android 8 devices (SDK=25). See this platform issue
+        // https://issuetracker.google.com/issues/110848122 . Unfortunately there isn't much
+        // we could do, but since this is just a log message, we can simply add a placeholder
+        // value.
+        return try {
+            date.time.toString()
+        } catch (e: AssertionError) {
+            "<buggy Android 8>"
+        }
+    }
+
+    /**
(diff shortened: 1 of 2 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -62,6 +62,24 @@ internal class MetricsPingScheduler(
...
+    internal fun safeDateToString(date: Date): String
+    {
+        // Work around a bug in Android 8 devices (SDK=25). See this platform issue
+        // https://issuetracker.google.com/issues/110848122 . Unfortunately there isn't much
+        // we could do, but since this is just a log message, we can simply add a placeholder
+        // value.
+        return try {
+            date.time.toString()
+        } catch (e: AssertionError) {
+            "<buggy Android 8>"
```
badboy, line 79: I wanted to complain that this is a bit to simple, but I guess it's only used in logging, so we don't actually care.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
```diff
@@ -62,6 +62,24 @@ internal class MetricsPingScheduler(
         }
     }
 
+    /**
+     * Safely transforms a `Date` to a `String`.
```
badboy, line 66: Not sure if worth pointing out that it uses a placeholder string in case of that exception?

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: Heheh yeah, logging only :)

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/MetricsPingScheduler.kt
Dexterp37: It is! Fixing it!

PR #1136, opened 2020-08-06
Bug 1641941 - Remove the dependency on Android-Components

Dexterp37:
This was a test-only dependency, but helps making Glean a bit more independent.


glean-core/android/src/test/java/mozilla/telemetry/glean/TestUtil.kt
```diff
@@ -256,3 +256,39 @@ fun RecordedRequest.getPlainBody(): String {
...
+// want a dependency on A-C.
+
+/**
+ * Mockito matcher that matches <strong>anything</strong>, including nulls and varargs.
+ *
+ * (The version from Mockito doesn't work correctly with Kotlin code.)
+ */
+internal fun <T> any(): T {
+    Mockito.any<T>()
+    return uninitialized()
```
badboy, line 271: that is a weird looking function definition.

Dexterp37: Indeed!

PR #1138, opened 2020-08-06
Make the C# language bindings use the official Glean SDK version

Dexterp37:


bin/prepare-release.sh
```diff
@@ -140,6 +140,14 @@ run $SED -i.bak -E \
     "${WORKSPACE_ROOT}/${FILE}"
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
+### GLEAN C# BINDINGS ###
+
+FILE=glean-core/csharp/Glean/Glean.csproj
+run $SED -i.bak -E \
+    -e "s/<Version>[0-9A-Z.-]+<\/Version>/<Version>${NEW_VERSION}<\/Version>/" \
```
badboy, line 147: Did you try this? I'm _always_ unsure if `<` has a special meaning or not in sed :D
(I think it only has in vim, not in sed, but let's make sure.)

Dexterp37: Yes, I tried this after changing the version in the file to match the pattern we're looking for, in my git-bash.

PR #1140, opened 2020-08-07
Bug 1657758 - Update Gradle to version 6.5.1

Dexterp37:
This is more recent than what A-C uses, but it is what Fenix has been using. This is required in order for us to make the Glean Gradle plugin compatible to the Android Gradle Plugin 4 (will come as a separate PR).

Note that the content of first commit is automatically generated using [the commands here](https://gradle.org/install/#with-the-gradle-wrapper). The second commit contains fixups mostly from [this](https://docs.gradle.org/current/userguide/upgrading_version_4.html)

**DO NOT MERGE YET**


Dexterp37:
@pocmo would this have any effect on A-C? Fenix bumped to this version as well.

NotWoods:
AC will be upgrading to Gradle 6 soon as well

Dexterp37:
> 
> 
> AC will be upgrading to Gradle 6 soon as well

Ok. I'm also concerned about consumers of Glean that are not using A-C. As far as I can tell by testing the Glean SDK in A-C, consumers will not be able to update Glean unless they also update Gradle.

```
* What went wrong:
A problem occurred evaluating project ':glean:glean'.
> Failed to apply plugin [id 'com.android.internal.version-check']
   > Minimum supported Gradle version is 6.1.1. Current version is 5.6.4. If using the gradle wrapper, try editing the distributionUrl in D:\Mozilla\android-components\gradle\wrapper\gradle-wrapper.properties to gradle-6.1.1-all.zip

```

Dexterp37:
The changes in this PR are included in #1144

PR #1141, opened 2020-08-07
Bug 1656589 - Track the database size on initialization

badboy:


glean-core/metrics.yaml
```diff
@@ -391,3 +391,17 @@ glean.upload:
     expires: never
     no_lint:
       - COMMON_PREFIX
+
+glean.database:
+  size:
+    type: memory_distribution
+    description: >
+      The size of the database file at startup.
+    memory_unit: kilobyte
```
Dexterp37, line 400: Shouldn't this be `byte` since [len returns bytes](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len) and the [accumulate](https://searchfox.org/glean/rev/85c44c5f0bb947803afdc28563bc3b5c7296d7a5/glean-core/src/metrics/memory_distribution.rs#63-66) expects data in the same unit we declare here?

badboy: I ... had that before and I have no idea why I changed it 🤔

glean-core/src/internal_metrics.rs
```diff
@@ -88,3 +88,26 @@ impl UploadMetrics {
         }
     }
 }
+
+#[derive(Debug)]
+pub struct DatabaseMetrics {
+    pub size: MemoryDistributionMetric,
+}
+
+impl DatabaseMetrics {
```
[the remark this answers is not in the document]
badboy: yes

badboy:
> This LGTM, but how does it make the tests on Android fail!?

I think I know why: we now have a metric available for each metrics ping.
Previously our tests relied on it being empty most of the time and thus it not getting sent.
Now that it has a never-cleared metric it will always get sent.

That brings up the questions:
* Ping or Application lifetime? Ping means we only get it once, I guess that is still enough for analysis, given it doesn't change for the application lifetime anyway?
* Is the metrics ping even the right place for it?

badboy:
Choosing a lifetime=user would indeed make tests pass.

glean-core/metrics.yaml
```diff
@@ -391,3 +391,17 @@ glean.upload:
...
+    description: >
+      The size of the database file at startup.
+    memory_unit: byte
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1656589
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1656589#c7
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: "2020-09-30"
```
mdboom, line 407: Should the time horizon be so short?  Maybe we do like 3 months to make sure it has plenty of lead time on the main Fenix channel?

glean-core/src/internal_metrics.rs
```diff
@@ -88,3 +88,26 @@ impl UploadMetrics {
...
+    pub fn new() -> DatabaseMetrics {
+        DatabaseMetrics {
+            size: MemoryDistributionMetric::new(
+                CommonMetricData {
+                    name: "size".into(),
+                    category: "glean.database".into(),
+                    send_in_pings: vec!["metrics".into()],
+                    lifetime: Lifetime::Application,
```
mdboom, line 105: This doesn't match the lifetime in the `metrics.yaml` (which defaults to ping).

glean-core/metrics.yaml
badboy: No, it's actually never expiring. Forgot to update the PR

glean-core/src/internal_metrics.rs
badboy: Yeah, as before forgot to update, but we need to figure out above questions before anyway

Dexterp37:
> 
> 
> > This LGTM, but how does it make the tests on Android fail!?
> 
> I think I know why: we now have a metric available for each metrics ping.
> Previously our tests relied on it being empty most of the time and thus it not getting sent.
> Now that it has a never-cleared metric it will always get sent.
> 
> That brings up the questions:
> 
>     * Ping or Application lifetime? Ping means we only get it once, I guess that is still enough for analysis, given it doesn't change for the application lifetime anyway?

I think Ping lifetime is fine. If we restart multiple times before sending a metric ping, we're sending a timing distribution anyway, so we'll record multiple values.

>     * Is the metrics ping even the right place for it?

Yes, I believe it is, still. We make no guarantee or assumption about this ping _not_ being sent.

badboy:
I had to resort to sort of a small hack to get tests passing by ignoring the metrics ping (by essentially delaying it). That touches _far too many_ tests to my liking, but it's the unfortunate truth how it all works.

I'd like some review on that. It also means we now _always_ have a metrics ping because we are recording metrics (but in actual applications we might not be the only ones anyway, so probably not a real problem?)

badboy:
Ah! I also think the same has to be done for iOS, which I haven't done yet. I also never ran the tests for iOS yet.

PR #1143, opened 2020-08-07
Bug 1574424 - Run Rust tests on beta again

badboy:


badboy:
CircleCI tasks won't run on this because of skip-ci, I guess that's ok given that we don't touch code or CircleCI config

PR #1144, opened 2020-08-07
Bug 1657758 - Support the Android Gradle Plugin v4

Dexterp37:
Note that while this works in the Glean SDK, this seems to break any consumer not using a recent version of Gradle.

Note: I tested that this work by using the locally published Glean in @pocmo 's work from mozilla-mobile/android-components#8360


badboy (approved):
Overall that looks simple enough.

PR #1150, opened 2020-08-12
Bug 1654384 - Add a specific simulator to each of the distributions metrics on the book

brizental:
[doc only]

This addresses only the 3rd bullet point from this bugs description
> 3. Provide an interactive plot.

I modelled myself after [the other histogram simulator](https://telemetry.mozilla.org/histogram-simulator), although I did not copy over the whole code from that one.

Here is a screenshot of the page result -> 
Screenshot_2020-08-12 Histogram simulator - Glean - Cross-platform Telemetry library

[UPDATE] I put this up on a temporary place to be easier to see in action https://rainy-quartz.surge.sh/


docs/user/metrics/memory_distribution.md
```diff
@@ -9,6 +9,8 @@ That is, the function from a value \\( x \\) to a bucket index is:
 
 This makes them suitable for measuring memory sizes on a number of different scales without any configuration.
 
+> **Note** Check out how this bucketing algorithm would behave on our [Histogram Simulator](../../appendix/histograms.html)
```
brizental, line 12: We can add params to this URL to open the simulator on a plot that matches the bucketing algorithm for the memory distribution, but I honestly was not sure what these params were. We would need, lowerBound, upperBound and bucketCount.

docs/user/metrics/timing_distribution.md
```diff
@@ -11,6 +11,8 @@ That is, the function from a value \\( x \\) to a bucket index is:
 
 This makes them suitable for measuring timings on a number of time scales without any configuration.
 
+> **Note** Check out how this bucketing algorithm would behave on our [Histogram Simulator](../../appendix/histograms.html)
```
brizental, line 14: We can add params to this URL to open the simulator on a plot that matches the bucketing algorithm for the timing distribution, but I honestly was not sure what these params were. We would need, lowerBound, upperBound and bucketCount.

docs/book.toml
```diff
@@ -9,7 +9,7 @@ create-missing = false
 
 [output.html]
 additional-css = ["glean.css", "mermaid.css"]
-additional-js = ["tabs.js", "mermaid.min.js", "mermaid-init.js"]
+additional-js = ["tabs.js", "mermaid.min.js", "mermaid-init.js", "chart.min.js", "chart-histogram.js"]
```
brizental, line 12: Is there a way to add these JS files only to the histogram simulator? I know I can just add the script tags directly, but is there a proper way?

docs/user/metrics/memory_distribution.md
badboy: Appreciate the work here.
I think the whole point though is that the bucketing algorithm for memory and timing distributions  is significantly different from the old one used in Firefox Telemetry (and custom distributions).
The bounds are not configurable for example.

Thus we should hard-code these as the defaults on top of the linear/exponential histograms (that are available as custom distributions)

docs/book.toml
badboy: Nope, mdbook doesn't have per-chapter configuration

docs/user/metrics/memory_distribution.md
brizental: That was what I figured, but was not sure. Okay, I'll send a new commit adding these two to the simulator.

Dexterp37:
Some overall comment: Glean has no concept of "histogram", so we should really reshape the histogram simulator to be more meaningful with respect of the other metric types that are histogram-like.

Ideally, we should really get rid of the "histogram" word in this PR :)

brizental:
This is ready for review, the link on the description is up to date with the latest changes. I am not sure how to address the spellcheck issue since it is complaining about the html apparently.

docs/chart-distributions.js
```diff
@@ -0,0 +1,404 @@
...
+        if (!(bucketMinimum in result)) {
+            result[bucketMinimum] = 0
+        }
+    }
+
+    return result
+}
+
+// Copied over and adapted
+// from https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/histogram-simulator/simulator.js
```
Dexterp37, line 310: Let's use a permalink please!

docs/chart-distributions.js
```diff
@@ -0,0 +1,404 @@
...
+        }
+        value = value * sigma + mu
+
+        values.push(value)
+    }
+    return values.map(value => value >= 0 ? value : 0)
+}
+
+// Copied over and adapted
+// from https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/histogram-simulator/simulator.js
```
Dexterp37, line 338: Permalink :)

docs/chart-distributions.js
```diff
@@ -0,0 +1,404 @@
...
+    let values = []
+    for (var i = 0; i <= count; i++) {
+        values.push(Math.random() * (max - min) + min);
+    }
+
+    return values
+}
+
+// Copied over and adapted
+// from https://github.com/mozilla/telemetry-dashboard/blob/gh-pages/histogram-simulator/simulator.js
```
Dexterp37, line 375: Permalink!

docs/chart-distributions-ui.js
```diff
@@ -0,0 +1,147 @@
...
+// which may receive the values in a given unit, but transform them to a base one upon recording
+let TRANSFORMATION
+
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+// !!! Memory distribution specific !!!
+// !!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!!
+
+function memoryUnitToByte(unit) {
+    switch(unit) {
+        case "byte":
```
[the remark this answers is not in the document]
brizental: HA! Nice catch.

docs/chart-distributions-ui.js
```diff
@@ -0,0 +1,147 @@
+// Transformation function we may want to apply to each value before plotting
+//
```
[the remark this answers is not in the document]
brizental: I don't really, but I see that is used throughout our other code, so sure...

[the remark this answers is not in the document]
Dexterp37: Yes, this would allow us to catch some problems :)

brizental:
> Every time you change the unit, data is regenerated, which makes it hard to understand the difference between units. Data should only be regenerated when changing the "data options"

That happens because when we change the unit we change the "transform" function applied to all values before accumulating.

> We should probably always display all the buckets.

What do you mean? We are displaying all buckets, as far as I understand.

Dexterp37:
> 
> 
> > Every time you change the unit, data is regenerated, which makes it hard to understand the difference between units. Data should only be regenerated when changing the "data options"
> 
> That happens because when we change the unit we change the "transform" function applied to all values before accumulating.

Yes, and I believe that's wrong :) The original data is in "bytes", which means data can stay the same

> > We should probably always display all the buckets.
> 
> What do you mean? We are displaying all buckets, as far as I understand.

When regenerating data, the number of buckets displayed changes for me :o I guess the chart library is cropping out empty buckets

docs/chart-distributions-ui.js
```diff
@@ -0,0 +1,147 @@
...
+// The current use cases are memory distributions and timing distribution,
+// which may receive the values in a given unit, but transform them to a base one upon recording
+let TRANSFORMATION
```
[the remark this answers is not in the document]
Dexterp37: Still here :)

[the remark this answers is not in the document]
brizental: AAAAAAAAAA 🤯

brizental:
> Yes, and I believe that's wrong :) The original data is in "bytes", which means data can stay the same

Hm, I understand now. This is maybe a problem in UI, because you and me are getting the meaning of this differently.

My idea was that when you change the unit, it's like you are changing the unit of the input data, that is why it changes. Does that make sense? I added the note about generated data always being between 0-100, to try and make that clear. Apparently it is still not clear. Please let me know what you think.

> When regenerating data, the number of buckets displayed changes for me :o I guess the chart library is cropping out empty buckets

The generated buckets are dependant on the input data, when we change the unit, because the input data is transformed, the buckets change. I am guessing, because the data range changes, the number of buckets also changes. Or this could be a bug in my functional histograms implementation, hm. Let me do a little investigating.

mdboom:
This is great.  Clearly there's some design decisions to make here, because there's some different expectations. I imagine most of that can be cleared up with some better transparency into what's happening "behind the scenes".

>     * Every time you change the unit, data is regenerated, which makes it hard to understand the difference between units. Data should only be regenerated when changing the "data options"

I'm not sure I agree.  I think seeing the drop down change nothing would be more confusing.  Maybe a way to make it less confusing is to have a line beneath it that says "mean at 1,000,000 ns" when milliseconds is selected.  That makes it clear that the input data is changing, which IMHO is the right thing.

It would also be nice (but not critical) to be able to specify a mean directly.

> We should probably always display all the buckets.

I'm also not sure about this.  Yes, in the case of custom distribution.  But in the case of timing and memory dists, it will make the data look awfully small.  Having something like "0 ..." on the left and "... 2^20" on the right would be nice, but probably gets into things that would be too much effort with the plotting library.

One thing I'm noticing is that empty buckets are left out for the log-normal distribution (which feels right to me at least), but they aren't for the normal distribution.


Lastly, there's some sort of performance issue, maybe. When I change one of the drop downs, the new data is displayed more-or-less immediately, but then there is ~5s until the page is responsive again. This is on Nightly, so could be due to some transient bug in Nightly.

That's comments from playing with it -- going to do a code read-through now.

docs/user/metrics/custom_distribution.md
```diff
@@ -86,3 +88,60 @@ assertEquals(1, Graphics.checkerboardPeak.testGetNumRecordedErrors(ErrorType.Inv
...
+        <div class="input-group">
+            <label for="upper-bound">Range maximum (<code>range_max</code>)</label>
+            <input name="upper-bound" id="upper-bound" type="number" value="500" />
+        </div>
+        <div class="input-group">
+            <label for="bucket-count">Bucket count (<code>bucket_count</code>)</label>
+            <input name="bucket-count" id="bucket-count" type="number" value="20" />
+        </div>
+    </div>
+</div>
```
mdboom, line 147: Since this HTML is copied mostly 3 times, I wonder if we should generate it from JS instead...  Only if the tradeoff makes sense.

mdboom:
Oh, also noticed that the custom distribution one is throwing an exception:

image

Dexterp37:
> > ```
> > * Every time you change the unit, data is regenerated, which makes it hard to understand the difference between units. Data should only be regenerated when changing the "data options"
> > ```
> 
> I'm not sure I agree. I think seeing the drop down change nothing would be more confusing. Maybe a way to make it less confusing is to have a line beneath it that says "mean at 1,000,000 ns" when milliseconds is selected. That makes it clear that the input data is changing, which IMHO is the right thing.

For timing distributions, I can see why this would make more sense, given that the resolution changes and we're supposed to provide "data in the unit of the metric". However, for memory distributions, the resolution stays the same and data is _always provided in bytes_. If data changes whenever you change the unit, you can't really appreciate what's going on, IHMO. But I don't have a strong opinion on this.

> > We should probably always display all the buckets.
> 
> I'm also not sure about this. Yes, in the case of custom distribution. But in the case of timing and memory dists, it will make the data look awfully small. Having something like "0 ..." on the left and "... 2^20" on the right would be nice, but probably gets into things that would be too much effort with the plotting library.

Yes, I believe that what you suggest or some sentence displayed like "Showing up to bucket X (last bucket is 2^20)" would work as well. Contrary to what I said in my previous comment, I don't really want to display all the buckets, but give users a sense of the range that the type is able to capture.

brizental:
> I'm not sure I agree. I think seeing the drop down change nothing would be more confusing. Maybe a way to make it less confusing is to have a line beneath it that says "mean at 1,000,000 ns" when milliseconds is selected. That makes it clear that the input data is changing, which IMHO is the right thing.

👍 

> It would also be nice (but not critical) to be able to specify a mean directly.

I can file a follow up :)

> One thing I'm noticing is that empty buckets are left out for the log-normal distribution (which feels right to me at least), but they aren't for the normal distribution.

Hm, I'll take a look.

> Lastly, there's some sort of performance issue, maybe. When I change one of the drop downs, the new data is displayed more-or-less immediately, but then there is ~5s until the page is responsive again. This is on Nightly, so could be due to some transient bug in Nightly.

Ah yes, I noticed that, but it is nightly only. I can't reproduce on stable, Chrome or Safari. We could maybe file a bug, I am not totally sure on how or where.

> I'm also not sure about this. Yes, in the case of custom distribution. But in the case of timing and memory dists, it will make the data look awfully small. Having something like "0 ..." on the left and "... 2^20" on the right would be nice, but probably gets into things that would be too much effort with the plotting library.

It really does make data look awfully small. This is one of the reasons I decided to generate data between 1-100 for timing and memory distributions, the smaller the variance between min and max the better the plot looks, otherwise you can't even really understand what is going on. We could file a follow up to add some sort of zoom to the chart to make it better when there are too many buckets, Chart.js has [a plugin](https://github.com/chartjs/chartjs-plugin-zoom) for that. Having A LOT of buckets doesn't really introduce a noticeable performance change on the actual rendering, so I believe it can be done.

mdboom:
> For timing distributions, I can see why this would make more sense, given that the resolution changes and we're supposed to provide "data in the unit of the metric". However, for memory distributions, the resolution stays the same and data is _always provided in bytes_. If data changes whenever you change the unit, you can't really appreciate what's going on, IHMO. But I don't have a strong opinion on this.

Timing distributions are the same here -- data is always provided in nanoseconds.  The unit only controls the maximum.  The only scaling happens through GeckoView, which I think we shouldn't optimize for.

I agree with you that changing the input data when changing the unit might be confusing.  Leaving it the same would probably also be confusing.  ¯\_(ツ)_/¯

Maybe the additional information proposed: (1) showing the mean of the input data and (2) showing the maximum bucket as changing, will help.

mdboom:
> Ah yes, I noticed that, but it is nightly only. I can't reproduce on stable, Chrome or Safari. We could maybe file a bug, I am not totally sure on how or where.

If it's just Nightly, I'd say let's not worry too much.

docs/user/metrics/custom_distribution.md
brizental: Honestly, I think this is not worth ATM. Maybe when we have more distributions.

brizental:
> Maybe the additional information proposed: (1) showing the mean of the input data and (2) showing the maximum bucket as changing, will help.

I added this and I also added a little note to try and clarify this even further. I also addressed the console error pointed in the custom distributions page and the buckets with 0 not showing up on the graph. I also updated the demo.

This is ready for a final round of reviews :)

docs/chart-distributions-ui.js
```diff
@@ -0,0 +1,172 @@
...
+inputs.forEach(input => {
+    setInputValueFromSearchParam(input);
+    input.addEventListener("change", event => {
+        let input = event.target;
+        setURLSearchParam(input.name, input.value);
+        buildChartFromInputs();
+    });
+});
+
+buildChartFromInputs();
```
badboy, line 133: This file is now included by default on all pages.
We should detect if there is any graph and otherwise bail out early to avoid any work if we know we don't display it.
We could wrap this in a function we can return from and put in a check on top

brizental: Fixed this in eb70508, really should have done this from the beginning.

brizental:
> It's also internally using floating point for values, which the API doesn't accept, so maybe we should normalize that to integers (at least for the "custom data" part where values are directly visible to users).

In eb70508 I apply `Math.floor` to all generated values. I'll update the demo so that you can verify it is still working as it should and then we can merge this. After that I can file all the follow ups.

PR #1151, opened 2020-08-13
Bug 1656038 - Use unified logging on iOS from the Rust side

badboy:
The unified logging system is recommended by Apple: https://developer.apple.com/documentation/os/logging/
We connect the Rust logging side to OSLog, all logs will be available in
the system log (and can be viewed through Console.app).
The log levels can be configured through the `log` utility.

---

This uses the [oslog crate](https://crates.io/crates/oslog).
While just being a 0.0.3 version I think it's ok to use it.
I reviewed the code and it's a minimal wrapper around the underlying `os_log_*` functionality.
It's 300 lines of (Rust) code, half of that are actually basic tests. It's MIT licensed.
`unsafe` code is limited to calling the C API (so required).

I'd like to run the iOS tests to get the logs for a final look that it works as expected (now that #1133 landed)


badboy:
ah, this requires a doc change, I'll add that.

badboy:
This is looking good:

image

badboy:
Yeah, that's what I mean, need to update that.

badboy:
More precisely we have iOS specific instructions: https://mozilla.github.io/glean/book/user/debugging/ios.html#glean-log-messages

I'll change them to be more specific now that we can properly filter all messages.

travis79:
Looks good to me, that was way easier than I expected :)

badboy:
Reflagging for documentation

PR #1154, opened 2020-08-17
Increase rate limiter to 15 pings per minute.

mdboom:
This will be introduced as an experiment in a patch release and shipped to Fenix quickly to see if this is the issue with not getting all migration pings.

(A successful migration sends 11 pings, so stopping with 10 does seem to point at the rate limiter, but this experiment will help us confirm.)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v32.0.0...main)
 
 * General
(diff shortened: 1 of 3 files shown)


mdboom:
> Two (four) questions:
> 
>     * Does this require a doc change? (And if not, do we need a bug to add doc coverage)

Yes -- should have thought of that.  I can only find one place to update, and I'll add it here.

>     * Does this require a test change? (And if not, do we need a bug to add test coverage)

Yes -- I'm leaning on having CI tell me that, and will update what needs to be.

PR #1157, opened 2020-08-18
Document the CI setup and expectations

badboy:
After the trouble Mike had yesterday I thought it would be fair to properly document it now.

---

I also slightly re-ordered some chapters. I _think_ that way it fits better, but I'd be okay to take that out.
While I was at it I also fixed the link-checker docs.


badboy:
(While I am at it I also use this to check if the android tests are running again)

PR #1160, opened 2020-08-19
Bug 1656934 - Upload manager: ignore files if metadata can't be read

badboy:
There's a rare case where this races against a parallel deletion of all pending ping files.
This could therefore fail, in which case we don't care about the result and can ignore the ping, it's already been deleted.

### DIFF
--- a/glean-core/src/upload/directory.rs
+++ b/glean-core/src/upload/directory.rs
@@ -221,12 +221,21 @@ impl PingDirectoryManager {
                         return None;
                     }
                     if let Some(data) = self.process_file(file_name) {
-                        // This may panic in case the path does not exist
-                        // or the user does not have permissions to read the file.
-                        // Both are extremely unlike so we can panic! here if that happens.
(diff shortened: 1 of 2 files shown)


glean-core/src/upload/directory.rs
```diff
@@ -221,12 +221,21 @@ impl PingDirectoryManager {
                         return None;
                     }
                     if let Some(data) = self.process_file(file_name) {
-                        // This may panic in case the path does not exist
-                        // or the user does not have permissions to read the file.
-                        // Both are extremely unlike so we can panic! here if that happens.
```
brizental, line 226: > Both are extremely unlike so we can panic! here if that happens.

Oops

brizental (approved):

glean-core/src/upload/directory.rs
travis79: Good thing it's in the removed lines ;)

PR #1167, opened 2020-08-24
Xcode 12 support (currently Beta5)

garvankeeley:
On Firefox iOS we are now supporting xcode 12 on main and need release for this version.

xref to a-s lib https://github.com/mozilla/application-services/pull/3532

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -464,7 +464,7 @@ jobs:
 
   Check Swift formatting:
     macos:
-      xcode: "11.5.0"
+      xcode: "12.0.0"
     steps:
       - checkout
       - run:
@@ -484,7 +484,7 @@ jobs:
 
   iOS build and test:
     macos:
-      xcode: "11.5.0"
+      xcode: "12.0.0"
     steps:
       - checkout
       - run:
@@ -608,7 +608,7 @@ jobs:
 
   iOS integration test:
     macos:
-      xcode: "11.5.0"
+      xcode: "12.0.0"
     steps:
       - checkout
       - skip-if-doc-only
@@ -674,7 +674,7 @@ jobs:
 
   Carthage release:
     macos:
-      xcode: "11.5.0"
+      xcode: "12.0.0"
     steps:
       - checkout
       - attach_workspace:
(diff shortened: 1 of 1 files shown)


garvankeeley:
Carthage currently has a bug with xcode 12 tooling, it may need to be run with 
```echo 'EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_simulator__NATIVE_ARCH_64_BIT_x86_64=arm64 arm64e armv7 armv7s armv6 armv8' > /tmp/tmp.xcconfig
echo 'EXCLUDED_ARCHS=$(inherited) $(EXCLUDED_ARCHS__EFFECTIVE_PLATFORM_SUFFIX_$(EFFECTIVE_PLATFORM_SUFFIX)__NATIVE_ARCH_64_BIT_$(NATIVE_ARCH_64_BIT))' >> /tmp/tmp.xcconfig
echo 'IPHONEOS_DEPLOYMENT_TARGET=11.4' >> /tmp/tmp.xcconfig
echo 'SWIFT_TREAT_WARNINGS_AS_ERRORS=NO' >> /tmp/tmp.xcconfig
echo 'GCC_TREAT_WARNINGS_AS_ERRORS=NO' >> /tmp/tmp.xcconfig
export XCODE_XCCONFIG_FILE=/tmp/tmp.xcconfig```

badboy:
Carthage issue is tracked here: https://github.com/Carthage/Carthage/issues/3019

Xcode 12 is still in beta. How quickly do you need us building on that?

badboy:
> I'm okay with bumping to a beta version of Xcode, but it does look like it's breaking on building Gzip framework. We may have a little work to get it to build on CI.

that's what I had locally as well and is related to the mentioned Carthage issue

garvankeeley:
Our team is on Beta 5 already, and our main branch is switching today. We can use builds from a fork if this is blocked due to build errors.

badboy:
> Our team is on Beta 5 already, and our main branch is switching today. We can use builds from a fork if this is blocked due to build errors. (at least temporarily we can do that, in a few days we will be entering a QA cycle where we should be using official builds)

👍 -- days sounds doable to us IMO

PR #1171, opened 2020-08-26
Bug 1597249 - Configure dependabot to check Rust and Python dependencies daily

badboy:
Let's try this

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,0 +1,31 @@
+version: 2
+updates:
+  # glean-core
+  - package-ecosystem: "cargo"
+    directory: "/glean-core"
+    schedule:
+      interval: "daily"
+    ignore:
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -0,0 +1,27 @@
...
+  # glean benchmark
+  - package-ecosystem: "cargo"
+    directory: "/glean-core/benchmark"
+    schedule:
+      interval: "daily"
+
+  # glean-python
+  - package-ecosystem: "pip"
+    directory: "/glean-core/python"
+    schedule:
```
mdboom, line 26: Let add: `ignore: glean_parser`, since I think we always want to update glean_parser in lock step in a few different places, and we don't want dependabot telling us when we want to do that one.

badboy: Ah, I forgot that it is a dependency, will add it

PR #1172, opened 2020-08-26
Bug 1661036 - Check if upload is enabled before recording a metric

brizental:
I added this check to the metrics code that wasn't checking, to avoid running unnecessary code when the metric won't be recorded after all.

I also added checks to the `record_*` functions on the databases, just to be sure we are not recording when we don't want to. Note that I only check if upload is disabled on the databases.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -590,10 +590,10 @@ impl Database {
 #[cfg(test)]
 mod test {
     use super::*;
-    use std::collections::HashMap;
-    use tempfile::tempdir;
     use crate::tests::new_glean;
     use crate::CommonMetricData;
+    use std::collections::HashMap;
+    use tempfile::tempdir;
 
     #[test]
     fn test_panicks_if_fails_dir_creation() {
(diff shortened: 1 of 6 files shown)


glean-core/src/database/mod.rs
```diff
@@ -274,6 +274,11 @@ impl Database {
 
     /// Records a metric in the underlying storage system.
     pub fn record(&self, glean: &Glean, data: &CommonMetricData, value: &Metric) {
+        // If upload is disabled we don't want to record.
+        if !glean.is_upload_enabled() {
```
mdboom, line 278: Random thought -- when we disable telemetry, we set the client-id to `c0ffee`.  Does that still work?

brizental: Interesting, let me check.

brizental: Should still work, TIL about this behaviour, but we set the client-id to `c0ffee` on the `clear_metrics` function, which [is called before we set `upload_enabled` to `false`](https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L418-L419)*.

\* At first sight that is what we do, but actually from looking closer at the code we set `upload_enabled` to `false` last thing on the `clear_metrics` function  and we assume `upload_enabled` is already `false` upon entering this function (see: https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L457-L478). Still, the changes in this PR don't affect this behaviour.

badboy (approved):
IMO a reasonable change.
I'm pretty pleased that this doesn't affect any existing tests, so I'm reasonably sure that the old behavior at least was correct.

PR #1179, opened 2020-08-26
Bug 1661276 - Raise limit for max wait attempts

brizental:
This is just a quick fix so that intermittents stop for today.

A more sustainable approach for this would be having a way to turn limits off for tests. I am almost at EOD for today so, quick fix it is for now and tomorrow I work on the more involved fix.


brizental:
> WFM (Let's just not close the bug so we remember to switch it back...) :)

I have already even made the commit switching it back :D Tomorrow I open a new PR fixing this properly.

PR #1191, opened 2020-08-27
Bug 1661276 - Make upload policies easily configurable for tests

brizental:
BONUS: Make uploader tests sync by default


brizental:
This also reverts what was done #1179

brizental:
>  Out of curiosity, are we going to expose this for consumers to set custom policies?

I am not aware of any use cases of this for now, so no. But if you have one, please file bug :D

PR #1195, opened 2020-08-28
iOS: Disable code coverage 

badboy:
This avoids linker errors due to missing errors such as as "___llvm_profile_runtime".
We only use code coverage in development for our manual checks, but we
should not put that in released binary frameworks.


badboy:
I actually need to properly test this once more against Firefox iOS

PR #1196, opened 2020-08-28
Bug 1661036 - Check if upload is enabled before recording a metric [ALTERNATE]

brizental:
This is an alternative to #1172 

The main difference is that, instead of directly getting the `upload_enabled` state from the Glean object to decide whether to record or not on the databases, this solution propagates the `upload_enabled` state to the different Database objects.

The main advantage of this is that it would unblock us on not having to pass the Glean object around to record metrics, and instead just pass the database objects. See: [Bug 1589675](https://bugzilla.mozilla.org/show_bug.cgi?id=1589675)

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -9,6 +9,7 @@ use std::num::NonZeroU64;
 use std::path::Path;
 use std::str;
 use std::sync::RwLock;
+use std::sync::atomic::{AtomicBool, Ordering};
 
 use rkv::{Rkv, SingleStore, StoreOptions};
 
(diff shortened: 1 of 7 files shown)


badboy (commented):
I've only taken a brief look at how you implemented it, not the full patch set though.
IMO I like the other solution slightly more, as I'm worried a bit about duplicating the state across the different objects.
It's easy to see where we need to toggle them now, but it still makes me icky about it.

brizental:
It's the same patch set FWIW, the only addition is the state in the databases. 

I also don't like the duplicated state. The advantage of not relying on the Glean object is a nice one, but there are many other things that rely on it and I did some trials for passing the Database instead of the Glean object to the metrics code and it has many more obstacles than just getting the upload enabled state. It would require kind of a massive change to remove it.

I'll close this PR and leave a comment on that bug with with details to what I investigated. Thanks for the review.

PR #1201, opened 2020-09-08
Bug 1662584 - Submitting a ping with upload disabled no longer errors

Dexterp37:
The error message is now just an info message, so that it doesn't clutter users logs.


mdboom (approved):
LGTM!

PR #1202, opened 2020-09-08
Bug 1655749 - Mention ABI split [doc only]

Dexterp37:
This makes the "enable app id" requirement more visible and adds mention of the ABI split/size concern for android projects.

### DIFF
--- a/docs/user/adding-glean-to-your-project.md
+++ b/docs/user/adding-glean-to-your-project.md
@@ -16,7 +16,9 @@ Products (applications or libraries) using the Glean SDK to collect telemetry **
 
 Additionally, applications (but not libraries) **must**:
 
-5. Provide a way for users to turn data collection off (e.g. providing settings to control `Glean.setUploadEnabled()`). The exact method used is application-specific.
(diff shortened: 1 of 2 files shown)


docs/user/adding-glean-to-your-project.md
```diff
@@ -58,6 +60,7 @@ For example, if version *33.0.0* is used, then the include directive becomes:
 implementation "org.mozilla.components:service-glean:33.0.0"
 ```
 
+> **Size Impact on Android APKs**: the Glean SDK APK ships binary libraries for all the supported platforms. Each library file weights about 600kB. If the final APK size of the consuming project is a concern, please enable [ABI splits](https://developer.android.com/studio/build/configure-apk-splits#configure-abi-split).
```
travis79, line 63:
```suggestion
> **Size Impact on Android APKs**: the Glean SDK APK ships binary libraries for all the supported platforms. Each library file measures about 600kB. If the final APK size of the consuming project is a concern, please enable [ABI splits](https://developer.android.com/studio/build/configure-apk-splits#configure-abi-split).
```

travis79 (approved):
r+wc

PR #1203, opened 2020-09-08
1660120: Make Python bindings installable from source

mdboom:
This will ship a Python source package that should be installable on platforms that we don't provide wheels for (if the destination has `cargo` on the path).  Of particular note to support Apple Silicon and more niche OSes like *BSD.

### DIFF
--- a/glean-core/python/setup.py
+++ b/glean-core/python/setup.py
@@ -4,8 +4,10 @@
 
 """The setup script."""
 
+from distutils.command.build import build as _build
 import os
 import shutil
+import subprocess
 import sys
 
 from setuptools import setup, Distribution, find_packages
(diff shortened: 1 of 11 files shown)


glean-core/python/setup.py
```diff
@@ -126,6 +118,34 @@ def finalize_options(self):
...
 
+class build(_build):
+    def run(self):
+        try:
+            subprocess.run(["cargo"])
+        except subprocess.CalledProcessError:
+            print("glean_sdk requires 'cargo' to build its Rust extension.")
+            sys.exit(1)
+
+        command = ["cargo", "build", "--all"]
```
badboy, line 129: Probably better to build only glean-ffi (`--all` probably also buils glean-preview, which we don't need)

```suggestion
        command = ["cargo", "build", "--package", "glean-ffi"]
```

setup.py
```diff
@@ -0,0 +1,16 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+"""
+A basic top-level setup.py script that delegates to the real one in
```
badboy, line 6: Why is this necessary?
(our top-level directory gets crowded already)

glean-core/python/setup.py
```diff
@@ -126,6 +118,34 @@ def finalize_options(self):
             self.install_lib = self.install_platlib
 
 
+class build(_build):
+    def run(self):
+        try:
+            subprocess.run(["cargo"])
+        except subprocess.CalledProcessError:
+            print("glean_sdk requires 'cargo' to build its Rust extension.")
```
badboy, line 126: Maybe expand this, e.g.

"Install Rust and Cargo through Rustup: https://rustup.rs/. Need help with installing the glean_sdk? Contact the Glean Team: ..."

badboy (commented):
Looks not too complicated.

One thing I don't understand though: is it bundling ALL source code in the repository, or only the Python + Rust parts (and where does it do that?)

mdboom:
> Looks not too complicated.
> 
> One thing I don't understand though: is it bundling ALL source code in the repository, or only the Python + Rust parts (and where does it do that?)

It's bundling the whole code repository (everything checked into git).  *Not* doing that got really complicated (because the rule specification language is kind of broken).  Given that the whole package is ~1.3MB compressed, I didn't think it was worth the effort to save a few hundred kb by not including Android, iOS etc.  Maybe we file a follow-up bug?

setup.py
mdboom: Installing a Python package from source requires a `setup.py` at the top-level of the archive, unfortunately.

badboy:
Oh, if it is that small I don't care much. I just couldn't see it in the code how it does it.

badboy (approved):

PR #1205, opened 2020-09-10
Bug 1656934 - Scan pending pings directories after dealing with upload status

brizental:
The bug that was causing that intermittent, was due to scanning the pings directories **before** running the `on_upload_disabled` function that deleted non-deletion request pings.

I fix that in this PR by moving the scanning to **after** function the `on_upload_disabled` function. 

As I mentioned in the bug, I was able to reproduce the intermittent reliably in a specific branch of mine, with the fix I don't get the errors anymore on that branch either.

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -187,7 +187,7 @@ impl Glean {
     ///
     /// Importantly, this will not send any pings at startup, since that
     /// sort of management should only happen in the main process.
-    pub fn new_for_subprocess(cfg: &Configuration) -> Result<Self> {
+    pub fn new_for_subprocess(cfg: &Configuration, scan_directories: bool) -> Result<Self> {
         log::info!("Creating new Glean v{}", GLEAN_VERSION);
 
         let application_id = sanitize_application_id(&cfg.application_id);
@@ -205,6 +205,12 @@ impl Glean {
         upload_manager.set_rate_limiter(
             /* seconds per interval */ 60, /* max tasks per interval */ 15,
         );
+
+        // We only scan the pending ping sdirectories when calling this from a subprocess,
+        // when calling this from ::new we need to scan the directories after dealing with the upload state.
+        if scan_directories {
+            let _scanning_thread = upload_manager.scan_pending_pings_directories();
+        }
 
         Ok(Self {
             upload_enabled: cfg.upload_enabled,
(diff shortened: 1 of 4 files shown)


glean-core/src/lib.rs
```diff
@@ -263,6 +262,12 @@ impl Glean {
             }
         }
 
+        // We only scan the pendings pings directories **after** dealing with the upload state.
+        // If upload is disabled, we delete all pending pings files
+        // and we need to do that **before** scanning the pending pings folder
+        // to ensure we don't enqueue pings before their files are deleted.
+        let _scanning_thread = glean.upload_manager.scan_pending_pings_directories();
```
Dexterp37, line 269: Eugh. So this was an actual bug and not a test error. We should add a changelog entry. Do we have any way to estimate how frequently we sent pings out?

I'd suspect that's not super frequent, but it's important to better understand. We could also draft a PSA email, if we find the impact being big.

brizental: > Do we have any way to estimate how frequently we sent pings out?

Hm, I need to think about this. When we have this bug it is usually followed by an error on deleting the ping file. Because we send the request to the caller -> it uploads succesfully (usually) -> and then it tries to delete an already deleted file, which logs an error. But that is all we have, from what I remember on the top of my head.

I'll think a bit further, if you have any ideas, let me know.

brizental: I have an idea to validate this: check if any pings are sent from a client_id after this client_id sent a deletion-request ping. Will come back to this thread with results when I have them.

brizental: https://sql.telemetry.mozilla.org/queries/74658/source

glean-core/src/lib.rs
```diff
@@ -201,12 +201,17 @@ impl Glean {
...
 
         // Create an upload manager with rate limiting of 10 pings every 60 seconds.
-        let mut upload_manager =
-            PingUploadManager::new(&cfg.data_path, &cfg.language_binding_name, false);
+        let mut upload_manager = PingUploadManager::new(&cfg.data_path, &cfg.language_binding_name);
         upload_manager.set_rate_limiter(
             /* seconds per interval */ 60, /* max tasks per interval */ 15,
         );
 
+        // We only scan the pending ping sdirectories when calling this from a subprocess,
```
brizental, line 209: Having this makes me question the name of this function. If you have better ideas I'd be up for changing it.

Dexterp37 (approved):

glean-core/src/lib.rs
travis79, line 209: Oops: `pending ping sdirectories`

PR #1209, opened 2020-09-11
Fix pypi-source-release script

brizental:
Found this small bug while making v32.2.2 release ;)


mdboom (approved):
Thanks for catching this!

mdboom (changes requested):
Actually, I think this isn't right. We want to make a source release from the top-level.  Let me investigate further...

mdboom:
Closing this in favor of #1210, which I think will do the right thing.

PR #1214, opened 2020-09-15
Improve spellchecker output in CI

mdboom:
The spellchecking CI *does* highlight the misspelled word using colorizing when
run locally, but this is somehow broken when displayed in the CircleCI log
viewer online.

There doesn't appear to be a way to put custom delimiters around the misspelled
word, but it is possible to *only* display the misspelled word without the
context from the rest of the line.  This is probably more helpful to understand
which words the spellchecker doesn't like.

Running the spellchecker locally through `make spellcheck` doesn't change --
that will still use the interactive version of aspell.


chutten (approved):

mdboom:
CI failing due to [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1665067)

PR #1215, opened 2020-09-15
Bug 1664196, Bug 1664198 - Fix errors in docs

badboy:


mdboom (approved):
LGTM.  The CI error is probably due to https://bugzilla.mozilla.org/show_bug.cgi?id=1665067.  I'm fine with ignoring that for now.

PR #1217, opened 2020-09-17
Bug 1665041 - Don't backoff on wait and add count limit for pending pings

badboy:


glean-core/src/upload/policy.rs
```diff
@@ -9,6 +9,14 @@ const MAX_WAIT_ATTEMPTS: u32 = 3;
 const MAX_PING_BODY_SIZE: usize = 1024 * 1024; // 1 MB
 const MAX_PENDING_PINGS_DIRECTORY_SIZE: u64 = 10 * 1024 * 1024; // 10MB
 
+// The average number of baseline pings per client (on Fenix) is at 15 pings a day.
+// The P99 value is ~110.
+// With a maximum of (a nice round) 250 we can store about 2 days worth of pings.
```
brizental, line 14: What is the average size of ping files? If Sebastian had more than 5k files and never hit the 10MB limit for ping folder, adding the 250 limit on the number of pings may make that max size almost obsolete, no?

I still haven't gotten to https://bugzilla.mozilla.org/show_bug.cgi?id=1657852, but that should give us some insight into this. 

We can re-evaluate that on a follow up, but I really think we should at some point. Maybe it's just remove that logic altogether.

brizental (approved):
LGTM!

glean-core/src/upload/policy.rs
badboy: Uh, you're right. I should have done the calculations. The baseline pings I have here from my tests are ~600 bytes.
At 10 MB for the directory that's 16k pings, so yeah, we never reach that.

Now metrics and events pings are most likely bigger, a reevaluation after your analysis would be good.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -51,6 +52,7 @@ internal inline fun <reified W : Worker> buildWorkRequest(tag: String): OneTimeW
 class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(context, params) {
     companion object {
         internal const val PING_WORKER_TAG = "mozac_service_glean_ping_upload_worker"
+        internal const val THROTTLED_BACKOFF_MILLIS = 60_000L
```
travis79, line 55: uber nit: I think most of the other places we refer to milliseconds in a variable name using `MS`

travis79 (approved):
This looks good to me!  Let's go ahead and file a bug to revisit the directory size limit after Bea's analysis.  Thanks for doing this!

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
badboy: True, not many places, but enough to convince me (google on the other hand uses `_MILLIS`)

glean-core/src/upload/policy.rs
badboy: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1665653

PR #1222, opened 2020-09-22
Use Kotlin 1.4.10

Dexterp37:
Nothing surprising, but see the individual commits for more info.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v33.0.0...main)
 
+* Android
+  * Update the Kotlin Gradle Plugin to version 1.4.10.
+
 # v33.0.0 (2020-09-22)
 
 [Full changelog](https://github.com/mozilla/glean/compare/v32.4.0...v33.0.0)
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
@@ -2,7 +2,7 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
 * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
-/* This file is based on the tests in the Glean android-components implentation.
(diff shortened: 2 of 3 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v33.0.0...main)
 
+* Android
+  * Update the Kotlin Gradle Plugin to version 1.4.10.
```
badboy, line 6: For my own understanding: this means we compile and test against that version of Kotlin and don't promise any lower Kotlin version still works.
But consumers _could_ compile their own code against other Kotlin versions and still rely on us?

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -450,12 +450,12 @@ class EventMetricTypeTest {
         // Trigger worker task to upload the pings in the background
         triggerWorkManager(context)
 
-        var request = server.takeRequest(20L, TimeUnit.SECONDS)
-        var docType = request.path.split("/")[3]
+        val request = server.takeRequest(20L, TimeUnit.SECONDS)
+        val docType = request.path.split("/")[3]
```
badboy, line 454: what's the reason for this? no more mutability needed? or it was never needed, but it didn't produce warning/errors?

badboy (approved):
No blocker, just a couple of questions.

CHANGELOG.md
Dexterp37: Yes, that's my understanding as well

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
Dexterp37: It was never needed, for some reason it did not produce a warning at compile time, but it produced warnings within Android Studio

CHANGELOG.md
Dexterp37: I double checked with @pocmo and he confirmed this. This is also discussed in the [Kotlin Components Stability](https://kotlinlang.org/docs/reference/evolution/components-stability.html) page!

PR #1227, opened 2020-09-24
[ci full] Change Dockerfile to force a rebuild of the Docker image

badboy:


badboy:
No docker changes needed. #1229 looks like the solution.

PR #1230, opened 2020-09-25
Document the special ci-full behavior

badboy:
[doc only]

### DIFF
--- a/docs/dev/ci.md
+++ b/docs/dev/ci.md
@@ -78,7 +78,7 @@ On release the full TaskCluster suite takes up to 30 minutes to run.
 4. Find the "Decision task" and click "Details"
 5. From the list of tasks pick the one you want to look at and follow the "View task in Taskcluster" link.
 
-## Special keywords
+## Special behavior
 
 ### Documentation-only changes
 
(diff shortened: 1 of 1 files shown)


docs/dev/ci.md
```diff
@@ -103,6 +103,23 @@ To skip tasks on TaskCluster add the following to the title of your pull request
 
 This should only be used for metadata files, such as those in `.github`, `LICENSE` or `CODE_OF_CONDUCT.md`.
 
+### Release-like Android builds
+
+Release builds are done on [TaskCluster].
+As it builds for multiple platforms this task is quite long and does not run on pull requests by default.
+
+It is possible trigger the task on pull requests.
+Add the following literal string to the pull request **title**:
```
brizental, line 112: Interesting, I always thought these flags should be added to the commit message or body. Or is that the case for `[doc only]` but not the case for `[ci full]`?

brizental (approved):
LGTM!

docs/dev/ci.md
badboy: It's true for all the others (because CircleCI looks at commit messages).
But TaskCluster is (currently) built to look at the PR.

We _could_ unify that at some point, but for now we copied this behavior from others and didn't adjust it.

PR #1236, opened 2020-09-28
Bug 1657852 - Transform ping dir size to kB before accumulating

brizental:
Found this bug during vlidation of these metrics. 

See: https://sql.telemetry.mozilla.org/queries/75022/source?p_date=2020-09-16--2020-09-28#187737

### DIFF
--- a/glean-core/src/upload/mod.rs
+++ b/glean-core/src/upload/mod.rs
@@ -399,7 +399,7 @@ impl PingUploadManager {
             cached_pings.pending_pings.reverse();
             self.upload_metrics
                 .pending_pings_directory_size
-                .accumulate(glean, pending_pings_directory_size as u64);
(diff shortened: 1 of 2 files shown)


glean-core/src/upload/mod.rs
```diff
@@ -399,7 +399,7 @@ impl PingUploadManager {
             cached_pings.pending_pings.reverse();
             self.upload_metrics
                 .pending_pings_directory_size
-                .accumulate(glean, pending_pings_directory_size as u64);
+                .accumulate(glean, pending_pings_directory_size as u64 / 1024);
```
Dexterp37, line 402: Let's add a changelog entry, please.

Dexterp37 (approved):
r+wc. I wish the "note" at the bottom of [this page](https://mozilla.github.io/glean/book/user/metrics/memory_distribution.html) was at the top :)

PR #1238, opened 2020-09-29
Split book into more explicit groups (and upgrade mdbook to make it happen)

badboy:
Old table of contents:

<details>

image

image

</details>

New table of contents:

<details>

image

image

</details>


Dexterp37 (approved):
Nice one!

PR #1240, opened 2020-09-30
Bug 1661240 - Have same throttle backoff time for all bindings

brizental:


glean-core/python/glean/glean.py
```diff
@@ -230,6 +230,9 @@ def _initialize_with_tempdir_for_testing(
         The temporary directory will be destroyed when Glean is initialized
         again or at process shutdown.
         """
+        # Lower throttle backoff time for testing
+        PingUploadWorker._throttle_backoff_time = 1
```
badboy, line 234: Why is Python the only one where we lower this value?

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -108,6 +109,7 @@ public class HttpPingUploader {
                     glean_process_ping_upload_response(&incomingTask, result.toFfi())
                 }
             case .wait:
+                sleep(Constants.throttleBackoffMs)
```
badboy, line 112: huh, I just looked into the code for iOS and we call `HttpPingUploader.process` synchronously in the task of the `initialize` function.
This means we essentially _block_ full initialization until we uploaded all pings. Now previously we were probably lucky in that if we're throttled we quickly get to 3 waits in a row and bail out, but now this could lead to blocking Glean initialization for an extended amount of time.

I think we need to look into that and ensure we're not doing that.
The intended behavior should be the same across platforms:
We launch a background worker to upload what's there. On Android we already do that.

badboy (changes requested):
See inline comment, I think this introduces a problem on iOS (well, the problem was already there, this would just amplify it)

glean-core/python/glean/glean.py
brizental: Because it is the only one that takes a long time to run the test suite with this change.

glean-core/ios/Glean/Net/HttpPingUploader.swift
brizental: Okay, I opened a bug about this and will leave this PR blocked for now, until that is fixed.

brizental: The bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1668820

badboy:
Can you rebase that on latest `main`? then we can trigger the iOS CI jobs again

badboy (approved):
Can you add a changelog entry, please?
Otherwise this is now good to go.

PR #1247, opened 2020-10-05
[ci full] Bug 1667410 - Check the contents of produced Maven artifacts

badboy:
[doc only]

### DIFF
--- a/bin/check-artifact.sh
+++ b/bin/check-artifact.sh
@@ -1,0 +1,71 @@
+#!/usr/bin/env bash
+#
+# A simple check that our Maven release files contain the shared libraries as expected.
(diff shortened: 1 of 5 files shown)


brizental (approved):
r+wc

bin/check-artifact.sh
```diff
@@ -0,0 +1,71 @@
+#!/usr/bin/env bash
+#
+# A simple check that our release .aar files are a reasonable size.
```
brizental, line 3: I am not the most well versed in shell script, but isn't this script just checking that the required files _exist_? Or is there a size check that I missed?

badboy: Woops, the comment there is wrong. We only check for the contents, not the size

PR #1252, opened 2020-10-09
Bug 1668820 - Always launch the ping uploader on a background thread

badboy:
The uploader is always started after pings are already submitted (and assembled).
The ping uploader can potentially block as it uploads several pings and
even goes to sleep if told to wait (soon).
By putting it on a background thread we ensure this doesn't stop any
other work.

This was already manually done in one place, but not in the others.
This should be equivalent to Android's way of pushing this work to a
workmanager job.

Unfortunately this is near impossible to add a test for, as we would somehow need to instruct the uploader to block so that we can encounter the problems. Which we can't.
So we have to live with the existing tests (they pass) & integration tests (on CI).

---

This should unblock #1240


brizental (approved):
Looks good to me.

travis79 (approved):

PR #1259, opened 2020-10-12
Don't update homebrew to install swiftlint

badboy:
It is super slow and at the moment causes quite some failures due to
timeouts or other network problems.
The image should get us a suitable version of swiftlint already without
requiring an update.

[doc only]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -468,13 +468,9 @@ jobs:
     steps:
       - checkout
       - run:
-          name: Update brew
-          command: |
-            # we want the latest versions of the tools
-            brew update
-      - run:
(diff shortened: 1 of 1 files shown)


Dexterp37 (approved):

Dexterp37:
(please run iOS test before merging)

badboy:
> (please run iOS test before merging)

This only affects the `Check Swift formatting` task and [that one](https://app.circleci.com/jobs/github/mozilla/glean/98942) is green (and super-fast with just 23s runtime).

We don't require installing new tools for the iOS tests, as everything we need comes pre-bundled with the image.

PR #1260, opened 2020-10-12
Bug 1670184 - Upgrade to rkv 0.15.0 and make the backend selectable

badboy:
The default is still Rkv's LMDB storage backend and we don't have any
intention to change that at the moment.

### DIFF
--- a/glean-core/Cargo.toml
+++ b/glean-core/Cargo.toml
@@ -27,7 +27,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.44"
-rkv = "0.10.3"
+rkv = "0.15.0"
 bincode = "1.3.1"
 log = "0.4.8"
 uuid = { version = "0.8.1", features = ["v4"] }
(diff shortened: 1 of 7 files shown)


glean-core/src/database/mod.rs
```diff
@@ -81,7 +110,12 @@ impl Database {
         let path = Path::new(data_path).join("db");
         log::debug!("Database path: {:?}", path.display());
 
+        // FIXME: This is probably more knowledge than we should have about the database internals.
+        // We could instead iterate over the directory and sum up the file sizes.
```
Dexterp37, line 114: This seems like a reasonable fix. Would you kindly file a bug and add the number here?

glean-core/Cargo.toml
```diff
@@ -41,3 +41,7 @@ env_logger = { version = "0.7.1", default-features = false, features = ["termcol
 tempfile = "3.1.0"
 iso8601 = "0.4"
 ctor = "0.1.12"
+
+[features]
+# Enable the "safe-mode" Rust storage backend instead of the default LMDB one.
+rkv-safe-mode = []
```
Dexterp37, line 47: Does this mean the new glean-core version, once we cut a release, will use the new storage system? What happens with the legacy files? Do we need migration?

badboy: No, it explicitly means we're staying on LMDB.

This feature allows us to toggle it on if we want to (which means we can more easily test this)

glean-core/src/database/mod.rs
badboy: yup, will do

Dexterp37 (approved):

PR #1262, opened 2020-10-15
Bug 1671192 - Document ping info in metrics yaml

badboy:
All of these "metrics" land in the `ping_info` section.
Internally this is considered the `glean_internal_info` ping.

One note: `experiments` is actually a key-value dictionary.
This does not exist as a metric type, we therefore label it a `string`.
The pipeline won't like this when it generates a schema purely based on
what's annoated in metrics.yaml

@fbertsch Will this cause trouble for the schema generator?

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -238,6 +238,78 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+  seq:
+    type: string
+    lifetime: ping
+    send_in_pings:
+      - glean_internal_info
+    description: |
+      A running counter of the number of times pings of this type have been sent
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1556964
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1512938#c3
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+
+  experiments:
+    type: string  # It's not, but we don't have a way to specify its type
+    lifetime: ping
+    send_in_pings:
+      - glean_internal_info
+    description: |
+      Optional. A dictionary of active experiments.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1552471
+    data_reviews:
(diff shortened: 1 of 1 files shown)


glean-core/metrics.yaml
```diff
@@ -238,8 +238,83 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
-  ping_reason:
+  seq:
     type: string
+    lifetime: ping
+    send_in_pings:
+      - glean_internal_info
```
mdboom, line 245: It's too bad we have `glean_internal_info` and couldn't call it `glean_ping_info` (as we have for `glean_client_info`).  Is that enough of an implementation detail that we could change it without too much pain?  That would make things more obvious for users over at the probe_scraper / Glean dictionary end of things...

badboy: We do persist `experiments` using this. The `experiments` annotation is `application` only so we _could_ migrate away from that over to `glean_ping_info` with relative ease. I'd rather do that in a follow-up though.

badboy: I spoke to soon: we also store the start_time with a `user` lifetime, so it's a bit more involved.

mdboom (approved):

fbertsch (commented):
@badboy aren't labeled strings stored as k/v arrays? Could we use that type here for experiments?

badboy:
> @badboy aren't labeled strings stored as k/v arrays? Could we use that type here for experiments?

the experiments annotation is a nested object, see https://mozilla.github.io/glean/book/user/pings/index.html#the-experiments-object

badboy:
@fbertsch what do you think? I can leave out the `experiments` part for now and at least we get the metric docs for the other parts for now.

fbertsch:
@badboy I see two options here:
1. We use `string` as the type, and MSG handles this field as a special case when creating schemas. It would be a bit confusing that we're changing the type during schema generation, but then we don't have to change the SDK
2. We add an internal type to the SDK that supports this. We would still need to figure out how to find out this type during schema generation.

I'm leaning towards 1. right now. If you agree, then let's merge this as-is.

badboy:
I'm leaning towards 1.
An "internal" type requires glean_parser to know it. We can still go that route afterwards and eventually remove special-casing it in MSG.

PR #1264, opened 2020-10-16
Bug 1671628 - Add traits to describe expected operations on Glean types

Dexterp37:
This exposes the description of the metric types in glean_core using Rust traits. All the traits live in the `traits` module.

### DIFF
--- a/glean-core/src/metrics/boolean.rs
+++ b/glean-core/src/metrics/boolean.rs
@@ -26,6 +26,10 @@ impl MetricType for BooleanMetric {
     }
 }
 
+// IMPORTANT:
+//
+// When changing this implementation, make sure all the operations are
+// also declared in the related trait in `../traits/`.
 impl BooleanMetric {
     /// Creates a new boolean metric.
     pub fn new(meta: CommonMetricData) -> Self {
--- a/glean-core/src/metrics/custom_distribution.rs
+++ b/glean-core/src/metrics/custom_distribution.rs
@@ -41,6 +41,10 @@ impl MetricType for CustomDistributionMetric {
     }
 }
 
+// IMPORTANT:
+//
+// When changing this implementation, make sure all the operations are
+// also declared in the related trait in `../traits/`.
 impl CustomDistributionMetric {
     /// Creates a new memory distribution metric.
     pub fn new(
--- a/glean-core/src/metrics/mod.rs
+++ b/glean-core/src/metrics/mod.rs
@@ -38,11 +38,11 @@ use crate::Glean;
 
 pub use self::boolean::BooleanMetric;
 pub use self::counter::CounterMetric;
+pub use self::custom_distribution::CustomDistributionMetric;
 pub use self::datetime::DatetimeMetric;
 pub use self::event::EventMetric;
 pub(crate) use self::experiment::ExperimentMetric;
 pub use crate::histogram::HistogramType;
-pub use self::custom_distribution::CustomDistributionMetric;
 // Note: only expose RecordedExperimentData to tests in
 // the next line, so that glean-core\src\lib.rs won't fail to build.
 #[cfg(test)]
--- a/glean-core/src/metrics/mod.rs
+++ b/glean-core/src/metrics/mod.rs
@@ -42,9 +42,9 @@ pub use self::datetime::DatetimeMetric;
 pub use self::event::EventMetric;
 pub(crate) use self::experiment::ExperimentMetric;
 pub use crate::histogram::HistogramType;
+pub use self::custom_distribution::CustomDistributionMetric;
 // Note: only expose RecordedExperimentData to tests in
 // the next line, so that glean-core\src\lib.rs won't fail to build.
-pub use self::custom_distribution::CustomDistributionMetric;
 #[cfg(test)]
 pub(crate) use self::experiment::RecordedExperimentData;
 pub use self::jwe::JweMetric;
--- a/glean-core/src/traits/boolean.rs
+++ b/glean-core/src/traits/boolean.rs
@@ -1,0 +1,23 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+/// A description for the `BooleanMetric` type.
+///
+/// When changing this trait, make sure all the operations are
+/// implemented in the related type in `../metrics/`.
+pub trait Boolean {
+    /// Sets to the specified boolean value.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - the value to set.
+    fn set(&self, value: bool);
+
+    /// **Exported for test purposes.**
+    ///
+    /// Gets the currently stored value as a boolean.
+    ///
+    /// This doesn't clear the stored value.
+    fn test_get_value(&self, storage_name: &str) -> Option<bool>;
+}
--- a/glean-core/src/traits/event.rs
+++ b/glean-core/src/traits/event.rs
@@ -1,0 +1,44 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::collections::HashMap;
+
+use crate::event_database::RecordedEvent;
+
+/// A description for the `EventMetric` type.
+///
+/// When changing this trait, make sure all the operations are
+/// implemented in the related type in `../metrics/`.
+pub trait Event {
+    /// Records an event.
+    ///
+    /// # Arguments
+    ///
+    /// * `extra` - A HashMap of (key, value) pairs. The key is an index into
+    ///   the metric's `allowed_extra_keys` vector where the key's string is
+    ///   looked up. If any key index is out of range, an error is reported and
+    ///   no event is recorded.
+    fn record<M: Into<Option<HashMap<i32, String>>>>(&self, extra: M);
+
+    /// **Exported for test purposes.**
+    ///
+    /// Tests whether there are currently stored events for this event metric.
+    ///
+    /// This doesn't clear the stored value.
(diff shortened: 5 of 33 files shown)


glean-core/src/metrics/mod.rs
```diff
@@ -31,37 +31,36 @@ mod uuid;
...
 };
-pub use self::memory_distribution::MemoryDistributionMetric;
+pub use self::memory_distribution::{MemoryDistribution, MemoryDistributionMetric};
 pub use self::memory_unit::MemoryUnit;
 pub use self::ping::PingType;
-pub use self::quantity::QuantityMetric;
-pub use self::string::StringMetric;
-pub use self::string_list::StringListMetric;
+pub use self::quantity::{Quantity, QuantityMetric};
+pub use self::string::{/*String, THIS FAILS HARD TODO */ StringMetric};
```
Dexterp37, line 58: Now, if we retain the `String` name for the trait (or for `Datetime`, or `Uuid`) then we start colliding with type names from other crates and that makes the build blow up.

Bikeshedding time:

- `StringContract`?
- `StringOperations`?

@badboy @chutten ... thoughts?

mdboom: It's verbose and Java-ey, but maybe `StringMetricTrait`?  `StringMetricRecorder`?

Dexterp37: > 
> 
> It's verbose and Java-ey, but maybe `StringMetricTrait`? `StringMetricRecorder`?

Jan-Erik vetoed the `Trait` suffix (for good reason, and I agree!). @mdboom what's the idea behind "recorder"?

mdboom: It's the interface that allows "recording" to a metric...  (Just grasping for something...)

mdboom: Thinking my thoughts on this, I think I like having `Metric` in it, since it really has to do with "string metrics" not "strings" in general.  But still don't have a good suggestion for the suffix.

There is the convention to put an `I` in front of interfaces, but that's obviously deeply not Rust like.

Dexterp37: > 
> 
> It's the interface that allows "recording" to a metric... (Just grasping for something...)

Oh! That's a good point. I'm up for a `StringRecorder` then! Let's see what others think.

glean-core/src/metrics/boolean.rs
```diff
@@ -16,6 +16,26 @@ pub struct BooleanMetric {
     meta: CommonMetricData,
 }
 
+/// A description for the `BooleanMetric` type.
+pub trait Boolean {
+    /// Creates a new instance of the metric.
+    fn new(meta: CommonMetricData) -> Self;
```
badboy, line 22: Do we really need to expose `new` as a trait method? This is not necessarily "metric type behaviour"

glean-core/src/metrics/custom_distribution.rs
```diff
@@ -21,6 +21,49 @@ pub struct CustomDistributionMetric {
...
+        bucket_count: u64,
+        histogram_type: HistogramType,
+    ) -> Self;
+
+    /// Accumulates the provided signed samples in the metric.
+    ///
+    /// This is required so that the platform-specific code can provide us with
+    /// 64 bit signed integers if no `u64` comparable type is available. This
+    /// will take care of filtering and reporting errors for any provided negative
+    /// sample.
```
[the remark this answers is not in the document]
badboy: It might need to stay with us to make FOG's IPC work efficiently.

glean-core/src/metrics/mod.rs
badboy: I think we should use modules to differentiate between metric types, their traits (& libstd's types).

That way `glean_core::metrics::String` can be the type, `glean_core::traits::String` can be the trait.
In code we can `use glean_core::metrics` and `glean_core::traits` and then use the more-qualified path to the type like `metrics::String` instead. That avoids awkward (and redudant) suffixes on type or trait names.

badboy: (This would mean we create a new directory `glean-core/src/traits/`, add `mod.rs` there and several new `counter.rs`, `string.rs`, etc. Makes looking for just a file name a bit harder, but at least I can still see the path in my editor, so can differentiate them if I open both the metric type and trait.(

badboy (changes requested):
Haven't looked closely at each metric trait, but I assume the methods are just copied from the actual implementation.

Left inline comments for the controversial stuff.

glean-core/src/metrics/boolean.rs
Dexterp37: You're right, we don't need it.

Dexterp37: Thinking of this again, one reason to define `new`: think of [custom distributions](https://searchfox.org/glean/rev/02b67f683104ceff34c24936525f496195b3abbe/glean-core/src/metrics/custom_distribution.rs#46-51) for example, they expect a specific set of parameters to be passed. If we bubbled up these expectations to the trait, then the "glean_parser" would always know what to expect, because we'd define this on the trait.

On the other hand, we control the glean_parser. So it's not too much of a big deal.

glean-core/src/traits/boolean.rs
```diff
@@ -0,0 +1,23 @@
...
+/// implemented in the related type in `../metrics/`.
+pub trait Boolean {
+    /// Sets to the specified boolean value.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - the value to set.
+    fn set(&self, value: bool);
+
+    /// **Test-only API (exported for FFI purposes).**
```
badboy, line 17: The FFI part is not true anymore. This is becoming a general "exported for test purposes" part.

badboy: (applies to all traits)

glean-core/src/traits/event.rs
```diff
@@ -0,0 +1,47 @@
...
+    /// # Arguments
+    ///
+    /// * `timestamp` - A monotonically increasing timestamp, in milliseconds.
+    ///   This must be provided since the actual recording of the event may
+    ///   happen some time later than the moment the event occurred.
+    /// * `extra` - A HashMap of (key, value) pairs. The key is an index into
+    ///   the metric's `allowed_extra_keys` vector where the key's string is
+    ///   looked up. If any key index is out of range, an error is reported and
+    ///   no event is recorded.
+    fn record<M: Into<Option<HashMap<i32, String>>>>(&self, timestamp: u64, extra: M);
```
badboy, line 25: Heh, after a FOG review I'm actually tasked to file a bug to get this changed. Wouldn't it be great if we don't need to take i32 but can take the enum instead? :) 

I'm also not super happy that we need to expose the `timestamp` argument here, but we can approach that in a follow-up eventually

glean-core/src/traits/jwe.rs
```diff
@@ -0,0 +1,41 @@
...
+    /// Builds a JWE value from its elements and set to it.
+    ///
+    /// # Arguments
+    ///
+    /// * `header` - the JWE Protected Header element.
+    /// * `key` - the JWE Encrypted Key element.
+    /// * `init_vector` - the JWE Initialization Vector element.
+    /// * `cipher_text` - the JWE Ciphertext element.
+    /// * `auth_tag` - the JWE Authentication Tag element.
+    fn set<S: Into<String>>(&self, header: S, key: S, init_vector: S, cipher_text: S, auth_tag: S);
```
badboy, line 26: Uh oh, this actually means all parameters need to be of the same type.
Not a problem for our current use, but we should consider changing that.
(actually a nicer Rust-y API would probably to take a `JweToken` struct or something, I'll file a bug here as well)

glean-core/src/traits/mod.rs
```diff
@@ -0,0 +1,40 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! The different metric types supported by the Glean SDK to handle data.
```
badboy, line 5: Can we call out that they are considered `unstable` / `experimental` right now?
I have some ideas how we might improve them to better accomodate for a nice Rust-y API, so I wouldn't want us to have them be considered stable yet (and yes, I know, no one should be using glean-core, but this also acts as docs for us)

(In _theory_ we should hide them behind a feature flag, but that will make integration in this repository with the RLB harder)

badboy (approved):
r+wc

some notes on follow-ups that I'd like to look us into, but this can (nearly) land now.

glean-core/src/traits/event.rs
Dexterp37: Oh wait, that's a mistake, we don't need to expose `timestamp`.

glean-core/src/traits/mod.rs
Dexterp37: Sure! Let me flag these as unstable / experimental here.

glean-core/src/traits/event.rs
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1672179

chutten: Unfortunately I think we have to expose timestamp since we need to providing timing from other processes somehow.

badboy: re timestamp: we kinda do need that API.
For FOG's IPC we need to get the timestamp immediately, pass that through the IPC layer and then later record the full event.
What we could do though is having a special\* `record_with_timestamp` API on the RLB only, not on the trait

\* I already dislike the "special" API, but I don't see a way around it

glean-core/src/traits/jwe.rs
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1672183

glean-core/src/traits/event.rs
Dexterp37: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1672198

PR #1267, opened 2020-10-19
Dependabot updates

badboy:
Dependabot should be able to handle the cargo workspace as one thing.

Otherwise I just ran `cargo update` to upgrade dependencies with minor/patch version releases.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -1,23 +1,8 @@
 version: 2
 updates:
-  # glean-core
+  # glean
   - package-ecosystem: "cargo"
-    directory: "/glean-core"
-    schedule:
-      interval: "daily"
-    ignore:
-      # Pinned to old version due to breaking changes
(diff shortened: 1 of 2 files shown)


.github/dependabot.yml
```diff
@@ -1,23 +1,8 @@
...
-  # glean-ffi
-  - package-ecosystem: "cargo"
-    directory: "/glean-core/ffi"
-    schedule:
-      interval: "daily"
-
-  # glean benchmark
-  - package-ecosystem: "cargo"
-    directory: "/glean-core/benchmark"
+    directory: "/"
```
brizental, line 5: Is it intentional that you do not ignore rkv anymore?

brizental (approved):
r+wc

.github/dependabot.yml
badboy: Yes! We recently upgraded it (#1260)

PR #1270, opened 2020-10-20
Bug 1671915 - Change the 'preview' crate to the 'rlb' crate

Dexterp37:
This requires #1264 to be merged first (and to be rebased after that happens)

This lays the foundations to develop the Rust Language Bindings: it renames the 'preview' crate, which is no longer useful, and makes sure we properly update the version numbers when cutting new releases.


docs/dev/cut-a-new-release.md
```diff
@@ -87,6 +87,8 @@ When CI has finished and is green for your specific release branch, you are read
     cargo publish --verbose
     cd ffi
     cargo publish --verbose
+    cd.. && cd rlb
```
badboy, line 90:
```suggestion
    cd .. && cd rlb
```

Gnaaaah, I'm going to automate this away in the coming weeks.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1598291

bin/prepare-release.sh
```diff
@@ -102,21 +102,13 @@ run $SED -i.bak -E \
     "${WORKSPACE_ROOT}/${FILE}"
 run rm "${WORKSPACE_ROOT}/${FILE}.bak"
 
-### GLEAN-PREVIEW ###
+### GLEAN RLB ###
 
 # Update the version of the glean-core dependency
 
-FILE=glean-core/preview/Cargo.toml
```
badboy, line 109: glean-preview was versioned independently of glean-core (it's at 0.0.5).
The code previously here ensured we update the glean-core version it depends on.

This needs changes to update both the glean-core version it depends on and its own version.

glean-core/rlb/Cargo.toml
```diff
@@ -1,8 +1,8 @@
 [package]
-name = "glean-preview"
+name = "glean"
 version = "0.0.5"
```
badboy, line 3: As said above, do we immediately bump this to the same version as glean-core (33.y.z)?

glean-core/rlb/Cargo.toml
```diff
@@ -1,8 +1,8 @@
...
 authors = ["Jan-Erik Rediger <jrediger@mozilla.com>", "The Glean Team <glean-team@mozilla.com>"]
-description = "Nice Glean SDK Rust API"
+description = "Glean SDK Rust language bindings"
```
badboy, line 5: I can't comment on a moved-only file, but the glean-core/rlb/README.md needs changes too

badboy (changes requested):

bin/prepare-release.sh
Dexterp37: Yes, this is what the file is doing now: i tested it locally and it updates both the glean-core version and its own version.

badboy: uh fancy, I can now see why. it's both a line with just `version = ...`

glean-core/rlb/Cargo.toml
Dexterp37: Done!

badboy (approved):
r+wc

The `glean-core/rlb/CHANGELOG.md` file can be dropped now. going forward we will document changes in the top-level changelog.

Otherwise good to go.

PR #1271, opened 2020-10-21
Bug 1672351 - Add a `BooleanMetric` to the RLB

Dexterp37:
This requires #1270 and to be rebased once that's merged.

The PR implements a basic `BooleanMetric` and exposes it in the RLB.

It does not add the dispatcher (will take this as a separate PR) nor any test (all the unit tests are covering the metric in `glean_core` already!).


glean-core/rlb/src/private/boolean.rs
```diff
@@ -0,0 +1,49 @@
...
+#[inherent(pub)]
+impl glean_core::traits::Boolean for BooleanMetric {
+    /// Sets to the specified boolean value.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - the value to set.
+    fn set(&self, value: bool) {
+        crate::with_glean(|glean| {
+            // Important note: this will need to use the
```
Dexterp37, line 35: I will take the dispatcher as a separate PR, unless there's a strong reason to do it here. I wanted to make this PR as small as possible to help reviewers!

badboy (approved):
Yup, I think this works. (and if not ... well, this is alpha software, we'll fix it)

glean-core/rlb/src/private/boolean.rs
Dexterp37: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1672371

PR #1273, opened 2020-10-21
Bug 1672371 - Copy the dispatcher from mozilla-central to the RLB

Dexterp37:
This imports the 'dispatcher' module revision from [mozilla-central](https://hg.mozilla.org/mozilla-central/rev/fbe0ea62f4bb50bfc5879a56667945697b2c90e7).

It additionally adds tests (some of them came from @badboy 's stash).


chutten (approved):
Definitely needs tests. Apparently I asked about tests during the impl and didn't follow-up on that : |

Dexterp37:
> 
> 
> Definitely needs tests. Apparently I asked about tests during the impl and didn't follow-up on that : |

No worries! I'll take care of it in this PR.

glean-core/rlb/src/dispatcher/mod.rs
```diff
@@ -247,3 +273,288 @@ impl Dispatcher {
...
+        }
+
+        // Ensure that no task ran.
+        assert_eq!(0, thread_canary.load(Ordering::SeqCst));
+
+        // Flush the queue and wait for the tasks to complete.
+        dispatcher
+            .flush_init()
+            .expect("Failed to get out of preinit queue mode");
+        dispatcher.block_on_queue();
```
chutten, line 348: This shouldn't be necessary because flush_init blocks on flushing the preinit operations, right?

glean-core/rlb/src/dispatcher/mod.rs
```diff
@@ -247,3 +273,288 @@ impl Dispatcher {
...
+        // Validate that we have the expected canary value.
+        assert_eq!(3, thread_canary.load(Ordering::SeqCst));
+    }
+
+    #[test]
+    fn queued_tasks_are_flushed_off_the_main_thread() {
```
chutten, line 354: Is this test the same as the above test except it also has a "this thread isn't the main thread" assert? If so, I think we can combine them.

glean-core/rlb/src/dispatcher/mod.rs
```diff
@@ -247,3 +273,288 @@ impl Dispatcher {
...
+                .unwrap();
+        }
+
+        dispatcher.block_on_queue();
+
+        assert_eq!(&*result.lock().unwrap(), &[1, 2, 3, 4, 5]);
+    }
+
+    #[test]
+    fn preinit_tasks_are_processed_after_flush() {
```
chutten, line 412: This test is a superset of the above test, is it not? We can probably do without the smaller test.

glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -74,3 +73,47 @@ pub fn flush_init() -> Result<(), DispatchError> {
...
+    use super::*;
+
+    // We can only test this once, as it is a global resource which we can't reset.
+    #[test]
+    fn global_fills_up_in_order_and_works() {
+        let _ = env_logger::builder().is_test(true).try_init();
+
+        let result = Arc::new(Mutex::new(vec![]));
+
+        for i in 1..=100 {
```
chutten, line 90: Should use `GLOBAL_DISPATCHER_LIMIT` here to be clear.

glean-core/rlb/Cargo.toml
```diff
@@ -25,12 +25,14 @@ path = ".."
 version = "33.0.4"
 
 [dependencies]
+crossbeam-channel = "0.4.3"
```
badboy, line 28: We're landing updates in #1279. Let's rebase on that after and go up to v0.5.0

badboy (approved):
r+wc -- what chutten said, and then the rebase once we updated dependencies.

PR #1283, opened 2020-10-23
Bug 1672984 - Recoverable network error is noisily logged at error level

travis79:
Reduce Sentry noise for Fx-iOS by reducing the logging level of the recoverable network errors in the upload module.


badboy (approved):
I could have sworn there was another bug we recently filed on looking into dropping some log levels here.

PR #1284, opened 2020-10-23
Bug 1672440 - Add a basic General API for the RLB

Dexterp37:
This does a bunch of things:

- Makes the RLB init async (by dispatching the init task).
- Changes `set_upload_enabled` to dispatch the state flip as well.
- Stubs out the full Glean initialization.


glean-core/rlb/src/glean_metrics.rs
```diff
@@ -0,0 +1,10 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+// ** IMPORTANT **
+//
+// This file is *temporary*, it will be generated by 'glean-parser'
```
badboy, line 7: Only sort of. We generate the `pings.rs`, we still need to include it somewhere.
So `glean_metrics.rs` won't go away fully.

glean-core/rlb/src/lib.rs
```diff
@@ -105,29 +112,139 @@ where
...
+            // Register builtin pings.
+            // Unfortunately we need to manually list them here to guarantee
+            // they are registered synchronously before we need them.
+            // We don't need to handle the deletion-request ping. It's never touched
+            // from the language implementation.
+            glean.register_ping_type(&glean_metrics::pings::baseline.ping_type);
+            glean.register_ping_type(&glean_metrics::pings::metrics.ping_type);
+            glean.register_ping_type(&glean_metrics::pings::events.ping_type);
+            /*
+                        // If any pings were registered before initializing, do so now.
```
badboy, line 181: Better turn this into a proper `TODO` comment instead of copy-pasting kotlin code.

glean-core/rlb/src/lib.rs
```diff
@@ -105,29 +112,139 @@ where
...
+            // Set up information and scheduling for Glean owned pings. Ideally, the "metrics"
+            // ping startup check should be performed before any other ping, since it relies
+            // on being dispatched to the API context before any other metric.
+            // TODO: start the metrics ping scheduler, will happen in bug 1672951.
+
+            // Check if the "dirty flag" is set. That means the product was probably
+            // force-closed. If that's the case, submit a 'baseline' ping with the
+            // reason "dirty_startup". We only do that from the second run.
+            if !is_first_run && dirty_flag {
+                // TODO: submitPingByNameSync("baseline", "dirty_startup")
```
badboy, line 218: Same here, is there a bug for it?

glean-core/rlb/src/lib.rs
```diff
@@ -196,7 +329,7 @@ pub fn register_ping_type(ping: &private::PingType) {
 /// # Returns
```
badboy, line 329: This isn't correct anymore now.

glean-core/rlb/src/lib.rs
```diff
@@ -207,8 +340,12 @@ pub fn submit_ping(ping: &private::PingType, reason: Option<&str>) -> bool {
...
```
badboy, line 340: As above, we're removing the return value.

glean-core/rlb/src/private/ping.rs
```diff
@@ -42,7 +42,7 @@ impl PingType {
     /// # Returns
```
badboy, line 42: As above

badboy: though actually ... we should keep the return value here.
We should just strip it out in the functions above where we dispatch it.

reason: we need to know if something is successfully submitted so we can actually trigger an upload.

badboy (changes requested):

glean-core/rlb/src/private/ping.rs
Dexterp37: We can still handle as we do for other bindings (within the upload mechanism / the glean singleton)

glean-core/rlb/src/lib.rs
Dexterp37: Yup, sorry, I just added another commit that links to the related bugs.

Dexterp37:
Note that the failure is an intermittent related to bug https://bugzilla.mozilla.org/show_bug.cgi?id=1672722

glean-core/rlb/src/private/ping.rs
badboy: Right, let's fix this up the way we need it when we need it.

badboy (approved):
r+!

I won't find time to look into the intermittent today but I take this on tomorrow.

glean-core/rlb/src/private/ping.rs
Dexterp37: Cool, it will happen as part of https://bugzilla.mozilla.org/show_bug.cgi?id=1672958, which I'm taking next.

PR #1290, opened 2020-10-29
Add library-loading configuration hooks for megazording.

rfk:
The application-services team is experimenting with including Glean
as part of their "megazord" library, a collection of Rust components
that are all compiled together into a single unit. See the PR for
more details:

  https://github.com/mozilla/application-services/pull/3669

This commit adds some library-loading configuration hooks so that Glean
can choose between loading either the standalone `libglean_ffi.so` or
the combined appservices megazord, based on build-time config.

I'm pushing this as a PR as well for early visibility. Feedback most welcome
of course, but this will stay squarely in "draft" until the work on the appservices
side has been reviewed.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
@@ -13,6 +13,7 @@ import com.sun.jna.StringArray
 import java.lang.reflect.Proxy
 import mozilla.telemetry.glean.config.FfiConfiguration
 import mozilla.telemetry.glean.net.FfiPingUploadTask
+import mozilla.telemetry.glean.BuildConfig
 
 // Turn a boolean into its Byte (u8) representation
 internal fun Boolean.toByte(): Byte = if (this) 1 else 0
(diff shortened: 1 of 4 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -65,10 +66,21 @@ internal fun Pointer.getRustString(): String {
...
-        private val JNA_LIBRARY_NAME = "glean_ffi"
-
         internal var INSTANCE: LibGleanFFI = try {
-            val lib = Native.load(JNA_LIBRARY_NAME, LibGleanFFI::class.java) as LibGleanFFI
+            var libName = BuildConfig.JNA_LIBRARY_NAME
+            // If we're part of the application-services megazord, we may need to dynamically
+            // look up the library name in system properties, to work with custom megazords.
+            // Ref https://github.com/mozilla/application-services/blob/main/docs/design/megazords.md
+            if (libName == "megazord") {
+                val mzLibName = System.getProperty("mozilla.appservices.megazord.library")
```
rfk, line 75: AppServices actually publishes a helper module for finding the right native library, which has [significantly more logic](https://github.com/mozilla/application-services/blob/b7839356f2fe07dee03d593e9d9d12ae91b3e062/components/support/android/src/main/java/mozilla/appservices/support/native/Helpers.kt#L160) around checking that we're loading the correct version of the compiled native code. We could take a dependency on `org.mozilla.appservices:native-support` here and use it, but that would in turn pull in a dependency on `com.google.protobuf:protobuf-javalite` which you may not want.

IMHO much of the logic in the `native-support` helper is that that useful these days because we're much better at preventing version mismatches at build time.

badboy: Hm, the protobuf part of native-support doesn't seem to be used for the megazord handling part, so we could still strip that out if we really need to unify the loading code.

rfk: Right. We may be able to factor it into a separate dependency on our side as well.

rfk: > is that that

Uh, this was supposed to be "is not that"...

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -65,11 +66,30 @@ internal fun Pointer.getRustString(): String {
...
+                            "but found $mzLibVersion"
+                        )
+                    }
+                    libName = mzLibName
+                }
+            }
+            val lib = Native.load(libName, LibGleanFFI::class.java) as LibGleanFFI
+            // If we're part of the application-services megazord, it will initialize logging for us.
+            if (! isMegazorded) {
+                lib.glean_enable_logging()
```
rfk, line 91: We also need to avoid initializing the logger when megazorded, since the logger config is now shared with appservices.

rfk:
I tried this out in a local build of glean+appservices+a-c+fenix and it seemed to work as expected. I'll add some more general notes from that experience to [the google doc](https://docs.google.com/document/d/1oUDbQTKzJuapzHyPvtAgFgpCKe3V-dhsYSCI62Mw4Zo) but I think the code here is ready for consideration.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -65,11 +66,30 @@ internal fun Pointer.getRustString(): String {
...
+            var libName = BuildConfig.JNA_LIBRARY_NAME
+            val isMegazorded = (libName == "megazord")
+            // If we're part of the application-services megazord, we may need to dynamically
+            // look up the library name in system properties, to work with custom megazords.
+            // Ref https://github.com/mozilla/application-services/blob/main/docs/design/megazords.md
+            if (isMegazorded) {
+                val mzLibName = System.getProperty("mozilla.appservices.megazord.library")
+                if (mzLibName != null) {
+                    val mzLibVersion = System.getProperty("mozilla.appservices.megazord.version")!!
+                    if (mzLibVersion != BuildConfig.LIBRARY_VERSION) {
```
badboy, line 79: Huh, I guess we never really bothered with that [build variable](https://github.com/mozilla/glean/blob/dc60dc2d370fe0107f959bc74de10c3e0ddfce98/glean-core/android/build.gradle#L42)

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
badboy: Seems reasoanble

PR #1292, opened 2020-10-29
Bug 1672958 - Introduce the RLB `net` module

Dexterp37:
This introduces the 'net' module for the RLB, fixes the ping submission functions and allows RLB consumers to
specify uploader implementations.

Thanks @badboy for the help on the flaky tests.

### DIFF
--- a/glean-core/rlb/src/configuration.rs
+++ b/glean-core/rlb/src/configuration.rs
@@ -2,10 +2,15 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use crate::net::PingUploader;
+
+/// The default server pings are sent to.
+pub(crate) const DEFAULT_GLEAN_ENDPOINT: &str = "https://incoming.telemetry.mozilla.org";
+
 /// The Glean configuration.
 ///
 /// Optional values will be filled in with default values.
-#[derive(Debug, Clone)]
+#[derive(Debug)]
 pub struct Configuration {
     /// Whether upload should be enabled.
     pub upload_enabled: bool,
@@ -19,4 +24,8 @@ pub struct Configuration {
     pub delay_ping_lifetime_io: bool,
     /// The release channel the application is on, if known.
     pub channel: Option<String>,
+    /// The server pings are sent to.
+    pub server_endpoint: Option<String>,
+    /// The instance of the uploader used to send pings.
+    pub uploader: Option<Box<dyn PingUploader + 'static>>,
 }
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -41,8 +41,8 @@ use once_cell::sync::OnceCell;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
-use configuration::DEFAULT_GLEAN_ENDPOINT;
 pub use configuration::Configuration;
+use configuration::DEFAULT_GLEAN_ENDPOINT;
 pub use core_metrics::ClientInfoMetrics;
 pub use glean_core::{global_glean, setup_glean, CommonMetricData, Error, Glean, Lifetime, Result};
 
@@ -86,11 +86,27 @@ fn global_state() -> &'static Mutex<RustBindingsState> {
     STATE.get().unwrap()
 }
 
-/// Set or replace the global Glean object.
+/// Set or replace the global bindings State object.
 fn setup_state(state: RustBindingsState) {
+    // The `OnceCell` type wrapping our state is thread-safe and can only be set once.
+    // Therefore even if our check for it being empty succeeds, setting it could fail if a
+    // concurrent thread is quicker in setting it.
+    // However this will not cause a bigger problem, as the second `set` operation will just fail.
+    // We can log it and move on.
+    //
+    // For all wrappers this is not a problem, as the State object is intialized exactly once on
+    // calling `initialize` on the global singleton and further operations check that it has been
+    // initialized.
     if STATE.get().is_none() {
-        STATE.set(Mutex::new(state)).unwrap();
+        if STATE.set(Mutex::new(state)).is_err() {
+            log::error!(
+                "Global Glean state object is initialized already. This probably happened concurrently."
+            );
+        }
     } else {
+        // We allow overriding the global State object to support test mode.
+        // In test mode the State object is fully destroyed and recreated.
+        // This all happens behind a mutex and is therefore also thread-safe.
         let mut lock = STATE.get().unwrap().lock().unwrap();
         *lock = state;
     }
@@ -108,10 +124,14 @@ fn get_upload_manager() -> &'static Mutex<net::UploadManager> {
     UPLOAD_MANAGER.get().unwrap()
 }
 
-/// Set or replace the global Glean object.
+/// Set or replace the global upload object.
 fn setup_upload_manager(upload_manager: net::UploadManager) {
     if UPLOAD_MANAGER.get().is_none() {
-        UPLOAD_MANAGER.set(Mutex::new(upload_manager));
+        if UPLOAD_MANAGER.set(Mutex::new(upload_manager)).is_err() {
+            log::error!(
+                "Global upload state object is initialized already. This probably happened concurrently."
+            );
+        }
     } else {
         let mut lock = UPLOAD_MANAGER.get().unwrap().lock().unwrap();
         *lock = upload_manager;
@@ -163,9 +183,10 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
 
         let glean = match Glean::new(core_cfg) {
             Ok(glean) => glean,
-            // glean-core already takes care of logging errors: other bindings
-            // simply do early returns, as we're doing.
-            Err(_) => return,
+            Err(err) => {
+                log::error!("Failed to initialize Glean: {}", err);
+                return;
+            }
         };
 
         // glean-core already takes care of logging errors: other bindings
@@ -183,13 +204,12 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
         });
 
         // Initialize the ping uploader.
-        setup_upload_manager(
-            net::UploadManager::new(
-                cfg.server_endpoint.unwrap_or(DEFAULT_GLEAN_ENDPOINT.to_string()),
-                cfg.uploader
-                    .unwrap_or_else(|| Box::new(net::HttpUploader) as Box<dyn net::PingUploader>)
-            )
-        );
+        setup_upload_manager(net::UploadManager::new(
+            cfg.server_endpoint
+                .unwrap_or_else(|| DEFAULT_GLEAN_ENDPOINT.to_string()),
+            cfg.uploader
+                .unwrap_or_else(|| Box::new(net::HttpUploader) as Box<dyn net::PingUploader>),
+        ));
 
         let upload_enabled = cfg.upload_enabled;
 
@@ -381,14 +401,14 @@ pub fn submit_ping_by_name(ping: &str, reason: Option<&str>) {
 }
 
 /// Collect and submit a ping (by its name) for eventual upload, synchronously.
-/// 
+///
 /// The ping will be looked up in the known instances of `private::PingType`. If the
 /// ping isn't known, an error is logged and the ping isn't queued for uploading.
-/// 
+///
 /// The ping content is assembled as soon as possible, but upload is not
 /// guaranteed to happen immediately, as that depends on the upload
 /// policies.
-/// 
+///
 /// If the ping currently contains no content, it will not be assembled and
 /// queued for sending, unless explicitly specified otherwise in the registry
 /// file.
@@ -419,6 +439,35 @@ pub(crate) fn submit_ping_by_name_sync(ping: &str, reason: Option<&str>) {
         let uploader = get_upload_manager().lock().unwrap();
         uploader.trigger_upload();
     }
+}
+
+/// TEST ONLY FUNCTION.
+/// Resets the Glean state and triggers init again.
+#[allow(dead_code)]
+pub(crate) fn reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clear_stores: bool) {
+    // Destroy the existing glean instance from glean-core.
+    if was_initialize_called() {
+        // We need to check if the Glean object (from glean-core) is
+        // initialized, otherwise this will crash on the first test
+        // due to bug 1675215 (this check can be removed once that
+        // bug is fixed).
+        if global_glean().is_some() {
+            with_glean_mut(|glean| {
+                if clear_stores {
+                    glean.test_clear_all_stores()
+                }
+                glean.destroy_db()
+            });
+        }
+        // Allow us to go through initialization again.
+        INITIALIZE_CALLED.store(false, Ordering::SeqCst);
+        // Reset the dispatcher.
+        dispatcher::reset_dispatcher();
+    }
+
+    // Always log pings for tests
+    //Glean.setLogPings(true)
+    initialize(cfg, client_info);
 }
 
 #[cfg(test)]
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -435,7 +435,7 @@ pub(crate) fn submit_ping_by_name_sync(ping: &str, reason: Option<&str>) {
         glean.submit_ping_by_name(&ping, reason.as_deref()).ok()
     });
 
-    if submitted_ping.unwrap() {
+    if let Some(true) = submitted_ping {
         let uploader = get_upload_manager().lock().unwrap();
         uploader.trigger_upload();
     }
@@ -443,6 +443,7 @@ pub(crate) fn submit_ping_by_name_sync(ping: &str, reason: Option<&str>) {
 
 /// TEST ONLY FUNCTION.
 /// Resets the Glean state and triggers init again.
+#[cfg(test)]
 #[allow(dead_code)]
 pub(crate) fn reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clear_stores: bool) {
     // Destroy the existing glean instance from glean-core.
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -25,6 +25,8 @@
 //!     max_events: None,
 //!     delay_ping_lifetime_io: false,
 //!     channel: None,
+//!     server_endpoint: None,
+//!     uploader: None,
 //! };
 //! glean::initialize(cfg, ClientInfoMetrics::unknown());
 //!
@@ -39,6 +41,7 @@ use once_cell::sync::OnceCell;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
 
+use configuration::DEFAULT_GLEAN_ENDPOINT;
 pub use configuration::Configuration;
 pub use core_metrics::ClientInfoMetrics;
 pub use glean_core::{global_glean, setup_glean, CommonMetricData, Error, Glean, Lifetime, Result};
@@ -47,6 +50,7 @@ mod configuration;
 mod core_metrics;
 pub mod dispatcher;
 mod glean_metrics;
+mod net;
 pub mod private;
 mod system;
 
@@ -92,9 +96,31 @@ fn setup_state(state: RustBindingsState) {
     }
 }
 
+/// An instance of the upload manager.
+///
+/// Requires a Mutex, because in tests we can actual reset this.
+static UPLOAD_MANAGER: OnceCell<Mutex<net::UploadManager>> = OnceCell::new();
+
+/// Get a reference to the global upload manager.
+///
+/// Panics if no global state object was set.
+fn get_upload_manager() -> &'static Mutex<net::UploadManager> {
+    UPLOAD_MANAGER.get().unwrap()
+}
+
+/// Set or replace the global Glean object.
+fn setup_upload_manager(upload_manager: net::UploadManager) {
+    if UPLOAD_MANAGER.get().is_none() {
+        UPLOAD_MANAGER.set(Mutex::new(upload_manager));
+    } else {
+        let mut lock = UPLOAD_MANAGER.get().unwrap().lock().unwrap();
+        *lock = upload_manager;
+    }
+}
+
 fn with_glean<F, R>(f: F) -> R
 where
-    F: Fn(&Glean) -> R,
+    F: FnOnce(&Glean) -> R,
 {
     let glean = global_glean().expect("Global Glean object not initialized");
     let lock = glean.lock().unwrap();
@@ -103,7 +129,7 @@ where
 
 fn with_glean_mut<F, R>(f: F) -> R
 where
-    F: Fn(&mut Glean) -> R,
+    F: FnOnce(&mut Glean) -> R,
 {
     let glean = global_glean().expect("Global Glean object not initialized");
     let mut lock = glean.lock().unwrap();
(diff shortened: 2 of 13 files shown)


glean-core/rlb/src/net/mod.rs
```diff
@@ -0,0 +1,55 @@
...
+
+            match incoming_task {
+                PingUploadTask::Upload(request) => {
+                    let doc_id = request.document_id.clone();
+                    let result = self.uploader.upload(&request);
+                    // Process the upload response.
+                    with_glean(|glean| {
+                        // TODO: this works but it's not ideal
+                        //glean.process_ping_upload_response(&doc_id, self.uploader.upload(&request))
+                        glean.process_ping_upload_response(&doc_id, result)
```
Dexterp37, line 42:
```
error[E0507]: cannot move out of result, a captured variable in an Fn closure
--> glean-core\rlb\src\net\mod.rs:39:84
|
37 | let result = self.uploader.upload(&request);
| ------ captured outer variable
38 | // Process the upload response.
39 | with_glean(|glean| glean.process_ping_upload_response(&doc_id, result));
| ^^^^^^ move occurs because result has type glean_core::upload::UploadResult, which does not implement the Copy trait

error[E0507]: cannot move out of result, a captured variable in an Fn closure
--> glean-core\rlb\src\net\mod.rs:39:84
|
37 | let result = self.uploader.upload(&request);
| ------ captured outer variable
38 | // Process the upload response.
39 | with_glean(|glean| glean.process_ping_upload_response(&doc_id, result));
| ^^^^^^ move occurs because result has type glean_core::upload::UploadResult, which does not implement the Copy trait

error: aborting due to previous error; 2 warnings emitted

For more information about this error, try rustc --explain E0507.
error: aborting due to previous error; 2 warnings emitted

For more information about this error, try rustc --explain E0507.
error: could not compile glean.

To learn more, run the command again with --verbose.
warning: build failed, waiting for other jobs to finish...
error: build failed

```

@badboy @brizental , this puzzles me a bit. I think I understand what's going on: `result` has a different lifetime than the scope within `with_glean` and its ownership cannot be transferred.

If I uncomment the previous line and comment this one, everything works just fine. However, that would not be ideal: I don't want to hold `glean` while doing networking stuff.

Moreover, I see that [FOG seems to do it right](https://searchfox.org/mozilla-central/rev/c409dd9235c133ab41eba635f906aa16e050c197/toolkit/components/glean/api/src/ping_upload.rs#47), but I can't understand why that works. Is it because of the `Send`?

badboy: Did you try with `with_glean(move |glean| { ... }`?
(Otherwise I need to take a closer look later)

Dexterp37: Yes, i did, it doesn't help with that error apparently.

badboy: Ah, `with_glean` takes `F: Fn` in Rlb, whereas it takes `F: FnOnce` in FOG: https://searchfox.org/mozilla-central/source/toolkit/components/glean/api/src/lib.rs#29-31

The latter is what we want here to give the most freedom.

Dexterp37: Thanks! That was it!

glean-core/rlb/src/configuration.rs
```diff
@@ -19,4 +24,8 @@ pub struct Configuration {
     pub delay_ping_lifetime_io: bool,
     /// The release channel the application is on, if known.
     pub channel: Option<String>,
+    /// The server pings are sent to.
+    pub server_endpoint: Option<String>,
+    /// The client implementation to use for uploading pings.
+    pub http_client: Option<Box<CustomUploaderType>>,
```
Dexterp37, line 30: @badboy @brizental I'm trying to extend the `Configuration` to allow for custom uploader to be passed in. What I'd love to achieve:

- The developers should be able to pass in an instance of any object implementing the `net::PingUploader` trait;
- If no custom uploader is provided, we'll provide a default simple HTTPS uploader (coming as a separate PR).

The problem with the current implementation is that, by using generics, `glean.initialize` requires us to tell the type for `Configuration` (see my other comment)

glean-core/rlb/src/lib.rs
```diff
@@ -119,7 +127,7 @@ where
 /// * `cfg` - the `Configuration` options to initialize with.
 /// * `client_info` - the `ClientInfoMetrics` values used to set Glean
 ///   core metrics.
-pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
+pub fn initialize(cfg: Configuration<FakeUploader>, client_info: ClientInfoMetrics) {
```
Dexterp37, line 130: @brizental @badboy that's my other comment. See how this becomes `Configuration<FakeUploader>` instead of just `Configuration`.

I can probably make `initialize` generic, too, but then that becomes a nightmare of `"T" cannot be sent between threads safely`. Thoughts?

glean-core/rlb/src/configuration.rs
badboy: You can use a trait object here, given that it's boxed anyway.

```rust
http_client: Option<Box<dyn net::PingUploader>
```

glean-core/rlb/src/lib.rs
brizental: Maybe if you make the generic a bit less generic? Like, `T: PingUploader + Send`? Then you would probably be able to send `T` between threads. The error message you got maybe lays out the traits it is missing.

glean-core/rlb/src/configuration.rs
Dexterp37: That was my first attempt: unfortunately I got shouted at with this:

```
`(dyn net::PingUploader + 'static)` doesn't implement `std::fmt::Debug`

`(dyn net::PingUploader + 'static)` cannot be formatted using `{:?}` because it doesn't implement `std::fmt::Debug`

help: the trait `std::fmt::Debug` is not implemented for `(dyn net::PingUploader + 'static)`
note: required because of the requirements on the impl of `std::fmt::Debug` for `std::boxed::Box<(dyn net::PingUploader + 'static)>`
note: required because of the requirements on the impl of `std::fmt::Debug` for `std::option::Option<std::boxed::Box<(dyn net::PingUploader + 'static)>>`
...
```

Which got me wandering down the path of figuring out if there's a way to exclude specific fields from the `Debug` and `Clone` directives, with no luck.

glean-core/rlb/src/lib.rs
Dexterp37: Yup, that would fix it (as long as I use `T: 'static +  net::PingUploader + Send` as trait bounds). However, this changes the API so that `initialize` will always require a type to be passed. It would be great to have something along the lines of what @badboy  suggests above

brizental: Hm, IIRC there is a way that you could only allow to implement a given trait on something that already implements other traits. Then you would be able to just say `T: net::PingUploader` and that would have all the other trait dependencies implicitly? Let me check that... Indeed there is a way: https://doc.rust-lang.org/book/ch19-03-advanced-traits.html#using-supertraits-to-require-one-traits-functionality-within-another-trait I have never tried this out, heh, but worth a test.

glean-core/rlb/src/configuration.rs
badboy: There's no automatic way, but we can implement `Debug` manually then. I'll take a closer look after the glean weekly on what need to do here.

glean-core/rlb/src/lib.rs
Dexterp37: It still would require users to always call `initialize<SomeUploader>(cfg, client_metrics)`.

One potential way around this would be to slightly change the General API:

```rust
pub fn initialize(cfg: Configuration<net::HttpUploader>, client_info: ClientInfoMetrics) {
    initialize_with_uploader::<net::HttpUploader>(cfg, client_info);
}

pub fn initialize_with_uploader<T: 'static +  net::PingUploader + Send>(cfg: Configuration<T>, client_info: ClientInfoMetrics) {
  // ... all the init code
}
```

@brizental @badboy would this be idiomatic API design? It feels a bit awkward to me compared to the failed example from the other comment

brizental: I don't think this is a bad choice, I am not sure it is idiomatic honestly, but it definitely is clear. I'd go with that, as I can't think of any other workarounds for this issue.

badboy: Having two separate functions is kinda idiomatic in Rust.
However I don't think _not_ passing in the actual uploader is idiomatic. It should be up to the caller to create and allocate an uploader.

glean-core/rlb/src/configuration.rs
badboy: We can also make `trait PingUploader: Debug`, thus requiring all ping uploaders to implement it, then the derive still works.

badboy: This way there's no more generics in `Configuration` and we can store a boxed uploader. It's a trait object and thus dynamic dispatch, but I don't think this is any sort of issue.

Dexterp37: > 
> 
> This way there's no more generics in `Configuration` and we can store a boxed uploader. It's a trait object and thus dynamic dispatch, but I don't think this is any sort of issue.

@badboy what do you mean, here?

glean-core/rlb/src/lib.rs
```diff
@@ -82,19 +86,57 @@ fn global_state() -> &'static Mutex<RustBindingsState> {
...
+        // This all happens behind a mutex and is therefore also thread-safe.
         let mut lock = STATE.get().unwrap().lock().unwrap();
         *lock = state;
     }
 }
 
+/// An instance of the upload manager.
+///
+/// Requires a Mutex, because in tests we can actual reset this.
+static UPLOAD_MANAGER: OnceCell<Mutex<net::UploadManager>> = OnceCell::new();
```
Dexterp37, line 118: @badboy what do you think about this? Would it make sense for this to live in `RustBindingsState` instead? (assuming that's possible)

glean-core/rlb/src/lib.rs
```diff
@@ -137,9 +179,10 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
...
-            // glean-core already takes care of logging errors: other bindings
-            // simply do early returns, as we're doing.
-            Err(_) => return,
+            Err(err) => {
+                log::error!("Failed to initialize Glean: {}", err);
+                return
+            },
         };
 
         // glean-core already takes care of logging errors: other bindings
```
Dexterp37, line 188: Funny thing: until we get to this line, anything that calls `with_glean` may crash.

glean-core/rlb/src/net/http_uploader.rs
```diff
@@ -0,0 +1,25 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `url` - the URL path to upload the data to.
+    /// * `body` - the serialized text data to send.
+    /// * `headers` - a vector of tuples containing the headers to send with
+    ///   the request, i.e. (Name, Value).
+    fn upload(&self, url: String, _body: Vec<u8>, _headers: Vec<(String, String)>) -> UploadResult 
+    {
+        log::debug!("TODO bug xxx: submitting to {:?}", url);
```
Dexterp37, line 22: Reminder: I will need to file a bug for this.

glean-core/rlb/src/test.rs
```diff
@@ -102,12 +156,16 @@ fn initialize_must_not_crash_if_data_dir_is_messed_up() {
...
+        server_endpoint: Some("invalid-test-host".into()),
+        uploader: None
     };
 
-    initialize(cfg, ClientInfoMetrics::unknown());
-
-    // TODO: is this test working? is it waiting for init to finish?
-    dispatcher::block_on_queue();
+    reset_glean(cfg, ClientInfoMetrics::unknown(), false);
+    // Glean init is async and, for this test, it bails out early due to
```
Dexterp37, line 164: This is really ugly. We either need to make the dispatcher behave like the other RLB or introduce some new test-only function to track the progress of the async init. @badboy what do you think?

I'm worried about the dispatchers diverging in semantics, because other discrepancies may be lurking under the hood.

glean-core/rlb/src/net/http_uploader.rs
badboy: yes, you should.

glean-core/rlb/src/net/http_uploader.rs
```diff
@@ -0,0 +1,24 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use crate::net::{PingUploader, UploadResult};
+
+/// A description of a component used to upload pings.
```
badboy, line 7: this comment seems weird.

glean-core/rlb/src/net/mod.rs
```diff
@@ -0,0 +1,112 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! A module responsible for the Glean upload logic.
```
badboy, line 5:
```suggestion
//! Handling the Glean upload logic.
```

glean-core/rlb/src/net/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+use crate::with_glean;
+use glean_core::upload::PingUploadTask;
+pub use glean_core::upload::{PingRequest, UploadResult};
+
+pub use http_uploader::*;
+
+mod http_uploader;
+
+/// The number of milliseconds to make the uploader thread, when told to,
+/// by glean-core.
```
badboy, line 26:
```suggestion
/// The duration the uploader thread should sleep, when told to by glean-core.
```

glean-core/rlb/src/net/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+const THROTTLE_BACKOFF_TIME_MS: u64 = 60_000;
```
badboy, line 27: Use proper types and not variable suffixes to type things (instantiation functions of `Duration` are all const)
```suggestion
const THROTTLE_BACKOFF_TIME: Duration = Duration::from_secs(60);
```

glean-core/rlb/src/net/mod.rs
```diff
@@ -0,0 +1,112 @@
...
+                    PingUploadTask::Upload(request) => {
+                        let doc_id = request.document_id.clone();
+                        let upload_url = format!("{}{}", inner.server_endpoint, request.path);
+                        let headers: Vec<(String, String)> = request.headers.into_iter().collect();
+                        let result = inner.uploader.upload(upload_url, request.body, headers);
+                        // Process the upload response.
+                        with_glean(|glean| glean.process_ping_upload_response(&doc_id, result));
+                    }
+                    PingUploadTask::Wait => {
+                        thread::sleep(Duration::from_millis(THROTTLE_BACKOFF_TIME_MS));
```
badboy, line 100: With the change above this becomes:

```suggestion
                        thread::sleep(THROTTLE_BACKOFF_TIME);
```

glean-core/rlb/src/private/ping.rs
```diff
@@ -15,7 +12,7 @@ pub struct PingType {
 impl PingType {
     /// Creates a new ping type.
     ///
-    /// # Arguments
+    /// ## Arguments
```
badboy, line 15: As per https://mozilla.github.io/glean/book/dev/core/documentation-guidelines.html:

```suggestion
    /// # Arguments
```

glean-core/rlb/src/private/ping.rs
```diff
@@ -34,10 +31,30 @@ impl PingType {
...
+    /// and the ping currently contains no content,
+    /// it will not be queued for upload.
+    /// If the ping is configured with `send_if_empty = true`
+    /// it will be queued for upload even if otherwise empty.
+    ///
+    /// Pings always contain the `ping_info` and `client_info` sections.
+    /// See [ping sections](https://mozilla.github.io/glean/book/user/pings/index.html#ping-sections)
+    /// for details.
+    ///
+    /// ## Parameters
```
badboy, line 55:
```suggestion
    /// # Arguments
    ///
```

glean-core/rlb/src/test.rs
```diff
@@ -20,7 +20,54 @@ const GLOBAL_APPLICATION_ID: &str = "org.mozilla.rlb.test";
...
+            url: String,
+            _body: Vec<u8>,
+            _headers: Vec<(String, String)>,
+        ) -> net::UploadResult {
+            self.sender.send(url).unwrap();
+            net::UploadResult::HttpStatus(200)
+        }
+    }
+
+    // Createa  custom configuration to use a fake uploader.
```
badboy, line 70:
```suggestion
    // Create a  custom configuration to use a fake uploader.
```

glean-core/rlb/src/test.rs
badboy: As discussed making them more equal again seems preferrable.
This can be done in a followup, so please add a TODO and file a bug.

badboy: Ah, the bug is at the end. I'd say add it above:
```suggestion
    // TODO(bug 1675215): ensure initialize runs through dispatcher.
    // Glean init is async and, for this test, it bails out early due to
```

glean-core/rlb/src/test.rs
```diff
@@ -102,12 +160,16 @@ fn initialize_must_not_crash_if_data_dir_is_messed_up() {
...
+        uploader: None,
     };
 
-    initialize(cfg, ClientInfoMetrics::unknown());
-
-    // TODO: is this test working? is it waiting for init to finish?
-    dispatcher::block_on_queue();
+    reset_glean(cfg, ClientInfoMetrics::unknown(), false);
+    // Glean init is async and, for this test, it bails out early due to
+    // an error: we can do nothing but wait. Tests in other bindings use
```
badboy, line 169: Hm, which error? Shouldn't it still run initialize to completion, thus unblocking the queue, which means we could wait for that?

glean-core/rlb/src/test.rs
```diff
@@ -118,20 +180,53 @@ fn queued_recorded_metrics_correctly_record_during_init() {
...
+            channel: Some("testing".into()),
+            server_endpoint: Some("invalid-test-host".into()),
+            uploader: None,
+        },
+        ClientInfoMetrics::unknown(),
+        true,
+    );
+
+    // Glean init is async and we need to wait for it to complete. There
+    // is nothing we can do but wait. Tests in other bindings use
```
badboy, line 205: Waiting for the queue to unblock?

glean-core/rlb/tests/schema.rs
```diff
@@ -48,39 +52,63 @@ fn new_glean() -> tempfile::TempDir {
...
-
-    // 1. line: the URL
-    // 2. line: the JSON body
-    let file = File::open(&pending_pings[0]).unwrap();
-    let mut lines = BufReader::new(file).lines();
-    let _url = lines.next().unwrap().unwrap();
-    let body = lines.next().unwrap().unwrap();
+    let (s, r) = crossbeam_channel::bounded::<Vec<u8>>(1);
+
+    // Define a fake uploader that reports back the submission URL
```
badboy, line 57: It reports back the body, not the URL

```suggestion
    // Define a fake uploader that reports back the submitted payload
```

glean-core/rlb/tests/schema.rs
```diff
@@ -48,39 +52,63 @@ fn new_glean() -> tempfile::TempDir {
...
+    let mut gzip_decoder = GzDecoder::new(&raw_body[..]);
+    let mut s = String::with_capacity(raw_body.len());
+
+    let data = gzip_decoder
+        .read_to_string(&mut s)
+        .ok()
+        .map(|_| &s[..])
+        .or_else(|| std::str::from_utf8(&raw_body).ok())
+        .and_then(|payload| serde_json::from_str(payload).ok())
+        .unwrap();
```
badboy, line 109: Yikes, good opportunity to factor out later.

badboy (approved):
r+wc.
Overall this looks good. A couple of documentation things and things we need to clarify.
Testing currently isn't optimal, but we can fix that in followups when we fix up the dispatcher I think

glean-core/rlb/src/net/http_uploader.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1675468

glean-core/rlb/src/lib.rs
Dexterp37: That's true !

glean-core/rlb/src/lib.rs
```diff
@@ -329,9 +396,79 @@ pub fn submit_ping_by_name(ping: &str, reason: Option<&str>) {
...
+            // This won't actually return from `submit_ping_by_name`, but
+            // returning `false` here skips spinning up the uploader below,
+            // which is basically the same.
+            return Some(false);
+        }
+
+        glean.submit_ping_by_name(&ping, reason.as_deref()).ok()
+    });
+
+    if submitted_ping.unwrap() {
```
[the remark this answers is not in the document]
Dexterp37: How would this panic? [glean.submit_ping_by_name](https://searchfox.org/glean/rev/3999e6a5ed45d7b5ed73877c14d4ea4c9855006f/glean-core/src/lib.rs#661) will always return `Ok(false)` for unknown pings, which means this line is unwrapped to `false` and we never enter the if branch? Or am I thinking about this wrong?

glean-core/rlb/src/lib.rs
```diff
@@ -329,9 +396,79 @@ pub fn submit_ping_by_name(ping: &str, reason: Option<&str>) {
...
+        let uploader = get_upload_manager().lock().unwrap();
+        uploader.trigger_upload();
+    }
+}
+
+/// TEST ONLY FUNCTION.
+/// Resets the Glean state and triggers init again.
+#[allow(dead_code)]
```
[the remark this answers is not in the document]
Dexterp37: TIL! Lovely!

glean-core/rlb/src/lib.rs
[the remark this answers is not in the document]
badboy: oh, _I_ got it wrong.
So when _do_ we return Err? 
When we fail to write a file. Which is not under our control, so we probably shouldn't panic either?

glean-core/rlb/src/test.rs
Dexterp37: Nope, this test checks that we don't crash if the data directory can't be created. If we can't do that, the glean object instantiation is not created and fails [here](https://searchfox.org/glean/rev/3999e6a5ed45d7b5ed73877c14d4ea4c9855006f/glean-core/rlb/src/lib.rs#142), not ever flushing the queue, as done in the other bindings. let me be more specific in the comment.

glean-core/rlb/src/test.rs
Dexterp37: Right, I think we can do this here as well, given your latest fix to the dispatcher. Looks like I wasn't able to due to the `block_on_queue` waiting on an old dispatcher :)

glean-core/rlb/src/lib.rs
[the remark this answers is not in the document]
Dexterp37: We cleared this up on Element. Changes incoming.

glean-core/rlb/src/lib.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1675475

glean-core/rlb/tests/schema.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1675476

PR #1294, opened 2020-11-02
Don't run Miniconda install in parallel

mdboom:
This should address the intermittent build failures in android-components [described here](https://github.com/gradle/gradle/issues/7047#issuecomment-430139316).

We will need to make a Glean release and ship it to android-components to fix the build failures there.  I was unable to confirm the existing build failure in a-c, but I did confirm that this patch seems to do no harm there.

Cc @pocmo

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * Android
   * Update the JNA dependency from 5.2.0 to 5.6.0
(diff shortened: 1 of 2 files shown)


badboy (approved):
I guess this is as good as any other.

pocmo:
Great, thank you for investigating this!

PR #1296, opened 2020-11-03
Bug 1673392 - Audit log levels of log messages in glean-core

brizental:
Creating this as a draft initially, for now I have only added the docs with the logging level guidelines I propose we adopt for glean-core. Wanted to discuss this prior to actually going ahead and making the changes to all our logs ;)

### DIFF
--- a/docs/dev/core/logging-levels.md
+++ b/docs/dev/core/logging-levels.md
@@ -1,0 +1,35 @@
+# Logging
+
+On the Glean SDK's Rust crates we use the [log](https://github.com/rust-lang/log) crate
+to display log messages.
+
+## Logging Levels
+
+### [trace](https://docs.rs/log/0.4.11/log/macro.info.html)
+
+- Used for logging very granular information about the flow of glean-core internals. This logging level is meant for helping with bug diagnostics and may be very noisy.
+    - **Example**: `log::trace!("{} pings left in the queue (only deletion-request expected)", queue.len());`
+
+### [debug](https://docs.rs/log/0.4.11/log/macro.debug.html)
+
+- Used for logging messages that contain internal glean-core information. This logging level is meant for helping with bug diagnostics and may be very noisy.
+    - **Example**: `log::debug!("Database path: {:?}", path.display());`
+
+### [info](https://docs.rs/log/0.4.11/log/macro.info.html)
+
+- Used for logging informational messages. These messages should be enough for glean-core developers and users to verify that Glean is working as expected during development.
+    - **Example**: `log::info!("Ping {} successfully sent {}.", document_id, status);`
+
+### [warn](https://docs.rs/log/0.4.11/log/macro.warn.html)
+
+- Used for logging internal glean-core errors. These errors do not require further action of a Glean consumer, but may be informative to diagnose bugs.
+    - **Example**: `log::warn!("IO error writing event to store '{}': {}", store_name, err);`
+
+> **Note** We prefer to put such errors on the `warn` level also because otherwise they are sent to error monitoring tools, such as Sentry, and may pollute our end users dashboards.
+
+### [error](https://docs.rs/log/0.4.11/log/macro.error.html)
+
+- Used for logging messages that indicate wrongful usage of the Glean API.
+    - **Example**: `log::error!("Unexpected 'uuid' value coming from platform code '{}'", value);`
+- Used for logging messages that may be of interest to a consumer of the Glean SDK.
+    - **Example**: `log::error!("Failed to submit deletion-request ping on optout: {}", err);`
--- a/docs/dev/core/logging-levels.md
+++ b/docs/dev/core/logging-levels.md
@@ -5,7 +5,7 @@ to display log messages.
 
 ## Logging Levels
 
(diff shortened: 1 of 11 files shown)


docs/dev/core/logging-levels.md
```diff
@@ -0,0 +1,35 @@
+# Logging
+
+On the Glean SDK's Rust crates we use the [log](https://github.com/rust-lang/log) crate
+to display log messages.
+
+## Logging Levels
+
+### [trace](https://docs.rs/log/0.4.11/log/macro.info.html)
```
badboy, line 8:
```suggestion
### [trace](https://docs.rs/log/0.4.11/log/macro.trace.html)
```

docs/dev/core/logging-levels.md
```diff
@@ -0,0 +1,35 @@
...
+    - **Example**: `log::debug!("Database path: {:?}", path.display());`
+
+### [info](https://docs.rs/log/0.4.11/log/macro.info.html)
+
+- Used for logging informational messages. These messages should be enough for glean-core developers and users to verify that Glean is working as expected during development.
+    - **Example**: `log::info!("Ping {} successfully sent {}.", document_id, status);`
+
+### [warn](https://docs.rs/log/0.4.11/log/macro.warn.html)
+
+- Used for logging internal glean-core errors. These errors do not require further action of a Glean consumer, but may be informative to diagnose bugs.
```
badboy, line 25: I think I would rephrase that to make it clear these are not "user-actionable" errors, but really warnings
Maybe:

Used for logging internal glean-core warning. These could be internal errors or warnings, that do not require further action by a Glean consumer, but may be informative to diagnose bugs.

badboy (commented):
Woops, missed this. I think this is good, we can refine it if needed.

docs/dev/core/logging-levels.md
```diff
@@ -0,0 +1,35 @@
...
+    - **Example**: `log::warn!("IO error writing event to store '{}': {}", store_name, err);`
+
+> **Note** We prefer to put such errors on the `warn` level also because otherwise they are sent to error monitoring tools, such as Sentry, and may pollute our end users dashboards.
+
+### [error](https://docs.rs/log/0.4.11/log/macro.error.html)
+
+- Used for logging messages that indicate wrongful usage of the Glean API.
```
mdboom, line 32: Should we note the thing about Sentry on iOS?  As in:

"Errors should be used sparingly. They are often surfaced in runtime error monitoring software, such as Sentry, so really should only be used when an action by a Glean user *must* be taken."

brizental: Hm, I added a similar not on the `warn` level part. do you think that would be better positioned on this `error` part?

brizental: In hindsight, now that I applied @badboy 's suggestions to the text I think it is better to but this note on the error part :)

glean-core/ffi/src/handlemap_ext.rs
```diff
@@ -41,7 +41,7 @@ pub fn log_if_error(error: ExternError) {
     if !error.get_code().is_success() {
         // in practice this should never panic, but you never know...
         ffi_support::abort_on_panic::call_with_output(|| {
-            log::error!(
+            log::warn!(
```
badboy, line 44: IMO this can stay an error. It's an exceptional case and we better know about this early (even if it occasionally results in noise we can't track down)

```suggestion
            log::error!(
```

glean-core/ffi/src/lib.rs
```diff
@@ -177,11 +177,11 @@ pub extern "C" fn glean_enable_logging() {
     #[cfg(all(not(target_os = "android"), not(target_os = "ios")))]
     {
         match env_logger::try_init() {
-            Ok(_) => log::debug!("stdout logging should be hooked up!"),
+            Ok(_) => log::trace!("stdout logging should be hooked up!"),
             // Please note that this is only expected to fail during unit tests,
             // where the logger might have already been initialized by a previous
             // test. So it's fine to print with the "logger".
-            Err(_) => log::debug!("stdout was already initialized"),
+            Err(_) => log::warn!("stdout was already initialized"),
```
badboy, line 184: while we're here:

```suggestion
            Err(_) => log::warn!("stdout logging was already initialized"),
```

glean-core/src/lib.rs
```diff
@@ -678,7 +678,7 @@ impl Glean {
     /// Register a new [`PingType`](metrics/struct.PingType.html).
     pub fn register_ping_type(&mut self, ping: &PingType) {
         if self.ping_registry.contains_key(&ping.name) {
-            log::debug!("Duplicate ping named '{}'", ping.name)
+            log::error!("Duplicate ping named '{}'", ping.name)
```
badboy, line 681: We only [just recently](https://github.com/mozilla/glean/commit/b61b60d4977cffca57b09a32190638b5b163bc05) dropped this log level, as it was surfaced in Sentry on iOS, but is of no help.

```suggestion
            log::debug!("Duplicate ping named '{}'", ping.name)
```

badboy (changes requested):

badboy (approved):

PR #1297, opened 2020-11-03
Bug 1672722 - Ignore return value of a launch that may fail or it might not.

badboy:
Just before the launch we send a shutdown command.
If processing is faster on the main thread it might still be able to
send the following command.
But it also might not.

In any case a task that is sent after a shutdown command should not be executed.
And that's what we test for.

Note: the *global* dispatcher API does not return a `Result`, so users
shouldn't even need to deal with this.


Dexterp37 (approved):
HA! Good catch!

PR #1299, opened 2020-11-03
Bug 1674786 - Remove 10MB quota for pending pings directory size

brizental:


brizental:
Oh wow, how did my tests pass locally...

No need to review this until I have fixed my mess :D

CHANGELOG.md
```diff
@@ -1,5 +1,9 @@
 # Unreleased changes
 
+* General
+  * Remove 10MB quota for pending pings directory size. ([#1299](https://github.com/mozilla/glean/pull/1299))
+    * This quota became obsolete after we added another quota of 250 ping files on the same directory (see: [v32.4.0](https://github.com/mozilla/glean/compare/v32.3.2...v32.4.0)). The latter is always reached before the 10MB quota.
```
Dexterp37, line 5: @brizental , is this true? What about products having big pings, almost one MB big, e.g. 0.9mb , and having many of them? It would take only 12 pings to break that 10MB limit and 250 pings would be about 225MB...

brizental: I hadn't considered that. Have you observed that in any product? The conclusion that this was obsolete was based on fenix metrics.

Dexterp37: > 
> 
> I hadn't considered that. Have you observed that in any product? The conclusion that this was obsolete was based on fenix metrics.

Firefox :-) :trollface:

brizental: Heh, that would be as easy as not merging this PR. @badboy Had agreed to remove this as well, do you have any counter arguments?

badboy: I actually ... didn't think that far ahead, but I agree, nothing really stops pings from growing bigger.
I'm therefore inclined to just leave the limit in.
We can revisit that in the future

badboy (changes requested):
As said in the comment we actually might still want to keep that and just keeping it in is the simpler way than bringing it back in the future.
We can however stop tracking this in our Fenix monitoring.

I'd say close this PR and close the bug with a WONTFIX. Thanks for doing the work here, sorry you had to.

PR #1304, opened 2020-11-05
Calculate the database size from all files in the directory

yvonmanzi:
@badboy, I just fixed the issue. Thanks a lot for the guidance you provided on Bugzilla. Looking forward to some feedback.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -86,45 +86,35 @@ impl std::fmt::Debug for Database {
     }
 }
 
-/// Get the file size of a file in the given path and file.
-///
-/// # Arguments
-///
-/// - `path` - The path
-///
-/// # Returns
-///
-/// Returns the non-zero file size in bytes,
-/// or `None` on error or if the size is `0`.
-fn file_size(path: &Path) -> Option<NonZeroU64> {
-    log::trace!("Getting file size for path: {}", path.display());
-    fs::metadata(path)
-        .ok()
-        .map(|stat| stat.len())
-        .and_then(NonZeroU64::new)
-}
-
-/// Determine the size of all the files in the given file directory.
+/// Calculate the  database size from all the files in the directory.
 ///
 ///  # Arguments
 ///
-///  -`path` - The path to the directory
+///  *`path` - The path to the directory
 ///
 ///  # Returns
 ///
-/// Returns the non-zero combined size of all files,
+/// Returns the non-zero combined size of all files in a directory,
 /// or `None` on error or if the size is `0`.
 fn database_size(dir: &Path) -> Option<NonZeroU64> {
     let mut total_size = 0;
     if let Ok(entries) = fs::read_dir(dir) {
         for entry in entries {
             if let Ok(entry) = entry {
-                let path = entry.path();
-                // the getter is returning the u64 value from NonZeroU64
-                total_size += file_size(&path)?.get();
+                if let Ok(file_type) = entry.file_type() {
+                    if file_type.is_file() {
+                        let path = entry.path();
+                        if let Ok(metadata) = fs::metadata(path) {
+                            total_size += metadata.len();
+                        } else {
+                            continue;
+                        }
+                    }
+                }
             }
         }
     }
+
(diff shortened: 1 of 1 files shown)


glean-core/src/database/mod.rs
```diff
@@ -104,6 +104,30 @@ fn file_size(path: &Path) -> Option<NonZeroU64> {
         .and_then(NonZeroU64::new)
 }
 
+/// Determine the size of all the files in the given file directory.
+///
+///  # Arguments
+///
+///  -`path` - The path to the directory
```
badboy, line 111:
```suggestion
///  * `path` - The path to the directory
```

glean-core/src/database/mod.rs
```diff
@@ -104,6 +104,30 @@ fn file_size(path: &Path) -> Option<NonZeroU64> {
...
+/// Returns the non-zero combined size of all files,
+/// or `None` on error or if the size is `0`.
+fn database_size(dir: &Path) -> Option<NonZeroU64> {
+    let mut total_size = 0;
+    if let Ok(entries) = fs::read_dir(dir) {
+        for entry in entries {
+            if let Ok(entry) = entry {
+                let path = entry.path();
+                // the getter is returning the u64 value from NonZeroU64
+                total_size += file_size(&path)?.get();
```
badboy, line 124: We can savae some work here. You can read the [`metadata`](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.metadata) of an entry directly, that one then has the [`len`](https://doc.rust-lang.org/std/fs/struct.Metadata.html#method.len). You can then also remove the `file_size` function as its not used anymore.

If there's no metadata for whatever reason, we shouldn't bail out early here. If we can't read the file size of a single file, we should still read the remaining files.

glean-core/src/database/mod.rs
```diff
@@ -104,6 +104,30 @@ fn file_size(path: &Path) -> Option<NonZeroU64> {
...
 
+/// Determine the size of all the files in the given file directory.
+///
+///  # Arguments
+///
+///  -`path` - The path to the directory
+///
+///  # Returns
+///
+/// Returns the non-zero combined size of all files,
```
badboy, line 115:
```suggestion
/// Returns the non-zero combined size of all files in a directory,
```

glean-core/src/database/mod.rs
```diff
@@ -104,6 +104,30 @@ fn file_size(path: &Path) -> Option<NonZeroU64> {
...
+/// or `None` on error or if the size is `0`.
+fn database_size(dir: &Path) -> Option<NonZeroU64> {
+    let mut total_size = 0;
+    if let Ok(entries) = fs::read_dir(dir) {
+        for entry in entries {
+            if let Ok(entry) = entry {
```
badboy, line 121: An entry could be another directory.
We don't use nested directories here, so it _shouldn't happen. 

Let's still handle it by ignoring it explicitly. You can get an entries [`file_type`](https://doc.rust-lang.org/std/fs/struct.DirEntry.html#method.file_type) and check that `is_file` is true.

badboy (approved):
This is coming along nicely, but will need some adjustements.
See the inline comments.

Also please change the commit message: `Calculate the database size from all files in the directory` or something similar.

yvonmanzi:
Awesome, makes sense. Thanks for the feedback! Lemme send another PR for you to see if it all looks good.

badboy:
> Awesome, makes sense. Thanks for the feedback! Lemme send another PR for you to see if it all looks good.

You can update this PR (might require force-pushing to your branch)

yvonmanzi:
Hi @badboy, I am not sure why this is happening, but I see that one of the checks is failing. I'd appreciate it if you helped me figure out what the issue is. Thanks.

badboy:
> Hi @badboy, I am not sure why this is happening, but I see that one of the checks is failing. I'd appreciate it if you helped me figure out what the issue is. Thanks.

That seems unrelated and you can ignore it.
I somehow missed that you already pushed new code though, let me take a look.

badboy (approved):

PR #1305, opened 2020-11-06
[doc only] Bug 1675792: Document that payload format is implementation

mdboom:


docs/dev/core/internal/payload.md
```diff
@@ -2,7 +2,9 @@
 
 The main sections of a Glean ping are described in [Ping Sections](../../../user/pings/index.md#ping-sections).
 This **Payload format** chapter describes details of the ping payload that are relevant for decoding Glean pings in the pipeline.
-This is less relevant for end users of the Glean SDK.
+
+> NOTE: The payload format is an implementation detail of the Glean SDK and subject to change at any time.
```
Dexterp37, line 6: I was thinking about this a bit more: this page lives within the [`Internal implementation details` section](https://mozilla.github.io/glean/book/dev/core/internal/index.html). Should we add a more general note to the index page, since it applies to all the other pages under this section?

mdboom: Good point.  I think we should probably have a note in both places (here and the index page).

Dexterp37 (approved):

PR #1311, opened 2020-11-10
Test that nested dispatcher calls work as expected

badboy:
Unfortunately this test can't run by default because it influences the
global dispatcher. We can fix that by resetting the dispatcher.

The test can be manually run with:

    cargo test -p glean -- --ignored global_nested_calls

### DIFF
--- a/glean-core/rlb/src/dispatcher/global.rs
+++ b/glean-core/rlb/src/dispatcher/global.rs
@@ -129,4 +129,42 @@ mod test {
         expected.push(200);
         assert_eq!(&*result.lock().unwrap(), &expected);
     }
+
+    #[test]
+    #[ignore] // We can't reset the queue at the moment, so flushing it breaks other tests.
+    fn global_nested_calls() {
+        let _ = env_logger::builder().is_test(true).try_init();
+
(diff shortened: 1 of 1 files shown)


glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -129,4 +129,42 @@ mod test {
         expected.push(200);
         assert_eq!(&*result.lock().unwrap(), &expected);
     }
+
+    #[test]
+    #[ignore] // We can't reset the queue at the moment, so flushing it breaks other tests.
```
Dexterp37, line 134: Is this still true? We do have `reset_dispatcher()` now!

Dexterp37 (approved):

glean-core/rlb/src/dispatcher/global.rs
badboy: I guess it's not the way it is written, but because of multi-threaded tests this then needs to hold the global lock.
IMO we can do that in a followup (e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1675476)

PR #1317, opened 2020-11-12
Do not require default-features on rkv and downgrade bincode

badboy:
~~**NOTE**: hold off on merging while I do another test please.~~

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -67,9 +67,9 @@ dependencies = [
 
 [[package]]
 name = "bincode"
-version = "1.3.1"
+version = "1.2.1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "f30d3a39baa26f9651f17b375061f3233dde33424a8b72b0dbe93a68a0bc896d"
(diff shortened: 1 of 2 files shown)


Dexterp37 (approved):
Hopeful r+!

badboy:
Ok, here's the test I ran to ensure rkv works as expected with the LMDB backend for us.
Code is here: https://github.com/mozilla/glean/commit/0f4e924dea373aa05446f16831974cdfb8c1f401

I then ran:

```
DATABASE_DIR=$(pwd)/tmp cargo test -p glean-core --lib -- --test-threads 1 --ignored write
```

This should create a new database in the provided directory, using rkv _with_ default features.

Then I changed the `Cargo.toml` to have the new line with rkv (`default-features = false`), then ran:

```
DATABASE_DIR=$(pwd)/tmp cargo test -p glean-core --lib -- --test-threads 1 --ignored read
```

It succeeded.

Just to double-check I looked at the code: `db-dup-sort` enables the `MultiStore`, which we don't use, `db-int-key` enables `IntegerStore`, which we also don't use.
the LMDB file format is unaffected.

We're good to go!

PR #1322, opened 2020-11-13
Bug 1676866 - Migrate to safe mode

badboy:
This automatically migrates an existing LMDB database of Glean to a safe-mode database, if and only if the `rkv-safe-mode` feature is activated on the glean-core crate.

The plan will be to enable this in mozilla-central.
I already took this for a manual test using my local m-c checkout and migration succeeds as expected and keeps the `client_info.client_id`.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -27,12 +27,18 @@ mod backend {
     pub fn rkv_new(path: &Path) -> Result<Rkv, rkv::StoreError> {
         Rkv::new::<rkv::backend::Lmdb>(path)
     }
+
+    /// No migration necessary when staying with LMDB.
+    pub fn migrate(_path: &Path, _dst_env: &Rkv) {
+        // intentionally left empty
+    }
 }
 
 // Select the "safe mode" storage backend when the feature is activated.
 #[cfg(feature = "rkv-safe-mode")]
 mod backend {
-    use std::path::Path;
+    use rkv::migrator::Migrator;
+    use std::{fs, path::Path};
 
     /// cbindgen:ignore
     pub type Rkv = rkv::Rkv<rkv::backend::SafeModeEnvironment>;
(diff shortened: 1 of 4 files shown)


glean-core/src/database/mod.rs
```diff
@@ -44,6 +50,91 @@ mod backend {
...
+                std::io::ErrorKind::NotFound => {
+                    // silently drop this error, the file was already non-existing
+                }
+                _ => log::warn!("{}", msg),
+            }
+        }
+    }
+
+    fn delete_lmdb_database(path: &Path) {
+        let datamdb = path.join("data.mdb");
```
Dexterp37, line 66: Do we need to adjust or fix [this](https://searchfox.org/glean/rev/dc60dc2d370fe0107f959bc74de10c3e0ddfce98/glean-core/src/database/mod.rs#123)?

glean-core/src/database/mod.rs
```diff
@@ -27,12 +27,18 @@ mod backend {
     pub fn rkv_new(path: &Path) -> Result<Rkv, rkv::StoreError> {
         Rkv::new::<rkv::backend::Lmdb>(path)
     }
+
+    /// No migration necessary when staying with LMDB.
+    pub fn migrate(_path: &Path, _dst_env: &Rkv) {
+        // intentionally left empty
```
Dexterp37, line 33:
```suggestion
        // Intentionally left empty.
```

Super nit. Feel free to ignore, but this makes it consistent with the rest of the comments in the codebase.

glean-core/src/database/mod.rs
```diff
@@ -44,6 +50,91 @@ mod backend {
     pub fn rkv_new(path: &Path) -> Result<Rkv, rkv::StoreError> {
         Rkv::new::<rkv::backend::SafeMode>(path)
     }
+
+    fn delete_and_log(path: &Path, msg: &str) {
+        if let Err(err) = fs::remove_file(path) {
+            match err.kind() {
+                std::io::ErrorKind::NotFound => {
+                    // silently drop this error, the file was already non-existing
```
Dexterp37, line 58:
```suggestion
                    // Silently drop this error, the file was already non-existing.
```

glean-core/src/database/mod.rs
badboy: We will have "wrong" data exactly once (on the first start that migrates): database size will be 0 (because it checks the wrong file).
#1304 should fix this though.

Dexterp37 (approved):

PR #1323, opened 2020-11-13
Bug 1673667 - Add the counter metric type to the RLB

Dexterp37:
This implements a new type for the RLB. The existing [boolean type](https://searchfox.org/glean/rev/dc60dc2d370fe0107f959bc74de10c3e0ddfce98/glean-core/rlb/src/private/boolean.rs#32) can be used as a reference for reviewing.

The test is ported over from [Kotlin](https://searchfox.org/glean/rev/dc60dc2d370fe0107f959bc74de10c3e0ddfce98/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt#319).


glean-core/rlb/src/private/counter.rs
```diff
@@ -0,0 +1,93 @@
...
+    /// # Arguments
+    ///
+    /// * `amount` - The amount to increase by. Should be positive.
+    ///
+    /// ## Notes
+    ///
+    /// Logs an error if the `amount` is 0 or negative.
+    fn add(&self, amount: i32) {
+        let metric = Arc::clone(&self.0);
+        dispatcher::launch(move || crate::with_glean(|glean| metric.add(glean, amount)));
```
badboy, line 46: Wonder at which point it might make sense to refactor this pattern into a single function `launch_with_glean(|glean| ...)` (naming to be bikeshed'd)

glean-core/rlb/src/private/counter.rs
```diff
@@ -0,0 +1,93 @@
...
+    ///
+    /// * `ping_name` - represents the optional name of the ping to retrieve the
+    ///   metric for. Defaults to the first value in `send_in_pings`.
+    fn test_get_value<'a, S: Into<Option<&'a str>>>(&self, ping_name: S) -> Option<i32> {
+        dispatcher::block_on_queue();
+
+        let queried_ping_name = match ping_name.into() {
+            Some(name) => name,
+            None => self.0.meta().send_in_pings.first().unwrap(),
+        };
```
badboy, line 65: If we want to get really fancy I think this should work too:

```suggestion
        let queried_ping_name = ping_name.into().unwrap_or_else(|| &self.0.meta().send_in_pings[0]);
```

glean-core/rlb/src/test.rs
```diff
@@ -254,9 +255,37 @@ fn initialize_must_not_crash_if_data_dir_is_messed_up() {
...
+    // This will queue 3 tasks that will add to the metric value once Glean is initialized
+    for _ in 0..3 {
+        metric.add(1);
+    }
+
+    // TODO: To be fixed in bug 1677150.
+    // Ensure that no value has been stored yet since the tasks have only been queued
+    // and not executed yet
+
+    // Calling resetGlean here will cause Glean to be initialized and should cause the queued
```
badboy, line 282: This is wrong now, we're calling `new_glean`

badboy (approved):

glean-core/rlb/src/private/counter.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1677441

PR #1324, opened 2020-11-16
Bug 1677446 - Implement the string metric type for RLB

Dexterp37:
As the title says, this implements the string metric type for the Rust Language Bindings.

See the individual commit messages (especially the testing ones) for more information!


glean-core/rlb/src/private/string.rs
```diff
@@ -0,0 +1,96 @@
...
+
+use inherent::inherent;
+use std::sync::Arc;
+
+use glean_core::metrics::MetricType;
+use glean_core::ErrorType;
+
+use crate::dispatcher;
+
+// We need to wrap the glean-core type: otherwise if we try to implement
```
travis79, line 13:
```suggestion
// We need to wrap the glean-core type, otherwise if we try to implement
```

travis79 (approved):
One nit, but LGTM!

glean-core/rlb/src/private/string.rs
Dexterp37: Thanks!

PR #1325, opened 2020-11-17
Bug 1676712 - Use the correct time unit for sleeps in Swift and Python 

badboy:
Closes #1315

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@
 * Rust
   * Introduce the Counter metric type in the RLB.
   * Introduce the String metric type in the RLB.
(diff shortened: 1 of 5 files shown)


brizental (approved):
I _think_ C# also has this issue, but no need to include that fix on this PR I think. Maybe follow up?

badboy:
> I _think_ C# also has this issue, but no need to include that fix on this PR I think. Maybe follow up?

Not according to the docs: https://docs.microsoft.com/en-us/dotnet/api/system.threading.thread.sleep?view=net-5.0#System_Threading_Thread_Sleep_System_Int32_

PR #1327, opened 2020-11-17
Bug 1677450 - Implement labeled types in the RLB

Dexterp37:
This implements support for the labeled types (bool, string, counter) in the RLB.

A few interesting bits were carried over from the m-c implementation:

- we have a `ToLabeled` trait that metrics need to implement in order to be "labeled";
- the trait is not exposed outside of the RLB;
- we now have common testing functions;

Note: this was tested in m-c (we're able to use the trait and to call in the RLB type from m-c). The IPC details will require additional thinking but are outside of the scope of the RLB work/this PR.


docs/user/metrics/labeled_booleans.md
```diff
@@ -130,6 +130,39 @@ Assert.Equal(0, Accessibility.features.TestGetNumRecordedErrors(ErrorType.Invali
 
 </div>
 
+<div data-lang="Rust" class="tab">
+
+```rust
+use glean_metrics;
+
+accessibility::features.get("screen_reader").set(is_screen_reader_enabled());
```
mdboom, line 138: Should we implement the `Index` [trait](https://doc.rust-lang.org/std/ops/trait.Index.html) so we could say:

```rust
accessibility::features["screen_reader"].set(is_screen_reader_enabled());
```

This would be more in line with other language bindings.

I assume (but don't know) that Traits can require other Traits to make this happen...

Dexterp37: I think that's a really good idea. @badboy any reason why this wasn't implemented in m-c?

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,366 @@
...
+        private::BooleanMetric, private::CounterMetric, private::StringMetric, CommonMetricData,
+    };
+    use glean_core::traits::ToLabeled;
+    use std::sync::Arc;
+
+    // `LabeledMetric<BooleanMetric>` is possible.
+    //
+    // See [Labeled Booleans](https://mozilla.github.io/glean/book/user/metrics/labeled_booleans.html).
+    impl ToLabeled for BooleanMetric {
+        type InnerCoreType = glean_core::metrics::BooleanMetric;
```
Dexterp37, line 27: Note to self: let's drop the `Type` suffix. This should either be `InnerCore` or `InnerCoreMetric`.

docs/user/metrics/labeled_booleans.md
badboy: `Index` returns a reference to an existing object. As we don't have an existing object when the user just calls it we can't create a reference out of thin air, thus we can't implement `Index` for this.

Dexterp37: I see. Thanks for explaining.

Dexterp37:
The beta error is being fixed in #1331

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v33.4.0...main)
 
+* Rust
+  * Introduce the Labeled metric type in the RLB.
```
badboy, line 6:
```suggestion
  * Introduce the Labeled metric type in the RLB ([#1327](https://github.com/mozilla/glean/pull/1327)).
```

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+        private::BooleanMetric, private::CounterMetric, private::StringMetric, CommonMetricData,
+    };
+    use std::sync::Arc;
+
+    /// The sealed labeled trait.
+    ///
+    /// This also allows us to hide methods, that are only used internally
+    /// and should not be visible to users of the object implementing the
+    /// `Labeled<T>` trait.
+    pub trait ToLabeled {
```
badboy, line 27: I don't like the name here, even if it's internal only.
`ToLabeled` suggests that we convert something to something that is labeled, but that's not what the trait here is doing.

The implementation here really would match the `AllowLabeled` trait, in m-c we just moved the methods here, so we can do the general implementation in line 90 easily.

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+        /// The `glean_core` metric type representing the labeled metric.
+        type InnerCoreMetric: glean_core::metrics::MetricType + Clone;
```
badboy, line 29: Can't we have this just be `type Inner`? that it is _core_ and that it is a _metric_ is implied by the bound on  `glean_core::metrics::MetricType`.

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+/// This trait is sealed and cannot be implemented for types outside this crate.
+pub trait AllowLabeled: private::ToLabeled {}
+
+// Implement the trait for everything we marked as allowed.
+impl<T> AllowLabeled for T where T: private::ToLabeled {}
+
+// We need to wrap the glean-core type: otherwise if we try to implement
+// the trait for the metric in `glean_core::metrics` we hit error[E0117]:
+// only traits defined in the current crate can be implemented for arbitrary
+// types.
```
badboy, line 95: The name of this concept is [orphan rules](https://blog.mgattozzi.dev/orphan-rules/). If we want to keep this comment we might as well add that name in case we ever need to look it up.

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+
+/// This implements the developer facing API for recording labeled metrics.
```
badboy, line 97:
```suggestion
/// This implements the specific API for recording labeled metrics.
```

or 

```suggestion
/// This implements the user-facing API for recording labeled metrics.
```

or at least

```suggestion
/// This implements the developer-facing API for recording labeled metrics.
```

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+///
+/// Instances of this class type are automatically generated by the parsers
```
badboy, line 99:
```suggestion
/// Instances of this type are automatically generated by the parser
```

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+        let _lock = lock_test();
+
+        let _ = new_glean(None, true);
+
+        let metric: LabeledMetric<CounterMetric> = LabeledMetric::new(
+            CommonMetricData {
+                name: "labeled_counter".into(),
+                category: "labeled".into(),
+                send_in_pings: vec!["test1".into()],
+                disabled: false,
```
badboy, line 194:
```suggestion
```

`false` is the default.

badboy: (let's start doing this properly and only set the non-default fields here and further below)

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+    use crate::common_test::{lock_test, new_glean};
+    use crate::destroy_glean;
+    use crate::private::{BooleanMetric, CounterMetric, LabeledMetric, StringMetric};
+    use crate::CommonMetricData;
+
+    #[test]
+    fn test_labeled_counter_type() {
+        let _lock = lock_test();
+
+        let _ = new_glean(None, true);
```
badboy, line 187: This drops the returned `TempDir` immediately, thus removing it. This kinda works here, because we keep the database open, but really we should keep the directory around for this test.

```suggestion
        let _t = new_glean(None, true);
```

badboy: same in the following tests.

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+        metric.get("bar").add(1);
+        metric.get("not_there").add(1);
+        metric.get("also_not_there").add(1);
+        metric.get("not_me").add(1);
+
+        assert_eq!(3, metric.get("foo").test_get_value(None).unwrap());
+        assert_eq!(1, metric.get("bar").test_get_value(None).unwrap());
+        assert!(metric.get("baz").test_get_value(None).is_none());
+        // The rest all lands in the __other__ bucket.
+        assert_eq!(3, metric.get("not_there").test_get_value(None).unwrap());
```
badboy, line 234: Comment says `__other__` while the code tries to get `not_there`.
We either should ask for `__other__`, or call out that this maps to `__other__` anyway

glean-core/rlb/src/private/labeled.rs
```diff
@@ -0,0 +1,389 @@
...
+                send_in_pings: vec!["test1".into()],
+                disabled: false,
+                ..Default::default()
+            },
+            None,
+        );
+
+        let invalid_label = "!#I'm invalid#--_";
+        metric.get(invalid_label).set(true);
+        // TODO: Uncomment after bug 1677807 is fixed.
```
badboy, line 382: Logging an error AND recording to the `__other__` label is kinda the behavior we stuck with ever since, so I doubt that the bug gets fixed in a way to make the below line pass.

Maybe make the assert explicitly check the `__other__` label to signal that the data _does_ get recorded?

glean-core/rlb/src/private/string.rs
```diff
@@ -28,6 +28,11 @@ impl StringMetric {
     pub fn new(meta: glean_core::CommonMetricData) -> Self {
         Self(Arc::new(glean_core::metrics::StringMetric::new(meta)))
     }
+
+    /// The internal constructor used by `LabeledMetric`.
+    pub(crate) fn new_from_meta(metric: Arc<glean_core::metrics::StringMetric>) -> Self {
```
badboy, line 33: Do we really need this method?
We could just as well do `StringMetric(metric)` where it is called.

Or we could `impl From<glean_core::metrics::StringMetric> for StringMetric` that then puts it in the `Arc` and returns `StringMetric`.

At the very least the name here is wrong. It's not constructing something new from `meta` (throughout the codebase we use the name `meta` only for `CommonMetricData` so far)

glean-core/rlb/src/lib.rs
```diff
@@ -519,10 +521,9 @@ pub(crate) fn destroy_glean(clear_stores: bool) {
     }
 }
 
+/// TEST ONLY FUNCTION.
 /// Resets the Glean state and triggers init again.
-#[cfg(test)]
```
badboy, line 523: Why is this not `cfg(test)` anymore?

badboy (changes requested):
Comments inline, good way, mostly stylistic/naming nits.

glean-core/rlb/src/lib.rs
Dexterp37: Because when writing tests in m-c, we need to reset the Glean state as well (other bindings have fixtures that they offer).

badboy (approved):

PR #1328, opened 2020-11-19
1678151: Record an error if UUID is invalid (affects only Python)

mdboom:
Even though for statically typed languages it is impossible to set the UUID to something invalid, there is still a hole for dynamically-typed languages, since we pass it over FFI as a string.

This solution is a little "deeper" than we might otherwise like, but it is currently impossible to record errors through the FFI, and even the `glean-ffi` itself can't record errors (since the API for it is all private).  But I think it's not terrible (the `glean-core` API isn't the real public RLB API anyway, so adding a new method there for this use doesn't seem that bad...)

### DIFF
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -786,6 +786,10 @@ uint64_t glean_new_uuid_metric(FfiStr category,
                                Lifetime lifetime,
                                uint8_t disabled);
 
+int32_t glean_uuid_test_get_num_recorded_errors(uint64_t metric_id,
+                                                int32_t error_type,
+                                                FfiStr storage_name);
+
 void glean_uuid_set(uint64_t metric_id, FfiStr value);
 
 uint8_t glean_uuid_test_has_value(uint64_t metric_id, FfiStr storage_name);
--- a/glean-core/ffi/src/uuid.rs
+++ b/glean-core/ffi/src/uuid.rs
@@ -13,6 +13,7 @@ use crate::{
 
 define_metric!(UuidMetric => UUID_METRICS {
     new           -> glean_new_uuid_metric(),
+    test_get_num_recorded_errors -> glean_uuid_test_get_num_recorded_errors,
     destroy       -> glean_destroy_uuid_metric,
 });
 
@@ -21,14 +22,7 @@ pub extern "C" fn glean_uuid_set(metric_id: u64, value: FfiStr) {
     with_glean_value(|glean| {
         UUID_METRICS.call_with_log(metric_id, |metric| {
             let value = value.to_string_fallible()?;
-            if let Ok(uuid) = uuid::Uuid::parse_str(&value) {
-                metric.set(glean, uuid);
-            } else {
-                log::error!(
-                    "Unexpected `uuid` value coming from platform code '{}'",
-                    value
-                );
-            }
+            metric.set_from_str(glean, &value);
             Ok(())
         })
     })
--- a/glean-core/src/metrics/uuid.rs
+++ b/glean-core/src/metrics/uuid.rs
@@ -4,6 +4,7 @@
 
 use uuid::Uuid;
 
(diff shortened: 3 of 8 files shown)


glean-core/ffi/glean.h
```diff
@@ -786,6 +786,10 @@ uint64_t glean_new_uuid_metric(FfiStr category,
                                Lifetime lifetime,
                                uint8_t disabled);
 
+int32_t glean_uuid_test_get_num_recorded_errors(uint64_t metric_id,
```
chutten, line 789: Why is the return type signed? Are negative amounts of errors expected?

glean-core/src/metrics/uuid.rs
```diff
@@ -54,6 +55,28 @@ impl UuidMetric {
...
+    /// Sets to the specified value, from a string.
+    ///
+    /// This should only be used from FFI. When calling directly from Rust, it
+    /// is better to use `set`.
+    ///
+    /// # Arguments
+    ///
+    /// * `glean` - The Glean instance this metric belongs to.
+    /// * `value` - The UUID to set the metric to.
+    pub fn set_from_str(&self, glean: &Glean, value: &str) {
```
chutten, line 67: I would like @badboy 's eyes on this. Maybe a better solution would be to preserve the strength of the typing of the metric API but open up a way to record errors from outside. (FOG would use such a capability since its errors might not be on the same process as the Glean singleton)

glean-core/ffi/glean.h
mdboom: Not sure -- but this is generated code from `cbindgen` for generated code in `macros.rs` (and thus is consistent with the other metrics).

glean-core/src/metrics/uuid.rs
mdboom: Yeah -- I thought we had a bug for that but I can't find it now.  It turns out to be tricky in a *generic* way that would work for all the metric types (without a new FFI for each of them).  Maybe that's the cost we are willing to pay / would be easier when uniffi is viable etc...

glean-core/ffi/glean.h
travis79: This is probably using a signed type because unsigned types are "experimental" in Kotlin.

badboy: Yup, Kotlin...

glean-core/src/metrics/uuid.rs
badboy: We have [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1589675) on a general refactor of the error reporting system, I also thought we had something more directly for exposing that over FFI.

Anyway, this probably requires a bit more work than we can afford right now to make this happen for the UUID type.

mdboom: @badboy: So you are saying we go forward with this (slightly) suboptimal solution now, or hold off until we have the time to do error reporting from the language bindings?

badboy: Go forward with this solution for now, UUIDs are already used and if this catches a potential bug in Python (and m-c?) code good.

chutten: It's not the worst in the world to have it be `pub` in `glean-core`. That's specifically not an API crate after all.

glean-core/ffi/src/uuid.rs
```diff
@@ -13,6 +13,7 @@ use crate::{
 
 define_metric!(UuidMetric => UUID_METRICS {
     new           -> glean_new_uuid_metric(),
+    test_get_num_recorded_errors -> glean_uuid_test_get_num_recorded_errors,
```
chutten, line 16: Can you remove the `dead_code` todo in `rlb/src/private/uuid.rs` now that `test_get_num_recorded_errors` is used?

chutten (approved):
One nit

PR #1329, opened 2020-11-19
bug 1677457 - Implement the UUID type in the RLB

chutten:


chutten:
May interfere with mdboom's #1328

glean-core/rlb/src/private/uuid.rs
```diff
@@ -0,0 +1,95 @@
...
+    ///
+    /// * `error` - The type of error
+    /// * `ping_name` - represents the optional name of the ping to retrieve the
+    ///   metric for. Defaults to the first value in `send_in_pings`.
+    ///
+    /// # Returns
+    ///
+    /// The number of errors reported.
+    #[allow(dead_code)] // Remove after mozilla/glean#1328
+    fn test_get_num_recorded_errors<'a, S: Into<Option<&'a str>>>(
```
Dexterp37, line 46: Let's make this public :)

glean-core/src/traits/uuid.rs
```diff
@@ -27,5 +27,5 @@ pub trait Uuid {
     ///
     /// * `ping_name` - represents the optional name of the ping to retrieve the
     ///   metric for. Defaults to the first value in `send_in_pings`.
-    fn test_get_value<'a, S: Into<Option<&'a str>>>(&self, ping_name: S) -> Option<String>;
+    fn test_get_value<'a, S: Into<Option<&'a str>>>(&self, ping_name: S) -> Option<uuid::Uuid>;
```
Dexterp37, line 30: We should also add `test_get_num_recorded_errors` to the trait

glean-core/src/metrics/uuid.rs
```diff
@@ -59,10 +59,9 @@ impl UuidMetric {
     /// # Arguments
     ///
     /// * `glean` - The Glean instance this metric belongs to.
-    pub fn generate_and_set(&self, storage: &Glean) -> Uuid {
+    pub fn generate_and_set(&self, storage: &Glean) {
```
Dexterp37, line 62: If we do this then this API suddenly becomes different with respect to the other bindings: we should either restore it here or change the other bindings, too (filing a bug should be ok if we don't have one already)

cc @badboy

Dexterp37 (changes requested):

glean-core/src/metrics/uuid.rs
chutten: Oops, right, I'd forgotten I'd only filed https://bugzilla.mozilla.org/show_bug.cgi?id=1673017 and it hadn't been processed/adopted. I definitely need to put this back for now.

I guess my Opinions were too strong when I wrote it. : )

chutten:
RFAL

Dexterp37 (approved):
Looks good! Would you kindly add a changelog entry and appease the CI gods before merging?

CHANGELOG.md
```diff
@@ -11,6 +11,7 @@
 * Rust
   * Introduce the Counter metric type in the RLB.
   * Introduce the String metric type in the RLB.
+  * Introduce the UUID metric type in the RLB.
```
mdboom, line 14: @chutten: Oops, I think this is in the wrong changelog section...

chutten: Oh bother.

chutten: https://github.com/mozilla/glean/pull/1333

PR #1330, opened 2020-11-20
Change the `Ping` trait to match the RLB API

Dexterp37:
This also adds a notice to the RLB type to mention to keep the trait updated.

### DIFF
--- a/glean-core/rlb/src/private/ping.rs
+++ b/glean-core/rlb/src/private/ping.rs
@@ -2,6 +2,8 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use inherent::inherent;
+
 /// A Glean ping.
 #[derive(Clone, Debug)]
 pub struct PingType {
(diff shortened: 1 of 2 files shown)


Dexterp37:
The beta error is being fixed in #1331

glean-core/rlb/src/private/ping.rs
```diff
@@ -2,13 +2,19 @@
...
 #[derive(Clone, Debug)]
 pub struct PingType {
     pub(crate) name: String,
     pub(crate) ping_type: glean_core::metrics::PingType,
 }
 
+// IMPORTANT:
+//
+// When changing this implementation, make sure all the operations are
+// also declared in the related trait in `../traits/`.
```
brizental, line 17: I don't understand this comment, 
1. I couldn't find the traits folder you are pointing to here; 
2. Shouldn't this comment be on line 47 before `impl glean_core::traits::Ping for PingType`? 
3. Wouldn't Rust error anyways if you don't implement a trait function that doesn't include a default?

brizental (approved):
r+wc

glean-core/rlb/src/private/ping.rs
Dexterp37: Gah, you're right. This comment should be removed. Copy pasta.

PR #1334, opened 2020-11-20
Bug 1677453: Implement the Quantity type in the RLB

mdboom:


glean-core/rlb/src/private/quantity.rs
```diff
@@ -0,0 +1,89 @@
...
+    /// # Returns
+    ///
+    /// The number of errors reported.
+    #[allow(dead_code)] // Remove after mozilla/glean#1328
+    fn test_get_num_recorded_errors<'a, S: Into<Option<&'a str>>>(
+        &self,
+        error: ErrorType,
+        ping_name: S,
+    ) -> i32 {
+        crate::with_glean_mut(|glean| {
```
Dexterp37, line 84: We should add `dispatcher::block_on_queue();` here, too, as we do for the [other types](https://github.com/mozilla/glean/blob/d13410ccc58398d584b61fc286f07f82e40e5d2f/glean-core/rlb/src/private/string.rs#L91).

Dexterp37 (approved):
Looks good! Let's merge with the commend below addressed :)

PR #1341, opened 2020-11-27
Bug 1679428 - Provide Glean debugging APIs for RLB consumers

Dexterp37:
This makes sure RLB consumers can directly enable logging, set debug tags or source tags by calling Glean APIs in addition to setting environment variables (e.g. this is needed for FOG's `about:glean`)

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -228,10 +228,8 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
 
             // The debug view tag might have been set before initialize,
             // get the cached value and set it.
-            if PRE_INIT_DEBUG_VIEW_TAG.get().is_some() {
-                // The following unwrap is safe, since we are checking that we have
-                // something right above.
-                let lock = PRE_INIT_DEBUG_VIEW_TAG.get().unwrap().try_lock();
+            if let Some(tag) = PRE_INIT_DEBUG_VIEW_TAG.get() {
+                let lock = tag.try_lock();
                 if let Ok(ref debug_tag) = lock {
                     glean.set_debug_view_tag(debug_tag);
                 }
@@ -244,10 +242,8 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
             }
             // The source tags might have been set before initialize,
             // get the cached value and set them.
-            if PRE_INIT_SOURCE_TAGS.get().is_some() {
-                // The following unwrap is safe, since we are checking that we have
-                // something right above.
-                let lock = PRE_INIT_SOURCE_TAGS.get().unwrap().try_lock();
+            if let Some(tags) = PRE_INIT_SOURCE_TAGS.get() {
+                let lock = tags.try_lock();
                 if let Ok(ref source_tags) = lock {
                     glean.set_source_tags(source_tags.to_vec());
                 }
@@ -587,26 +583,16 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
 ///
 /// # Returns
 ///
-/// This will return `false` in case `tag` is not a valid tag and `true` otherwise
-/// or if the tag is set before Glean is initialized.
+/// This will return `false` in case `tag` is not a valid tag and `true` otherwise.
+/// If called before Glean is initialized it will always return `true`.
 pub fn set_debug_view_tag(tag: &str) -> bool {
     if was_initialize_called() {
         with_glean_mut(|glean| glean.set_debug_view_tag(tag))
     } else {
         // Glean has not been initialized yet. Cache the provided tag value.
-        if PRE_INIT_DEBUG_VIEW_TAG.get().is_none() {
-            if PRE_INIT_DEBUG_VIEW_TAG
-                .set(Mutex::new(tag.to_string()))
-                .is_err()
-            {
-                log::error!(
-                    "Unable to set the debug view tag: already set. This probably happened concurrently."
-                );
-            }
-        } else {
-            let mut lock = PRE_INIT_DEBUG_VIEW_TAG.get().unwrap().lock().unwrap();
-            *lock = tag.to_string();
-        }
+        let m = PRE_INIT_DEBUG_VIEW_TAG.get_or_init(Default::default);
+        let mut lock = m.lock().unwrap();
+        *lock = tag.to_string();
         // When setting the debug view tag before initialization,
         // we don't validate the tag, thus this function always returns true.
         true
@@ -614,8 +600,6 @@ pub fn set_debug_view_tag(tag: &str) -> bool {
 }
 
 /// Sets the log pings debug option.
-///
-/// This will return `false` in case we are unable to set the option.
 ///
 /// When the log pings debug option is `true`,
 /// we log the payload of all succesfully assembled pings.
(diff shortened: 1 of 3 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -79,6 +79,11 @@ struct RustBindingsState {
 /// Note: The initialization might still be in progress, as it runs in a separate thread.
 static INITIALIZE_CALLED: AtomicBool = AtomicBool::new(false);
 
+/// Keep track of the debug features before Glean is initialized.
```
badboy, line 82: /o\ so much global state (I don't have a better idea at the moment though)

glean-core/rlb/src/lib.rs
```diff
@@ -221,6 +226,33 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
         with_glean_mut(|glean| {
             let state = global_state().lock().unwrap();
 
+            // The debug view tag might have been set before initialize,
+            // get the cached value and set it.
+            if PRE_INIT_DEBUG_VIEW_TAG.get().is_some() {
+                // The following unwrap is safe, since we are checking that we have
+                // something right above.
+                let lock = PRE_INIT_DEBUG_VIEW_TAG.get().unwrap().try_lock();
```
badboy, line 234: Couldn't this be:



```suggestion
            if let Some(tag) = PRE_INIT_DEBUG_VIEW_TAG.get() {
                let lock = tag.try_lock();
```

glean-core/rlb/src/lib.rs
```diff
@@ -221,6 +226,33 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
...
+            }
+            // The log pings debug option might have been set before initialize,
+            // get the cached value and set it.
+            let log_pigs = PRE_INIT_LOG_PINGS.load(Ordering::SeqCst);
+            if log_pigs {
+                glean.set_log_pings(log_pigs);
+            }
+            // The source tags might have been set before initialize,
+            // get the cached value and set them.
+            if PRE_INIT_SOURCE_TAGS.get().is_some() {
```
badboy, line 247: As above.

glean-core/rlb/src/lib.rs
```diff
@@ -541,5 +576,94 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
...
+/// value of the tag and are sent to the ["Ping Debug Viewer"](https://mozilla.github.io/glean/book/dev/core/internal/debug-pings.html).
+///
+/// # Arguments
+///
+/// * `tag` - A valid HTTP header value. Must match the regex: "[a-zA-Z0-9-]{1,20}".
+///
+/// # Returns
+///
+/// This will return `false` in case `tag` is not a valid tag and `true` otherwise
+/// or if the tag is set before Glean is initialized.
```
badboy, line 591: I find this sentence a bit ... harder to read than necessary.
What about changing it to cover the _exceptional_ case more clearly:

```suggestion
/// This will return `false` in case `tag` is not a valid tag and `true` otherwise.
/// If called before Glean is initialized it will always return `true`.
```

glean-core/rlb/src/lib.rs
```diff
@@ -541,5 +576,94 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
...
+                .is_err()
+            {
+                log::error!(
+                    "Unable to set the debug view tag: already set. This probably happened concurrently."
+                );
+            }
+        } else {
+            let mut lock = PRE_INIT_DEBUG_VIEW_TAG.get().unwrap().lock().unwrap();
+            *lock = tag.to_string();
+        }
```
badboy, line 609: Isn't this just `get_or_set`, then locking and setting the value?
E.g.

```suggestion
        let m = PRE_INIT_DEBUG_VIEW_TAG.get_or_init(Default::default);
        let mut lock = m.lock().unwrap();
        *lock = tag.to_string();
```

glean-core/rlb/src/lib.rs
```diff
@@ -541,5 +576,94 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
...
+    if was_initialize_called() {
+        with_glean_mut(|glean| glean.set_log_pings(value));
+    } else {
+        PRE_INIT_LOG_PINGS.store(value, Ordering::SeqCst);
+    }
+}
+
+/// Sets source tags.
+///
+/// Ping tags will show in the destination datasets, after ingestion.
```
badboy, line 636:
```suggestion
/// Overrides any existing source tags.
/// Source tags will show in the destination datasets, after ingestion.
```

glean-core/rlb/src/lib.rs
```diff
@@ -541,5 +576,94 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
...
+///
+/// # Arguments
+///
+/// * `value` - A vector of at most 5 valid HTTP header values. Individual
```
badboy, line 640: Should this be named `tags` instead?

glean-core/rlb/src/lib.rs
```diff
@@ -541,5 +576,94 @@ pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clea
...
+/// # Returns
+///
+/// This will return `false` in case `value` contains invalid tags and `true`
+/// otherwise or if the tag is set before Glean is initialized.
+pub fn set_source_tags(value: Vec<String>) -> bool {
+    if was_initialize_called() {
+        with_glean_mut(|glean| glean.set_source_tags(value))
+    } else {
+        // Glean has not been initialized yet. Cache the provided source tags.
+        if PRE_INIT_SOURCE_TAGS.get().is_none() {
```
badboy, line 652: As above the shorter `get_or_init` pattern seems better.

(I _think_ I initially came up with this `is_none` check then print if it still errors, but I'm not even sure why anymore)

glean-core/rlb/src/test.rs
```diff
@@ -354,6 +354,128 @@ fn test_dirty_flag_is_reset_to_false() {
...
 }
 
+#[test]
+fn setting_debug_view_tag_before_initialization_should_not_crash() {
+    let _lock = lock_test();
+
+    destroy_glean(true);
+    assert!(!was_initialize_called());
+
+    // Define a fake uploader that reports back the submission headers
```
badboy, line 364: Wonder if we should turn this pattern into a helper function eventually (not necessary to do now)

glean-core/rlb/src/test.rs
```diff
@@ -354,6 +354,128 @@ fn test_dirty_flag_is_reset_to_false() {
...
+
+    // Attempt to set a debug view tag before Glean is initialized.
+    set_debug_view_tag("valid-tag");
+
+    // Create a custom configuration to use a fake uploader.
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().display().to_string();
+
+    let cfg = Configuration {
+        data_path: tmpname,
```
badboy, line 392: only realizing it now, but .... whyyy did we make the data path a string? Aaanyway, not to be fixed here. ignore.

badboy (changes requested):

glean-core/rlb/src/lib.rs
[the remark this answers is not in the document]
Dexterp37: I don't have a good sense of this. It just makes me feel uncomfortable writing panicky code during the init path :(

glean-core/rlb/src/test.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1679793

glean-core/rlb/src/test.rs
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1679796

glean-core/rlb/src/lib.rs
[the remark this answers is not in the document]
badboy: Then keep this!

glean-core/rlb/src/lib.rs
```diff
@@ -221,6 +226,29 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
...
+                if let Ok(ref debug_tag) = lock {
+                    glean.set_debug_view_tag(debug_tag);
+                }
+            }
+            // The log pings debug option might have been set before initialize,
+            // get the cached value and set it.
+            let log_pigs = PRE_INIT_LOG_PINGS.load(Ordering::SeqCst);
+            if log_pigs {
+                glean.set_log_pings(log_pigs);
+            }
```
badboy, line 242: Hm, I guess it's not wrong to do the check here, because if you flip back and forth the `log pings` setting before enable you really only need to pass through a `true` value.

But not having the `if log_pings` would be just as correct.

badboy (approved):

glean-core/rlb/src/lib.rs
Dexterp37: You are correct!

PR #1342, opened 2020-11-30
Expose the core's `ErrorType` in the RLB

Dexterp37:
The `ErrorType` is required for consumers to use the testing APIs to count the number of reported errors.

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -45,7 +45,9 @@ use std::sync::Mutex;
 pub use configuration::Configuration;
 use configuration::DEFAULT_GLEAN_ENDPOINT;
 pub use core_metrics::ClientInfoMetrics;
-pub use glean_core::{global_glean, setup_glean, CommonMetricData, Error, Glean, Lifetime, Result};
+pub use glean_core::{
+    global_glean, setup_glean, CommonMetricData, Error, ErrorType, Glean, Lifetime, Result,
+};
 use private::RecordedExperimentData;
 
 mod configuration;


badboy (approved):

glean-core/rlb/src/lib.rs
```diff
@@ -45,7 +45,9 @@ use std::sync::Mutex;
 pub use configuration::Configuration;
 use configuration::DEFAULT_GLEAN_ENDPOINT;
 pub use core_metrics::ClientInfoMetrics;
-pub use glean_core::{global_glean, setup_glean, CommonMetricData, Error, Glean, Lifetime, Result};
+pub use glean_core::{
+    global_glean, setup_glean, CommonMetricData, Error, ErrorType, Glean, Lifetime, Result,
+};
```
chutten, line 50: Should we also be exporting the metric traits so (for example) FOG's metrics need not `use` `glean_core` at all?

Dexterp37: I have no strong opinion on this, but does it change anything at all other than using `glean_core::traits` vs `glean::traits`? @badboy , thoughts?

badboy: If we can drop a direct dependency on `glean_core` and instead only need to rely on `glean` we have less hassle of always updating the versions in sync.

Given that we just pulled a release we can do that in a followup though, no need to rush this out.

Dexterp37: Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1679963

PR #1344, opened 2020-11-30
Bug 1679857: Attempt to fix Python Windows tests

mdboom:


mdboom:
This was caused by an upstream networking error.  Closing.

mdboom:
The issue is that we can't download tarballs, build wheels from them and then install those wheels in this weird Windows-Python-running-inside-Wine configuration.  This never actually worked, it was just a warning before and these packages aren't actually necessary to run the unit tests.  The workaround is to install all of the wheels that were built when creating the host environment inside the target environment first.

Dexterp37 (approved):

PR #1346, opened 2020-12-01
Bug 1677807 - Document that invalid labels result in recordings going to the `__other__` label

badboy:
[doc only]


Dexterp37 (approved):

mdboom (changes requested):
Should we make these same changes to `labeled_counter.md` as well?

badboy:
> Should we make these same changes to `labeled_counter.md` as well?

I did ... and then didn't include it in the commit.

mdboom (approved):

PR #1347, opened 2020-12-01
Bug 1677455 - Implement a timespan for RLB. 

badboy:


badboy:
I'll take this for a spin in m-c to see that it actually works.

badboy:
Turns out we don't need the raw API because we already disallow timespans over IPC (and yet the raw API might be what we need if we ever allow timespans across IPC)

mdboom (changes requested):
I think this is missing changes to the `timespan.md` chapter in the book...

docs/user/metrics/timespan.md
```diff
@@ -288,6 +323,12 @@ TODO. To be implemented in [bug 1648442](https://bugzilla.mozilla.org/show_bug.c
 
 </div>
 
+<div data-lang="Rust" class="tab">
+
+The raw API is not supported in Rust.
```
badboy, line 328: cc @Dexterp37 -- I explicitly took out this API. I'm not sure we really need/should provide it on the Rust/FOG side right now. What do you think?

CHANGELOG.md
```diff
@@ -2,6 +2,10 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v33.5.0...main)
 
+* Rust
+  * BUGFIX: Negative timespans for the timespan metric now correctly record an `InvalidValue` error.
+  * Introduce the Timespan metric type in the RLB.
```
Dexterp37, line 7:
```suggestion
  * Introduce the Timespan metric type in the RLB [#1347](https://github.com/mozilla/glean/pull/1347).
```

glean-core/rlb/Cargo.toml
```diff
@@ -33,6 +33,7 @@ thiserror = "1.0.4"
 serde_json = "1.0.44"
 serde = { version = "1.0.104", features = ["derive"] }
 uuid = { version = "0.8.1", features = ["v4"] }
+time = "0.1.40" # Match version from mozilla-central
```
Dexterp37, line 36: Do we need the comment here? m-c needs to match basically with everything here, right?

glean-core/tests/timespan.rs
```diff
@@ -327,3 +327,27 @@ fn timespan_is_not_tracked_across_upload_toggle() {
         test_get_num_recorded_errors(&glean, metric.meta(), ErrorType::InvalidState, None)
     );
 }
+
+#[test]
+fn time_cannot_go_backwards() {
```
Dexterp37, line 332: Let's do the timeeee warppp agaaaaaaaaaaaaaain!! (nothing to do here)

glean-core/rlb/src/private/timespan.rs
```diff
@@ -0,0 +1,189 @@
...
+// only traits defined in the current crate can be implemented for arbitrary
+// types.
+
+/// This implements the developer facing API for recording timespan metrics.
+///
+/// Instances of this class type are automatically generated by the parsers
+/// at build time, allowing developers to record values that were previously
+/// registered in the metrics.yaml file.
+#[derive(Clone)]
+pub struct TimespanMetric(pub(crate) Arc<RwLock<glean_core::metrics::TimespanMetric>>);
```
Dexterp37, line 25: Why do we need a `RwLock` here? We're only expected to access the inner metric in the dispatched queue, FIFO-style.

Dexterp37 (changes requested):
Looks good, just minor changes requested below. I believe we should also still support `set_raw_nanos` given that this is required when using external clocks (we know that's a need and settled on this API a while back). See [usage in Fenix](https://searchfox.org/mozilla-mobile/source/fenix/app/src/main/java/org/mozilla/fenix/perf/StartupFrameworkStartMeasurement.kt#58) and [sync](https://searchfox.org/mozilla-mobile/source/android-components/components/support/migration/src/main/java/mozilla/components/support/migration/FennecMigrator.kt#709).

docs/user/metrics/timespan.md
Dexterp37: I agree on the "right now part", but we also know this covers the know use-case of using external clocks. We definitely know that sync will need it, for example. I'm fine with filing a bug for adding it later. I'd recommend reducing the amount of binding-specific APIs (or lack of APIs!) to not confuse consumers :)

glean-core/rlb/src/private/timespan.rs
badboy: glean-core's TimespanMetric requires `&mut self`, because it stores the start time internally. We therefore need a lock to even get a mutable reference.

glean-core/rlb/Cargo.toml
badboy:
```suggestion
time = "0.1.40"
```

docs/user/metrics/timespan.md
badboy:
```suggestion
The raw API is not supported in Rust. See [bug 1680225](https://bugzilla.mozilla.org/show_bug.cgi?id=1680225).
```

Dexterp37 (approved):

PR #1348, opened 2020-12-01
Bug 1680015: Rationalize Python logging

mdboom:


docs/user/debugging/python.md
```diff
@@ -17,15 +17,15 @@ If the `GLEAN_LOG_PINGS` environment variable is set to `true`, pings are
 logged to the console on `DEBUG` level whenever they are submitted.
 
 Make sure that when you configure logging in your application, you set the
-level for the `glean` logger to `DEBUG` or higher. Otherwise pings won't be
+level for the Python logger to `DEBUG` or higher. Otherwise pings won't be
```
badboy, line 20: Hm, the previous sentence was not completely wrong. As I understand it Glean uses a logger by the name of `glean.rust_core` which can be configured independently from what the application does for its own loggers.

docs/user/debugging/python.md
```diff
@@ -17,15 +17,15 @@ If the `GLEAN_LOG_PINGS` environment variable is set to `true`, pings are
...
 logged even if `GLEAN_LOG_PINGS` is set to `true`.
 
-You can set the logging level for Glean to `DEBUG` as follows:
+You can set the logging level for Python to `DEBUG` as follows:
 
 ```python
 import logging
 
-logging.getLogger("glean").setLevel(logging.DEBUG)
+logging.basicConfig(level=logging.DEBUG)
```
badboy, line 28: as above, this configures a basic logger (log to stdout) and changes the level to debug, whereas the old line only changed the glean logger itself and left the rest of the config to the application.

So maybe we need a combination here?
A "quick, give me everything to stdout" and a "here's how you configure only the glean part, the rest is up to you"

badboy (changes requested):

docs/user/debugging/python.md
mdboom: The messages coming from the Rust core use `glean.rust_core`, but other messages coming from Python come from loggers named after the Python module they come from (as is Python convention),e.g. `glean.metrics.string` etc.  The crappy bit about Python logging comes into play, though, because there is no way to say "for all loggers with a given prefix, set this level", so there's no way to set the level of all Glean logging at once.  So these old instructions never really worked (except for logging from the top-level module).

We could change the behavior so that everything uses a single logger called `glean` -- it's not really the Python convention, but it feels like the better fit for this library.

badboy (approved):
This will require a rebase so the changelog gets into the right place

badboy:
Rebased and merged as https://github.com/mozilla/glean/commit/2144da2021295c43d455b1d1ff722f01dcf11683

PR #1353, opened 2020-12-02
Upgrade to newer manylinux image

badboy:
manylinux10 is based on CentOS 6, which as of 2020-12-02 09:00 is dead.

From http://mirror.centos.org/centos/6/readme:

> The whole CentOS 6 is *dead* and *shouldn't* be used anywhere at *all*

We rely on installing an SSH client (probably to checkout the repository),
so CentOS 6 is no good for us.

manylinux2014 is CentOS 7.8, which is also unmaintained, but at least
continues to work for now.

[doc only]

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -852,20 +852,13 @@ jobs:
   pypi-linux-release:
     docker:
       # The official docker image for building manylinux2010 wheels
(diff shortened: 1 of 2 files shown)


badboy:
fwiw: for the latest 33.6.0 release I already ran the steps locally in a manylinux2014 container!

mdboom (approved):

PR #1355, opened 2020-12-02
Bug 1680230 - Upgrade rkv to v0.16.0 in preparation for m-c bump

badboy:


Dexterp37 (approved):
r+wc: would you kindly call this out in the changelog?

PR #1358, opened 2020-12-03
Bug 1675475 - Move the `UPLOAD_MANAGER` to the `RustBindingState` struct

yvonmanzi:


glean-core/rlb/src/lib.rs
```diff
@@ -256,8 +234,8 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
             // 1. Pings were submitted through Glean and it is ready to upload those pings;
             // 2. Upload is disabled, to upload a possible deletion-request ping.
             if pings_submitted || !upload_enabled {
-                let uploader = get_upload_manager().lock().unwrap();
-                uploader.trigger_upload();
+                let state = global_state().lock().unwrap();
+                state.upload_manager.trigger_upload();
```
badboy, line 238: We got a handle to the global state in line 200, locking it here will lead to a deadlock (and I think that's what might have happened in the test runs).
Fixing that is rather easy though: just use the existing state variable!

```suggestion
                state.upload_manager.trigger_upload();
```

glean-core/rlb/src/lib.rs
```diff
@@ -364,8 +342,8 @@ pub fn set_upload_enabled(enabled: bool) {
             if old_enabled && !enabled {
                 // If uploading is disabled, we need to send the deletion-request ping:
                 // note that glean-core takes care of generating it.
-                let uploader = get_upload_manager().lock().unwrap();
-                uploader.trigger_upload();
+                let state = global_state().lock().unwrap();
+                state.upload_manager.trigger_upload();
```
badboy, line 346: as above
```suggestion
                state.upload_manager.trigger_upload();
```

badboy (changes requested):

glean-core/rlb/src/lib.rs
yvonmanzi: Oh yeah you are right, there was a deadlock in tests! Thanks. I had assumed that it was the tests taking a little longer to run and pass, which is why I didn't think too much of it.

badboy (approved):

badboy:
I rebased this to pull it in, keeping your attribution of course.

PR #1360, opened 2020-12-04
Bug 1680477 - Provide names for the thread spawned in rust

Dexterp37:
This exclusively changes the way we spawn threads: instead of using `thread::spawn` we use a `Builder`.
Unfortunately, `cargo fmt` re-indents most of the code so the changes look much bigger.

### DIFF
--- a/glean-core/rlb/src/dispatcher/mod.rs
+++ b/glean-core/rlb/src/dispatcher/mod.rs
@@ -199,52 +199,57 @@ impl Dispatcher {
 
         let queue_preinit = Arc::new(AtomicBool::new(true));
 
-        let worker = thread::spawn(move || {
-            if block_receiver.recv().is_err() {
-                // The other side was disconnected.
-                // There's nothing the worker thread can do.
-                log::error!("The task producer was disconnected. Worker thread will exit.");
-                return;
-            }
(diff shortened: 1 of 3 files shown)


badboy:
Ignoring whitespace changes in diffs is possible on GitHub by appending `?w=1` to the url, e.g. https://github.com/mozilla/glean/pull/1360/files?w=1, which makes this much less scary

glean-core/rlb/src/dispatcher/mod.rs
```diff
@@ -199,52 +199,57 @@ impl Dispatcher {
...
-
-                        // The swap command MUST be the last one received on the preinit buffer,
-                        // so by the time we run this we know all preinit tasks were processed.
-                        // We can notify the other side.
-                        swap_done
-                            .send(())
-                            .expect("The caller of `flush_init` has gone missing");
-                    }
+        let worker = thread::Builder::new()
+            .name("glean dispatcher".into())
```
badboy, line 203: You use dots in the other names, so let's do it here too

```suggestion
            .name("glean.dispatcher".into())
```

badboy (changes requested):

badboy (approved):

PR #1361, opened 2020-12-04
Bug 1673668 - Implement the nice Rust API for events

badboy:
Pending some more opportunistic coding in m-c to make sure this API works.
Already creating the PR so I can get some early review in.

I did try this out in m-c already, so don't expect huge API changes requires anymore.


docs/user/metrics/event.md
```diff
@@ -144,7 +144,38 @@ Assert.Equal(2, snapshot.Length);
...
+use metrics::views;
+
+let mut extra = HashMap::new();
+extra.insert(views::LoginOpenedKeys::SourceOfLogin, "toolbar".into());
+views::login_opened.record(extra);
+```
+
+There are test APIs available too, for example:
+
+```C#
```
Dexterp37, line 166: Copy.. pasta! 🍝

glean-core/src/traits/event.rs
```diff
@@ -3,14 +3,86 @@
...
+    fn index(self) -> i32;
+}
+
+/// Default of no extra keys for events.
+///
+/// An enum with no values for convenient use as the default set of extra keys
+/// that an `EventMetric` can accept.
+///
+/// *Note*: There exist no values for this enum, it can never exist.
+/// It its equivalent to the [`never / !` type](https://doc.rust-lang.org/std/primitive.never.html).
```
Dexterp37, line 41: Thanks for providing this link!

glean-core/rlb/src/private/event.rs
```diff
@@ -0,0 +1,173 @@
...
+
+/// This implements the developer facing API for recording counter metrics.
+///
+/// Instances of this class type are automatically generated by the parsers
+/// at build time, allowing developers to record values that were previously
+/// registered in the metrics.yaml file.
+#[derive(Clone)]
+pub struct EventMetric<K> {
+    pub(crate) inner: Arc<glean_core::metrics::EventMetric>,
+    extra_keys: PhantomData<K>,
```
Dexterp37, line 28: TIL!

glean-core/rlb/src/private/event.rs
```diff
@@ -0,0 +1,173 @@
...
+use crate::{dispatcher, ErrorType, RecordedEvent};
+
+pub use glean_core::traits::NoExtraKeys;
+
+// We need to wrap the glean-core type: otherwise if we try to implement
+// the trait for the metric in `glean_core::metrics` we hit error[E0117]:
+// only traits defined in the current crate can be implemented for arbitrary
+// types.
+
+/// This implements the developer facing API for recording counter metrics.
```
Dexterp37, line 20:
```suggestion
/// This implements the developer facing API for recording event metrics.
```

glean-core/rlb/src/private/event.rs
```diff
@@ -0,0 +1,173 @@
...
+    }
+}
+
+#[inherent(pub)]
+impl<K: traits::ExtraKeys> traits::Event for EventMetric<K> {
+    type Extra = K;
+
+    fn record<M: Into<Option<HashMap<<Self as traits::Event>::Extra, String>>>>(&self, extra: M) {
+        const NANOS_PER_MILLI: u64 = 1_000_000;
+        let now = time::precise_time_ns() / NANOS_PER_MILLI;
```
Dexterp37, line 52: No strong opinion on this, but did you consider using `Duration::from_nanos(time::precise_time_ns())` and then call `as_millis` on it? Would mean one less constant :)

glean-core/rlb/src/private/event.rs
```diff
@@ -0,0 +1,173 @@
...
+            extra_keys: PhantomData,
+        }
+    }
+}
+
+#[inherent(pub)]
+impl<K: traits::ExtraKeys> traits::Event for EventMetric<K> {
+    type Extra = K;
+
+    fn record<M: Into<Option<HashMap<<Self as traits::Event>::Extra, String>>>>(&self, extra: M) {
```
Dexterp37, line 50: I really wonder what the sequence would be if we recorded events from both Kotlin and Rust. Aside from the dispatcher problem, I imagine timings would be different and things would look out of sequence in a weird way. I wish we could have "unified times"... provided by glean-core. Is this worth filing a bug?

Dexterp37 (approved):
r+wc, this looks really good. Only minor stuff in the comments.

glean-core/rlb/src/private/event.rs
badboy: I briefly did. IMO it's expressing the wrong thing. `precise_time_ns` doesn't specify a _duration_ = `a span of time`. It's just a timestamp (against an unknown but fixed epoch).

Now luckily the timestamps in events are always relative to the first event, so on ping assembly these get adjusted anyway.

glean-core/rlb/src/private/event.rs
badboy: Hm, that's going to be a problem for FOG as well for events in child processes.
Filing a bug.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1680739

PR #1362, opened 2020-12-04
Bug 1680474- Join on the dispatcher thread after shutting it down

Dexterp37:
This makes sure tools (e.g. TSAN) properly account for the thread being shut down and do not flag this as a pending
thread error.

### DIFF
--- a/glean-core/rlb/src/dispatcher/global.rs
+++ b/glean-core/rlb/src/dispatcher/global.rs
@@ -3,7 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use once_cell::sync::Lazy;
-use std::{mem, sync::RwLock};
+use std::sync::RwLock;
 
 use super::{DispatchError, DispatchGuard, Dispatcher};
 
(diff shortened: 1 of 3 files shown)


glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -63,9 +63,20 @@ pub fn flush_init() -> Result<(), DispatchError> {
...
+        .write()
+        .unwrap()
+        .as_mut()
+        .map(|dispatcher| {
+            if let Some(worker) = dispatcher.worker.take() {
+                return worker.join().map_err(|_| DispatchError::WorkerPanic);
+            }
+            Ok(())
+        })
+        .unwrap()
```
badboy, line 79:
```suggestion
    let dispatcher = GLOBAL_DISPATCHER
        .write()
        .unwrap()
        .as_mut()
        .expect("Global dispatcher has gone missing");
      
    if let Some(worker) = dispatcher.worker.take() {
        return worker.join().map_err(|_| DispatchError::WorkerPanic);
    }
    
    Ok(())
```

Dexterp37:
@badboy this is now ready for review. I tested this locally with a TSAN build and it no longer leaks when commenting [this line](https://hg.mozilla.org/mozilla-central/file/tip/mozglue/build/TsanOptions.cpp#l312) to suppress the .. suppression.

glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -63,9 +63,19 @@ pub fn flush_init() -> Result<(), DispatchError> {
 ///
 /// This will initiate a shutdown of the worker thread
 /// and no new tasks will be processed after this.
-/// It will not block on the worker thread.
 pub fn try_shutdown() -> Result<(), DispatchError> {
```
badboy, line 66: Should this be renamed to `shutdown`? We're not only trying anymore, we're actively waiting for it to shutdown.

glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -63,17 +63,27 @@ pub fn flush_init() -> Result<(), DispatchError> {
...
+    Ok(())
 }
 
 /// TEST ONLY FUNCTION.
 /// Resets the Glean state and triggers init again.
 pub(crate) fn reset_dispatcher() {
     // We don't care about shutdown errors, since they will
     // definitely happen if this
-    let _ = try_shutdown();
+    let _ = shutdown();
```
badboy, line 86: I can't comment on the lines below (damn you, github!), but line 92 (comment) is now kinda wrong and the code could be simplified I think.
As the worker thread is already `join`ed on, it's correctly shut down. I _think_ we don't even need the `mem::replace` anymore and a simple `*lock = new_dispatcher` should work now instead.

badboy (changes requested):

badboy (approved):

PR #1363, opened 2020-12-06
Unbreak Python on non-Linux ELF platforms (*BSD, Solaris/illumos)

myfreeweb:
This will fix not being able to pip install MozPhab on FreeBSD, etc.


badboy (approved):
Taking over review (so I can put it in a release). This looks good

PR #1368, opened 2020-12-07
Move the UPLOAD_MANAGER to the RustBindingState struct

badboy:
Rebase of #1358


badboy:
I already reviewed it initially so I'm going ahead with admin rights once CI is green

PR #1374, opened 2020-12-09
rkv: Handle all corrupted cases in the migration

badboy:
See upstream fix: https://github.com/mozilla/rkv/pull/211


Dexterp37 (approved):
Oh, nice catch.

PR #1375, opened 2020-12-09
Bug 1680789 - Properly shut down glean in tests (RLB)

Dexterp37:
FOG tests sometimes don't do anything other than initializing Glean. In these cases, nothing blocks on the dispatcher and tests might end up triggering init twice, in parallel, locking files on the disk or corrupting local data. This, in turn, makes FOG tests flaky and even RLB tests occasionally fail.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v33.7.0...main)
 
+* Rust
+  * Shut down Glean in tests before resetting to make sure they don't mistakenly init Glean twice in parallel.
+
 # v33.7.0 (2020-12-07)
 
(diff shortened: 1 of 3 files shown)


Dexterp37:
The code in this PR is currently being tested [in FOG CI](https://treeherder.mozilla.org/jobs?repo=try&revision=ca338f5bc9fe550c9317435d8d76b3a340566dc1). The BR jobs should not fail due to Glean problems :) Unfortunately, they can't be retriggered in a simple way, so Aryx is doing us the gigantic favour of triggering them manually. Moreover, they can't run in parallel, so Aryx will trigger them throughout his day. One by one.

badboy (approved):

Dexterp37:
All right, this indeed seems to fix the FOG problems!

PR #1376, opened 2020-12-09
Bug 1677451 - Implement Memory Distribution in the RLB

chutten:
Also, while I was here I added `test_get_num_recorded_errors` because any metric that records errors should expose this method (heck, maybe _all_ metrics should expose this method for consistency's sake)


glean-core/rlb/src/private/memory_distribution.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `ping_name` - represents the optional name of the ping to retrieve the
+    ///   metric for. Defaults to the first value in `send_in_pings`.
+    fn test_get_value_as_json_string<'a, S: Into<Option<&'a str>>>(
+        &self,
+        ping_name: S,
+    ) -> Option<String> {
+        self.test_get_value(ping_name)
```
Dexterp37, line 120: Let's call `dispatcher::block_on_queue();` here as well!

glean-core/rlb/src/private/memory_distribution.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// # Arguments
+    ///
+    /// * `sample` - The sample to be recorded by the metric. The sample is assumed to be in the
+    ///   configured memory unit of the metric.
+    ///
+    /// ## Notes
+    ///
+    /// Values bigger than 1 Terabyte (2<sup>40</sup> bytes) are truncated
+    /// and an `ErrorType::InvalidValue` error is recorded.
```
Dexterp37, line 47: Because @badboy found out that rustdocs correctly refer us to the trait docs if docstrings are missing here, he did not add docstrings to [events](https://searchfox.org/glean/source/glean-core/rlb/src/private/event.rs#47). Maybe we should to the same here? WDYT @badboy ? (and to other comments in this trait implementation.

glean-core/rlb/src/private/memory_distribution.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// # Returns
+    ///
+    /// The number of errors recorded.
+    fn test_get_num_recorded_errors<'a, S: Into<Option<&'a str>>>(
+        &self,
+        error: ErrorType,
+        ping_name: S,
+    ) -> i32 {
+        crate::with_glean_mut(|glean| {
```
Dexterp37, line 142: Let's call `dispatcher::block_on_queue();` here as well!

glean-core/rlb/src/private/memory_distribution.rs
```diff
@@ -0,0 +1,147 @@
...
+    ///
+    /// * `samples` - The vector holding the samples to be recorded by the metric.
+    ///
+    /// ## Notes
+    ///
+    /// Discards any negative value in `samples` and report an `ErrorType::InvalidValue`
+    /// for each of them.
+    /// Values bigger than 1 Terabyte (2<sup>40</sup> bytes) are truncated
+    /// and an `ErrorType::InvalidValue` error is recorded.
+    fn accumulate_samples_signed(&self, samples: Vec<i64>) {
```
Dexterp37, line 76: I wonder if we need this. We only really needed it to power the FFI layer for Kotlin. Maybe we should remove it from the bindings as well. @badboy , thoughts?

Dexterp37 (changes requested):
Thank you for doing this! It looks really good overall. There's a couple of things that we might simplify, I left a few comments above.

Would you kindly also add:

- a changelog entry;
- docs [here](https://searchfox.org/glean/rev/d77298da0e8026a9812c891d9a3b10c29a28a036/docs/user/metrics/memory_distribution.md#180)?

glean-core/rlb/src/private/memory_distribution.rs
chutten: Even though `test_get_value` calls it for me?

glean-core/rlb/src/private/memory_distribution.rs
chutten: If so, that'd help  me out because I have a broken link in the rustdocs as they're written : |

glean-core/rlb/src/private/memory_distribution.rs
chutten: Happy to remove it from the trait as part of this work if so.

glean-core/rlb/src/private/memory_distribution.rs
chutten: Good point!

glean-core/rlb/src/private/memory_distribution.rs
chutten: Hm. By that logic, should we also remove `test_get_value_as_json_string`? Rust consumers don't need/want it, to my knowledge.

Dexterp37: > 
> 
> Hm. By that logic, should we also remove `test_get_value_as_json_string`? Rust consumers don't need/want it, to my knowledge.

Yes, I believe we want to remove that as well, but let's see what @badboy thinks.

glean-core/rlb/src/private/memory_distribution.rs
Dexterp37: Oh, right. No need to do it here!

glean-core/rlb/src/private/memory_distribution.rs
badboy: Yup, let's do the minimal viable traits and expand them when we require new APIs

glean-core/rlb/src/private/memory_distribution.rs
badboy: Yup, rustdocs will give a link to the trait where the docs are then visible, that's why I left them out on events. I didn't go around removing them anywhere else, thought that could be done afterwards.
For new stuff IMO it's fine to leave them out

chutten:
RFAL

glean-core/rlb/src/private/memory_distribution.rs
mdboom: I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1681570) to do the follow-up work on removing docstrings where possible.

docs/user/metrics/memory_distribution.md
```diff
@@ -177,6 +177,37 @@ Assert.Equal(1, Memory.heapAllocated.TestGetNumRecordedErrors(ErrorType.InvalidV
...
+assert!(11, data.sum)
+// The actual buckets and counts live in `data.values`.
+
+// Were there any errors?
+assert!(1, memory::heap_allocated.test_get_num_recorded_errors(InvalidValue));
+
+```
+
+</div>
+
```
mdboom, line 210: Also add a link to the Rust API docs in the "Reference" section below.

mdboom (approved):
r+, wc

Dexterp37 (approved):

PR #1383, opened 2020-12-11
Bug 1679527 - Remove superfluous store iterations from snapshot_metric

brizental:


glean-core/src/storage/mod.rs
```diff
@@ -129,6 +129,7 @@ impl StorageManager {
         storage: &Database,
         store_name: &str,
         metric_id: &str,
+        metric_lifetime: Lifetime,
```
brizental, line 132: Looking at this again I wonder if it would be better to just pass &CommonMetricData directly to this function 🤔

badboy: But then we would also need to pass `glean` because otherwise you won't get the identifier (`self.meta().identifier(glean)`). But `StorageManager` shouldn't need to know about Glean, it should only know about the database.

brizental: Ah! You are right, better this way then.

badboy (approved):

PR #1384, opened 2020-12-11
Bug 1677448 - Implement the datetime metric in the RLB

Dexterp37:
This implements the datetime metric in the rust language bindings.

Nothing special, except for the addition of the `test_get_value` function: unfortunately, we need to do the truncation ourselves, which is a big quirky (and also the reason why we're not doing that for the other bindings!).


glean-core/src/metrics/datetime.rs
```diff
@@ -143,6 +143,46 @@ impl DatetimeMetric {
...
+                match tu {
+                    TimeUnit::Nanosecond => d.date().and_hms_nano_opt(time.hour(), time.minute(), time.second(), time.nanosecond()),
+                    TimeUnit::Microsecond =>d.date().and_hms_nano_opt(time.hour(), time.minute(), time.second(), time.nanosecond() / 1000),
+                    TimeUnit::Millisecond => d.date().and_hms_nano_opt(time.hour(), time.minute(), time.second(), time.nanosecond() / 1000000),
+                    TimeUnit::Second => d.date().and_hms_nano_opt(time.hour(), time.minute(), time.second(), 0),
+                    TimeUnit::Minute => d.date().and_hms_nano_opt(time.hour(), time.minute(), 0, 0),
+                    TimeUnit::Hour => d.date().and_hms_nano_opt(time.hour(), 0, 0, 0),
+                    TimeUnit::Day => d.date().and_hms_nano_opt(0, 0, 0, 0),
+                }
+            },
```
Dexterp37, line 181: @badboy is there any better way to do this?

glean-core/rlb/src/private/datetime.rs
```diff
@@ -0,0 +1,134 @@
...
+            },
+            TimeUnit::Day,
+        );
+
+        // Record a date: it will get truncated to Day resolution.
+        metric.set_with_details(2018, 2, 25, 11, 5, 0, 0, 0);
+
+        // Check that the value has the correct resolution.
+        let date = metric.test_get_value(None).unwrap();
+        assert_eq!(
```
Dexterp37, line 123: This is currently weirdly failing because it says that the dates are different. However, printing both of them yields the same result: `2018-02-25T00:00:00+00:00`. Any clue what could be the culprit, @badboy ?

glean-core/src/metrics/datetime.rs
badboy: Hm, not that I know off.
This seems long but IMO is very expressive and easy to read, so I'm not to worried about it

glean-core/rlb/src/private/datetime.rs
badboy: Is it failing? On CI the next assert is failing only

glean-core/src/traits/datetime.rs
```diff
@@ -4,6 +4,8 @@
 
 #![allow(clippy::too_many_arguments)]
 
+use crate::ErrorType;
```
badboy, line 7: Can't comment on unmodified lines, so put it here:

We should move the `fn set` function above the `set_with_details` one, so it's the first thing visible in the file/in the docs.

I wonder ... do we really need to make that `set_with_details` part of the _exposed_ API (the trait) or can we put it on the type only (thus making it more "internal", but FOG can still use it)?
What do you think?

glean-core/rlb/src/private/datetime.rs
Dexterp37: Ahhhh that's what happens when you stare at the same thing too much. And that's why it didn't make any sense in my brain :D

glean-core/src/traits/datetime.rs
Dexterp37: > 
> 
> Can't comment on unmodified lines, so put it here:
> 
> We should move the `fn set` function above the `set_with_details` one, so it's the first thing visible in the file/in the docs.

Good idea!

> I wonder ... do we really need to make that `set_with_details` part of the _exposed_ API (the trait) or can we put it on the type only (thus making it more "internal", but FOG can still use it)?
> What do you think?

Your call :-) But why do we need this in FOG? And can this be a follow-up?

badboy: Right now for the C++ API.
We _could_ leave it out, but then FOG would need to replicate [what [`glean-core`] is already doing](https://github.com/mozilla/glean/blob/141ae1da58b9da794ebbee5ac34f189fb9b4ff6d/glean-core/src/metrics/datetime.rs#L83-L94).

Dexterp37: I'm torn on this. On one hand, this is already done by glean-core, as you said. On the other, we're adding API surface just to not have FOG copy paste these lines :)

I'd be fore removing the API completely and have FOG copy/paste that :D But it's really your call.

docs/user/metrics/datetime.md
```diff
@@ -176,6 +176,36 @@ Assert.Equal(1, Install.firstRun.TestGetNumRecordedErrors(ErrorType.InvalidValue
...
+```
+
+There are test APIs available too.
+
+```Rust
+// Was anything recorded?
+assert!(metrics.install.first_run.test_get_value(None).is_some());
+
+// Was it the expected value?
+assert_eq!("2019-03-25-05:00", metrics.install.first_run.test_get_value(None));
```
mdboom, line 200: Does comparing a datetime to a string actually work?  (I'm not saying it doesn't, but the test below doesn't do that ... maybe we should add that to the test to make sure this example works...)

Dexterp37: Whoops, nope, the docs are wrong. Good catch.

glean-core/rlb/src/private/datetime.rs
```diff
@@ -0,0 +1,133 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+#![allow(clippy::too_many_arguments)]
```
badboy, line 5: Should we move this only to the corresponding function below?

glean-core/rlb/src/private/datetime.rs
```diff
@@ -0,0 +1,133 @@
...
+}
+
+#[inherent(pub)]
+impl glean_core::traits::Datetime for DatetimeMetric {
+    fn set(&self, value: Option<Datetime>) {
+        let metric = Arc::clone(&self.0);
+        dispatcher::launch(move || crate::with_glean(|glean| metric.set(glean, value)));
+    }
+
+    fn set_with_details(
```
badboy, line 45: Like this:
```suggestion
	#[allow(clippy::too_many_arguments)]
    fn set_with_details(
```

glean-core/src/traits/datetime.rs
badboy: Then I call it: let's remove that API! (... wait for chutten to complain)

glean-core/rlb/src/private/datetime.rs
badboy: (see below, I'm actually for removing this API)

badboy (changes requested):

docs/user/metrics/datetime.md
```diff
@@ -176,6 +176,36 @@ Assert.Equal(1, Install.firstRun.TestGetNumRecordedErrors(ErrorType.InvalidValue
...
+<div data-lang="Rust" class="tab">
+
+```Rust
+use glean::ErrorType;
+use glean_metrics;
+
+use chrono::{FixedOffset, TimeZone};
+
+// Records "now"
+install::first_run.set(None);
```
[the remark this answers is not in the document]
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1682244

docs/user/metrics/datetime.md
```diff
@@ -176,6 +176,37 @@ Assert.Equal(1, Install.firstRun.TestGetNumRecordedErrors(ErrorType.InvalidValue
...
+
+```Rust
+// Was anything recorded?
+assert!(metrics.install.first_run.test_get_value(None).is_some());
+
+// Was it the expected value?
+let expected_date = FixedOffset::east(0).ymd(2019, 3, 25).and_hms(0, 0, 0);
+assert_eq!(expected_date, metrics.install.first_run.test_get_value(None));
+// Was the value invalid?
+assert_eq!(1, install::first_run.test_get_num_recorded_errors(
```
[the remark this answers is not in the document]
Dexterp37: Let me fix that for all the examples on this page.

glean-core/rlb/src/private/datetime.rs
Dexterp37: You're right :)

badboy (approved):

PR #1385, opened 2020-12-11
Bug 1680224 - Return a specific wait interval value with PingUploadTask::Wait

brizental:
I did not include these changes in C#, as I was not sure it is necessary. 

Please let me know if I should :)

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -118,7 +118,7 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                     LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
                 }
                 is PingUploadTask.Wait -> SystemClock.sleep(action.time)
-                PingUploadTask.Done -> return Result.success()
+                is PingUploadTask.Done -> return Result.success()
             }
         } while (true)
         // Limits are enforced by glean-core to avoid an inifinite loop here.
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -52,7 +52,6 @@ internal inline fun <reified W : Worker> buildWorkRequest(tag: String): OneTimeW
 class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(context, params) {
     companion object {
         internal const val PING_WORKER_TAG = "mozac_service_glean_ping_upload_worker"
-        internal const val THROTTLED_BACKOFF_MS = 60_000L
 
         /**
          * Function to aid in properly enqueuing the worker in [WorkManager]
@@ -118,7 +117,7 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                     // Process the upload response
                     LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
                 }
-                PingUploadTask.Wait -> SystemClock.sleep(THROTTLED_BACKOFF_MS)
+                is PingUploadTask.Wait -> SystemClock.sleep(action.time)
                 PingUploadTask.Done -> return Result.success()
             }
         } while (true)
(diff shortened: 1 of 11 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -118,7 +117,7 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                     // Process the upload response
                     LibGleanFFI.INSTANCE.glean_process_ping_upload_response(incomingTask, result)
                 }
-                PingUploadTask.Wait -> SystemClock.sleep(THROTTLED_BACKOFF_MS)
+                is PingUploadTask.Wait -> SystemClock.sleep(action.time)
```
badboy, line 120: Why is it we use `is <item>` here but not on the line below? Can we make that the same on all match arms for consistency?

glean-core/src/upload/mod.rs
```diff
@@ -98,7 +106,12 @@ impl RateLimiter {
         }
 
         if self.count == self.max_count {
-            return RateLimiterState::Throttled;
+            let remaining = self.interval.as_millis() - self.elapsed().as_millis();
+            return RateLimiterState::Throttled(
+                remaining
+                    .try_into()
+                    .unwrap_or(self.interval.as_secs() * 1000),
```
badboy, line 113: Why doing the conversion manually instead of using `as_millis`?

glean-core/src/upload/mod.rs
```diff
@@ -98,7 +106,12 @@ impl RateLimiter {
         }
 
         if self.count == self.max_count {
-            return RateLimiterState::Throttled;
+            let remaining = self.interval.as_millis() - self.elapsed().as_millis();
```
badboy, line 109: Maybe worth calling out here that elapsed is always smaller than interval, because we already check that in `should_reset`?

glean-core/src/upload/mod.rs
```diff
@@ -135,6 +151,24 @@ pub enum PingUploadTask {
     Done,
 }
 
+impl PingUploadTask {
+    /// Whether the current task is an upload task.
+    pub fn is_upload(&self) -> bool {
+        if let PingUploadTask::Upload(_) = self {
+            return true;
+        }
+        false
```
badboy, line 160:
```suggestion
        matches!(self, PingUploadTask::Upload(_))
```

glean-core/src/upload/mod.rs
```diff
@@ -135,6 +151,24 @@ pub enum PingUploadTask {
...
+    }
+
+    /// Whether the current task is wait task.
+    pub fn is_wait(&self) -> bool {
+        if let PingUploadTask::Wait(_) = self {
+            return true;
+        }
+        false
```
badboy, line 168:
```suggestion
        matches!(self, PingUploadTask::Wait(_))
```

glean-core/src/upload/mod.rs
```diff
@@ -32,6 +32,8 @@ mod policy;
 mod request;
 mod result;
 
+const WAIT_TIME_FOR_PING_PROCESSING: u64 = 500; // in milliseconds
```
badboy, line 35: Python and Swift take the sleep time in seconds, so we convert that, but that results in a 0s sleep. Might still result in enough of a context switch to make this work in the general case, but it seems the 1s we have right now might be better?

badboy (changes requested):

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
brizental: Ugh, nice catch, this is a leftover from development. I has a bug and thought that was the fix, but it wasn't. I'll just remove it.

glean-core/src/upload/mod.rs
brizental: `as_millis` returns a `u128` which we need to `try_into` a `u64`. If that is unsuccesfull I make it default to the interval value through `as_secs` which already returns a `u64`.

glean-core/src/upload/mod.rs
brizental: TIL!

glean-core/src/upload/mod.rs
brizental: Ah! I thought it would becaose 0.5s. I'll change it to 1s, makes sense.

glean-core/src/upload/mod.rs
badboy: Ah, someone would need to use a huuuuge interval :D 
But that also means the first try_into is unlikely to fail, so we can keep it as is.

badboy (approved):
Can you file a bug so we explicitly take this for a test run in Fenix? Seems enough of a change to require manual verification.

brizental:
Filed! https://bugzilla.mozilla.org/show_bug.cgi?id=1682256

PR #1390, opened 2020-12-14
Fix broken tree

Dexterp37:
This adds the missing parameter to the snapshot function,
it was missing due to a wrong rebase.


mdboom (approved):
r+ (once tests pass)

PR #1391, opened 2020-12-14
Document that timespan doesn't overwrite and records `InvalidState`

badboy:
In [bug 1562859](https://bugzilla.mozilla.org/show_bug.cgi?id=1562859) we changed how this work: a timespan never overwrites a previous value, not even with `set_raw_nanos`.
We didn't document the error it records though.
We also didn't remove the unused parameter.

This tackles both.


Dexterp37 (approved):
Oh, great catch.

PR #1392, opened 2020-12-14
Bug 1677150: Panic before blocking

badboy:
This ensures a test case like the following doesn't hang, but explodes:

1. create a metric
2. check what it contains
3. init glean

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -320,6 +320,17 @@ pub fn shutdown() {
     }
 }
 
+/// Block on the dispatcher emptying.
+///
+/// This will panic if called before Glean is initialized.
+fn block_on_dispatcher() {
+    assert!(
+        was_initialize_called(),
+        "initialize was never called. Can't block on the dispatcher queue."
+    );
+    dispatcher::block_on_queue()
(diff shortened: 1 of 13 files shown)


Dexterp37:
@badboy did you check this in m-c? I think it might have repercussions there

badboy:
> @badboy did you check this in m-c? I think it might have repercussions there

will do

badboy:
Tests are fine on m-c against this branch.

glean-core/rlb/src/lib.rs
```diff
@@ -320,6 +320,17 @@ pub fn shutdown() {
     }
 }
 
+/// Block on the dispatcher emptying.
+///
+/// This will panic if called before Glean is initialized.
+fn block_on_queue() {
```
Dexterp37, line 326: Since this is now living in the RLB lib, what do you think about giving it a more specific name? For example `block_on_dispatcher`?

Dexterp37 (approved):
r+ with the optional name change.

glean-core/rlb/src/lib.rs
badboy: works for me.

PR #1393, opened 2020-12-14
Bug 1673850 - Allow ping registration before RLB init

Dexterp37:
This is needed so that products using Glean (and tests) could register pings at an arbitrary time, even if the Glean init is deferred a bit.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   * Introduce the String List metric type in the RLB. [#1380](https://github.com/mozilla/glean/pull/1380).
   * Introduce the `Datetime` metric type in the RLB [#1384](https://github.com/mozilla/glean/pull/1384).
   * Introduce the `CustomDistribution` and `TimingDistribution` metric type in the RLB [#1394](https://github.com/mozilla/glean/pull/1394).
-  * Support ping registration before Glean initializes.
+  * Support ping registration before Glean initializes [#1393](https://github.com/mozilla/glean/pull/1393).
 
 # v33.8.0 (2020-12-10)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   * Introduce the String List metric type in the RLB. [#1380](https://github.com/mozilla/glean/pull/1380).
   * Introduce the `Datetime` metric type in the RLB [#1384](https://github.com/mozilla/glean/pull/1384).
   * Introduce the `CustomDistribution` and `TimingDistribution` metric type in the RLB [#1394](https://github.com/mozilla/glean/pull/1394).
+  * Support ping registration before Glean initializes.
 
 # v33.8.0 (2020-12-10)
 
--- a/docs/user/pings/custom.md
+++ b/docs/user/pings/custom.md
@@ -220,6 +220,16 @@ Pings.search.Submit(
 
 </div>
 
+<div data-lang="Rust" class="tab">
+
+```Rust
+use glean::Pings;
+
+Pings::search.submit(Pings::searchReasonCodes::performed);
+```
+
+</div>
+
 {{#include ../../tab_footer.md}}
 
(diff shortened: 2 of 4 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -411,6 +422,13 @@ pub fn register_ping_type(ping: &private::PingType) {
                 glean.register_ping_type(&ping.ping_type);
             })
         })
+    } else {
+        // We need to keep track of pings, so they get re-registered after a reset.
+        // This state is kept across Glean resets, which should only ever happen in test mode.
+        // It's a set and keeping them around forever should not have much of an impact.
+        let m = PRE_INIT_PING_REGISTRATION.get_or_init(Default::default);
+        let mut lock = m.lock().unwrap();
+        (*lock).push(*ping);
```
Dexterp37, line 431: @badboy this really makes Rust unhappy, and I think I found a solution, but that got me into a different rabbit-hole, so I'm questioning my approach.

This line errors out because I'm trying to move `ping` (thanks borrow checker!).

If i store references in the global vector (i.e. `&private::PingType`), then this whole file gets happy. However, it requires me to require a `static` lifetime on the `register_ping_type` function, which in turns makes [this](https://searchfox.org/glean/rev/e9e1f0a83036cc62a63ea9adb34bff90f7a54a20/glean-core/rlb/src/private/ping.rs#38) unhappy. We know that, in practice, our generated code will always use `static` lifetime, so this change seems ok, but I can't find a way to communicate this to the compiler. Or is this whole approach wrong?

badboy: What if we clone the ping? We can worry about making it more efficient later (and it's not like we register hundreds of custom pings).
(You also don't need the `(*lock)`, `lock` should suffice)

Dexterp37: Good point.

Dexterp37:
I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1682538) to document testing pings.

CHANGELOG.md
```diff
@@ -5,6 +5,7 @@
 * Rust
   * Introduce the String List metric type in the RLB. [#1380](https://github.com/mozilla/glean/pull/1380).
   * Introduce the `Datetime` metric type in the RLB [#1384](https://github.com/mozilla/glean/pull/1384).
+  * Support ping registration before Glean initializes.
```
badboy, line 8:
```suggestion
  * Support ping registration before Glean initializes [#1393](https://github.com/mozilla/glean/pull/1393).
```

docs/user/pings/custom.md
```diff
@@ -220,6 +220,16 @@ Pings.search.Submit(
 
 </div>
 
+<div data-lang="Rust" class="tab">
+
+```Rust
+use glean::Pings;
+
+Pings::search.submit(Pings::searchReasonCodes::performed);
```
badboy, line 228:
```suggestion
pings::search.submit(pings::SearchReasonCodes::Performed);
```
we went for lowercase module and ping names, but the reason codes will be an enum, so they should be CamelCased.

glean-core/rlb/src/test.rs
```diff
@@ -481,3 +483,63 @@ fn setting_source_tags_before_initialization_should_not_crash() {
...
+            net::UploadResult::HttpStatus(200)
+        }
+    }
+
+    // Create a custom ping and attempt its registration.
+    #[allow(non_upper_case_globals)]
+    pub static SamplePing: Lazy<PingType> =
+        Lazy::new(|| PingType::new("pre-register", true, true, vec![]));
+
+    register_ping_type(&SamplePing);
```
badboy, line 519: For testing purposes we shouldn't need a static, it definitely doesn't need to be `pub`.
`register_ping_type` is also internally called on `PingType::new`, so I _think_ we can change this to just:

```suggestion
    let sample_ping = PingType::new("pre-register", true, true, vec![]);
```

badboy (changes requested):

badboy (approved):

PR #1395, opened 2020-12-15
Update identifier in string metric `set` docstring

brizental:
[doc only]

Randomly found this bug while reading through code.


badboy:
We _also_ have `MAX_STRING_LENGTH`, but that's in `string_list.rs` and it has a different length.
Also the same comment is used in `traits/string.rs`, so we should also change it there.

travis79 (approved):

PR #1398, opened 2020-12-16
Bug 1682638 - Multiple fixes: Don't let the queue run, handle empty databases 

badboy:
I identified 2 problems, that in combination cause the crashes in bug 1682638.

1. Panic on shutdown

See [document](https://hackmd.io/5yej-1q7THe023-GpQ9yxQ).
When init fails for some reason we never set a global Glean object. However on shutdown we try to unblock the dispatcher, which might have some recording tasks in there already. These should never be executed without a Glean.

The first 4 commits address that: we know when it is initialized, so we can avoid emptying the dispatcher if it hasn't.

2. Empty database files

For reasons unknown to me right now the database file was empty on at least the machine of one tester.
Rkv in safe-mode considers empty files as invalid files. I filed this [upstream](https://github.com/mozilla/rkv/issues/212).
The solution (for now) is to remove that file and start over again.

---

For both cases I added new tests.


glean-core/rlb/tests/fog_simple.rs
```diff
@@ -0,0 +1,89 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This test should be conceptually similar to how the RLB is used from e.g. FOG.
```
Dexterp37, line 5: Do we need to mention FOG at all here? It seems very out of context. We have a [full init example](https://searchfox.org/glean/source/glean-core/examples/sample.rs). Maybe we should remove the example and just turn that into a `simple_usage_test.rs`?

glean-core/rlb/tests/fog_simple.rs
```diff
@@ -0,0 +1,89 @@
...
+        Lazy::new(|| glean::private::PingType::new("fog-validation", true, true, vec![]));
+}
+
+/// Test scenario: A clean run
+///
+/// FOG is initialized, in turn Glean gets initialized without problems.
+/// Some data is recorded (before and after initialization).
+/// And later the whole process is shutdown.
+#[test]
+fn fog_lifecycle() {
```
Dexterp37, line 57: Was this breaking without your changes?

glean-core/rlb/tests/fog_init_fails.rs
```diff
@@ -0,0 +1,89 @@
...
+//! (different files run as different processes).
+
+mod common;
+
+use std::{thread, time::Duration};
+
+use glean::Configuration;
+
+/// Some user metrics.
+mod metrics {
```
Dexterp37, line 17: The FOG* tests feel a bit out of place in here. Can we remove the references to FOG or move them in m-c?

glean-core/rlb/tests/common/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+pub fn enable_test_logging() {
+    // When testing we want all logs to go to stdout/stderr by default,
+    // without requiring each individual test to activate it.
+    // This only applies to glean-core tests, users of the main library still need to call
+    // `glean_enable_logging` of the FFI component (automatically done by the platform wrappers).
+    let _ = env_logger::builder().is_test(true).try_init();
+}
+
+/// Exit the whole process when a panic occurs.
```
Dexterp37, line 21: Can you please call out why we need the panic handlers somewhere in this file?

glean-core/rlb/tests/common/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+use std::{panic, process};
+
+use glean::{ClientInfoMetrics, Configuration};
+
+pub fn enable_test_logging() {
+    // When testing we want all logs to go to stdout/stderr by default,
+    // without requiring each individual test to activate it.
+    // This only applies to glean-core tests, users of the main library still need to call
+    // `glean_enable_logging` of the FFI component (automatically done by the platform wrappers).
```
chutten, line 17: RLB doesn't.

glean-core/rlb/tests/common/mod.rs
chutten, line 21: Isn't this the same as running with panic=abort?

glean-core/rlb/tests/common/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+    let orig_hook = panic::take_hook();
+    panic::set_hook(Box::new(move |panic_info| {
+        // invoke the default handler and exit the process
+        orig_hook(panic_info);
+        process::exit(1);
+    }));
+}
+
+// Create a new instance of Glean with a temporary directory.
+// We need to keep the `TempDir` alive, so that it's not deleted before we stop using it.
```
chutten, line 32: This comment is interesting because TempDir isn't referenced at all in this function. What's its deal?

chutten (commented):
Drive by comment pile.

glean-core/rlb/tests/fog_simple.rs
badboy: integration tests run as part of `cargo test`, which is kinda nice.
(I see them similar to our sample apps in Kotlin and Swift).
FOG is mentioned because the patterns here are modeled the way we use them in FOG.

glean-core/rlb/tests/fog_simple.rs
badboy: Nope, the good path always worked.

glean-core/rlb/tests/fog_init_fails.rs
badboy: As above: I'd like to have these integration tests here. Much quicker to test _before_ we have to do the hard cycle of putting it into m-c, compiling for 20 minutes and then hope we trigger stuff there.

glean-core/rlb/tests/common/mod.rs
badboy: copy-pasta, will fix

glean-core/rlb/tests/common/mod.rs
badboy: copy-pasta, will fix

glean-core/rlb/tests/common/mod.rs
badboy: Yep, but we run this as part of `cargo test`. Without this the test might hang if we use `test_get_value` or other things that block on the dispatcher (which ... maybe I should also fix that -> followup)

glean-core/rlb/tests/common/mod.rs
```diff
@@ -0,0 +1,42 @@
...
+
+// #[allow(dead_code)] is required on this module as a workaround for
+// https://github.com/rust-lang/rust/issues/46379
+#![allow(dead_code)]
+
+use std::{panic, process};
+
+use glean::{ClientInfoMetrics, Configuration};
+
+pub fn enable_test_logging() {
```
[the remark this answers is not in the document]
badboy: Unfortunately that's in glean-core's test part. We can't reach across those crates, so I duplicated it. Given that it's _only_ logging and _only_ for tests IMO that's fine.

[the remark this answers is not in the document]
Dexterp37: Shoot, ok!

glean-core/rlb/tests/fog_init_fails.rs
Dexterp37: Sure, but can we drop the `fog` word? It's confusing because there's nothing about FOG in this repo. These, to me, are simply integration tests for the RLB.

Dexterp37 (changes requested):
@badboy would you consider dropping the FOG prefix from the filenames and the references in the test files?

Dexterp37 (approved):
Thank you :)

PR #1399, opened 2020-12-17
Merge back v33.9.1

badboy:


bin/prepare-release.sh
```diff
@@ -43,7 +43,7 @@ fi
 NEW_VERSION="$1"
 DATE=$(date +%Y-%m-%d)
 
-if ! echo "$NEW_VERSION" | grep --quiet --extended-regexp '^[0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9.-]+)?'; then
+if ! echo "$NEW_VERSION" | grep --quiet --extended-regexp '^[0-9]+\.[0-9]+\.[0-9]+(-[a-z0-9.-]+)?$'; then
```
Dexterp37, line 46: Ah, interesting. How did you find out about this/how did you trigger the problem?

Dexterp37 (approved):

bin/prepare-release.sh
badboy: I typed: `bin/prepare-release.sh 33.9.1.` by accident (trailing dot), which passed the regex but then failed when cargo stumbled upon it.

Dexterp37: Got it, thanks. I'm fixing this in Rally, too :)

PR #1413, opened 2021-01-04
Clippy fixes

mdboom:
Fixes to some various clippy checks that appeared over the break.

### DIFF
--- a/glean-core/rlb/tests/schema.rs
+++ b/glean-core/rlb/tests/schema.rs
@@ -59,7 +59,7 @@ fn validate_against_schema() {
     #[derive(Debug)]
     pub struct ValidatingUploader {
         sender: crossbeam_channel::Sender<Vec<u8>>,
-    };
+    }
     impl glean::net::PingUploader for ValidatingUploader {
         fn upload(
             &self,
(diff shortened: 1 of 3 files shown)


Dexterp37 (approved):

glean-core/rlb/tests/schema.rs
```diff
@@ -115,7 +115,7 @@ fn validate_against_schema() {
         Ok(()) => {}
         Err(e) => {
             let errors = e.map(|e| format!("{}", e)).collect::<Vec<_>>();
-            assert!(false, "Data: {:#?}\nErrors: {:#?}", data, errors);
+            unreachable!("Data: {:#?}\nErrors: {:#?}", data, errors);
```
badboy, line 118: Shouldn't this be more like `panic!` to communicate intent better? This code is clearly not _unreachable_, but given that it is a test we want it to blow up (`unreachable!` will also just result in a panic, but the error message will include "internal error: entered unreachable code")

mdboom: Sure -- no problem making this `panic!` instead.

badboy (approved):

PR #1415, opened 2021-01-05
Have each tab constrain its height based on its own content [doc only]

brizental:
Before, the tabs would all be the same height as the longest tab,
which would result in a lot of unnecessary white space at the bottom of
a tab, if a tab with very little content was sibling
to a tab with a lot of content.

See this problem, for example, by selecting the C# tab in this page: https://mozilla.github.io/glean/book/user/general-api.html

[UPDATE]

Found another bug on the tabs, where no tab would be selected in case the "Kotlin" (the default) tab was not present. I fixed it in 335e921. You can see this bug on the "Paralelism" section of the same link I posted above. If this behaviour was intentional, we can just delete that commit :)

### DIFF
--- a/docs/glean.css
+++ b/docs/glean.css
@@ -40,6 +40,10 @@
     padding: 6px 12px;
 }
 
+.tab.active {
+    display: block;
+}
+
 /* The footer with the "Open on GitHub" link */
 footer#open-on-gh {
   font-size: 0.8em;
--- a/docs/glean.css
+++ b/docs/glean.css
@@ -33,7 +33,8 @@
 
 /* The container for each individual language */
 .tab {
-    visibility: hidden;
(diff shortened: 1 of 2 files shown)


badboy:
If I'm not mistaken we had it like that before, but changed it to keep the heigth, because now content jumps around if you switch.
But maybe the previous solution was just causing awful jumps, whereas this makes it nicer between switching. Let me check that out.

badboy:
I know remember what my problem with it was previously.
If there's multiple boxes and you switch the lower one then you might end up somehwere else on the page because suddenly its page heigth increases.

See the video:


https://user-images.githubusercontent.com/2129/103665250-7fddae80-4f73-11eb-88a3-09fc0622da0a.mp4

Issue #1410, opened 2021-01-04
Support Alpine Linux


marco-c:
Currently, the released wheel is `glean_sdk-33.9.1-cp36-abi3-manylinux2014_x86_64.whl`.

badboy:
cc @mdboom 
I'm not entirely sure what we need to do. We're building this using Python 3.8, which is a still-supported version.
Do we also need to build with Python 3.9 to support that one?

marco-c:
>  I'm not entirely sure what we need to do. We're building this using Python 3.8, which is a still-supported version.

The only wheel you have is the one for 3.6, so anyone with a Python version other than 3.6 will have to build from source.

>  Do we also need to build with Python 3.9 to support that one?

For a Python library we have written in Rust, where we are using [maturin](https://github.com/PyO3/maturin), we simply build with [this PATH](https://github.com/mozilla/pyo3-parsepatch/blob/5592de516a90f91b5528f2216ac7cbcf6a1b5408/Dockerfile#L7) and maturin takes care of it for us. I'm not sure how you'd do it in your case.

badboy:
maturin looks interesting and might solve this for us too, but it requires some changes to our project first.

mdboom:
Because the wheel specifies `abi3`, the wheel works with any Python [supporting abi3](https://docs.python.org/3/c-api/stable.html), which is Python 3.2 and later.  

If I make a fresh environment with Python 3.9, for example, you can see it installs using the wheel:

```
$ python
Python 3.9.1 | packaged by conda-forge | (default, Dec 21 2020, 22:08:58) 
[GCC 9.3.0] on linux
Type "help", "copyright", "credits" or "license" for more information.
>>> quit()
```

```
$ pip install glean_sdk
Collecting glean_sdk                                                                                                  
  Downloading glean_sdk-33.9.1-cp36-abi3-manylinux2014_x86_64.whl (1.6 MB)                                                 |████████████████████████████████| 1.6 MB 2.9 MB/s                        
```

[...]

PR #1416, opened 2021-01-05
Support installing Glean Python on musl-based platforms

mdboom:
This adds the ability to install from a source Python package on musl-based platforms, such as Alpine Linux.  This still requires that a Rust compiler is installed (the usual `rustup` incantation should be sufficient).  This is a partial fix for #1410.

To test, in an Alpine Linux Docker image, such as `python:alpine-3.8`.  You can install the pip package from git from this branch:

```
# To build glean_sdk from a source package
apk add curl
curl --proto '=https' --tlsv1.2 -sSf https://sh.rustup.rs | sh
export PATH=~/.cargo/bin:$PATH

# To build cffi from a source package
apk add gcc musl-dev libffi-dev

pip install git+git://github.com/mdboom/glean.rs.git@3e4ff4d36f5aa2535965b51444d79aecc666a96e#egg=glean_sdk
```

Cc: @marco-c

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -767,6 +767,20 @@ jobs:
           root: glean-core/python/
           paths: .venv3.9
 
+  Python 3_9 on Alpine tests:
+    docker:
+      - image: python:3.9-alpine
+    shell: /bin/sh -leo pipefail
+    environment:
+      - BASH_ENV: /etc/profile
+    steps:
+      - run:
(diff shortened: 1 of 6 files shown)


badboy (commented):
oh wow, I like this.
We should make the Python task required (and some others listed there as well, let's do that when merged)

badboy (approved):

PR #1418, opened 2021-01-05
Bug 1683707: Add foreground count metric

mdboom:


chutten (commented):
Any tests?

mdboom:
> Any tests?

Oh, gee.  Of course, let me add some.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -192,6 +192,9 @@ class GleanTest {
             serverEndpoint = "http://" + server.hostName + ":" + server.port
         ))
 
+        // Send an "empty" metrics ping to reset the foreground_count
```
chutten, line 195: Oof.

chutten (approved):
Do we have to file a follow-up for adding a swift test to cover the swift implementation?

mdboom:
> Do we have to file a follow-up for adding a swift test to cover the swift implementation?

We can do that, unless one of our Mac devs has a few cycles to port the Kotlin test to Swift in this PR.  (Cc: @badboy, @brizental, @travis79)

brizental:
> We can do that, unless one of our Mac devs has a few cycles to port the Kotlin test to Swift in this PR.

Might be best to file a follow up because I don't think any of us can commit to your fork :) I'd be happy to take it as a follow up if necessary.

mdboom:
> > We can do that, unless one of our Mac devs has a few cycles to port the Kotlin test to Swift in this PR.
> 
> Might be best to file a follow up because I don't think any of us can commit to your fork :) I'd be happy to take it as a follow up if necessary.

Fair enough.  I filed https://bugzilla.mozilla.org/show_bug.cgi?id=1685712 for the follow-on work, and decided to remove all the Swift changes in this PR -- I think it better to commit the implementation and test at the same time.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
badboy: Shouldn't `resetGlean` take care of that? It has `clearStores = true`

mdboom: Yeah -- I really struggled to make this test work, and thought this was required, but it seems in the end it isn't.  Removing.

PR #1403, opened 2020-12-21
Bug 1680783: Increase the resolution of first_run_date

mdboom:

PR #1421, opened 2021-01-06
Fix the Swift tests to accomodate for the new metric

badboy:
Regressed in #1403


brizental (approved):

badboy:
Not sure what's going on on TC, the same test worked fine on CircleCI. I'm going to merge this for now.

PR #1423, opened 2021-01-06
Bug 1598291 - Publish the Rust crates from CI on every release

badboy:
Not yet ready -- I'm taking this for a test run with a sample crate.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -396,6 +396,33 @@ jobs:
             - docs/docs
             - docs/index.html
 
+  Publish Rust crates:
+    docker:
+      - image: circleci/rust:latest
+    steps:
+      - checkout
+      - run:
+          name: Publish Cargo Package
+          command: |
+            # Login to crates.io so the following commands work
+            cargo login -- "$CRATES_IO_TOKEN"
+
+            # Publish all crates from CI.
+            # The token is set in CircleCI settings.
+            # Need some sleep inbetween to ensure crates.io is updated.
+
+            pushd glean-core
+            cargo publish --verbose
+
+            sleep 30
+            pushd ffi
+            cargo publish --verbose
+            popd
+
+            sleep 30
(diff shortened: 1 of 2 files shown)


badboy:
Test crate: [test-gha](https://crates.io/crates/test-gha)
Repository: https://github.com/badboy/test-gha
Publish workflow: https://app.circleci.com/pipelines/github/badboy/test-gha/10/workflows/defb4af7-e375-43b5-aa6f-2ff7167db19a/jobs/14

It works \o/

Dexterp37 (approved):
Neat!

.circleci/config.yml
```diff
@@ -396,6 +396,33 @@ jobs:
...
+            cargo login -- "$CRATES_IO_TOKEN"
+
+            # Publish all crates from CI.
+            # The token is set in CircleCI settings.
+            # Need some sleep inbetween to ensure crates.io is updated.
+
+            pushd glean-core
+            cargo publish --verbose
+
+            sleep 30
```
mdboom, line 417: This is ... interesting.  I think I understand why having done this manually.  But maybe something that repeatedly tried to `curl` the new package would be more robust?

mdboom (commented):
Only take my suggestion if it's not too onerous -- it's mostly just my general-purpose aversion to `sleep`...

.circleci/config.yml
badboy: yeah, it's ... a hack.

`cargo-release` for example [waits for the index to update](https://github.com/sunng87/cargo-release/blob/f4400027579a795e790b3885e33228638a6b111a/src/cargo.rs#L59), but of course that could happen due to other crates as well, [so they still sleep](https://github.com/sunng87/cargo-release/commit/58f7c28f43a20e3df128e165fbbfa9406d519d4b).

IMO we can go with this solution for now and, if it fails to often, fix it better.

PR #1428, opened 2021-01-12
Bug 1686233 - Record a new error whenever we fail to write a pending ping to disk

badboy:


glean-core/src/internal_metrics.rs
```diff
@@ -56,6 +63,15 @@ impl CoreMetrics {
                 disabled: false,
                 dynamic_label: None,
             }),
+
+            io_errors: CounterMetric::new(CommonMetricData {
+                name: "io".into(),
+                category: "glean.error".into(),
+                send_in_pings: vec!["metrics".into()],
```
badboy, line 70: In the bug report you said:

> Maybe in the "baseline" ping? We want this failure to be observable since it might result in data loss.

For now I have it in the metrics ping. I'm not really sure where to put it though.

glean-core/src/internal_metrics.rs
```diff
@@ -10,6 +10,13 @@ pub struct CoreMetrics {
     pub first_run_date: DatetimeMetric,
     pub first_run_hour: DatetimeMetric,
     pub os: StringMetric,
+
+    /// The number of times we encountered an IO error
+    /// when writing a pending ping to disk.
+    ///
+    /// **Note**: Not a _core_ metric, but an error metric,
```
badboy, line 17: I don't like it, but that's the only place I could reasonably add it without introducing yet another substructure or renaming stuff.

glean-core/metrics.yaml
```diff
@@ -429,6 +429,22 @@ glean.error:
...
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1686233
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1686233
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+      - jrediger@mozilla.com
+    expires: never
```
badboy, line 446: Another note: Because the code is not auto-generated we don't have any automatic way to expire those, so the only real thing is to keep them as never expiring.

chutten: Oof. I wonder how much of this we'll end up mirroring for use counters in FOG.

glean-core/src/lib_unit_tests.rs
```diff
@@ -871,3 +871,38 @@ fn records_database_file_size() {
...
     assert!(data.sum > 0);
 }
+
+#[test]
+fn records_io_errors() {
+    use std::fs;
+    let _ = env_logger::builder().is_test(true).try_init();
+
+    let (glean, _data_dir) = new_glean(None);
+    let pending_pings_dir = glean.get_data_path().join(crate::PENDING_PINGS_DIRECTORY);
```
chutten, line 881: Not using `_data_dir`?

Also, the fact there's two `new_glean` test `fn` in the SDK is annoying. It was hard for me to find the right one.

chutten (approved):

glean-core/src/lib_unit_tests.rs
badboy: I was using `_data_dir` but then noticed there's a `get_data_path` function, so it seemed better to just use that.

About the duplication I agree, maybe it's time to clean up these tests.

glean-core/src/internal_metrics.rs
badboy: We don't have it yet (it's a bit tricky, we will need a template that's different from the one for the RLB, because we actually need to generate for the _internal_ metric types, not the RLB types, also right now they are not global statics)

PR #1431, opened 2021-01-14
bug 1686726 - Fix lock order inversion in RLB Timing Distribution

chutten:


glean-core/rlb/src/private/timing_distribution.rs
```diff
@@ -65,12 +65,15 @@ impl glean_core::traits::TimingDistribution for TimingDistributionMetric {
     ) -> Option<DistributionData> {
         crate::block_on_dispatcher();
 
-        let inner = self.0.read().unwrap();
-        let queried_ping_name = ping_name
-            .into()
-            .unwrap_or_else(|| &inner.meta().send_in_pings[0]);
+        crate::with_glean(|glean| {
+            // The order of taking these locks matter. Glean must be first.
+            let inner = self.0.read().unwrap();
```
Dexterp37, line 70:
```suggestion
            let inner = self.0.read().expect("Lock poisoned for timing distribution metric on test_get_value.");
```

Dexterp37 (approved):
r+wc. Before merging, would you kindly do two things:

1. make sure this fixes the problem detected by TSAN
2. make sure the suggested change doesn't break linting :)

chutten:
Tested that it fixed the TSan-discovered fault by using https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html#patching-third-party-crates and https://firefox-source-docs.mozilla.org/tools/sanitizer/tsan.html. With this fix, no inversion. Without this fix, inversion.

PR #1432, opened 2021-01-14
Bug 1673672 - User RLB types for core metrics and add more tests

Dexterp37:
RLB was using glean-core types instead of the new RLB types, which made writing tests a bit more complicated. This PR uses the binding types for the core metrics (as we do for the other bindings) and then ports back some tests from the Kotling bindings.

Note that using the RLB types also pave the way for using generated metrics :-)


glean-core/rlb/src/core_metrics.rs
```diff
@@ -23,76 +26,90 @@ impl ClientInfoMetrics {
...
-                name: "device_model".into(),
-                category: "".into(),
-                send_in_pings: vec!["glean_client_info".into()],
-                lifetime: Lifetime::Application,
-                disabled: false,
-                dynamic_label: None,
-            }),
-        }
-    }
+    #[allow(non_upper_case_globals)]
```
badboy, line 32: You could move this to the top of the file as:

```rust
#![allow(non_upper_case_globals)]
```

glean-core/rlb/src/private/string.rs
```diff
@@ -28,6 +29,11 @@ impl StringMetric {
     pub fn new(meta: glean_core::CommonMetricData) -> Self {
         Self(Arc::new(glean_core::metrics::StringMetric::new(meta)))
     }
+
+    /// Internal only, synchronous API for setting a string value.
+    pub(crate) fn set_sync<S: Into<std::string::String>>(&self, glean: &Glean, value: S) {
```
badboy, line 34: Yey for internal-only APIs and the compiler ensuring it stays that way

glean-core/rlb/src/test.rs
```diff
@@ -478,9 +668,76 @@ fn setting_source_tags_before_initialization_should_not_crash() {
 }
 
 #[test]
-#[ignore] // TODO: To be done in bug 1673672.
 fn flipping_upload_enabled_respects_order_of_events() {
-    todo!()
+    // NOTES(janerik):
```
badboy, line 672: I make the best notes.

badboy (approved):
Looks good to me!

badboy:
Oh, this will need a rebase because of the changelog changes. You can resolve that on GitHub using the UI if you want.

glean-core/rlb/src/test.rs
Dexterp37: You sure do ;-)

glean-core/rlb/src/core_metrics.rs
Dexterp37: Oh right. If it doesn't bother you too much, I'd save another change and do that as a follow-up: we're going to re-generate this file anyway!

badboy: Sure

PR #1434, opened 2021-01-15
Upgrade rkv to 0.17

upsuper:
Upgrade rkv to 0.17 to remove the dependency to failure in mozilla/rkv#210 as part of [effort](https://bugzilla.mozilla.org/show_bug.cgi?id=1681898) of removing dependencies to `failure` in m-c, as `failure` has been [officially deprecated](https://rustsec.org/advisories/RUSTSEC-2020-0036.html).

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -337,27 +337,6 @@ dependencies = [
 ]
 
 [[package]]
-name = "failure"
-version = "0.1.8"
-source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "d32e9bd16cc02eae7db7ef620b392808b89f6a5e16bb3497d159c6b92a0f4f86"
-dependencies = [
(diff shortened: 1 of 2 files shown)


upsuper:
I looked into the failure on android-build-pr, but couldn't figure out what's going wrong there, as I can't even get the test to work locally without this PR.

It's basically just upgrading rkv without any real functionality change, so I don't really expect it to cause any issue.

badboy:
> I looked into the failure on android-build-pr, but couldn't figure out what's going wrong there, as I can't even get the test to work locally without this PR.
> 
> It's basically just upgrading rkv without any real functionality change, so I don't really expect it to cause any issue.

yeah, that PR is failing for unrelated reasons. The same test is running on CircleCI and passed there.

badboy (approved):

PR #1438, opened 2021-01-19
Bug 1686736 - (RLB) Record the number of tasks queued in the pre-initialization buffer.

badboy:
Based on #1437.
The last 3 commits are the new implementation.

I went for an integration test to avoid having to deal with resetting the dispatcher and Glean.

### DIFF
--- a/glean-core/rlb/src/dispatcher/mod.rs
+++ b/glean-core/rlb/src/dispatcher/mod.rs
@@ -29,7 +29,7 @@
 use std::{
     mem,
     sync::{
-        atomic::{AtomicBool, Ordering},
+        atomic::{AtomicBool, AtomicUsize, Ordering},
         Arc,
     },
     thread::{self, JoinHandle},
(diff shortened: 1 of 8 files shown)


badboy:
I guess that will need a quick data review because it extends it to new clients?

badboy:
From the original data request:

> 7) What populations will you measure?
> All products using Glean.


So I gues not.

glean-core/rlb/src/dispatcher/mod.rs
```diff
@@ -107,6 +107,9 @@ struct DispatchGuard {
     /// Whether to queue on the preinit buffer or on the unbounded queue
     queue_preinit: Arc<AtomicBool>,
 
+    /// The number of items that were added to the queue after it filled up.
+    overflow_count: Arc<AtomicUsize>,
```
Dexterp37, line 111: Can you kindly add a changelog entry?

Dexterp37 (approved):
r+wc

glean-core/rlb/src/dispatcher/mod.rs
badboy: I did ... then I rebased ... and lost it -_-

Dexterp37: You almost had a perfect PR. Almost.

PR #1440, opened 2021-01-20
Disable C# tests

badboy:
The C# implementation is deprecated. We don't need to use up CI resources for it anymore.


Dexterp37 (approved):
2nd sad r+ of the day.

PR #1442, opened 2021-01-20
Bug 1685664 - Try to unblock the RLB dispatcher from bindings

badboy:
This exposes a new symbol `rlb_flush_dispatcher` from the RLB code.
glean-ffi exposes a new method to be used in language bindings.
Internally it tries to look up the new RLB symbol and, if found, calls
it to unblock the RLB dispatcher.
This allows using both the RLB and the Kotlin/Swift/Python bindings and still rely on
2 dispatchers.


Cargo.toml
```diff
@@ -7,6 +7,10 @@ members = [
   "glean-core/benchmark",
 ]
 
+exclude = [
```
Dexterp37, line 10: OH! TIL!

Does it make sense to exclude the tests as well, so that they don't get vendored, for example, in m-c? (not for this PR)

glean-core/ffi/Cargo.toml
```diff
@@ -46,3 +46,6 @@ oslog = { version = "0.0.3", default-features = false, features = ["logger"] }
 
 [target.'cfg(not(target_os = "android"))'.dependencies]
 env_logger = { version = "0.7.1", default-features = false, features = ["termcolor", "atty", "humantime"] }
+
+[target.'cfg(unix)'.dependencies]
+libc = "0.2.82"
```
Dexterp37, line 51: Why do we need this?

glean-core/ffi/src/weak.rs
```diff
@@ -0,0 +1,57 @@
+//! Based on `library/std/src/sys/unix/weak.rs` from the Rust libstd code base.
```
Dexterp37, line 1: Should we note down the revision we pulled this from?

glean-core/ffi/src/lib.rs
```diff
@@ -366,6 +370,35 @@ pub extern "C" fn glean_clear_application_lifetime_metrics() {
     with_glean_value(|glean| glean.clear_application_lifetime_metrics());
 }
 
+/// Try to unblock the RLB dispatcher to start processing queued tasks.
+///
+/// **Note**: glean-core does not have its own dispatcher at the moment.
+/// This tries to detect the RLB and, if loaded, instructs the RLB dispatcher to flush.
+/// This allows the usage of both the RLB and other language bindings (e.g. Kotlin/Swift)
+/// within the same application.
```
Dexterp37, line 378: Super smart. I like it.

glean-core/ios/Glean/Glean.swift
```diff
@@ -188,6 +188,7 @@ public class Glean {
             }
 
             // Signal Dispatcher that init is complete
+            glean_flush_rlb_dispatcher()
```
Dexterp37, line 191: As above, can we add the same small comment about why we're doing this?

glean-core/python/glean/glean.py
```diff
@@ -207,6 +207,7 @@ def initialize():
                 _ffi.lib.glean_clear_application_lifetime_metrics()
                 cls._initialize_core_metrics()
 
+            _ffi.lib.glean_flush_rlb_dispatcher()
```
Dexterp37, line 210: nit: the lil' comment :)

samples/multi-bindings/README.md
```diff
@@ -0,0 +1,59 @@
+# Multi-binding example
+
+This example showcases the interplay of using a megazord build of a Rust library, the Glean RLB and glean-ffi,
```
Dexterp37, line 3: Can we linkify megazord to point at some repo?

samples/multi-bindings/app.c
```diff
@@ -0,0 +1,68 @@
+#include <stdint.h>
```
Dexterp37, line 1: This is missing a license header.

samples/multi-bindings/src/lib.rs
```diff
@@ -0,0 +1,25 @@
+pub use glean_ffi;
```
Dexterp37, line 1: Missing a license header

Dexterp37 (changes requested):
Looks very good to me overall, just a few minor things to take care of.

Dexterp37:
@badboy can we also add a CHANGELOG entry, please?

Cargo.toml
badboy: This line here only excludes it from the workspace. Otherwise `cargo build` won't work inside the sample directory.
Reason I did not put it in the workspace was that it should be more of a standalone example (but then again I think initially there was another blocker which I got rid off and now it can be reasonably part of the workspace without real issues, then it would share build artifacts). I'll check that again.

Re bundled tests: One _can_ exclude that in the respective `Cargo.toml` of a crate (`include` and `exclude` lists), but there's one advantage of bundling them:
The Rust project regurlarly runs ALL tests of ALL crates on crates.io to find potential breakage in the compiler. It'd be good if our crates are part of that.

glean-core/ffi/Cargo.toml
badboy: We use `dlsym` from `libc` to look up symbols. That works on Unix platforms (macOS, Linux), I'm not sure it would actually work on Windows the same way. Therefore we only need this dependency on unix platforms.

I'll file a followup to check Windows (it's less important because this feature is mainly to support Android right now)

glean-core/ffi/src/weak.rs
badboy: That file upstream is essentially unchanged since 2016. I can add a Rust version number maybe.

glean-core/ffi/src/lib.rs
badboy: HACK!

samples/multi-bindings/README.md
badboy: Good point.

Cargo.toml
Dexterp37: > The Rust project regurlarly runs ALL tests of ALL crates on crates.io to find potential breakage in the compiler. It'd be good if our crates are part of that.

TIL2 :)

glean-core/ffi/Cargo.toml
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1687957

Dexterp37 (approved):

Cargo.toml
badboy: Gonna leave that as is:

We need to run with `lto` enabled, but these need to be configured for the whole workspace.
We could run the multi-bindings example in release mode, but that takes longer and I don't think it's worth it.
The majority of people won't run that example anyway, so the additional compilation and storage overhead is negligible for devs.

PR #1443, opened 2021-01-21
Update cbindgen in use on CI

badboy:
Based on #1442 which needs to land first.

I set that other branch as the base. Let's see if that then auto-updates when the other PR lands.

* [ ] Needs iOS test run, will do that after #1442 lands.


glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -47,7 +47,7 @@ internal open class FfiPingUploadTask(
     // NOTE: We need to provide defaults here, so that JNA can create this object.
     @JvmField var tag: Byte = UploadTaskTag.Done.ordinal.toByte(),
     @JvmField var upload: UploadBody = UploadBody(),
-    @JvmField var time: Long = 60_000
+    @JvmField var wait: Long = 60_000
```
Dexterp37, line 50: <3

Dexterp37 (approved):
r+ - this can be merged if iOS is happy about it, after the rebase.

PR #1444, opened 2021-01-21
Explicitly set linker for Big Sur

boek:
Currently the project fails to compile on Big Sur.

I was able to fix it by explicitly setting the linkers in the cargo config


badboy:
(close-reopen cycle to trigger CircleCI)

badboy:
Android build failure because the branch is not up to date. That can be ignored.

badboy:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1688178. Merging.

PR #1447, opened 2021-01-22
Bug 1670261 - Implement metric types through the (dynamic) FFI to support geckoview 

badboy:
That was some work ....

I wrote up lengthy notes [in this implementation notes document](https://docs.google.com/document/d/1LWJu3Qx6KSKzUODm7vMdJiN_YOltxm3jBRERncku8JI/edit#).

I will also push up the Gecko side of the code. That one is even hackier though as I found _a__ point to initialize Glean, probably not the right place in the code though.

I did test that in a local build of Fenix and it worked, but it will require much more testing (especially because of the additional dispatcher, that's currently just unblocked. I still need to look how to handle that)


glean-core/rlb/Cargo.toml
```diff
@@ -35,6 +43,7 @@ serde = { version = "1.0.104", features = ["derive"] }
 uuid = { version = "0.8.1", features = ["v4"] }
 chrono = { version = "0.4.10", features = ["serde"] }
 time = "0.1.40"
+libloading = "0.6.6"
```
chutten, line 46: Do we need `libloading` if not `glean-dynamic`?

glean-core/rlb/src/private/dynamic/counter.rs
```diff
@@ -0,0 +1,56 @@
...
+    }
+
+    /// Internal only, synchronous API for incremeting a counter
+    pub(crate) fn add_sync(&self, glean: &Glean, amount: i32) {
+        let id = self.0;
+        crate::sys::with_glean(|glean| unsafe { glean.glean_counter_add(id, amount) })
+    }
+}
+
+impl Default for CounterMetric {
```
chutten, line 28: Took me a while to notice that we only need Default because of `new_metric!`. Could use a comment (or figuring out some way to generate the whole type by macro, maybe. In the future, of course.).

chutten (commented):
It's very interesting, this. And complex.

I'd love to see the impl notes turned into dev docs.

glean-core/rlb/src/private/dynamic/counter.rs
badboy: Yeah, this was a last minute fix for the labeled counters.

glean-core/rlb/Cargo.toml
badboy: Nope. missed that.

glean-core/rlb/src/private/dynamic/timespan.rs
```diff
@@ -0,0 +1,75 @@
...
+
+impl TimespanMetric {
+    /// The public constructor used by automatically generated metrics.
+    pub fn new(meta: glean_core::CommonMetricData, time_unit: TimeUnit) -> Self {
+        let metric = Arc::new(AtomicU64::new(0));
+
+        let deferred = Arc::clone(&metric);
+        dispatcher::launch(move || {
+            let metric = new_metric!(glean_new_timespan_metric, meta, time_unit as i32);
+            deferred.store(metric, SeqCst);
```
badboy, line 25: See commit message: c92e04140b3e84bdad5cb463d794dd2f638b13c8

glean-core/rlb/src/lib.rs
```diff
@@ -385,6 +388,14 @@ pub fn shutdown() {
...
     log::trace!("FLushing RLB dispatcher through the FFI");
 
+    #[cfg(feature = "glean-dynamic")]
+    {
+        if sys::global_glean_sys().is_none() {
+            sys::NEEDS_FLUSH.store(true, Ordering::SeqCst);
+            return;
+        }
+    }
+
```
badboy, line 398: See commit message 546ffaaa8b4a6359439726b23d0c709bb8f8f231

badboy (commented):
Expanded some commit messages to explain stuff further. Gecko patchset coming too.

Will update the doc and see how we can integrate that into the dev docs (if we move forward with this)

badboy:
Gecko stack at https://phabricator.services.mozilla.com/D105307

glean-core/ffi-sys/examples/ffi-usage.rs
```diff
@@ -0,0 +1,34 @@
+use std::ffi::CString;
+
+use cfg_if::cfg_if;
+use glean_ffi_sys::*;
+
+cfg_if! {
```
chutten, line 6: Oh, so this is sugar so we can have `else` instead of a big `#[cfg(not(any(` like [here](https://searchfox.org/glean/rev/ed1b6d3bcf54a82e8f43bfeb4fd9f1af3d758382/glean-core/src/system.rs#70)?

glean-core/ffi-sys/src/lib.rs
```diff
@@ -0,0 +1,33 @@
...
+#![allow(non_snake_case)]
+#![allow(clippy::style)]
+#![allow(clippy::too_many_arguments)]
+#![allow(clippy::transmute_ptr_to_ptr)]
+
+// Re-import/export some enums.
+//
+// `bindgen` butchers their defintions by having both the enum and a typedef,
+// thus breaking the build.
+// `glean-core` already defines them and marks them `repr(i32)`, therefore FFI-compatible.
```
chutten, line 16: Point to this file from the bindgen build command? These explanations are very important context.

glean-core/rlb/src/lib.rs
```diff
@@ -760,5 +771,17 @@ pub fn set_source_tags(tags: Vec<String>) -> bool {
     }
 }
 
+/// Load the dynamic library and dispatch all metric recordings to that library using the Glean
+/// FFI.
```
chutten, line 775: We're gonna need a more distinct name than "FFI" for this since we have (at last count) 2 (specific metrics FFI for underpinning non-RLB LBs, MLA FFI for FOG) plus this one... "dynFFI"? I dunno, I'm not a marketer.

glean-core/rlb/src/private/dynamic/string.rs
```diff
@@ -0,0 +1,63 @@
...
+#[inherent(pub)]
+impl glean_core::traits::String for StringMetric {
+    fn set<S: Into<std::string::String>>(&self, value: S) {
+        let id = self.0;
+        let new_value = value.into();
+        dispatcher::launch(move || {
+            let value = CString::new(new_value).unwrap();
+            crate::sys::with_glean(|glean| unsafe { glean.glean_string_set(id, value.as_ptr()) });
+        });
+        todo!()
```
chutten, line 44: ? Looks like the impl is complete.

glean-core/rlb/src/private/dynamic/timespan.rs
chutten: Oh, so everything will need to be two-step constructed: ctor then init. 

Okay, I think I've got it.

chutten (commented):
Let's see if I can explain the approach correctly.

The dynamic loading remains more or less the same in its mechanics. We're being a little more specific about its timing. But ultimately we're still generating a stable cross-platform FFI for the syslib via rust bindgen, and we're still using ffi support to provide a handlemap for metric use a la cbingden FFI underpinning the LBs who aren't RLB.

(( Speaking of which, if it's a handlemap, maybe we call them `handle` instead of `id` to limit the number of metric ids we need to keep straight in our brains? ))

In order to have a metric instance to call prior to init we need to split all metric init into two steps: ctor then init. Init will happen in an ordered way in the dispatcher before any calls to instance methods. Metric ids (handles) are assigned at init, just in time for any pending operations to have them.

Macros. Macros everywhere. We are adding yet another Rust implementation of all these types, but at least this one is swapped out for the static impls at compile time based on cargo feature. Lemme just increment the "We need code generation." counter for you.

But then things get a little more straightforward once the call against the dynamic lib actually reaches the lib. We pull a metric instance from the handlemap, Do some type coercion. Hit the db.

Not 100% sure what the upload flow will look like since that goes the other direction... but I'm guessing it's more of the same principles, just applied differently.

glean-core/ffi-sys/Cargo.toml
```diff
@@ -0,0 +1,19 @@
+[package]
+name = "glean-ffi-sys"
+version = "0.1.0"
+authors = ["Jan-Erik Rediger <jrediger@mozilla.com>"]
```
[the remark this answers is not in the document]
badboy: Woops, default-generated data.

glean-core/ffi-sys/examples/ffi-usage.rs
badboy: Yup, though this is just the example and I think I even broke this with recent changes.

badboy:
> Let's see if I can explain the approach correctly.

Seems about right.

> The dynamic loading remains more or less the same in its mechanics. We're being a little more specific about its timing. But ultimately we're still generating a stable cross-platform FFI for the syslib via rust bindgen, and we're still using ffi support to provide a handlemap for metric use a la cbingden FFI underpinning the LBs who aren't RLB.

Correct.

> (( Speaking of which, if it's a handlemap, maybe we call them `handle` instead of `id` to limit the number of metric ids we need to keep straight in our brains? ))

(( yes ))


> In order to have a metric instance to call prior to init we need to split all metric init into two steps: ctor then init. Init will happen in an ordered way in the dispatcher before any calls to instance methods. Metric ids (handles) are assigned at init, just in time for any pending operations to have them.

Yes, it means there's some overhead now, worth some testing how much that's noticeable. But there's no way around this at the moment.


> Macros. Macros everywhere. We are adding yet another Rust implementation of all these types, but at least this one is swapped out for the static impls at compile time based on cargo feature. Lemme just increment the "We need code generation." counter for you.

Right now it's a lot of manual implementation, but yes, it can be hidden behind macros better.
 
> But then things get a little more straightforward once the call against the dynamic lib actually reaches the lib. We pull a metric instance from the handlemap, Do some type coercion. Hit the db.

Yes, that's the nice part. It just works.

> Not 100% sure what the upload flow will look like since that goes the other direction... but I'm guessing it's more of the same principles, just applied differently.

Actually ... I'm not sure either. I didn't think that far yet. Right now submitting a ping will correctly assemble it and store it on disk. But it will not trigger the Kotlin-side uploader, but might trigger the internal one? I'll better test that, even if a final solution becomes followup work.

badboy:
Blocked by current discussions around Better Builds/Megazording with GeckoView

badboy:
We're tackling this now using a GeckoView bundle build.

PR #1452, opened 2021-01-25
Bug 1679963 - Expose glean_core::traits through the RLB

badboy:
That way wrapping implementations such as FOG only need to depend on
`glean` and are guaranteed to get the correct `glean-core` version.

---

This should be the last piece to make sure we don't need to depend directly on `glean-core` anymore from FOG.


glean-core/rlb/src/lib.rs
```diff
@@ -49,7 +49,8 @@ pub use core_metrics::ClientInfoMetrics;
 pub use glean_core::{
     global_glean,
     metrics::{DistributionData, MemoryUnit, RecordedEvent, TimeUnit},
-    setup_glean, CommonMetricData, Error, ErrorType, Glean, HistogramType, Lifetime, Result,
+    setup_glean, traits, CommonMetricData, Error, ErrorType, Glean, HistogramType, Lifetime,
+    Result,
```
Dexterp37, line 53: Should `global_glean` and `setup_glean` be just `use` instead of `pub use`?

Dexterp37 (approved):
r+

glean-core/rlb/src/lib.rs
badboy: Huh, I think we can drop that now from exposing it, you are right.
Let me test that in m-c though to save us from one more unnecessary roundtrip if I overlook something

badboy: ok, I only grepped for it and now there's no more setup_glean or global_glean left in m-c!

PR #1453, opened 2021-01-25
Update docs on which places in Fenix to modify for a local build

badboy:
[doc only]


Dexterp37 (approved):
r+ assuming you've tested this :D

badboy:
> r+ assuming you've tested this :D

```
BUILD SUCCESSFUL in 9m 29s
39 actionable tasks: 39 executed
```

I just did!

PR #1454, opened 2021-01-25
Manual test that we get the right values from the FFI on upload

badboy:


badboy:
draft so I don't forget to: 
* add a changelog entry
* probably adjust a comment or two

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -27,7 +27,7 @@ enum class UploadTaskTag {
 @Structure.FieldOrder("tag", "documentId", "path", "body", "headers")
 internal class UploadBody(
     // NOTE: We need to provide defaults here, so that JNA can create this object.
-    @JvmField val tag: Byte = UploadTaskTag.Done.ordinal.toByte(),
+    @JvmField val tag: Byte = UploadTaskTag.Upload.ordinal.toByte(),
```
brizental, line 30: Why was it necessary to change the default here?

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
```diff
@@ -43,11 +43,18 @@ internal class UploadBody(
     }
 }
 
+@Structure.FieldOrder("tag", "time")
+internal class WaitBody(
+    @JvmField var tag: Byte = UploadTaskTag.Wait.ordinal.toByte(),
+    @JvmField var time: Long = 60_000
+) : Structure() {
+}
+
```
brizental, line 52: Sheesh, what an oversight. Good thing you caught this...

brizental (approved):
Looks good to me.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
badboy: It wasn't. I just noticed it when I wrote the `WaitBody` and IMO it felt wrong to have a tag of `Done` on an `UploadBody`.
We need default values, but won't actually use a default-constructed `UploadBody` further, so this change is not really noticable.

glean-core/android/src/main/java/mozilla/telemetry/glean/net/Upload.kt
badboy: I didn't notice in review either. That's why we do the testing.

PR #1455, opened 2021-01-25
Bug 1685522 - Add the 'clients activity API' and implement the baseline ping for the RLB

Dexterp37:
This PR does a bunch of things:

- it implements the client activity API in glean-core;
- it implements the startup `baseline` ping with reason `dirty_startup` for the RLB;
- it updates the other language bindings to use the clients activity API

TODO:

- [x] Fix the RLB test.
- [x] Add changelog entries (glean-core and each LB)
- [ ] ~Swift!~ (filed a bug)

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -563,8 +563,6 @@ open class GleanInternalAPI internal constructor () {
      * Handle the foreground event and send the appropriate pings.
      */
     internal fun handleForegroundEvent() {
-        GleanValidation.foregroundCount.add(1)
-
         // Note that this is sending the length of the last foreground session
         // because it belongs to the baseline ping and that ping is sent every
         // time the app goes to background.
@@ -580,6 +578,7 @@ open class GleanInternalAPI internal constructor () {
 
         // Start the timespan for the new activity period.
         GleanBaseline.duration.start()
+        GleanValidation.foregroundCount.add(1)
     }
 
     /**
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -18,6 +18,7 @@ import com.sun.jna.StringArray
 import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
+import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
 import mozilla.telemetry.glean.config.Configuration
 import mozilla.telemetry.glean.config.FfiConfiguration
 import mozilla.telemetry.glean.utils.getLocaleTag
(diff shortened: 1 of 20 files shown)


Dexterp37:
I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1688842) for taking care of the Python bindings and [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1688845) for the Swift implementation.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -562,16 +563,42 @@ open class GleanInternalAPI internal constructor () {
      * Handle the foreground event and send the appropriate pings.
      */
     internal fun handleForegroundEvent() {
-        Pings.baseline.submit(Pings.baselineReasonCodes.foreground)
         GleanValidation.foregroundCount.add(1)
```
badboy, line 566: This slightly changes counting.
Before we submitted the ping first, then counted foreground.

Now it's the other way around. Feels like a bug fix, so just asking if that was intentionally.

glean-core/pings.yaml
```diff
@@ -29,14 +29,14 @@ baseline:
...
+      killed before the Glean SDK had the chance to generate this ping, before
+      becoming inactive, in the last session.
 
       *Note*: this ping will not contain the `glean.baseline.duration` metric.
-    background: |
-      The ping was submitted before going to background.
-    foreground: |
-      The ping was submitted when the application went to foreground, which
+    inactive: |
+      The ping was submitted before becoming inactive.
```
badboy, line 37: While "before" is technically correct, what do you think about making it more explicit that this is triggered by becoming inactive? 

```suggestion
      The ping was submitted when becoming inactive.
```

glean-core/pings.yaml
```diff
@@ -94,8 +94,8 @@ events:
       The ping was submitted at startup. The events ping is always sent if there
       are any pending events at startup, because event timestamps can not be
       mixed across runs of the application.
-    background: |
-      The ping was submitted before going to background.
+    inactive: |
+      The ping was submitted before becoming inactive.
```
badboy, line 98:
```suggestion
      The ping was submitted when becoming inactive.
```

glean-core/rlb/src/test.rs
```diff
@@ -171,15 +171,141 @@ fn test_experiments_recording_before_glean_inits() {
...
+    handle_client_active();
+
+    // We expect a baseline ping to be generated here (reason: 'active').
+    let url = r.recv().unwrap();
+    assert!(url.contains("baseline"));
 }
 
 #[test]
-#[ignore] // TODO: To be done in bug 1672958.
 fn test_sending_of_startup_baseline_ping() {
```
badboy, line 254: As above: drop the `test_` prefix

badboy (changes requested):
Implementation-wise this looks good.
Few things I'd like some rephrasing on.

One thing I realized is that we don't really have docs on how to integrate the RLB, that is: when to call what function.
It doesn't need to be solved in this PR (except maybe a short call out where/when handle_(in)active needs to happen)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
Dexterp37: Whoops, no, not intentional. Let me roll it back.

glean-core/rlb/src/test.rs
```diff
@@ -171,15 +171,143 @@ fn test_experiments_recording_before_glean_inits() {
...
+
+    // Simulate becoming inactive
+    handle_client_inactive();
+
+    // Wait for the pings to arrive.
+    let mut expected_pings = vec!["baseline", "events"];
+    for _ in 0..2 {
+        let url = r.recv().unwrap();
+        // If the url contains the expected reason, remove it from the list.
+        expected_pings.retain(|&name| url.contains(name));
```
badboy, line 242: This doesn't actually does what we want, does it?
We want to check that we receive the 2 expected pings and no other.
Right now we check that ... barely anything.
Easy to fix:

```suggestion
        expected_pings.retain(|&name| !url.contains(name));
```

Longer details: 

<details>

`retain` removes elements when the closure returns `false`.
If the first ping that gets in is `baseline`, then:
1. it checks if the url contains `baseline`. This is true. The element is kept.
2. it checks if the url contains `events`. This is false. `events` is removed.
We're left with `expected_pings = ["baseline"]`
If we would now get a `deletion-request` ping, then:
1. it checks if the url contains `baseline`. This is false. `baseline` is removed.
We're left with `expected_pings = []`

By inverting the retain check we keep everything that's not ping we currently look at.
</details>

badboy (changes requested):

glean-core/rlb/src/test.rs
Dexterp37: good catch!

glean-core/ffi/src/lib.rs
```diff
@@ -409,6 +409,16 @@ pub extern "C" fn glean_is_dirty_flag_set() -> u8 {
     with_glean_value(|glean| glean.is_dirty_flag_set())
 }
 
+#[no_mangle]
+pub extern "C" fn handle_client_active() {
```
badboy, line 413: This needs to be `glean_` prefixed:

```suggestion
pub extern "C" fn glean_handle_client_active() {
```

glean-core/ffi/src/lib.rs
```diff
@@ -409,6 +409,16 @@ pub extern "C" fn glean_is_dirty_flag_set() -> u8 {
...
+    with_glean_value_mut(|glean| glean.handle_client_active());
+}
+
+#[no_mangle]
+pub extern "C" fn handle_client_inactive() {
```
badboy, line 418:
```suggestion
pub extern "C" fn glean_handle_client_inactive() {
```

badboy (changes requested):

glean-core/ffi/src/lib.rs
badboy: which brings up the questions once again: does it need the `client` in there?

Dexterp37: `glean_handle_active` seems to imply glean becomes active. But we're really looking for the product/client activity. I'm open to better wording :)

badboy: hm, why don't you want an active glean?!?!?!

Dexterp37: Too much sport :) Your call!

badboy: After discussion in the channel we settle with the chosen name here.

PR #1459, opened 2021-01-26
Revert "Bump criterion from 0.3.3 to 0.3.4"

badboy:
This reverts commit 2eae7c941c665a614abd2d10a9d38dc7468ead65.

---

No changelog, no git tag on the repo.
This definitely doesn't seem to be a patch release.
It's causing intermittents on CI, probably because it requires more memory during compilation/linking.

I should have been more careful in merging #1451

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -194,16 +194,16 @@ dependencies = [
 
 [[package]]
 name = "criterion"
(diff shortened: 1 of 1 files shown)


badboy:
skipping review, I did the initial merge myself too

PR #1465, opened 2021-01-28
Swift: use the client activity API to handle foreground/background events

badboy:


Dexterp37 (changes requested):
Why are tests failing?

I can't seem to find where in the code the events ping gets generated, may be due to that? (if we record > 500 events)

Moreover, this is lacking a changelog entry :)

badboy:
> Why are tests failing?

I have to find out.

> I can't seem to find where in the code the events ping gets generated, may be due to that? (if we record > 500 events)

That's internal and shouldn't change 🤔

> Moreover, this is lacking a changelog entry :)

True that, will add it

travis79 (approved):
Kind of annoying to have to ignore all the inital "active" pings for tests, but beats intermittent test failures.  LGTM!

Dexterp37 (approved):

PR #1470, opened 2021-01-29
Remove the minimal but unused benchmarking

badboy:
It's starting to cause problems on CI, probably because of linking
issues or running out of memory.
It's been untouched and unused in a long while now, let's just remove
it.


mdboom (approved):
Hah.  I forgot we even had this.  I suspect our performance issues (if any) are all at a higher-level (integration into async startup of our apps etc...) so this probably didn't bring a ton of value.

PR #1471, opened 2021-01-29
Use Python 3.9 on macOS on release

badboy:
---

Now we release everything else with Python 3.8.
Is that gonna be a problem?

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -953,10 +953,10 @@ jobs:
           name: Build macOS wheel
           command: |
             cd glean-core/python
-            .venv3.8/bin/python3 setup.py bdist_wheel
+            .venv3.9/bin/python3 setup.py bdist_wheel
             # Requires that the TWINE_USERNAME and TWINE_PASSWORD environment
             # variables are configured in CircleCI's environment variables.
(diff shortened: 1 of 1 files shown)


mdboom (approved):
I assume this is just because the macos images now have Python 3.9 as a default?  I guess to be more bullet-proof, we could detect that and adjust these paths accordingly, but maybe not worth the effort for something that changes every year or so.

It shouldn't be any issue what Python we use to build and submit the wheels -- it's mainly just packaging files up into a zip.  And we test all supported versions of Python elsewhere...

badboy:
> I assume this is just because the macos images now have Python 3.9 as a default? I guess to be more bullet-proof, we could detect that and adjust these paths accordingly, but maybe not worth the effort for something that changes every year or so.
> 
> It shouldn't be any issue what Python we use to build and submit the wheels -- it's mainly just packaging files up into a zip. And we test all supported versions of Python elsewhere...

Yup, we* recently upgraded the Xcode image we build on, which now has a newer Python.
The newest release thus didn't fully happen.

I don't think making it bullet-proof is a priority right now. It's a bit annoying if we forget that but not breaking too much.

PR #1477, opened 2021-02-01
Swift: replace foreground notification with activation notification

badboy:
`applicationWillEnterForeground`[1] is posted _before_ the app is actually
in the foreground, when it leaves the background stae.
This notification is NOT posted on first app launch (because it doesn't
leave a _background_ state).

Therefore we previously triggered `handleForegroundEvent` in the
constructor to work accordingly.
That's not wrong for all current uses of Glean (I assume),
but has two things worth pointing out:

1. It's different from Android behavior, where we do
   `handleForegroundEvent` only on an event (`ON_START`).
2. Because of that Android tests won't trigger `handleForegroundEvent`
   in unit tests, but only within a real app. We therefore can observe
   state after `Glean.initialize` but before `handleForegroundEvent`.
   We can't observe that currently in iOS.

`didBecomeActiveNotification`[2] on the other hand is posted when the
application is really active.
It can be observed on first startup.
Which should more closely match the behavior we also have on Android.

See also #1476, which we are currently unable to test.
We might be able to test it with this change.

[1]: https://developer.apple.com/documentation/uikit/uiapplication/1622944-willenterforegroundnotification
[2]: https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification

### DIFF
--- a/glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
+++ b/glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
@@ -8,8 +8,8 @@ class GleanLifecycleObserver {
     init() {
         NotificationCenter.default.addObserver(
             self,
-            selector: #selector(GleanLifecycleObserver.appWillEnterForeground(notification:)),
-            name: UIApplication.willEnterForegroundNotification,
+            selector: #selector(GleanLifecycleObserver.didBecomeActiveNotification(notification:)),
+            name: UIApplication.didBecomeActiveNotification,
             object: nil
         )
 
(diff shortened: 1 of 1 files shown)


travis79 (changes requested):
I believe I recall why we didn't use this before...  From the docs on `didBecomeActiveNotification`: "loses focus when an overlay window pops up".  Using `didBecomeActive` fires if any dialog pops up (like a notification, etc), if I remember correctly, which was triggering startup pings without the background pings or something like that.  Would you mind verifying that this will still get us a "pair" of active/inactive pings and not active/active/inactive or something?

badboy:
> I believe I recall why we didn't use this before... From the docs on `didBecomeActiveNotification`: "loses focus when an overlay window pops up". Using `didBecomeActive` fires if any dialog pops up (like a notification, etc), if I remember correctly, which was triggering startup pings without the background pings or something like that. Would you mind verifying that this will still get us a "pair" of active/inactive pings and not active/active/inactive or something?

Hm ... I didn't really look if we ever documented why we choose what we choose.
It's gonna be hard to properly verify that and I don't think I will have the time to do a thorough investigation on this.

If anything rejecting this PR acts as some sort of documentation.

travis79:
To follow up, I've agreed to look into this PR this week to see if my memory serves me regarding the `didBecomeActive` notification's behavior. I will report back once I have more info.

badboy:
Due to the unclear behavior with overlays/etc. as called out above it's probably not worth it.
To note: we didn't test this out again, however we trust the _current_ behaviour enough to not touch it yet.

PR #1478, opened 2021-02-01
Bug 1586764: Record the number of pings sent

mdboom:
This still needs data review, but I thought it might be worth getting some review on the implementation first.


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v34.0.0...main)
 
+* General
+  * A new metric `glean.validation.pings_submitted` tracks the number of pings sent. It is included on both the `metrics` and `baseline` pings.
```
badboy, line 6:
```suggestion
  * A new metric `glean.validation.pings_submitted` tracks the number of pings sent. It is included in both the `metrics` and `baseline` pings.
```

glean-core/tests/ping.rs
```diff
@@ -101,3 +102,89 @@ fn empty_pings_with_flag_are_sent() {
     assert_eq!(false, ping2.submit(&glean, None).unwrap());
     assert_eq!(1, get_queued_pings(glean.get_data_path()).unwrap().len());
 }
+
+#[test]
+fn test_pings_submitted_metric() {
+    let (mut glean, _temp) = new_glean(None);
+
+    let pings_submitted = LabeledMetric::new(
```
badboy, line 110: Can you add a comment saying that we reconstruct it here to check its values instead of reaching within the library internals?

glean-core/tests/ping.rs
```diff
@@ -101,3 +102,89 @@ fn empty_pings_with_flag_are_sent() {
...
+            .get("metrics")
+            .test_get_value(&glean, "metrics")
+    );
+    assert_eq!(
+        None,
+        pings_submitted
+            .get("baseline")
+            .test_get_value(&glean, "metrics")
+    );
+    assert_eq!(
```
badboy, line 151: On first read I kinda missed the `test_get_value` parameter.
Can we add a newline and maybe a comment here, making it clear that we check once for the metrics ping storage and once for baseline ping storage?

glean-core/tests/ping.rs
```diff
@@ -101,3 +102,89 @@ fn empty_pings_with_flag_are_sent() {
...
+            .get("metrics")
+            .test_get_value(&glean, "metrics")
+    );
+    assert_eq!(
+        Some(1),
+        pings_submitted
+            .get("baseline")
+            .test_get_value(&glean, "metrics")
+    );
+    assert_eq!(
```
badboy, line 178: as above

glean-core/tests/ping.rs
```diff
@@ -101,3 +102,89 @@ fn empty_pings_with_flag_are_sent() {
...
+            .test_get_value(&glean, "baseline")
+    );
+    assert_eq!(
+        None,
+        pings_submitted
+            .get("baseline")
+            .test_get_value(&glean, "baseline")
+    );
+
+    assert!(baseline_ping.submit(&glean, None).unwrap());
```
badboy, line 164: Can we trigger 2 baseline pings here? Likely to happen out in the wild and then we can test that we indeed get a `2` further below.

badboy: Actually maybe a comment here about which values we expect to reset would be helpful.

badboy (changes requested):
Implementation wise this seems easy enough.
Couple of small nits to better explain our testing around it would be helpful I think.

glean-core/metrics.yaml
```diff
@@ -596,3 +596,27 @@ glean.validation:
       - glean-team@mozilla.com
     expires:
       never
+  pings_submitted:
+    type: labeled_counter
+    description: >
```
badboy, line 601: Always getting confused by all the various YAML multi-line string options.
I think `>` will remove newlines (unless it's double lines, which turn into a single newline).
`|` keeps newlines as written. Is that what we should use here to keep the formatting?

badboy (changes requested):

docs/user/pings/metrics.md
```diff
@@ -12,6 +12,16 @@ Additionally, it is undesirable to mix metric recording from different versions
 
 > **Note:** As the `metrics` ping was specifically designed for mobile operating systems, it is not sent when using the Glean Python bindings.
 
+## Contents
+
+The metrics ping includes the following fields:
```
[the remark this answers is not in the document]
mdboom: Yeah -- I think these predate the generated metrics docs.  I'm going to update this to just link to the generated docs.

docs/user/pings/baseline.md
```diff
@@ -22,6 +22,7 @@ The baseline ping includes the following fields:
 | Field name | Type | Description |
 |---|---|---|
 | `duration` | Timespan | The duration, in seconds, of the last foreground session. Only available if `reason: inactive`. [^1] |
+| `glean.validation.pings_submitted` | Labeled counter | A count of the pings submitted, by ping type. The counts include the number of pings sent since the last metrics ping (including the last metrics ping itself). |
```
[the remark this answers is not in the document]
mdboom: Ditto -- will link to the generated docs.

glean-core/metrics.yaml
mdboom: Yeah, we should use `|` throughout to avoid surprises.

badboy (approved):

badboy:
(close/re-open in the hope this would rekick all of CI)

badboy:
that didn't help. Manually retriggered CI

PR #1479, opened 2021-02-01
Go into more detail on metric expiry

wlach:
Noticed this when working on mozilla/glean-dictionary#357. I'm not sure if my additions are actually correct, but that's part of the point of this PR. 😄


mdboom (approved):
Yes, this is correct.

One *very* corner detail, that may even not be worth mentioning, is that you can get:

- app collects on metric A
- app updates, the new version has expired metric A
- app starts up, sending data containing metric A

In this case, that last ping containing metric A will still have an app_version field from the previous version of the app, so from a DS perspective, it looks like it came from the old version, but in fact the submitting and uploading happened in the new version.

In other words, there's some smoke and mirrors to make this sentence appear like reality, one in fact it's not quite the case...

wlach:
That feels like quite the edge case! I suspect we should document that somewhere else.

PR #1482, opened 2021-02-03
Bug 1690780 - Coverage support

mdboom:
Just a playground to see if I can get the pieces working together.

Depends on https://github.com/mozilla/glean_parser/pull/272 for the glean_parser side of things.

Example coverage uploaded to codecov.io here: https://codecov.io/gh/mozilla/glean/src/ed8704ac50bcdfcb44ddd3a1fd6004d0eef57675/glean-core/metrics.yaml

Unfortunately, this requires forking the codecov.io uploader, since it explicitly filters out coverage reporting on `yaml` files.  (In fairness "coverage" on yaml files is unusual, but seems like it shouldn't be out-and-out disallowed).  Anyway, we might have to address that upstream in order for codecov.io to be a viable solution for our users of this feature.

Other than that, the implementation here turned out much simpler than in the proposal (reminder to go back and update the proposal with where we ended up).  By using only a environment variable to turn the feature on (thanks @brizental for the suggestion) and just writing to a file as we go (thanks @chutten for the suggestion), there's no need to hook into the unit testing framework, and thus there's no need for an FFI or language bindings, which makes this all over quite a small change.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -60,12 +60,19 @@ commands:
           rust-version: <<parameters.rust-version>>
       - run:
           name: Test
-          command: cargo test --all --verbose -- --nocapture
+          command: GLEAN_TEST_COVERAGE=$(realpath glean_coverage.txt) cargo test --all --verbose -- --nocapture
       - run:
           name: Test Glean with rkv safe-mode
           command: |
             cd glean-core
             cargo test -p glean-core --features rkv-safe-mode -- --nocapture
+      - run:
+          name: Upload coverage report
+          command: |
+            sudo apt install python3-pip
+            pip3 install glean_parser
+            glean_parser coverage --allow-reserved -c glean_coverage.txt -f codecovio -o codecov.json glean-core/metrics.yaml
+            bash <(curl -s https://codecov.io/bash) -X yaml -f codecov.json
 
   install-rustup:
     steps:
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,7 +7,7 @@
     * The new API exists for all language bindings (Kotlin, Swift, Rust, Python).
   * Updated `glean_parser` version to 2.5.0
   * Change the `fmt-` and `lint-` make commands for consistency ([#1526](https://github.com/mozilla/glean/pull/1526))
-  * The Glean SDK can now produce testing coverage reports for your metrics.
+  * The Glean SDK can now produce testing coverage reports for your metrics ([#1482](https://github.com/mozilla/glean/pull/1482/files)).
 * Python
   * Update minimal required version of `cffi` dependency to 1.13.0 ([#1520](https://github.com/mozilla/glean/pull/1520)).
 * RLB
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,7 @@
     * The new API exists for all language bindings (Kotlin, Swift, Rust, Python).
   * Updated `glean_parser` version to 2.5.0
   * Change the `fmt-` and `lint-` make commands for consistency ([#1526](https://github.com/mozilla/glean/pull/1526))
+  * The Glean SDK can now produce testing coverage reports for your metrics.
 * Python
   * Update minimal required version of `cffi` dependency to 1.13.0 ([#1520](https://github.com/mozilla/glean/pull/1520)).
 * RLB
--- a/docs/dev/code_coverage.md
+++ b/docs/dev/code_coverage.md
@@ -5,7 +5,7 @@
 > which suggests it has a lower chance of containing undetected software bugs compared to a program with low test coverage.
 > ([Wikipedia](https://en.wikipedia.org/wiki/Code_coverage))
 
-This chapter describes how to generate a traditional code coverage report over the Kotlin, Rust, Swift and Python code in the Glean SDK repository. To learn how to generate a coverage report about what metrics your project is testing, see the user documentation on generating testing coverage reports.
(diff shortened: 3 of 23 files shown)


.circleci/config.yml
```diff
@@ -60,12 +60,19 @@ commands:
...
       - run:
           name: Test Glean with rkv safe-mode
           command: |
             cd glean-core
             cargo test -p glean-core --features rkv-safe-mode -- --nocapture
+      - run:
+          name: Upload coverage report
+          command: |
+            sudo apt install python3-pip
+            pip3 install git+https://github.com/mdboom/glean_parser.git@coverage-analysis
```
mdboom, line 73: This just installs a branch of glean_parser.  No longer necessary once that branch is merged and we make a release.

.circleci/config.yml
```diff
@@ -60,12 +60,19 @@ commands:
           rust-version: <<parameters.rust-version>>
       - run:
           name: Test
-          command: cargo test --all --verbose -- --nocapture
+          command: GLEAN_TEST_COVERAGE=$(realpath glean_coverage.txt) cargo test --all --verbose -- --nocapture
```
mdboom, line 63: Have to use `realpath` here since `cargo test` changes the working directory.  This makes sure there is a single file (even though there are multiple crates tested) and it ends up where we expect it to.

.circleci/config.yml
```diff
@@ -60,12 +60,19 @@ commands:
...
           command: |
             cd glean-core
             cargo test -p glean-core --features rkv-safe-mode -- --nocapture
+      - run:
+          name: Upload coverage report
+          command: |
+            sudo apt install python3-pip
+            pip3 install git+https://github.com/mdboom/glean_parser.git@coverage-analysis
+            glean_parser coverage --allow-reserved -c glean_coverage.txt -f codecovio -o codecov.json glean-core/metrics.yaml
+            bash <(curl -s https://raw.githubusercontent.com/mdboom/codecov-bash/master/codecov) -f codecov.json
```
mdboom, line 75: Had to fork codecov.io's upload script so it wouldn't ignore `yaml` files...

badboy:
initial commit that strips out yaml files in the uploader is here: https://github.com/codecov/codecov-bash/commit/aa5f5839776e02769eac09aa02f3360799da6eae

glean-core/src/coverage.rs
```diff
@@ -0,0 +1,32 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
badboy, line 3: Can we add some module docs here (`//!`) of what this does. Calling out the environment variable and the format in which the file is written ("one metric id per line")

glean-core/src/storage/mod.rs
```diff
@@ -142,6 +143,8 @@ impl StorageManager {
 
         storage.iter_store_from(metric_lifetime, &store_name, None, &mut snapshotter);
 
+        record_coverage(metric_id);
```
badboy, line 146: The proposal said `snapshot_metric` can be called from one non-test place. Do we need to handle this here?

badboy (commented):
This is ... too simple. :D 

I guess the difference to the proposal is that we don't need to keep any state beyond the `GLEAN_TEST_COVERAGE` file part?
When we want to integrate this into test harnesses, we could just say "set this before you run your tests" or we can build out some helpers (such as our GleanTestRule) that sets that environment variable, right?)

mdboom:
> When we want to integrate this into test harnesses, we could just say "set this before you run your tests" or we can build out some helpers (such as our GleanTestRule) that sets that environment variable, right?)

Yeah, I think depending on the details, they could just set the environment variable before running the test harness (as here).  I haven't tested that with Gradle/JUnit yet because my Android build environment seems to have broken...

mdboom:
The upstream changes in glean_parser and codecov.io bash uploader have landed, so this is ready for a final review.

CHANGELOG.md
```diff
@@ -7,6 +7,7 @@
     * The new API exists for all language bindings (Kotlin, Swift, Rust, Python).
   * Updated `glean_parser` version to 2.5.0
   * Change the `fmt-` and `lint-` make commands for consistency ([#1526](https://github.com/mozilla/glean/pull/1526))
+  * The Glean SDK can now produce testing coverage reports for your metrics.
```
badboy, line 10:
```suggestion
  * The Glean SDK can now produce testing coverage reports for your metrics ([#1482](https://github.com/mozilla/glean/pull/1482/files)).
```

badboy: Should we add a link to the book here?

docs/dev/code_coverage.md
```diff
@@ -5,6 +5,8 @@
 > which suggests it has a lower chance of containing undetected software bugs compared to a program with low test coverage.
 > ([Wikipedia](https://en.wikipedia.org/wiki/Code_coverage))
 
+This chapter describes how to generate a traditional code coverage report over the Kotlin, Rust, Swift and Python code in the Glean SDK repository. To learn how to generate a coverage report about what metrics your project is testing, see the user documentation on generating testing coverage reports.
```
badboy, line 8: Should this link to the user documentation?

docs/user/user/testing-metrics.md
```diff
@@ -266,3 +268,53 @@ TODO. To be implemented in [bug 1648448](https://bugzilla.mozilla.org/show_bug.c
...
+
+```bash
+GLEAN_TEST_COVERAGE=$(realpath glean_coverage.txt) make test
+```
+
+### Post-processing the results
+
+A post-processing step is required to convert the raw output in the file specified by `GLEAN_TEST_COVERAGE` into usable output for coverage reporting tools.
+
+This post-processor is available in the `coverage` subcommand in the [`glean_parser`](https://github.com/mozilla/glean_parser) tool.
```
badboy, line 291: Should this call out that (right now) we only support the codecov format?

badboy (approved):
Good from my side, couple of questions, mostly around documentation.

PR #1483, opened 2021-02-04
Bug 1690751 - Clear the dirty ping on shutdown in the RLB

Dexterp37:
This makes sure no baseline ping with reason `dirty_startup` is generated when starting RLB consumers after a clean shutdown.

### DIFF
--- a/glean-core/rlb/src/test.rs
+++ b/glean-core/rlb/src/test.rs
@@ -311,6 +311,67 @@ fn sending_of_startup_baseline_ping() {
 }
 
 #[test]
+fn no_dirty_baseline_on_clean_shutdowns() {
+    let _lock = lock_test();
+
+    // Create an instance of Glean, wait for init and then flip the dirty
+    // bit to true.
+    let data_dir = new_glean(None, true);
+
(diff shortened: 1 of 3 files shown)


glean-core/rlb/src/test.rs
```diff
@@ -310,6 +310,67 @@ fn sending_of_startup_baseline_ping() {
...
+        ) -> net::UploadResult {
+            self.sender.send(url).unwrap();
+            net::UploadResult::HttpStatus(200)
+        }
+    }
+
+    // Create a custom configuration to use a fake uploader.
+    let tmpname = data_dir.path().display().to_string();
+
+    // Now reset Glean: it should still send a baseline ping with reason
```
badboy, line 351: Wait, isn't it that we explicitly _don't_ expect a baseline ping now?

badboy (changes requested):

glean-core/rlb/src/test.rs
Dexterp37: And Here I thought I had the perfect PR.

badboy (approved):

PR #1485, opened 2021-02-04
Bug 1688178 - Big Sur/M1 updates: xcframeworks, using Xcode clang, lipo

badboy:
This is my first try at building Glean iOS on a M1 (arm) machine.

It essentially requires us to use xcframeworks, because the Carthage workaround is not suitable on M1 machines (as we can't exclude the arm64 target, because that IS the host target).
I can build Glean iOS now on my M1.
I can't run the tests though, because [Rust doesn't know the difference between the host system and the simulator](https://github.com/rust-lang/rust/issues/81632).

It's also likely that our sample app won't work the same anymore, at least not if I also want to get that compilable and runnable on an M1. It will need to start using xcframeworks too.
Carthage currently cannot produce xcframework binary releases, so the sample will always need to compile Glean.
Consumers wouldn't be affected, as long as they don't require xcframeworks. It just means I can't build those projects.

* [x] Test it on an Intel Mac
* [x] Ensure tests on CI work
* [x] Ensure sample app on CI works
* [x] Ensure our release still works for consumers
	* Actually probably not as important given that the only active consumer (firefox-ios) gets it through application-services.
* [x] Ensure that application-services doesn't break

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -656,10 +656,8 @@ jobs:
             xcrun instruments -w list || true
             # See https://circleci.com/docs/2.0/testing-ios/#pre-starting-the-simulator
             xcrun instruments -w "iPhone 11 (14" || true
-      - attach_workspace:
-          at: .
       - run:
-          name: Use binary build of Glean
+          name: Use current commit of Glean
           command: |
             GLEAN_PATH="$(pwd)"
             CARTFILE_PATH="${GLEAN_PATH}/samples/ios/app/Cartfile"
@@ -1118,8 +1116,6 @@ workflows:
             branches:
               only: main
       - iOS integration test:
-          requires:
-            - iOS build and test
           filters:
             branches:
               only: main
@@ -1170,7 +1166,7 @@ workflows:
               ignore: main
       - iOS integration test:
           requires:
-            - iOS build and test
+            - hold
           filters:
             branches:
               ignore: main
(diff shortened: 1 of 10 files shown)


travis79 (approved):
havewebeenappledyet.com?   LGTM!

.circleci/config.yml
```diff
@@ -667,18 +661,14 @@ jobs:
...
             CARTFILE_PATH="${GLEAN_PATH}/samples/ios/app/Cartfile"
-            # The previous step generated a binary file and the corresponding JSON manifest
-            JSON_URL="https://circleci.com/api/v1.1/project/github/mozilla/glean/${IOS_BUILD_NUM}/artifacts/0/dist/mozilla.glean.json"
 
             echo "Current Cartfile:"
             cat "${CARTFILE_PATH}"
             echo "================="
             echo "New Cartfile:"
-            sed -i.bak "/mozilla\/glean/s#.*#binary \"$JSON_URL\" ~> 0.0.1-SNAPSHOT#" "$CARTFILE_PATH"
+            sed -i.bak "/mozilla\/glean/s#.*#github \"mozilla/glean\" \"${CIRCLE_SHA1}\"#" "$CARTFILE_PATH"
```
travis79, line 671: Eliminating the "snapshot" really cleaned up this section of the script

travis79 (approved):
Still looks good to me!

.circleci/config.yml
badboy: ... but also slower :(

badboy:
Now that the 2 iOS tasks don't depend on each other we can run them in parallel! The summed up runtime will be the same, but the actual runtime should be shorter!

PR #1486, opened 2021-02-04
Bug 1690887: Fix Android build for Java 11

mdboom:
My system Java recently upgraded to Java 11, breaking the Glean build.
CI is still on Java 8 so is not seeing these issues, but if we can find
a way to be compatible across these versions, I guess we should try.

One issue was just requiring a new jacoco config flag (which I admit I don't
really understand, but it works):
https://github.com/gradle/gradle/issues/5184#issuecomment-457865951

Behind that is the old version of okhttp3.mockwebserver is no longer compatible.
Upgrading that lead to some incompatible API changes, mostly that the API is now
much more shouty -- resulting in most of the changes here.

~It remains to be seen whether these fixes will work on CI with Java 8.~  They work!

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -18,7 +18,7 @@ buildscript {
         jna: '5.6.0',
         junit: '4.12',
         mockito: '2.28.2',  // This is different than a-c, but we're fine, it's only tests.
-        mockwebserver: '3.10.0',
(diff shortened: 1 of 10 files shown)


travis79 (approved):
All those `!!` make the tests look edgy!!

badboy:
you!! really!! want!! to!! make!! this!! work!! don't!! you??

badboy (approved):

PR #1487, opened 2021-02-05
Docs: Fix timespan example

pocmo:
Used the sample and noticed that the linter complained about `milliseconds` not being right. :)


badboy (approved):

badboy:
It's a minimal doc change, so I'm not gonna rekick CI and wait for it.

PR #1490, opened 2021-02-08
Bug 1691080 - Check for glean.validation keys as well

badboy:
We recently got a new `glean.validation.pings_submitted` labeled counter
that now shows up.
We need to take care of that in tests.


mdboom:
Thanks!  Sorry, I forgot to run the iOS test on that PR, which would have caught this...

PR #1491, opened 2021-02-08
Bug 1655112: Add messages to NullPointerExceptions in testing API

mdboom:
This just adds a better informational message when a `NullPointerException` is thrown from the Kotlin testing APIs.

As the bug contemplates, we could throw a different exception, but that's strictly speaking a breaking API change...  How much that matters in practice is perhaps up for debate, however.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -478,7 +478,7 @@ open class GleanInternalAPI internal constructor () {
 
         val ptr = LibGleanFFI.INSTANCE.glean_experiment_test_get_data(
             experimentId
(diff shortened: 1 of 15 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -478,7 +478,7 @@ open class GleanInternalAPI internal constructor () {
 
         val ptr = LibGleanFFI.INSTANCE.glean_experiment_test_get_data(
             experimentId
-        )!!
+        ) ?: throw NullPointerException("experiment data is not set")
```
Dexterp37, line 481:
```suggestion
        ) ?: throw NullPointerException("Experiment data is not set")
```

Dexterp37 (approved):
r+wc

PR #1494, opened 2021-02-10
Bug 1677441: Add launch_with_glean and launch_with_glean_mut for readibility

ChinYing-Li:
Two functions were added: `launch_with_glean` & `launch_with_glean_mut`; test cases passes.
When running `make clippy`, clippy says `extra` in the following snippet is a redundant clone:
```
pub fn set_experiment_active(
     branch: String,
     extra: Option<HashMap<String, String>>,
 ) {
     dispatcher::launch_with_glean(move |glean| {
     glean.set_experiment_active(experiment_id.to_owned(), branch.to_owned(), extra.to_owned)
     })
 }
```
while the original `set_experiment_active`, which doesn't causes the "redundant clone" error, looks like this:
```
pub fn set_experiment_active(
     branch: String,
     extra: Option<HashMap<String, String>>,
 ) {
    dispatcher::launch(move || {
        with_glean(|glean| {
            glean.set_experiment_active(
                experiment_id.to_owned(),
                branch.to_owned(),
                extra.to_owned(),
            )
        });
     })
 }
```
I am still trying to figure out clippy's rule on this, and had removed `to_owned()` for now to make clippy happy.
@badboy Any suggestion is appreciated!

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -151,6 +151,17 @@ where
     f(&mut lock)
 }
 
+/// Launches a new task on the global dispatch queue with a reference to the Glean singleton.
+fn launch_with_glean(callback: impl FnOnce(&Glean) + Send + 'static) {
+    dispatcher::launch(|| crate::with_glean(callback));
+}
+
+/// Launches a new task on the global dispatch queue with a mutable reference to the
+/// Glean singleton.
+fn launch_with_glean_mut(callback: impl FnOnce(&mut Glean) + Send + 'static) {
+    dispatcher::launch(|| crate::with_glean_mut(callback));
+}
+
 /// Creates and initializes a new Glean object.
 ///
 /// See [`glean_core::Glean::new`] for more information.
@@ -352,10 +363,8 @@ pub fn shutdown() {
         return;
     }
 
-    dispatcher::launch(move || {
-        with_glean_mut(|glean| {
-            glean.set_dirty_flag(false);
-        })
+    crate::launch_with_glean_mut(|glean| {
+        glean.set_dirty_flag(false);
     });
 
     if let Err(e) = dispatcher::shutdown() {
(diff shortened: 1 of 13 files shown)


badboy:
> I am still trying to figure out clippy's rule on this, and had removed `to_owned()` for now to make clippy happy.

That's probably the right approach. If it compiles without `to_owned` then that's the better way.
Probably clippy wasn't able to look through 2 closures before.

glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -46,6 +50,17 @@ pub fn launch(task: impl FnOnce() + Send + 'static) {
...
+/// Launches a new task on the global dispatch queue with a reference to the Glean singleton.
+pub fn launch_with_glean(callback: impl FnOnce(&Glean) + Send + 'static) {
+    dispatcher::launch(|| crate::with_glean(callback));
+}
+
+/// Launches a new task on the global dispatch queue with a mutable reference to the
+/// Glean singleton.
+pub fn launch_with_glean_mut(callback: impl FnOnce(&mut Glean) + Send + 'static) {
+    launch(|| crate::with_glean_mut(callback));
+}
```
badboy, line 62: Right now the dispatcher module has no knowledge about Glean, which is a good thing.

Let's move these functions into the top level module, e.g. [right after `with_glean_mut` here](https://github.com/mozilla/glean/blob/4445cf3a1bcec8c696264dbcc02c8aae17c355db/glean-core/rlb/src/lib.rs#L152).

Whereever we call it then we can use `crate::launch_with_glean` instead of `dispatcher::launch(...)`.

badboy (changes requested):
Good work, just one detail I'd like to see changed. See inline.

glean-core/rlb/src/dispatcher/global.rs
ChinYing-Li: Thanks! Just made the change and squashed the two commits into one.

glean-core/rlb/src/lib.rs
```diff
@@ -151,6 +151,17 @@ where
     f(&mut lock)
 }
 
+/// Launches a new task on the global dispatch queue with a reference to the Glean singleton.
+pub fn launch_with_glean(callback: impl FnOnce(&Glean) + Send + 'static) {
```
badboy, line 155: These don't need to be `pub` (as that would make them accessible outside of this crate).

```suggestion
fn launch_with_glean(callback: impl FnOnce(&Glean) + Send + 'static) {
```

glean-core/rlb/src/lib.rs
```diff
@@ -151,6 +151,17 @@ where
...
+    dispatcher::launch(|| crate::with_glean(callback));
+}
+
+/// Launches a new task on the global dispatch queue with a mutable reference to the
+/// Glean singleton.
+pub fn launch_with_glean_mut(callback: impl FnOnce(&mut Glean) + Send + 'static) {
```
badboy, line 161:
```suggestion
fn launch_with_glean_mut(callback: impl FnOnce(&mut Glean) + Send + 'static) {
```

badboy (changes requested):
One last thing but then this is ready to land!

glean-core/rlb/src/lib.rs
ChinYing-Li: Yep that makes sense! Thank you for reviewing.

badboy (approved):

PR #1495, opened 2021-02-11
Bug 1690088: Use explicitly passed-in buildinfo, if provided

mdboom:
Depends on https://github.com/mozilla/glean_parser/pull/276.

The sample application demonstrates the straightforward changes that would be required of Fenix to use this.  This shouldn't be a breaking change, as the new parameter is optional.  Ideally, though, once we've confirmed this is a better approach, we would make it required and then remove the package manager code path.


badboy (commented):
(please re-request review when turning this from draft into full PR)

Just as straight forward as the glean_parser changes.

I'd like @Dexterp37's eyes on this before we land it.
Can we also do a quick check in Fenix that what our `buildInfo` reports and what we currently report is the same? I could do that tomorrow on my local dev build by adding some logging.

mdboom:
@badboy: This is using a real glean_parser release and passing CI.  I agree @Dexterp37 should have a look.  I have not integrated this into Fenix locally to double check that the data matches.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -129,8 +137,14 @@ open class GleanInternalAPI internal constructor () {
...
         uploadEnabled: Boolean,
-        configuration: Configuration = Configuration()
+        configuration: Configuration = Configuration(),
+        buildInfo: BuildInfo? = null
     ) {
+        this.buildInfo = buildInfo
+        if (buildInfo == null) {
+            Log.w(LOG_TAG, "Passing a non-null BuildInfo object will be required in the future. See bug 1691953")
+        }
+
```
Dexterp37, line 147: If we wanted to get fancy, we could probably create an overload for `Glean.initialize` without `buildInfo` (i.e. the current signature) and mark it as [deprecated](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/) (which talks nicely to the compiler and Android Studio!).

It's a super nit, not very important, though.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -814,4 +814,15 @@ class GleanTest {
         val docType = request.path!!.split("/")[3]
         assertEquals("deletion-request", docType)
     }
+
+    @Test
+    fun `test passing in explicit BuildInfo`() {
```
Dexterp37, line 819: Can we also add a similar test to the [Java compatibility test](https://searchfox.org/glean/source/glean-core/android/src/test/java/mozilla/telemetry/glean/GleanFromJavaTest.java)?

docs/user/general-api.md
```diff
@@ -61,13 +62,14 @@ class SampleApplication : Application() {
             applicationContext,
             // Here, `settings()` is a method to get user preferences, specific to
             // your application and not part of the Glean SDK API.
-            uploadEnabled = settings().isTelemetryEnabled
+            uploadEnabled = settings().isTelemetryEnabled,
+            buildInfo = GleanBuildInfo.buildInfo
```
Dexterp37, line 66: I wonder if future versions of the language bindings could even do without this, with Glean automatically fetching the generated code.

Dexterp37 (approved):
r+wc! This looks nice, I agree with @badboy , let's make sure this works in Fenix before triggering the domino-updates-effect!

docs/user/general-api.md
badboy: Not really? We need to generate this in the final application because only there the build info is valid (or even existent?)
The only way we get to that data is if someone gives it to us (this is the same problem as with `registerPings`: We don't know about them unless they get passed in)

badboy:
Oh one more thing I just realized: Is the build info stuff valid for libraries? If we generate code that doesn't work for libraries that would be bad.

mdboom:
> Oh one more thing I just realized: Is the build info stuff valid for libraries? If we generate code that doesn't work for libraries that would be bad.

From the docs, I think it *should*, but we will know when we integrate this in android-components.  This definitely wouldn't work if the library is built with something other than Gradle (since the `BuildConfig` object this uses is injected by Gradle).  But Glean isn't really supported on non-Gradle build systems at the moment anyway...

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
mdboom: This is a good idea.  Might as well...

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
mdboom: Yep.  Good idea.

mdboom: I struggled to use the Glean testing API from those Java tests, and it doesn't seem like anything else there does that already.  I filed [this bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1692506) for that, and maybe we don't need to tackle that here?

Dexterp37: Oh, interesting. Sure that's fine!

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanFromJavaTest.java
```diff
@@ -76,4 +78,20 @@ public void testCanSendPing() {
...
+
+    @Test
+    public void testPassingExplicitBuildInfo() {
+        Configuration config =
+            new Configuration(Configuration.DEFAULT_TELEMETRY_ENDPOINT, "test-channel");
+        BuildInfo buildInfo =
+            new BuildInfo("foo", "c0ffee");
+        Glean.INSTANCE.initialize(appContext, true, config, buildInfo);
+
+        // The testing API doesn't seem to be available from these Java tests, so the best
```
Dexterp37, line 90: nit: can you reference the bug you just filed about the testing API in java, within this comment?

Dexterp37 (approved):

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanFromJavaTest.java
mdboom: Done

CHANGELOG.md
```diff
@@ -2,10 +2,12 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v34.1.0...main)
 
-* Kotlin
+* Android
+  * `Glean.initialize` can now take a `buildInfo` parameter to pass in build time version information, and avoid calling out to the Android package manager at runtime. A suitable instance is generated by `glean_parser` in `${PACKAGE_ROOT}.GleanMetrics.GleanBuildInfo.buildInfo`.
```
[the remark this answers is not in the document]
badboy: Can we call out that this makes _not passing it_ deprecated?

badboy (approved):
Looks good.

Can we maybe squash this into 2 reasonable commits?
1. Upgrade glean_parser
2. The rest of this PR

PR #1498, opened 2021-02-12
iOS: Don't exclude arm64 build for the simulator

badboy:
This currently requires a local build of rustc with https://github.com/rust-lang/rust/pull/81966 to enable the target (and also LLVM 12, LLVM 11 untested). Thus we can't really land that before that becomes stable.

I think it would also require direct consumers to switch over to xcframeworks.
It's untested in app-services, I guess it _could_ work for Intel builds there, but this would need to be tested.

No need for a formal review or test yet.

* [ ] Wait for [upstream changes in rustc](https://github.com/rust-lang/rust/pull/81966)
* [ ] Test it on an Intel machine
* [ ] Test it in app-services

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -555,7 +555,6 @@ jobs:
           name: Setup build environment
           command: |
             rustup target add aarch64-apple-ios x86_64-apple-ios
-            cargo install cargo-lipo
 
             # Bootstrap dependencies
             bin/bootstrap.sh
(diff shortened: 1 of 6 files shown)


travis79:
I'll be prototyping the "Better Builds" proposal next week, this may help me with packaging the iOS stuff as I'm sure that app-services will face a similar issue with this (as will Nimbus when we have iOS for it).

badboy:
`aarch64-apple-ios-sim` is available now on Nightly ... sort of. One needs to build `libstd`, but cargo (nightly) can easily do that.
As this only affects developers on M1 machines (like me) we can land this in our build script and I can move my dev environment off a fork of our own code base that already used these changes.

travis79 (approved):
Note: reminder to file a bug to change this once the feature has made it to release in rustc.

badboy:
> Note: reminder to file a bug to change this once the feature has made it to release in rustc.

That might or might not happen anytime soon unfortunately (we'll see, I'm in the rust zulip to see how we can push it forward)

badboy:
Filed https://bugzilla.mozilla.org/show_bug.cgi?id=1694605

PR #1499, opened 2021-02-12
Bug 1682244: Add set_now method to DatetimeMetric

ChinYing-Li:
A small patch to address the bug; any suggestion is appreciated!
I don't see these `set` methods in the test cases, or am I missing something?

### DIFF
--- a/glean-core/src/metrics/datetime.rs
+++ b/glean-core/src/metrics/datetime.rs
@@ -122,6 +122,16 @@ impl DatetimeMetric {
         glean.storage().record(glean, &self.meta, &value)
     }
 
+    /// Sets the metric to the current date/time with the timezone offset.
+    /// See [`self.set`] for details.
+    ///
+    /// # Arguments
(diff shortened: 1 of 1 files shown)


badboy:
Thanks for the work, unfortunately I have to reject it.
First of all, if we'd do this, the API should be tackled onto the RLB (`glean-core/rlb/`), not on the internal `glean-core` crate.

But more importantly: I'm not sure we really need it right now. I'd rather not expose more API surface on the RLB, especially given that this wraps something that's already possible.
The bug was backlogged because we didn't make a decision yet on whether we actually want that.

I'm happy to discuss what other bugs we could use your help with.
To avoid unnecessary work on your end I propose you leave a comment on the bug before you start working on it. I'll make sure we get back to you quickly to not block you.

ChinYing-Li:
Thanks for letting me know! Right, it's better to ask whether a bug is valid before starting to work on it.

PR #1504, opened 2021-02-17
Bug 1693103 - CI(iOS): Use the local git for Carthage instead of re-fetching it remotely

badboy:


badboy:
the yamllint check actually succeeded: https://app.circleci.com/pipelines/github/mozilla/glean/7104/workflows/7b01ab15-f149-4d9b-918f-7f61bf5ef169/jobs/122971
the failure here is from a previous run.

.circleci/config.yml
```diff
@@ -666,7 +666,7 @@ jobs:
             cat "${CARTFILE_PATH}"
             echo "================="
             echo "New Cartfile:"
-            sed -i.bak "/mozilla\/glean/s#.*#github \"mozilla/glean\" \"${CIRCLE_SHA1}\"#" "$CARTFILE_PATH"
+            sed -i.bak "/mozilla\/glean/s#.*#git \"${GLEAN_PATH}\" \"${CIRCLE_SHA1}\"#" "$CARTFILE_PATH"
```
travis79, line 669: I suppose that it is no longer a problem to have to point to github to get the prebuilt binary since we are being built/distributed via the megazord?  (That and I think that the iOS team has moved away from BuddyBuild, so maybe their current CI could build Rust if needed)

travis79 (approved):

.circleci/config.yml
badboy: This is only for our own sample app. We always depended on the current build, it's just that we managed to share that between the 2 iOS tasks on CircleCI, which saved some time.
I had to change that in #1485 to make xcframeworks work already, our builds are slightly longer (but we can run both tasks in parallel).

If you built the sample app locally you already used a git checkout.
Consumers can still use the binary built (as long as they are not on an M1 that is), but also firefox iOS uses the appservice's release, not ours anyway.

In our CI we can now use the local checkout instead of depending on GitHub, which fixes the issue that the commit of a PR might come from outside the repository.

PR #1506, opened 2021-02-18
Bug 1659706 - CI: Use up-to-date CircleCI-provided Android image again

badboy:
This brings Java 11.
We fixed support for that in commit 3ff81f0794696491cc6b094d74550fffe4e3211d


mdboom (approved):
LGTM, since the tests are passing.  Seems like a real improvement to be using a tag rather than a commit hash, too.

PR #1507, opened 2021-02-18
Bug 1685401 - Implement `before_next_submit` API

badboy:


CHANGELOG.md
```diff
@@ -2,11 +2,13 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v34.1.0...main)
 
+* General
+  * Introduce a new API `Ping#test_before_next_submit` to run a callback right before a custom ping is submitted.
```
chutten, line 6: Annotate with PR# and a list of languages you're supporting, if you please.

glean-core/ios/Glean/Metrics/Ping.swift
```diff
@@ -82,6 +96,15 @@ public class Ping<ReasonCodesEnum: ReasonCodes>: PingBase {
     /// - parameters:
     ///     * reason: The reason the ping is being submitted.
     public func submit(reason: ReasonCodesEnum? = nil) {
+        if let cb = self.testCallback {
+            do {
+                try cb(reason)
+            } catch {
+                assert(false, "Callback threw before submitting ping \(name).")
```
chutten, line 103: Huh. Well this is an interesting wrinkle in the semantics. Should we try to submit irrespective of thrown exceptions (like swift), or should we allow thrown exception to abort the submit (like Kotlin, Python)?

Either way we should be consistent. I'm leaning towards always submitting because I don't think granting an ability to suppress submission (even if solely in some nebulous "test mode") is a good idea.

glean-core/rlb/src/private/ping.rs
```diff
@@ -36,15 +45,43 @@ impl PingType {
...
+
+        let cb = Box::new(cb);
+        *test_callback = Some(cb);
+    }
 }
 
 #[inherent(pub)]
 impl glean_core::traits::Ping for PingType {
     fn submit(&self, reason: Option<&str>) {
+        // TODO: Ensure this is only usable in test mode.
```
chutten, line 78: Again: We have a test mode?!

glean-core/rlb/src/private/ping.rs
```diff
@@ -2,15 +2,24 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use std::sync::{Arc, Mutex};
+
 use inherent::inherent;
 
+type BoxedCallback = Box<dyn FnOnce(Option<&str>) + Send + 'static>;
```
chutten, line 9: Why does it need to be Send? I think I understand why everything else is there.

Makefile
```diff
@@ -26,7 +26,7 @@ $(GLEAN_PYENV)/bin/python3:
 	python3 -m venv $(GLEAN_PYENV)
 	$(GLEAN_PYENV)/bin/pip install --upgrade pip
 	$(GLEAN_PYENV)/bin/pip install --use-feature=2020-resolver -r glean-core/python/requirements_dev.txt
-	sh -c "if [ \"$(GLEAN_PYDEPS)\" == \"min\" ]; then \
+	sh -c "if [ \"$(GLEAN_PYDEPS)\" = \"min\" ]; then \
```
chutten, line 29: What's this about?

chutten (changes requested):
Some questions and comments, but I'm really liking how it looks.

glean-core/ios/Glean/Metrics/Ping.swift
badboy: I added this merely to make it easier on users.
In Swift one needs to explicitly annotate functions as `throwable` if they do.
`testGetValue` can throw, so users need to use `try metric.testGetValue()` everywhere. That is fine in test methods, because they are throwable to begin with and will mark the test as failed.

Now when folks use `testBeforeNextSubmit` they should be able to use similar code to test their metrics. They are only allowed to call `testBeforeNextSubmit` in test mode (as enforced by the `assertInTestingMode()` call).

If we say that the closure a user passes is not throwable, than users need to wrap their `try metric.testGetValue()` in a `do {} catch {}` block themselves, only to then also fail the test.
Therefore I think the better ergonomics are if we do the wrapping and assert to fail tests if that happens.

glean-core/rlb/src/private/ping.rs
badboy: No, we don't in the RLB.
We do in other bindings and are thus able to assert when using `testBeforeNextSubmit`. No such protection is given in the RLB.

Makefile
badboy: Side fix I stumbled upon when I saw the test fail.
`sh` is a POSIX shell, so it uses `[` (aka `test`) in its single-bracket form (because ... posix), and in posix-`test` one uses a single equal sign (`=`) for comparison (whereas `==` is a bash thing and not posix compatible)

Source: https://pubs.opengroup.org/onlinepubs/9699919799/utilities/test.html, why I know this? Good you asked! I once maintained scripts to install Redis _everywhere_ and thus the only thing I could ever rely on was POSIX and not bash. Trust me, you don't want to maintain that.

CHANGELOG.md
badboy: Will do

glean-core/rlb/src/private/ping.rs
badboy: Because [sometimes we send around pings](https://github.com/mozilla/glean/blob/before-next-submit/glean-core/rlb/src/lib.rs#L492-L494).
And because we store the closure within the ping we need to ensure it can be safely send (that is: it doesn't hold on to data that can't be send, such as local references).

glean-core/ios/Glean/Metrics/Ping.swift
chutten: Ah, so we have equivalent ergo without having equivalent semantics. Fair enough.

chutten (approved):

badboy:
I'm starting to think the Python failures are legit and I need to look into them.

PR #1508, opened 2021-02-18
Fix link to the dev book

badboy:
[doc only]


brizental (approved):
Phew, thanks for doing this :D

PR #1510, opened 2021-02-19
Bug 1654207: Remove LEGACY_TAG_PINGS from the Android GleanDebugActivity

ChinYing-Li:
This ready-for-review PR aims to resolve bug [1654297](https://bugzilla.mozilla.org/show_bug.cgi?id=1654207).
One test case had been removed to reflect the change.
Any suggestion is appreciated! @badboy


badboy (changes requested):
Looks good. Can you add an entry to the CHANGELOG.md?
This is a breaking change, so we should explicitly call it out.

badboy (approved):
Thanks!

PR #1513, opened 2021-02-22
Bumped version to 35.0.0

mdboom:


chutten (approved):
Breaking changes means major version update. Seems legit.

PR #1516, opened 2021-02-23
Bug1645166 - Add `rate` metric

chutten:


glean-core/src/metrics/mod.rs
```diff
@@ -173,6 +180,9 @@ impl Metric {
             Metric::Datetime(d, time_unit) => json!(get_iso_time_string(*d, *time_unit)),
             Metric::Experiment(e) => e.as_json(),
             Metric::Quantity(q) => json!(q),
+            Metric::Rate(num, den) => {
+                json!({"numerator": num, "denominator": den})
```
badboy, line 184: Didn't we want to explore if we can skip this metric if denominator is 0, because that's not particular helpful?

badboy (changes requested):
Overall this looks good already.
Minor things, see inline.

Should this also upgrade the glean_parser we require? Even though FOG needs to upgrade its own parser anyway, but for documentation sakes we should keep the parser version in sync)

docs/user/user/metrics/rate.md
```diff
@@ -0,0 +1,129 @@
...
+* How often did an HTTP connection error?
+* How many documents used a given CSS Property?
+
+## Recorded errors
+
+* `invalid_value`: If either numerator or denominator is incremented by a negative value.
+
+## Reference
+
+* [Rust API docs](../../../docs/glean/private/counter/struct.RustMetric.html)
```
badboy, line 129:
```suggestion
* [Rust API docs](../../../docs/glean/private/counter/struct.RateMetric.html)
```

docs/dev/core/new-metric-type/rust.md
```diff
@@ -0,0 +1,182 @@
...
+
+```rust
+mod counter;
+...
+pub use counter::Counter;
+```
+
+Then add the trait in e.g. `counter.rs`:
+
+```rust
```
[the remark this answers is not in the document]
chutten: Good point. I'll `...` it. I was copying the approach in the other docs (see e.g. [Kotlin](https://mozilla.github.io/glean/dev/core/new-metric-type/kotlin.html)) which has some large code samples in it.

docs/dev/core/new-metric-type/rust.md
```diff
@@ -0,0 +1,182 @@
...
+you must always take the metric lock and Glean in the same order.
+
+## Tests
+
+Add at least a "smoke test" (a simple confirmation of the API's behaviour)
+to the RLB implementation.
+
+## Documentation
+
+Don't forget to document the new Metric Type in the Book.
```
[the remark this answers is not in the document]
chutten: Nah, this is supposed to be about adding the Rust API tab to the Book. I'll make it specific.

docs/user/user/metrics/rate.md
```diff
@@ -0,0 +1,129 @@
...
+You can think of it like a fraction, with a numerator and a denominator.
+
+All rates start without a value.
+A rate with a numerator of 0 is valid and will be sent to ensure we capture the
+"no errors happened" or "no use counted" cases.
+
+> **IMPORTANT:** When using a rate metric, it is important to let the Glean metric do the counting.
+  Using your own variable for counting and setting the metric yourself could be problematic:
+  ping scheduling will make it difficult to ensure the metric is at the correct value at the correct time.
+  Instead, count to the numerator and denominator as you go.
```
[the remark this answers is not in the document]
chutten: This is copypasta from Counter. The concern remains that, unless it's in a custom ping, you don't know when to flush your temporary partial counts to ensure they end up in pings.

I think it's better to leave this explanation at the top and optimize perf for cases that need it, later. Make it look like it's a hard rule, but supply exceptions under the table.

glean-core/src/metrics/mod.rs
chutten: If the numerator is 0, I think you mean. And yeah, we talked about it, but I thought we'd settled on "keep it simple for now, and optimize it later as needed". If we get 2093 use counters with 0 for their numerators and it's blowing up our user disk/user bandwidth/moz bandwidth/moz disk budget, then we can adjust then.

chutten (commented):
Alrighty, I've got some changes to add.

glean-core/src/metrics/mod.rs
badboy: No, I meant the denominator.
If that's never increased we end up with a rate such as 17/0. Is that useful? (I would need to re-read the proposal, but we discussed it there)

chutten: Then the next ping has 0/17. We need 17/0 from the first ping to aggregate them together.

badboy: Oh.

docs/user/user/metrics/rate.md
```diff
@@ -0,0 +1,129 @@
...
+    description: >
+      Total number of http connections made.
+    ...
+
+  http_connection_error:
+    type: rate
+    description: >
+      How many HTTP connections error out out of the total connections made.
+    denominator_metric: network.http_connections
+    ...
```
mdboom, line 51: *Maybe* we add a second numerator just to drive home the point about sharing the denominator?

docs/user/user/metrics/rate.md
```diff
@@ -0,0 +1,129 @@
...
+
+```rust
+use glean_metrics;
+
+if connection_had_error {
+    network::http_connection_error.add_to_numerator(1);
+}
+
+// network::http_connection_error has no `add_to_denominator` method.
+network::http_connections.add(1);
```
mdboom, line 84: The proposal says the API is the same whether the denominator is internal or external.  Did that change?

docs/user/user/metrics/rate.md
```diff
@@ -0,0 +1,129 @@
...
+* Numerator and Denominator saturate at the largest value that can be represented as a 32-bit signed integer (`2147483647`).
+
+## Examples
+
+* How often did an HTTP connection error?
+* How many documents used a given CSS Property?
+
+## Recorded errors
+
+* `invalid_value`: If either numerator or denominator is incremented by a negative value.
```
mdboom, line 125: Should we also do `invalid_overflow` if it saturates?

docs/user/user/metrics/rate.md
chutten: Yes. To Jan-Erik and I it seemed likely (upon review) that use of `add_to_denominator` with a shared denominator would result in misuse. `network::http_connection_error.add_to_denominator` and `network::http_connection_success.add_to_denominator` do not immediately highlight that they're talking about the same denominator. But if we get rid of the API, it forces the user to use `network::http_connections.add` which should be harder to misuse.

docs/user/user/metrics/rate.md
chutten: Counter doesn't, and I've been patterning off of it.

docs/user/user/metrics/rate.md
mdboom: Yeah, it's definitely more transparent as to what's happening.  And I can't really think of the magical use case where you want to be sort of "polymorphic" over internal and external denominators.  So seems fine -- mostly just me not being up to speed since the proposal.  (I would say updating the proposal to reflect this is a non-goal -- the docs in this PR become the canonical description of what *is* -- but that's a grey area).

docs/user/user/metrics/rate.md
mdboom: Fair enough.  In this case we can detect the error from the data -- since we saturate, rather than actually integer overflow, we can detect these cases.  So not urgent, but we could file a bug to do that if we ever saw a high frequency of max u32 values in real data.

.dictionary
```diff
@@ -88,6 +89,7 @@ backend
 backoff
 backport
 barcode
+behaviour
```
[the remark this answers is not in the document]
badboy: Hah, did he try to sneak in the other English again?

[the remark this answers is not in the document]
mdboom: It's definitely a nit, though.  The different spellings are mutually comprehensible.

I still have mixed feelings about removing this other Canadianism: https://github.com/mozilla/glean_parser/commit/72a10a6efd853ecc117af08992250f56e89c3a1b

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v35.0.0...main)
 
+* General
+  * Updated `glean_parser` version to 2.5.0
+* RLB
+  * Added `rate` metric type ([#1516](https://github.com/mozilla/glean/pull/1516)).
```
badboy, line 8: Will need rebasing now.

CHANGELOG.md
```diff
@@ -12,7 +17,7 @@
   * **Breaking change:** `LEGACY_TAG_PINGS` is removed from `GleanDebugActivity` ([#1510](https://github.com/mozilla/glean/pull/1510))
 * RLB
   * **Breaking change:** `Configuration.data_path` is now a `std::path::PathBuf`([#1493](https://github.com/mozilla/glean/pull/1493)). 
-  
+
```
badboy, line 20: (I think I now also fixed that)

badboy (approved):
r+ from my side, will require a rebase now.
@mdboom's point above about a second example up for your digression.

docs/user/user/metrics/rate.md
chutten: I've thrown a comment in the doc anyway so that context lives on. It's a low-energy-enough ask. But I'm a firm believer that these API docs (and the developer comments, and the code) are the living product of the proposal process and should be taken as canonical once they land.

(( I noodled an explanatory note in the docs to editorialize on how it could be misused, but I couldn't make it clear so I left it out for brevity. The API of add_to_denominator doesn't even exist for external-denominator rates, so it felt weird talking about how it could be misused. ))

PR #1521, opened 2021-02-25
bug 1694999 - Update payload doc to include "rate" metric

chutten:
Also update the payload docs' links to not be broken.


mdboom:
Hmm...  I wonder why the linkchecker CI didn't catch this.  Possibly broken by the big doc reorg...

badboy:
> Hmm... I wonder why the linkchecker CI didn't catch this. Possibly broken by the big doc reorg...

I think it actually only starts from the user book, meaning it traces links back to the dev book, but not going from the dev book, so stuff that's not linked to from the user book is never checked.
I'll update the config to take care of that.

badboy (approved):

PR #1529, opened 2021-03-01
Bug 1694715: Add testing against schema for the rate metric type

mdboom:
We should block this on https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/661, since this pulls in the new schema from there.

This just adds rate metric type data to the existing RLB schema validation unit test to make sure the output matches the schema.

No `CHANGELOG` entry since this is purely testing...


glean-core/rlb/tests/schema.rs
```diff
@@ -88,8 +89,62 @@ fn validate_against_schema() {
...
+                ..Default::default()
+            },
+            CommonMetricData {
+                name: "num2".into(),
+                category: "test".into(),
+                send_in_pings: vec![PING_NAME.into()],
+                ..Default::default()
+            },
+        ],
+    );
```
mdboom, line 141: @chutten pointed out in Matrix that maybe we should document how to make these in the developer docs.  First, to confirm that this is correct, though...

glean-core/rlb/tests/schema.rs
```diff
@@ -88,8 +89,62 @@ fn validate_against_schema() {
...
     const PING_NAME: &str = "test-ping";
+
+    // Test each of the metric types, just for basic smoke testing against the
+    // schema
+
+    // TODO: 1695762 Test all of the metric types against the schema from Rust
+
+    let rate_metric: RateMetric = RateMetric::new(CommonMetricData {
+        name: "rate".into(),
+        category: "test".into(),
```
chutten, line 101: Can we put the `category` first so I can read it in the same order ( `test.rate` ) that it appears in the payload?

glean.1.schema.json
```diff
@@ -276,7 +300,22 @@
         "labeled_rate": {
```
chutten, line 300: A labeled what now?

chutten (commented):
There seem to be a lot of changes not related to the `rate` type in this local schema copy. Where are they from?

glean-core/rlb/tests/schema.rs
mdboom: Sure thing.

glean.1.schema.json
mdboom: Yeah, the schema has placeholders for a bunch of things that were in the original design doc from 2018.  In hindsight maybe not the greatest.  Maybe we file a separate bug to remove the cruft?

mdboom:
> There seem to be a lot of changes not related to the `rate` type in this local schema copy. Where are they from?

The vendored copy of the schema hasn't been updated in a while.  Something to add to the developer docs...

glean-core/rlb/tests/schema.rs
chutten: This looks correct to me.

chutten (approved):

PR #1534, opened 2021-03-03
Add wheels for arm64 macos

glandium:
This also refactors wheels creation to allow cross-compilation in a smoother manner.

### DIFF
--- a/glean-core/python/setup.py
+++ b/glean-core/python/setup.py
@@ -34,18 +34,6 @@ import wheel.bdist_wheel
 sys.dont_write_bytecode = True
 
 
-platform = sys.platform
-
-if os.environ.get("GLEAN_PYTHON_MINGW_I686_BUILD"):
-    mingw_arch = "i686"
-elif os.environ.get("GLEAN_PYTHON_MINGW_X86_64_BUILD"):
-    mingw_arch = "x86_64"
-else:
-    mingw_arch = None
-
-if mingw_arch is not None:
-    platform = "windows"
-
 if sys.version_info < (3, 6):
     print("glean requires at least Python 3.6", file=sys.stderr)
     sys.exit(1)
@@ -81,24 +69,6 @@ setup_requirements = ["cffi>=1.13.0"]
 # The environment variable `GLEAN_BUILD_VARIANT` can be set to `debug` or `release`
 buildvariant = os.environ.get("GLEAN_BUILD_VARIANT", "debug")
 
-if mingw_arch == "i686":
-    shared_object_build_dir = SRC_ROOT / "target" / "i686-pc-windows-gnu"
-elif mingw_arch == "x86_64":
-    shared_object_build_dir = SRC_ROOT / "target" / "x86_64-pc-windows-gnu"
-else:
-    shared_object_build_dir = SRC_ROOT / "target"
-
-
-if platform == "darwin":
-    shared_object = "libglean_ffi.dylib"
-elif platform.startswith("win"):
-    # `platform` can be both "windows" (if running within MinGW) or "win32"
-    # if running in a standard Python environment. Account for both.
-    shared_object = "glean_ffi.dll"
-else:
-    # Anything else must be an ELF platform - Linux, *BSD, Solaris/illumos
-    shared_object = "libglean_ffi.so"
-
 
 class BinaryDistribution(Distribution):
     def is_pure(self):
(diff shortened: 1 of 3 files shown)


glandium:
The failure on Android tests don't seem related.

badboy:
> The failure on Android tests don't seem related.

Yeah, ignore those. Will need to look into them separately.

glandium:
I just realized there was temporary garbage in the history. Removed them.

badboy (approved):
I like it! The part for arm64 macos looks good.
Moving around where and how we build the wheel looks fine as well.
As long as from-source-and-build-it continues to work, we can land this.

Defering to @mdboom for a final look, because he's the Python expert here.

glean-core/python/setup.py
```diff
@@ -116,20 +92,24 @@ def has_ext_modules(self):
...
+            elif cpu == "x86_64":
+                plat_name = "win_amd64"
             else:
                 raise ValueError("Unsupported Windows platform")
         else:
             # Keep local wheel build on BSD/etc. working
-            return super().get_tag()
+            _, __, plat_name = super().get_tag()
+
+        return ("cp36", "abi3", plat_name)
```
mdboom, line 112: Have you confirmed this works with Windows wheels?  The pattern matching by pip is slightly different there, hence the weird hard coding of all of the supported options that was here before.  (Related, let's update the comment above this function if this new approach *does* work).

mdboom (commented):
Thanks for this.  Generally a nice cleanup of the `setup.py`, especially the standard way to cross compile anything.

Before merging, we should confirm that:
1) The new Windows wheels produced by this work with the different pattern matching values
2) Source installation through a non-wheel-provided platform (e.g. OpenBSD) still works.

glean-core/python/setup.py
glandium: It worked for me locally, but it also obviously works more generally, since Windows python tests on CI use the wheels.

glandium: Note the comment above get_tag is still relevant as-is.

mdboom: I just reminded myself of the details of the issue (which we should probably put in a comment so it isn't lost again).

Windows Python prior to 3.7 does not have a stable ABI, so therefore doesn't support the `abi3` tag.  Therefore, it won't "match" that wheel and it falls back to installing from source.  You can test this locally by putting the wheels in a local directory and forcing `pip` to look there instead of PyPI with the `-f directory` option.  It won't work with the wheel named as this PR does it.

Of course, we don't actually use the Python3 ABI since we use CFFI, so the best way to make a wheel that will work on Windows Python 3.6 and up is to mark it with `py3-none` rather than `cp36-abi3`.

glandium: I'm not sure this is the full story. Because while installing a fresh Windows Python 3.6 and running `pip install glean_sdk-35.0.0-cp36-abi3-win_amd64.whl` doesn't work, pip tells you it's version 18.1 and that you should upgrade to 21.0.1, and after doing that, it works. Trying a few versions, it started working with pip 20.0.1.

mdboom: Installing directly from a file path always works.  What doesn't is the pattern matching engine. So you need to do:

```
pip install -f directory_containing_wheels glean_sdk
```

to match what would happen in normal situations when installing from pypi.

This doesn't work for me on Python 3.6, even when upgrading pip.

glandium: > Installing directly from a file path always works

That's not my experience:
```
# ./AppData/Local/Programs/Python/Python36/python -m pip install glean_sdk-35.0.0-cp36-abi3-win_amd64.whl
glean_sdk-35.0.0-cp36-abi3-win_amd64.whl is not a supported wheel on this platform.
You are using pip version 18.1, however version 21.0.1 is available.
You should consider upgrading via the 'python -m pip install --upgrade pip' command.
```

> This doesn't work for me on Python 3.6, even when upgrading pip.

That doesn't look to be true (after upgrading to pip 21.0.1):
```
# ./AppData/Local/Programs/Python/Python36/python -m pip install -f . glean_sdk
Looking in links: .
Processing c:\users\glandium\glean_sdk-35.0.0-cp36-abi3-win_amd64.whl
(...)
```

mdboom: Thanks for that.  I think I wasn't upgrading `pip` correctly.

So then I think the only thing to confirm is that we aren't held back to a particular version of `pip` by how things are vendored in m-c.  Assuming that checks out, :+1: from me.

glandium: Thank you for making me look. Things are grim, unfortunately, as the common way to install and upgrade moz-phab on Windows uses pip 19.3 :(

mdboom (approved):

glandium:
Any idea yet when the next release is going to be?

badboy:
> Any idea yet when the next release is going to be?

This week. I'll probably get one done today or tomorrow.

badboy:
Upload failed. Tracking here: https://bugzilla.mozilla.org/show_bug.cgi?id=1698860

We can manually upload that wheel when we figure out what platform tag we need.

PR #1537, opened 2021-03-04
[doc only] [Bug 1696188] Document adding new metric type to data platform

mdboom:
So, I'm still not 100% sure on *where* this belongs, but I thought in the meantime we could review the content itself.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 212 utf-8
+personal_ws-1.1 en 213 utf-8
 AAR
 AARs
 ABI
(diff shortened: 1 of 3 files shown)


acmiyaguchi (approved):
Looks good to me, in terms of the surfaces that need to be covered before the metric is ingested into the schema repos.

mdboom:
Thanks, @acmiyaguchi.  Would appreciate @badboy's review as to whether this belongs here or ... somewhere else...?  I think I'm less concerned about confusing Glean SDK with "Glean platform" as this is internal developer docs, but still...

chutten (commented):
Content looks good (though I thought the SDK docs practised Semantic Linefeeds). I think here's just fine to have it for at least now, if not forever.

PR #1538, opened 2021-03-06
Bug 1679837: Set internal_metrics::os_version for specific os

ChinYing-Li:
At the first glance, two existing crates could be used: `sys_info` and `os_info`.
`os_info` returns the distribution version instead of the kernel version for Linux systems, so we are only left with `sys_info`.

Following the [proposal](https://docs.google.com/document/d/1VD1z8njHu6A_vvojyIQhKcUmvhlQHVjVAlug2jitqbM/), `get_os_version()` is added to `glean_core::system.rs`, however I doubt this is the right place, since `internal_metrics::os_version` is initialized in `rlb`. 
Given the fact that we have to manipulate the `String` returned by `sys_info` in case of Linux, I failed to come up with a way to create something like `pub const OS_VERSION: String`.

Any suggestion is appreciated!

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -54,6 +54,7 @@
   * Send the `baseline` ping with reason `dirty_startup`, if needed, at startup.
   * Expose all required types directly ([#1452](https://github.com/mozilla/glean/pull/1452)).
     * Rust consumers will not need to depend on `glean-core` anymore.
+  * Set `internal_metrics::os_version` for MacOS, Windows and Linux ([#1538](https://github.com/mozilla/glean/pull/1538))
 * Android
   * BUGFIX: Don't crash the ping uploader when throttled due to reading too large wait time values ([#1454](https://github.com/mozilla/glean/pull/1454)).
   * Use the client activity API ([#1455](https://github.com/mozilla/glean/pull/1455)).
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -434,7 +434,7 @@ fn initialize_core_metrics(
     if let Some(app_channel) = channel {
         core_metrics::internal_metrics::app_channel.set_sync(glean, app_channel);
     }
-    core_metrics::internal_metrics::os_version.set_sync(glean, "unknown".to_string());
+    core_metrics::internal_metrics::os_version.set_sync(glean, system::get_os_version());
     core_metrics::internal_metrics::architecture.set_sync(glean, system::ARCH.to_string());
     core_metrics::internal_metrics::device_manufacturer.set_sync(glean, "unknown".to_string());
     core_metrics::internal_metrics::device_model.set_sync(glean, "unknown".to_string());
--- a/glean-core/rlb/src/system.rs
+++ b/glean-core/rlb/src/system.rs
@@ -53,3 +53,37 @@ pub const ARCH: &str = "x86_64";
 )))]
 /// `target_arch` when building this crate: unknown!
 pub const ARCH: &str = "unknown";
+
+#[cfg(any(target_os = "macos", target_os = "windows"))]
+/// Returns Darwin kernel version for MacOS, or NT Kernel version for Windows
+pub fn get_os_version() -> String {
+    whatsys::kernel_version().unwrap_or_else(|| "unknown".to_owned())
+}
+
+#[cfg(not(any(target_os = "linux", target_os = "macos", target_os = "windows")))]
+/// Returns "unknown" for platforms other than Linux, MacOS or Windows
+pub fn get_os_version() -> String {
+    "unknown".to_owned()
+}
+
+#[cfg(target_os = "linux")]
+/// Returns Linux kernel version, in the format of <Major>.<Minor> e.g. 5.8
+pub fn get_os_version() -> String {
+    parse_linux_os_string(whatsys::kernel_version().unwrap_or_else(|| "unknown".to_owned()))
+}
+
+#[cfg(target_os = "linux")]
+fn parse_linux_os_string(os_str: String) -> String {
+    os_str.split('.').take(2).collect::<Vec<&str>>().join(".")
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn parse_fixed_linux_os_string() {
+    let alpine_os_string = "4.12.0-rc6-g48ec1f0-dirty".to_owned();
+    assert_eq!(parse_linux_os_string(alpine_os_string), "4.12");
+    let centos_os_string = "3.10.0-514.16.1.el7.x86_64".to_owned();
+    assert_eq!(parse_linux_os_string(centos_os_string), "3.10");
+    let ubuntu_os_string = "5.8.0-44-generic".to_owned();
+    assert_eq!(parse_linux_os_string(ubuntu_os_string), "5.8");
+}
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -268,6 +268,7 @@ dependencies = [
  "thiserror",
  "time",
  "uuid",
+ "whatsys",
 ]
 
 [[package]]
@@ -893,6 +894,17 @@ name = "wasi"
 version = "0.10.0+wasi-snapshot-preview1"
 source = "registry+https://github.com/rust-lang/crates.io-index"
 checksum = "1a143597ca7c7793eff794def352d41792a93c481eb1042423ff7ff72ba2c31f"
+
+[[package]]
+name = "whatsys"
+version = "0.1.2"
+source = "registry+https://github.com/rust-lang/crates.io-index"
+checksum = "c24fff5aa1e0973964ba23a995e8b10fa2cdeae507e0cbbbd36f8403242a765d"
+dependencies = [
+ "cc",
+ "cfg-if 1.0.0",
+ "libc",
+]
 
 [[package]]
 name = "winapi"
--- a/glean-core/rlb/Cargo.toml
+++ b/glean-core/rlb/Cargo.toml
@@ -35,6 +35,7 @@ serde = { version = "1.0.104", features = ["derive"] }
 uuid = { version = "0.8.1", features = ["v4"] }
 chrono = { version = "0.4.10", features = ["serde"] }
 time = "0.1.40"
+whatsys = "0.1.2"
 
 [dev-dependencies]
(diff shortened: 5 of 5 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -434,10 +434,12 @@ fn initialize_core_metrics(
     if let Some(app_channel) = channel {
         core_metrics::internal_metrics::app_channel.set_sync(glean, app_channel);
     }
-    core_metrics::internal_metrics::os_version.set_sync(glean, "unknown".to_string());
+    core_metrics::internal_metrics::os_version
+        .set_sync(glean, glean_core::system::get_os_version());
     core_metrics::internal_metrics::architecture.set_sync(glean, system::ARCH.to_string());
     core_metrics::internal_metrics::device_manufacturer.set_sync(glean, "unknown".to_string());
     core_metrics::internal_metrics::device_model.set_sync(glean, "unknown".to_string());
+    core_metrics::internal_metrics::device_model.set_sync(glean, "unknown".to_string());
```
chutten, line 442: This seems unintentional.

chutten (commented):
I'd love to see a regression test against that linux-specific string parsing. Grab a bunch of kernel versions from the wild (and make up some) and ensure it generates expected information each time. (Maybe factor the string construction to its own cross-platform-available function for testing)

From a FOG perspective, I've looked at `sys-info` and it depends on only `libc 0.2` which mozilla-central already has in tree. `sys-info`'s under the MIT license which is compatible, but we'll need to eyeball the code and ensure it's well tested (see [the guide](https://firefox-source-docs.mozilla.org/build/buildsystem/rust.html?highlight=adding%20rust%20code#third-party-crate-dependencies)). At a quick glance, there's a lot of C code in there, and mozilla-central folks might wonder why Glean needs its own implementation of the `system-info` xpcom utility when it's being used in-tree.

Definitely some @badboy questions in here, but regardless of what the answers are, a good solid list of test strings and correct answers will be useful, so that'd be a good thing to look into while we wait.

badboy:
Just to give feedback: I saw it and will review it tomorrow in detail.

glean-core/src/system.rs
```diff
@@ -79,3 +81,31 @@ pub const OS: &str = "Solaris";
     target_os = "solaris",
 )))]
 pub const OS: &str = "unknown";
+
+#[cfg(any(target_os = "macos", target_os = "windows"))]
+/// Returns Darwin kernel version for MacOS, or NT Kernel version for Windows
+pub fn get_os_version() -> String {
+    sys_info::os_release().unwrap()
```
badboy, line 88: This could panic if it can't read the version. We should probably turn this into "unknown" in case of errors.

```suggestion
    sys_info::os_release().unwrap_or_else(|| "unknown".to_string())
```

glean-core/src/system.rs
```diff
@@ -79,3 +81,31 @@ pub const OS: &str = "Solaris";
...
+/// Returns "unknown" for platforms other than Linux, MacOS or Windows
+pub fn get_os_version() -> String {
+    "unknown".to_owned()
+}
+
+#[cfg(target_os = "linux")]
+/// Returns Linux kernel version, in the format of <Major>.<Minor> e.g. 5.8
+pub fn get_os_version() -> String {
+    sys_info::os_release()
+        .unwrap()
```
badboy, line 101: Same here, this could panic.

glean-core/src/system.rs
```diff
@@ -79,3 +81,31 @@ pub const OS: &str = "Solaris";
...
+        .split('.')
+        .take(2)
+        .fold("".to_owned(), |lhs, rhs| {
+            if lhs.is_empty() {
+                lhs.add(rhs)
+            } else {
+                lhs.add(".").add(rhs)
+            }
+        })
```
badboy, line 110: This is just `.join(".")`

glean-core/src/system.rs
```diff
@@ -79,3 +81,31 @@ pub const OS: &str = "Solaris";
     target_os = "solaris",
 )))]
 pub const OS: &str = "unknown";
+
+#[cfg(any(target_os = "macos", target_os = "windows"))]
+/// Returns Darwin kernel version for MacOS, or NT Kernel version for Windows
```
badboy, line 86: Hm, I guess this is really the only file where we have cfg before the comment.
I was about to suggest to swap it around, but let's not do it here now.

badboy (changes requested):
I still need to review sys_info before we can land this. However, the way we use it even if we need to swap it out this code here should be mostly fine already.

> Following the [proposal](https://docs.google.com/document/d/1VD1z8njHu6A_vvojyIQhKcUmvhlQHVjVAlug2jitqbM/), `get_os_version()` is added to `glean_core::system.rs`, however I doubt this is the right place, since `internal_metrics::os_version` is initialized in `rlb`.

Hm, I agree. We can move that into a `system` module in the RLB for now.

> Given the fact that we have to manipulate the `String` returned by `sys_info` in case of Linux, I failed to come up with a way to create something like `pub const OS_VERSION: String`.

That's fine. It's a dynamic value after all and not something we know statically at build time.

You should transform the manual tests you posted into a `#[test]` in the (then new) `system` module, so we always run them.
Easiest way would be to refactor the parsing into its own function, which you can then test against the fixed values you already used above.

badboy:
I gave sys-info a look. It does much more than we need, but that would be fine.
I _think_ I found a (mostly theoretical) issue in their macOS code for getting the kernel version, which makes me a bit uncomfortable using it as is (I send a PR upstream of course)

glean-core/src/lib.rs
```diff
@@ -39,7 +39,7 @@ mod internal_pings;
 pub mod metrics;
 pub mod ping;
 pub mod storage;
-mod system;
+pub mod system;
```
badboy, line 42: That's not needed anymore now, is it?

ChinYing-Li: You are right, I oversaw that!

glean-core/rlb/src/system.rs
```diff
@@ -53,3 +53,37 @@ pub const ARCH: &str = "x86_64";
...
+#[cfg(target_os = "linux")]
+fn parse_linux_os_string(os_str: String) -> String {
+    os_str.split('.').take(2).collect::<Vec<&str>>().join(".")
+}
+
+#[test]
+#[cfg(target_os = "linux")]
+fn parse_fixed_linux_os_string() {
+    let alpine_os_string = "4.12.0-rc6-g48ec1f0-dirty".to_owned();
+    assert_eq!(parse_linux_os_string(alpine_os_string), "4.12".to_owned());
```
badboy, line 84: Tip: in theory you should be able to compare a `&str` with `String` here. Might need to swap things around though, but it's also not super important for testing.

```suggestion
    assert_eq!("4.12", parse_linux_os_string(alpine_os_string));
```

glean-core/rlb/src/system.rs
```diff
@@ -53,3 +53,37 @@ pub const ARCH: &str = "x86_64";
...
+    parse_linux_os_string(sys_info::os_release().unwrap_or_else(|_| "unknown".to_owned()))
+}
+
+#[cfg(target_os = "linux")]
+fn parse_linux_os_string(os_str: String) -> String {
+    os_str.split('.').take(2).collect::<Vec<&str>>().join(".")
+}
+
+#[test]
+#[cfg(target_os = "linux")]
```
badboy, line 81: I was about to mention this :D

badboy (changes requested):
Hey, this looks good already. 
We're gonna replace sys-info with a smaller crate I already wrote, mostly because sys-info does much more and I'm not super confident in how it does stuff. It will be a simple swap though, I just still need to do a final test with that other crate.
I'll let you know when that's ready.

In the meantime can you add a changelog entry?
In the `RLB` something like "`os.version` now reports the underlying OS/Kernel version on Windows, macOS and Linux again."

ChinYing-Li:
> In the meantime can you add a changelog entry?
> In the `RLB` something like "`os.version` now reports the underlying OS/Kernel version on Windows, macOS and Linux again."

Just wondering about the criteria for a changelog entry since there's no `Contributing.md` or similar: does any change, including internal changes like this PR, need to be added to changelog?

badboy:
> Just wondering about the criteria for a changelog entry since there's no `Contributing.md` or similar: does any change, including internal changes like this PR, need to be added to changelog?

We have a [Contributing page here](https://mozilla.github.io/glean/dev/contributing.html). I should probably mirror that into a CONTRIBUTING.md to follow github patterns.
Still, that also doesn't mention the changelog.

We don't really have _strict_ rules about what to name there, but in general we should call out everything that might affect users. Reporting a more precise value for a built-in metric seems important enough to mention (as it affects the analysis that can be done)

CHANGELOG.md
```diff
@@ -54,6 +54,7 @@
   * Send the `baseline` ping with reason `dirty_startup`, if needed, at startup.
   * Expose all required types directly ([#1452](https://github.com/mozilla/glean/pull/1452)).
     * Rust consumers will not need to depend on `glean-core` anymore.
+  * Set internal_metrics::os_version for MacOS, Windows and Linux ([#1538](https://github.com/mozilla/glean/pull/1538))
```
badboy, line 57:
```suggestion
  * The os version is now correctly reported for macOS, Windows and Linux ([#1538](https://github.com/mozilla/glean/pull/1538)).
```

badboy (changes requested):
Nearly ready now!

And also we can switch over to the [`whatsys`](https://crates.io/crates/whatsys) crate now, which we control and only does the bare minimum.

glean-core/rlb/src/lib.rs
```diff
@@ -42,6 +42,7 @@ use once_cell::sync::OnceCell;
 use std::collections::HashMap;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::Mutex;
+pub use whatsys;
```
badboy, line 45: This shouldn't be necessary. As written this would re-export that crate, which we don't want to do.
Dependency crates are in the global namespace and thus already accessible without a `use` statement.

badboy (changes requested):

badboy (approved):

ChinYing-Li:
Ah @badboy I realized I haven't update the CHANGELOG entry as you suggested... Sorry about that.

badboy:
oh, :D didn't check that again. I think it's still fine as is.
Thanks for the work!

PR #1540, opened 2021-03-08
Bug 1695795: Remove unused metric types from schema

mdboom:


Dexterp37 (approved):
r+ since this matches exactly what's in the mozilla-pipeline-schemas repo.

PR #1542, opened 2021-03-09
Bug 1697100: Don't let environment affect subprocess module search path

mdboom:
@cpeterso: Can you confirm this addresses your issue by copying `_process_dispatcher.py` on this branch to `/Users/chris/Code/mozilla/firefox/obj-x86_64-apple-darwin20.3.0-clang-mozbuild/_virtualenvs/init_py3/lib/python3.8/site-packages/glean/`?

### DIFF
--- a/glean-core/python/glean/_process_dispatcher.py
+++ b/glean-core/python/glean/_process_dispatcher.py
@@ -120,8 +120,14 @@ class ProcessDispatcher:
                     Path(".coveragerc").absolute()
                 )
 
+            # Explicitly pass the contents of `sys.path` as `PYTHONPATH` to the
+            # subprocess so that there aren't any module search path
+            # differences.
+            python_path = ":".join(sys.path)[1:]
+
             p = subprocess.Popen(
-                [sys.executable, _process_dispatcher_helper.__file__, payload]
+                [sys.executable, _process_dispatcher_helper.__file__, payload],
+                env={"PYTHONPATH": python_path},
             )
 
             cls._last_process = p
--- a/glean-core/python/tests/test_dispatcher.py
+++ b/glean-core/python/tests/test_dispatcher.py
@@ -3,6 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
+from pathlib import Path
 import threading
 import sys
 import time
@@ -12,9 +13,11 @@ import pytest
 
 
 from glean._dispatcher import Dispatcher
+from glean._process_dispatcher import ProcessDispatcher
 from glean import Glean
 from glean import metrics
 from glean.metrics import Lifetime
+from glean.glean import _rmtree
 
 
 def test_launch_correctly_adds_tasks_to_queue_if_queue_tasks_is_true():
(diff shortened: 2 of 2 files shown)


badboy (approved):
Awaiting confirmation that it solves the issue.

cpeterso:
> @cpeterso: Can you confirm this addresses your issue by copying `_process_dispatcher.py` on this branch to `/Users/chris/Code/mozilla/firefox/obj-x86_64-apple-darwin20.3.0-clang-mozbuild/_virtualenvs/init_py3/lib/python3.8/site-packages/glean/`?

I copied your new https://github.com/mozilla/glean/blob/95fb6783762608800d70c2e071b665a8c094e8a6/glean-core/python/glean/_process_dispatcher.py to my objdir `obj-x86_64-apple-darwin20.3.0-clang-mozbuild/_virtualenvs/init_py3/lib/python3.8/site-packages/glean/_process_dispatcher.py`, but I still hit the same Python error:

```
firefox % mach mozregression
...
 0:00.78 INFO: Running mozilla-central build for 2020-03-09
  File "/Users/chris/Code/mozilla/firefox/obj-x86_64-apple-darwin20.3.0-clang-mozbuild/_virtualenvs/init_py3/lib/python3.8/site-packages/glean/_subprocess/_process_dispatcher_helper.py", line 36, in <module>
    success = func(*args)
TypeError: _process() missing 1 required positional argument: 'throttle_backoff_time_s'
```

cpeterso:
Looks like I just needed to re-run `mach bootstrap`. So you probably close this issue as WONTFIX. Sorry for the wild goose chase!

I don't have a `PYTHONPATH` environment variable set, but it looks like mach is picking up glean.py from my mach bootstrap directory here: `$HOME/.mozbuild/_virtualenvs/mach/lib/python3.8/site-packages/glean/glean.py`. That copy of glean.py still refers to `throttle_backoff_time_s`.

I reran `mach bootstrap` and then `mach mozregression` worked without hitting the `throttle_backoff_time_s` error.

cpeterso:
It would be helpful if mach had a clobber status so it could warn users when they need to re-run `mach bootstrap`.

badboy:
> Looks like I just needed to re-run `mach bootstrap`. So you probably close this issue as WONTFIX. Sorry for the wild goose chase!
> 
> I don't have a `PYTHONPATH` environment variable set, but it looks like mach is picking up glean.py from my mach bootstrap directory here: `$HOME/.mozbuild/_virtualenvs/mach/lib/python3.8/site-packages/glean/glean.py`. That copy of glean.py still refers to `throttle_backoff_time_s`.
> 
> I reran `mach bootstrap` and then `mach mozregression` worked without hitting the `throttle_backoff_time_s` error.

Hmpf, sounds like bad interaction between the 2 python programs.



> It would be helpful if mach had a clobber status so it could warn users when they need to re-run `mach bootstrap`.

IMO worth reporting on bugzilla. Would also play well with `ac_add_options --enable-bootstrap` then I guess.
Though I'm not sure it would have helped here because probably no one would have flagged this as clobber-worthy.

mdboom:
Thanks @cpeterso.  It feels like it should at least theoretically be possible to ensure that the code in the subprocess is the same as in the parent process somehow (even if this shot-in-the-dark didn't work).  Then we wouldn't have to worry about clobbering issues etc.  So, I'll close this PR as WONTFIX, but might keep the parent bug open for a while to investigate further.

kvark:
I was having a similar issue on NixOS, and grabbing the patch appears to fix it.
@mdboom can we land this please?

mdboom (commented):
I think this is still fine.  We had originally chosen to pass on it because we didn't full understand the root cause (and we still don't).  But it's definitely helping the minority that falls into this -- we should keep an eye out on any regressions when this lands in m-c, though.l

PR #1544, opened 2021-03-10
Use full path to Swift.Error to avoid namespace clashes

badboy:
See https://github.com/mozilla/application-services/pull/3913 for when
it went wrong.


travis79 (approved):
Wow, that's it?  🎉

PR #1545, opened 2021-03-11
Fix link to dev book

badboy:
Somehow we broke that link again, probably when we split the book or moved things around


brizental (approved):
Argh, sorry. I broke this link when I re-wrote this page's content.

PR #1546, opened 2021-03-11
Bug 1695167 - Use a single implementation of a monotonic clock across all bindings

badboy:


travis79 (approved):
Woosh!  I especially like the EventPingTest!

Dexterp37 (approved):
Instant r+ because of zeitstempel.

glean-core/python/glean/metrics/event.py
```diff
@@ -125,7 +124,7 @@ def record(self, extra: Optional[Dict[int, str]] = None) -> None:
         if self._disabled:
             return
 
-        timestamp = _util.time_ms()
+        timestamp = _ffi.lib.glean_get_timestamp_ms()
```
mdboom, line 127: Maybe we should also remove `_util.py::time_ms` since this is the only use of it in the code base?

PR #1547, opened 2021-03-11
Fix the misleading comment about label validation

Dexterp37:
I wish I found about this earlier :\

### DIFF
--- a/glean-core/src/metrics/labeled.rs
+++ b/glean-core/src/metrics/labeled.rs
@@ -11,18 +11,17 @@ const MAX_LABELS: usize = 16;
 const OTHER_LABEL: &str = "__other__";
 const MAX_LABEL_LENGTH: usize = 61;
 
-/// Checks whether the given value matches the label regex.
+/// Checks whether the given label is sane.
 ///
-/// This regex is used for matching against labels and should allow for dots,
-/// underscores, and/or hyphens. Labels are also limited to starting with either
-/// a letter or an underscore character.
+/// The check corresponds to the following regular expression:
 ///
-/// The exact regex (from the pipeline schema [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/glean/dot_separated_short_id.1.schema.json)) is:
+/// ```text
+/// ^[a-z_][a-z0-9_-]{0,29}(\\.[a-z_][a-z0-9_-]{0,29})*$
+/// ```
 ///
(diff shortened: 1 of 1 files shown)


glean-core/src/metrics/labeled.rs
```diff
@@ -11,18 +11,11 @@ const MAX_LABELS: usize = 16;
 const OTHER_LABEL: &str = "__other__";
 const MAX_LABEL_LENGTH: usize = 61;
 
-/// Checks whether the given value matches the label regex.
+/// Checks whether the given label is sane.
```
badboy, line 14: Maaaybe this function also shouldn't be called `matches_label_regex` anymore, but more like `is_valid_label` or something like that?

glean-core/src/metrics/labeled.rs
```diff
@@ -11,18 +11,11 @@ const MAX_LABELS: usize = 16;
...
-/// Checks whether the given value matches the label regex.
+/// Checks whether the given label is sane. 
 ///
-/// This regex is used for matching against labels and should allow for dots,
-/// underscores, and/or hyphens. Labels are also limited to starting with either
-/// a letter or an underscore character.
-///
-/// The exact regex (from the pipeline schema [here](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/master/templates/include/glean/dot_separated_short_id.1.schema.json)) is:
-///
-///    "^[a-z_][a-z0-9_-]{0,29}(\\.[a-z_][a-z0-9_-]{0,29})*$"
```
badboy, line 22: AHA! See, that used a `*`!

glean-core/src/metrics/labeled.rs
```diff
@@ -11,18 +11,11 @@ const MAX_LABELS: usize = 16;
...
-///
-/// The regex crate isn't used here because it adds to the binary size, and the
-/// Glean SDK doesn't use regular expressions anywhere else.
+/// While this could be implemented as a regex, the regex crate isn't
```
badboy, line 16: Would it make sense to still call out the old regex?
Then again it's not used in the pipeline either, so our examples below are probably more useful than the regex anyway.

badboy (approved):
Just minor comments.

glean-core/src/metrics/labeled.rs
badboy:
```suggestion
/// Checks whether the given label is sane.
```

PR #1550, opened 2021-03-15
[ci full] Bump macOS SDK to 10.12 (Sierra)

badboy:
We now use `clock_gettime_nsec_np`, which was only introduced in 10.12
(which was released in 2016).
This effectively drops support for 10.11 for Android builds.


Dexterp37 (approved):
r+ with Ci happy!

PR #1551, opened 2021-03-15
Bug 1675580: Use rkv SafeMode in Python

mdboom:
Posting this early to show the (straightforward) `setup.py` changes required to use SafeMode.

Unfortunately, safe mode isn't safe for multiple processes writing to the database at the same time.

In the general case, we can't guard against this short of some sort of lockfile (and all of the challenges that come with that). The use case of a command line tool that is run in parallel as part of a build is probably real enough that we should care someday, but is probably only a hazard waiting to be tripped on right now.

Beyond that, Glean require changes even for normal single-process usage, since the ping uploader runs in a separate process *and* sets network error metrics in the database.  This probably needs to be refactored so network error metrics are reported back to the main process through some sort of IPC, which is tricky because the ping uploader process is started and then not waited on, so it doesn't block the main process.

### DIFF
--- a/glean-core/python/glean/net/ping_upload_worker.py
+++ b/glean-core/python/glean/net/ping_upload_worker.py
@@ -111,7 +111,11 @@ def _process(data_dir: Path, application_id: str, configuration) -> bool:
         cfg = _ffi.make_config(
             data_dir,
             application_id,
-            True,
+            # Set upload enabled to False. The subprocess should not record
+            # telemetry. In the special `glean_initialize_for_subprocess` mode,
+            # this does not have any side effects like resetting the client_id
+            # or sending a deletion-request ping.
+            False,
             configuration.max_events,
         )
         if _ffi.lib.glean_initialize_for_subprocess(cfg) == 0:
--- a/glean-core/python/setup.py
+++ b/glean-core/python/setup.py
@@ -179,14 +179,23 @@ class build(_build):
         if target == "i686-pc-windows-gnu":
             env["RUSTFLAGS"] = env.get("RUSTFLAGS", "") + " -C panic=abort"
 
-        command = ["cargo", "build", "--package", "glean-ffi", "--target", target]
+        command = [
+            "cargo",
+            "build",
+            "--package",
+            "glean-ffi",
+            "--target",
+            target,
+            "--features",
+            "rkv-safe-mode",
+        ]
         if buildvariant != "debug":
             command.append(f"--{buildvariant}")
 
         if "-darwin" in target:
             env["MACOSX_DEPLOYMENT_TARGET"] = macos_compat(target)
 
-        subprocess.run(command, cwd=SRC_ROOT, env=env)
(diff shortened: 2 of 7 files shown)


glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -111,7 +111,7 @@ def _process(data_dir: Path, application_id: str, configuration) -> bool:
         cfg = _ffi.make_config(
             data_dir,
             application_id,
-            True,
+            False,
```
badboy, line 114: Can we document why and what we're passing here?
Should call out that we're disabling upload because the process should not write data and that this does _not_ reset the client id

glean-core/python/setup.py
```diff
@@ -177,14 +177,23 @@ def run(self):
...
+            "rkv-safe-mode",
+        ]
         if buildvariant != "debug":
             command.append(f"--{buildvariant}")
 
         if "-darwin" in target:
             env["MACOSX_DEPLOYMENT_TARGET"] = macos_compat(target)
 
-        subprocess.run(command, cwd=SRC_ROOT, env=env)
+        subprocess.check_call(command, cwd=SRC_ROOT / "glean-core", env=env)
```
badboy, line 196: Why do we change the working directory now?

badboy (changes requested):
Minimal nits, but I need to give this a fresh look tomorrow again.

Should we file a followup to refactor glean-core to avoid opening the database when called in a subprocess? In my tests that worked.

glean-core/python/setup.py
mdboom: I got this -->

```
error: --features is not allowed in the root of a virtual workspace
note: while this was previously accepted, it didn't actually do anything
help: change the current directory to the package directory, or use the --manifest-path flag to the path of the package
```

badboy: Oh right, but then we should switch directly to the `glean-core/ffi` directory. Not even sure why it works in `glean-core`. 
Weird interaction of cargo to not allow features in the workspace root, but then somehow in a subdirectory that's not even the final one.

... which now brings me to another point: `rkv-safe-mode` is a feature of `glean-core`, not of `glean-ffi`. We need to re-expose that on `glean-ffi` to actually enable it on the build, I think.
While we're at it `glean` (RLB) should also get it, because right now we need to explicitely reference `glean-core` in FOG.

I can do that tomorrow, before we than update this PR.

mdboom:
> Minimal nits, but I need to give this a fresh look tomorrow again.
> 
> Should we file a followup to refactor glean-core to avoid opening the database when called in a subprocess? In my tests that worked.

Ah, I didn't realize how easy that was to do, so might as well do it.

Part of me wants to change the `glean_initialize_for_subprocess` to `glean_initialize_for_python_subprocess` since it's deeply geared to the specifics of how the Python bindings work at this point.

glean-core/python/setup.py
mdboom: Updated this here (and rebased to get the feature delegation changes).

badboy (approved):

PR #1557, opened 2021-03-18
Bug 1697469 - Rename internal function to express what it does: validate a dynamic label

badboy:
Additionally this hides previously publicly exported functions.
Technically this would be a breaking change, but glean-core is not
considered a stable public API, so this can be handled as a patch change
only.


Dexterp37 (approved):
Thanks!

PR #1560, opened 2021-03-19
Bug 1699505 - Kotlin: Don't inline the closure when measuring

badboy:
This can change control flow, as the code is essentially copied 1:1 in
place, which means `return` inside will be an early return before our
measure wrapper gets to stop the timer.

Thanks to @mcomella for the detailed explanation in the bug report.
See: https://blog.mindorks.com/understanding-inline-noinline-and-crossinline-in-kotlin

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v36.0.0...main)
 
+* Android
+  * BUGFIX: `TimespanMetricType.measure` and `TimingDistributionMetricType.measure` won't get inlined anymore ([#1560](https://github.com/mozilla/glean/pull/1560)).
+    This avoids a potential bug where a `return` used inside the closure would end up not measuring the time.
+    Use `return@measure <val>` for early returns.
+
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v36.0.0...main)
 
+* Android
+  * BUGFIX: `TimespanMetricType.measure` and `TimingDistributionMetricType.measure` won't get inlined anymore ([#1560](https://github.com/mozilla/glean/pull/1560)).
+    This avoids a potential bug where a `return` used inside the closure would end up not measuring the time.
+    Use `return@measure <val>` for early returns.
```
Dexterp37, line 8: Should this make it to the docs?

badboy: It probably should, but right now we're not even calling out `measure` at all^^

Dexterp37: Ok, would you kindly file a follow up for this?

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1699653

PR #1561, opened 2021-03-19
Bug 1695228 - RLB: API to pass in raw samples for timing distributions

badboy:
Unfortunately this needs a new API instead of the already existing `accumulate_samples_signed`.
We want the RLB consumer (in this case: FOG) to not care about the unit,
so we expect nanosecond samples.
Internally we then do the right thing.

And because this is pure Rust we can rely on `u64` and just never need
to filter out negative samples.

### DIFF
--- a/glean-core/src/metrics/timing_distribution.rs
+++ b/glean-core/src/metrics/timing_distribution.rs
@@ -314,6 +314,63 @@ impl TimingDistributionMetric {
         }
     }
 
+    /// Accumulates the provided samples in the metric.
+    ///
+    /// # Arguments
+    ///
+    /// * `samples` - A list of samples recorded by the metric.
+    ///               Samples must be in nanoseconds.
+    /// ## Notes
+    ///
+    /// Reports an [`ErrorType::InvalidOverflow`] error for samples that
+    /// are longer than `MAX_SAMPLE_TIME`.
+    pub fn accumulate_raw_samples_nanos(&mut self, glean: &Glean, samples: &[u64]) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let mut num_too_long_samples = 0;
+        let min_sample_time = self.time_unit.as_nanos(1);
+        let max_sample_time = self.time_unit.as_nanos(MAX_SAMPLE_TIME);
+
+        glean.storage().record_with(glean, &self.meta, |old_value| {
+            let mut hist = match old_value {
+                Some(Metric::TimingDistribution(hist)) => hist,
+                _ => Histogram::functional(LOG_BASE, BUCKETS_PER_MAGNITUDE),
+            };
+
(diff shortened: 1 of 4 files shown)


glean-core/src/metrics/timing_distribution.rs
```diff
@@ -314,6 +314,64 @@ impl TimingDistributionMetric {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `samples` - A list of samples recorded by the metric.
+    ///               Samples must be in nanoseconds.
+    /// ## Notes
+    ///
+    /// Reports an [`ErrorType::InvalidOverflow`] error for samples that
+    /// are longer than `MAX_SAMPLE_TIME`.
+    /// Samples must be in nanoseconds, based on time from a monotonic clock.
```
chutten, line 327: Do they really have to be based on a monotonic clock? I mean, I guess they have to be nonnegative (because they're `u64`) which requires them to be from an increasing clock? Not sure this phrase is adding much.

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -314,6 +314,64 @@ impl TimingDistributionMetric {
...
+    pub fn accumulate_raw_samples_nanos(&mut self, glean: &Glean, samples: &[u64]) {
+        if !self.should_record(glean) {
+            return;
+        }
+
+        let mut num_too_long_samples = 0;
+        let min_sample_time = self.time_unit.as_nanos(1);
```
chutten, line 334: TIL zeros are forbidden

glean-core/tests/timing_distribution.rs
```diff
@@ -334,3 +334,51 @@ fn stopping_non_existing_id_records_an_error() {
...
+    assert_eq!(1, snapshot.values[&2_784_941_737]);
+
+    // No errors should be reported.
+    assert!(test_get_num_recorded_errors(
+        &glean,
+        metric.meta(),
+        ErrorType::InvalidValue,
+        Some("store1")
+    )
+    .is_err());
```
chutten, line 383: Please also test minimum value and the error case.

chutten (approved):
Could land as-is, but I'd prefer if it came with one more test that checks two more things.

glean-core/src/metrics/timing_distribution.rs
badboy: They should, but I guess we take whatever as long as you promise it's some duration.

badboy: (will change the sentence)

glean-core/src/metrics/timing_distribution.rs
badboy: Hey, I learned this as well.

PR #1563, opened 2021-03-22
Bug 1697214 - Clarify source tags documentation about validation

brizental:
Docs were not clear about the fact that **all** source tags
are ignored if one or more of the tags in the list are invalid.

[doc only]

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -552,12 +552,13 @@ public class Glean {
     /// Set the source tags to be applied as headers when uploading pings.
     ///
     /// If any of the tags is invalid nothing will be set and this function will
-    /// return `false`, although if we are not initialized yet, there won't be any validation.
+    /// return `false`.
+    /// If Glean is not initialized yet, tags will not be validated at this point.
     ///
     /// This is only meant to be used internally by the `GleanDebugActivity`.
     ///
-    /// @param tags A list of tags, which must be valid HTTP header values.
-    ///
+    /// - parameters:
+    ///    * tags: A list of tags, which must be valid HTTP header values.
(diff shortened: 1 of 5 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -549,6 +549,15 @@ public class Glean {
...
 
+    /// Set the source tags to be applied as headers when uploading pings.
+    ///
+    /// If any of the tags is invalid nothing will be set and this function will
+    /// return `false`, although if we are not initialized yet, there won't be any validation.
+    ///
+    /// This is only meant to be used internally by the `GleanDebugActivity`.
+    ///
+    /// @param tags A list of tags, which must be valid HTTP header values.
+    ///
```
badboy, line 560:
```suggestion
    /// - parameters:
    ///    * tags: A list of tags, which must be valid HTTP header values.
```

Yes, Swift is unusual here.

glean-core/ios/Glean/Glean.swift
```diff
@@ -549,6 +549,15 @@ public class Glean {
         }
     }
 
+    /// Set the source tags to be applied as headers when uploading pings.
+    ///
+    /// If any of the tags is invalid nothing will be set and this function will
+    /// return `false`, although if we are not initialized yet, there won't be any validation.
```
badboy, line 555:
```suggestion
    /// return `false`.
    /// If Glean is not initialized yet, tags will not be validated at this point.
```

Separates out the 2 cases, which makes it easier to understand IMO

badboy (approved):

glean-core/ios/Glean/Glean.swift
brizental: Oops, 🍝

PR #1565, opened 2021-03-22
Expose rkv-safe-mode feature on the crates that are directly consumed

badboy:
Otherwise e.g. Rust users need to awkwardly also depend on `glean-core`
to set that feature (FOG does just that).
Plus now that we might switch Python we need it on glean-ffi too to
activate it in a wheel build.

---

Prerequisite for #1551


Dexterp37 (approved):
ah! NEAT!

PR #1567, opened 2021-03-24
Bug 1693486 - New event extra properties

badboy:
So far this is an early draft that implements the new calling convention in Kotlin and Swift.
Stuff to do before this is ready though.

* [ ] Implement it in glean-parser
* [ ] Decide on the name of the new interface/protocol
* [ ] Prototype the C++/JavaScript API for FOG
* [ ] Python implementation (might be split out into a followup)
* [ ] Kotlin
	* [ ] Connection between the enum <-> interface, so the `EventMetricType` is generic over the right thing (necessary to avoid using the some extras on the wrong event)
* [ ] Swift
	* [ ] Connection between the enum <-> protocol, so the `EventMetricType` is generic over the right thing (necessary to avoid using the some extras on the wrong event)


badboy:
Only just now realized we're also changing the payload format. That's ... more work

badboy:
Turns out it's a bit more complicated than initially sketched out.
We also want to change the payload, which means we can't do the `$type -> string` conversion on the language binding side, but need to pass in the right values.
So we need to separate out the two APIs to handle them differently.

badboy:
Blocked by the discussion in https://bugzilla.mozilla.org/show_bug.cgi?id=1693486

badboy:
Closing this in favor of only changing the user-facing API (which will probably include this code, but let's start with a fresh PR)

PR #1568, opened 2021-03-26
Fix clippy suggestions (in beta)

badboy:
This is valid code in release, so we can apply them now.

### DIFF
--- a/glean-core/ffi/src/byte_buffer.rs
+++ b/glean-core/ffi/src/byte_buffer.rs
@@ -98,7 +98,7 @@ impl ByteBuffer {
         let data = buf.as_mut_ptr();
         let len = i32::try_from(buf.len()).expect("buffer length cannot fit into a i32.");
         std::mem::forget(buf);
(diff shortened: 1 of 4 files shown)


mdboom (approved):

badboy:
```
[task 2021-03-26T15:06:03.536Z] error: failed to execute compile
[task 2021-03-26T15:06:03.536Z] caused by: Compiler not supported: "error: Unrecognized option: \'E\'\n\n"
[task 2021-03-26T15:06:03.536Z] error: failed to execute compile
[task 2021-03-26T15:06:03.536Z] caused by: Compiler not supported: "error: Unrecognized option: \'E\'\n\n"
[task 2021-03-26T15:06:03.537Z] error: could not compile `glean-core`
```

I have no idea where that error is coming from.

badboy:
[sccache error](https://github.com/mozilla/sccache/pull/728). I'm going to remove our use of it, we don't store its cache so there's no benefit for us right now.

PR #1572, opened 2021-04-05
Update to 2.0.0 glean_parser schemas

mdboom:


chutten (approved):

mdboom:
Gah -- this reveals a bug in the glean_parser schema -- the `glean_*` ping names are no longer special-case allowed.  This is only a problem for glean itself, since these aren't allowed for third-party projects anyway, but we'll need to go ahead and update `glean_parser` and ship a release etc. to make this PR work.

mdboom:
Blocked by https://github.com/mozilla/glean_parser/pull/301

PR #1574, opened 2021-04-06
[doc only] Bug 1703280: Move metric parameter docs to the Glean Book

mdboom:
Relates to https://github.com/mozilla/glean_parser/pull/303

### DIFF
--- a/docs/user/user/metrics/memory_distribution.md
+++ b/docs/user/user/metrics/memory_distribution.md
@@ -13,6 +13,16 @@ This makes them suitable for measuring memory sizes on a number of different sca
 
 ## Configuration
 
+Memory distributions have a required `memory_unit` parameter, which specifies
+the unit the incoming memory size values are recorded in. The units are the
+power-of-2 units, so "kilobyte" is more correctly a "kibibyte".
+
+```
+- kilobyte == 2^10 ==         1,024 bytes
+- megabyte == 2^20 ==     1,048,576 bytes
+- gigabyte == 2^30 == 1,073,741,824 bytes
+```
+
 If you wanted to create a memory distribution to measure the amount of heap memory allocated, first you need to add an entry for it to the `metrics.yaml` file:
 
 ```YAML
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 216 utf-8
+personal_ws-1.1 en 218 utf-8
 AAR
 AARs
 ABI
@@ -146,6 +146,7 @@ io
 ios
 janerik
 js
+kibibyte
 ktlint
 lang
 latencies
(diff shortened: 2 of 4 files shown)


docs/user/SUMMARY.md
```diff
@@ -14,6 +13,7 @@
     - [Error reporting](user/error-reporting.md)
     - [Using the experiments API](user/experiments-api.md)
 - [Metric types](user/metrics/index.md)
+    - [Metric parameters](user/metrics/metric-parameters.md)
```
chutten, line 16: I'm not sure "Metric parameters" belongs under "Metric types" unless Testing and Adding metrics _also_ belong there.

docs/user/user/metrics/memory_distribution.md
```diff
@@ -13,6 +13,18 @@ This makes them suitable for measuring memory sizes on a number of different sca
...
+the unit the incoming memory size values are recorded in. The units are the
+power-of-2 units, so "kilobyte" is more correctly a "kibibyte".
+
+```
+- kilobyte == 2^10 ==         1,024 bytes
+- megabyte == 2^20 ==     1,048,576 bytes
+- gigabyte == 2^30 == 1,073,741,824 bytes
+```
+
+Values are automatically converted to and transmitted as bytes.
```
chutten, line 26: If we open the door to documenting how data is stored/transmitted in each metrics docs, we have a lot to do. I think we omit this in favour of leaving it consolidated in the Payload format docs.

chutten (changes requested):

docs/user/user/metrics/memory_distribution.md
mdboom: Sure.  Good point on this one.

docs/user/SUMMARY.md
mdboom: Yeah, I waffled on this a bit.  It also felt wrong in its old location at the top-level.  It's very closely related to the current contents of `user/metrics/index.md`, so maybe should be merged with that?

chutten: I don't have a strong holistic view of documentation organization beyond that putting it here feels wrong. I'm inclined to leave it where it was for now and file a bug for that sort of shift.

mdboom: Yeah makes sense.  Pinging @brizental for thoughts, since she's done a lot of documentation organization design work.  But if this isn't informed by that, yeah, probably the default position should be to move it back to its original location.

brizental: I am currently working on a proposal to address this organization. For the current organization though, I agree this should not be in here, but in the "Using the glean SDK" section.

chutten (approved):
:ship:

PR #1577, opened 2021-04-08
Bug 1691073, Bug 1703095: Record errors from bindings

mdboom:
This is just a PoC PR to solicit feedback before fanning it out to all of the
metric types.

This just handles the string metric type for RLB and Python.

The Python bindings add a generic decorator that can be used to catch all
exceptions when recording a metric and converting them to a Glean error.

### DIFF
--- a/glean-core/ffi/glean.h
+++ b/glean-core/ffi/glean.h
@@ -702,6 +702,11 @@ uint64_t glean_new_string_metric(FfiStr category,
                                  Lifetime lifetime,
                                  uint8_t disabled);
 
+void glean_string_record_error(uint64_t metric_id,
+                               int32_t error_type,
+                               FfiStr message,
+                               int32_t error_count);
+
 int32_t glean_string_test_get_num_recorded_errors(uint64_t metric_id,
                                                   int32_t error_type,
                                                   FfiStr storage_name);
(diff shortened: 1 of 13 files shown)


badboy (commented):
I remember us discussing this a long while ago and dropping it back then exactly because it would have required a new method per metric type.
I don't see a way around that just yet.

Is it worth exposing this on all metrics (and thus increasing the amount of things we have to implement everytime. I want a uniffi world _now_.)?
If not, should we provide this only on some selected metric types?

mdboom:
For the Python use case (bug 1703095), a `None` could be passed into anything, so probably would be weird/surprising to not be on every metric type.  Not an argument for doing this, because I share your concern about "a new FFI method for every metric type".  I don't know if there's some magic to look up the metric in *any* of the mappings instead ... I imagine not, but maybe I'm wrong...

mdboom:
Cc @badboy: Spent a little time looking at the alternative approach of "a single, generic FFI function for error reporting."  Unfortunately, the language binding side of a metric doesn't know its category/name.  We can duplicate and store that information on the language binding side.  Not difficult, but certainly suboptimal.  Or add FFI functions to retrieve the name and category, which is probably worse than the current approach in this PR, IMHO.

badboy:
We will tackle this again when UniFFI is done.

PR #1581, opened 2021-04-12
Bug 1611770 - Implement an additional check to avoid crash due to faulty timezone offset

badboy:
`Local::now` takes the user's timezone offset
and panics if it's not within a range of [-24, +24] hours.
This causes crashes in a small number of clients on Windows.

We can't determine the faulty clients
or the circumstancens under which this happens,
so the best we can do is have a workaround:

We try getting the time and timezone first,
then manually check that it is a valid timezone offset.
If it is, we proceed and use that time and offset.
If it isn't we fallback to UTC.

This has the small downside that it will use 2 calls to get the time,
but only on Windows.

See https://bugzilla.mozilla.org/show_bug.cgi?id=1611770.

---

Only downside I currently see is that I can't really test it.


glean-core/src/util.rs
```diff
@@ -48,6 +48,39 @@ pub fn get_iso_time_string(datetime: DateTime<FixedOffset>, truncate_to: TimeUni
...
+
+        use chrono::Utc;
+
+        // Get timespec, including the user's timezone.
+        let tm = time::now();
+        // Same as chrono:
+        // https://docs.rs/chrono/0.4.10/src/chrono/offset/local.rs.html#37
+        let offset = tm.tm_utcoff;
+        if let None = FixedOffset::east_opt(offset) {
+            let now: DateTime<Utc> = Utc::now();
```
chutten, line 79: Instrument it?

I agree, testing it would be nightmarish. Platform-specific, and somehow mocking `utcoff`... ugh. But with instrumentation we could at least see if we see the same number of folks with artificial UTC as the number we previously saw crashing.

chutten (approved):
Ultimately we could land as-is and I'd be okay with that. 

But I strongly suggest instrumentation here. At least temporarily (it doesn't have to be forever)

badboy:
I'm glad we have CI because I guess this code didn't work as expected.

glean-core/src/util.rs
badboy: Uff, I would like to instrument this, but it turns out instrumenting internals that are used in the init part of Glean is ... hard.

glean-core/src/internal_metrics.rs
```diff
@@ -85,6 +101,15 @@ impl CoreMetrics {
                 }),
                 None,
             ),
+
+            invalid_timezone_offset: CounterMetric::new(CommonMetricData {
+                name: "invalid_timezone_offset".into(),
+                category: "glean.time".into(),
+                send_in_pings: vec!["metrics".into()],
+                lifetime: Lifetime::Ping,
+                disabled: false,
```
chutten, line 110: What, no clever `#[cfg(build_time > '2021-06-30T00:00:00+00:00')]` cargo hack?

glean-core/src/lib.rs
```diff
@@ -213,24 +214,35 @@ impl Glean {
...
             ping_registry: HashMap::new(),
-            start_time: local_now_with_offset(),
+            start_time,
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
             debug: DebugOptions::new(),
-        })
+        };
+
+        // Can't use `into_inner_record` above, because we needed a valid `Glean` first.
```
chutten, line 238: What is `into_inner_record`?

glean-core/src/internal_metrics.rs
badboy: I _did_ think about it to be honest, but then I remembered that I should get this finished _today_

glean-core/src/lib.rs
badboy: A thing that doesn't exist and should have never existed (I'm glad I didn't push the code, it was like 50 lines more for no good reason).
I'll change the comment.

PR #1582, opened 2021-04-12
Add a warning about using events

wlach:
Per discussion today in #glam, we should warn about using events in this document as well (as we do here: https://mozilla.github.io/glean/book/user/adding-new-metrics.html#do-you-need-to-know-the-order-of-events-relative-to-other-events).

### DIFF
--- a/docs/user/user/metrics/event.md
+++ b/docs/user/user/metrics/event.md
@@ -1,7 +1,5 @@
 # Events
 
-> **Important:** events are the most expensive metric type to record, transmit, store and analyze, so they should be used sparingly, and only when none of the other metric types are sufficient for answering your question.
-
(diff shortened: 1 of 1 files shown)


docs/user/user/metrics/event.md
```diff
@@ -1,5 +1,7 @@
 # Events
 
+> **Important:** events are the most expensive metric type to record, transmit, store and analyze, so they should be used sparingly, and only when none of the other metric types are sufficient for answering your question.
+
```
brizental, line 4: nit: IMO this should come after the initial description at the end of this section, not at the start.

brizental (approved):

docs/user/user/metrics/event.md
wlach: Makes sense, I made the adjustment.

PR #1584, opened 2021-04-13
Bug 1702280 - Document the limit on number of statically-defined labels

badboy:
The limit was raised in https://github.com/mozilla/glean_parser/pull/297

[doc only]

---

This should be landed when we update glean_parser to include that limit.


travis79 (approved):

chutten (approved):
Could you also add a note to https://mozilla.github.io/glean/book/user/metrics/index.html#labeled-metrics while you're in the mood?

PR #1585, opened 2021-04-13
Update to Xcode 12.4

badboy:
We still require to update carthage though, as only 0.37.0 brings
xcframework support.

Supported software: https://circle-macos-docs.s3.amazonaws.com/image-manifest/v4519/index.html

---

I saw that the iOS builds just broken on `main`. Not looking into this for now, but just upgrading Xcode and see if that works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -528,7 +528,8 @@ jobs:
       - run:
           name: Install new carthage
           command: |
-            brew upgrade carthage
+            time brew update
+            time brew upgrade carthage
       - run:
           name: Set Ruby Version
           command: echo 'chruby ruby-2.7.2' >> ~/.bash_profile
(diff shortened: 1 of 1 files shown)


badboy:
I guess not, because the upstream URL is 404ing right now.

.circleci/config.yml
```diff
@@ -522,13 +522,14 @@ jobs:
...
     macos:
-      xcode: "12.3.0"
+      xcode: "12.4.0"
     steps:
       - checkout
       - run:
           name: Install new carthage
           command: |
-            brew upgrade carthage
+            time brew update
```
travis79, line 531: time?

travis79 (approved):

.circleci/config.yml
badboy: Wanted to know the time it takes without splitting this up into circleci tasks. Probably gonna remove this before landing, but right now it's still broken anyway for other reasons now.

travis79: Gotcha!

PR #1588, opened 2021-04-15
CI: Use local copy of the codecov script

badboy:


badboy:
Merging with admin privileges.

PR #1590, opened 2021-04-16
Bug 1705319 - Ensure at least pip 20.3 is available in iOS build

badboy:


glean-core/ios/sdk_generator.sh
```diff
@@ -139,6 +139,11 @@ fi
 VENVDIR="${SOURCE_ROOT}/.venv"
 
 [ -x "${VENVDIR}/bin/python" ] || python3 -m venv "${VENVDIR}"
+# We need at least pip 20.3 for Big Sur support, see https://pip.pypa.io/en/stable/news/#id48
+# Latest pip is 21.0.1
+if ! "${VENVDIR}"/bin/pip --version | grep -E 'pip (21|20.[3-9])'; then
+  "${VENVDIR}"/bin/pip install --upgrade pip
```
mdboom, line 145: You could just do `pip install "pip>=20.3"` here, which will upgrade only if too old (and remove the grep logic around it).  But this is also fine.

mdboom (approved):

glean-core/ios/sdk_generator.sh
badboy: Oh I like that.

mdboom (approved):

PR #1592, opened 2021-04-19
Gradle: Point to glean dictionary when generating metrics docs

wlach:
The idea is to give people skimming the markdown documentation a pointer to the Glean Dictionary, which
is usually a better way of accessing this information.

This depends on a pending change to the glean parser: https://github.com/mozilla/glean_parser/pull/298

### DIFF
--- a/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
+++ b/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
@@ -251,6 +251,8 @@ except:
                 args "--allow-missing-files"
                 args "-f"
                 args "markdown"
+                args "--option"
+                args "introduction_extra=\"For another view of these metrics, see the [Glean Dictionary](https://dictionary.protosaur.dev/)\""
                 args "-o"
                 args gleanDocsDirectory
 


gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -251,6 +251,8 @@ except:
                 args "--allow-missing-files"
                 args "-f"
                 args "markdown"
+                args "--option"
+                args "introduction_extra=\"For another view of these metrics, see the [Glean Dictionary](https://dictionary.protosaur.dev/)\""
```
wlach, line 255: It goes without saying that this is completely untested, but I don't see how this could be *too* wrong. 😄

mdboom (approved):
Looks good -- we should be able to test this locally (since it would force a regeneration of Glean's own metrics.md).

Note that this will only affect Android users -- we'll also need to handle iOS, and FOG may require changes in m-c (Cc @chutten).

chutten:
Correct. File a bug and we'll block it on vendoring a version of `glean_parser` that supports the `introduction_extra` option.

wlach:
Per discussion on `#glean-dev`, I think we can skip this. My understanding is that our guidance moving forward is for most Mozilla projects *not* to opt-in to markdown-based doc generation, and just point to https://dictionary.telemetry.mozilla.org instead. I'll file a different PR to just update the docs (as well as a few downstream PRs to update other things once dictionary.tmo is considered ready).

PR #1595, opened 2021-04-26
Bug 1670261 - Split up Glean & prepare for GeckoView-provided symbols [ci full]

badboy:
This is the Glean part of changes to provide Glean through GeckoView.
It's a co-production of @agi (who did the Gradle plugin changes) and me.

I put extensive explanations of each step into the commit descriptions.
This will need changes to Gecko/GeckoView, as well as some documentation changes and a tiny bit of adjustement for consumers.
It will need a coordinated release with GeckoView so we can get that into Fenix, etc.
Posting now to get early feedback already.

* [x] Post Gecko/GeckoView changes
* [ ] Adopt documentation as necessary to make it clear to consumers what they need to do
* [ ] Prepare PRs for consumers so they continue to work with this change


badboy:
Opening this up for review. I'm trying to coordinate the remaining work and thus the sooner this gets into a final stage, the sooner we can land all the work necessary.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -72,14 +72,22 @@ internal fun loadIndirect(libraryName: String): LibGleanFFI {
...
+            Proxy.newProxyInstance(
+                LibGleanFFI::class.java.classLoader,
+                arrayOf(LibGleanFFI::class.java)
+            ) { _, _, _ ->
+                throw IllegalStateException("Glean functionality not available", e)
+            } as LibGleanFFI
+        }
     }
 
+    Log.e("glean/loadIndirect", "Glean loaded. Continue.")
```
chutten, line 90: Logging an error on every app start seems odd to me. Leftover from debugging?

glean-core/bundle/src/lib.rs
```diff
@@ -3,3 +3,21 @@
...
+/// rexported.
+/// This is only required for debug builds (and `debug_assertions` is the closest thing we have to
+/// check that).
+/// In release builds we rely on LTO builds to take care of it.
+/// Our tests should ensure this actually happens.
+///
+/// See https://github.com/rust-lang/rust/issues/50007
+#[cfg(debug_assertions)]
+#[no_mangle]
+pub extern "C" fn _glean_force_reexport_donotcall() {
```
chutten, line 21: ick.

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
```diff
@@ -62,16 +63,21 @@ internal fun Pointer.getRustString(): String {
...
+ * If that also fails, give up and raise an exception.
+ */
+internal fun loadIndirect(libraryName: String): LibGleanFFI {
+    val lib = try {
+        Native.load(libraryName, LibGleanFFI::class.java) as LibGleanFFI
+    } catch (e: UnsatisfiedLinkError) {
+        Log.i("glean/loadIndirect", "Failed to load $libraryName")
+        try {
+            Log.i("glean/loadIndirect", "Trying to load libxul directly")
+            Native.load("xul", LibGleanFFI::class.java) as LibGleanFFI
```
chutten, line 78: Eventually we'll probably want to prefer libxul's Glean over libgleanffi for a couple reasons
1) Performance micro-optimization that we don't need to fail to load a lib on every startup in the "usual" case
2) In the event that we do have both, we probably want to prefer the libxul-present one since that's the one libxul itself will be using.

Modulo us shipping this to Android first and m-c second, of course. If we ship this code to Android first, it'll succeed at finding a libxul and not fall through to loading libgleanffi. (It should fail at finding symbols to satisfy LibGleanFFI... but I could see us not wanting to rely on that.)

chutten (approved):
Can't really say I understand all of the Android-ness, but I like the approach and I'm pretty sure I understand where m-c fits in. A couple of nits and commentary as per usual.

chutten:
Oh, I neglected to ask for docs and tests. We'll be needing both (presuming tests are at all possible other than "the build fails")

badboy:
> Oh, I neglected to ask for docs and tests. We'll be needing both (presuming tests are at all possible other than "the build fails")

Docs coming.
We do build the C sample ([example](https://app.circleci.com/pipelines/github/mozilla/glean/7687/workflows/628fb545-fb4e-4ca1-aa04-033a51377aec/jobs/136319) and run it (we don't test that it does the right stuff, but if it crashes CI would be red)

glean-core/android/src/main/java/mozilla/telemetry/glean/rust/LibGleanFFI.kt
badboy: Definitely.

glean-core/bundle/src/lib.rs
badboy: yes.

build.gradle
```diff
@@ -76,6 +76,7 @@ plugins {
 
 allprojects {
     repositories {
+        mavenLocal()
```
Dexterp37, line 79: Should this be removed?

glean-core/android-native/build.gradle
```diff
@@ -0,0 +1,117 @@
...
+android {
+    compileSdkVersion rootProject.ext.build.compileSdkVersion
+
+    defaultConfig {
+        minSdkVersion rootProject.ext.build['minSdkVersion']
+        targetSdkVersion rootProject.ext.build['targetSdkVersion']
+
+        // TODO: 1551691 Get the version from git tag...? Also, we need to select a
+        // version that won't conflict with legacy glean-ac versions.
+        buildConfigField("String", "LIBRARY_VERSION", "\"0.1\"")
```
Dexterp37, line 22: Should we use the same numbers that we use for the rest of the SDK?

glean-core/android-native/proguard-rules-consumer.pro
```diff
@@ -0,0 +1,7 @@
+# ProGuard rules for consumers of this library.
+
+# JNA specific rules
+# See https://github.com/java-native-access/jna/blob/master/www/FrequentlyAskedQuestions.md#jna-on-android
+-dontwarn java.awt.*
+-keep class com.sun.jna.* { *; }
+-keepclassmembers class * extends com.sun.jna.* { public *; }
```
Dexterp37, line 7: Should we retain the same things we retain for the [`android` package](https://searchfox.org/glean/source/glean-core/android/proguard-rules-consumer.pro#15)?

glean-core/android-native/src/main/AndroidManifest.xml
```diff
@@ -0,0 +1,5 @@
+<manifest xmlns:android="http://schemas.android.com/apk/res/android"
+    package="mozilla.telemetry.glean_native" >
+    <application>
```
Dexterp37, line 3: This shouldn't be exposing [debug activity](https://searchfox.org/glean/rev/0fc3f5092b3dd1015abb7bc725a36eea50b7ec32/glean-core/android/src/main/AndroidManifest.xml#4) as well, right? I don't think so because that's still something `android` needs to do

build.gradle
badboy: Oops, yes.

glean-core/android-native/build.gradle
badboy: This is copied. I don't think we _need_ it anymore really.

glean-core/android-native/proguard-rules-consumer.pro
badboy: I'll check, you might be right.

badboy: We don't need to, the native package doesn't export any namespaces.

glean-core/android-native/src/main/AndroidManifest.xml
badboy: It shouldn't (so this code here is correct). 
Consumers will depend on the glean package, which does expose the debug activity.

badboy:
close-open to re-trigger full build on taskcluster

PR #1596, opened 2021-04-26
Bug 1706411 - Use proper `Path`s throughout glean-core

badboy:
This might or might not avoid the crash in the linked bug.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -234,8 +234,8 @@ impl Database {
     ///
     /// It also loads any Lifetime::Ping data that might be
     /// persisted, in case `delay_ping_lifetime_io` is set.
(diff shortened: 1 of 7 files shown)


chutten (approved):
r+ Up to you if you'd like a CHANGELOG item for this.

glean-core/src/database/mod.rs
```diff
@@ -783,19 +783,19 @@ mod test {
     use crate::tests::new_glean;
     use crate::CommonMetricData;
     use std::collections::HashMap;
+    use std::path::Path;
     use tempfile::tempdir;
 
     #[test]
     fn test_panicks_if_fails_dir_creation() {
-        assert!(Database::new("/!#\"'@#°ç", false).is_err());
+        let path = Path::new("/!#\"'@#°ç");
```
mdboom, line 791: Is it possible to write a Windows-only test with a not-UTF-16-but-allowed path?

PR #1599, opened 2021-04-28
Bug 1672951 - Implement a Metrics Ping Scheduler in glean-core

chutten:
The MPS has actually been written for a little while. I strove for as direct a translation of `MetricsPingScheduler.{kt|swift}` as was practicable (we can always refactor it later). The hard part has been tests. 

I've tried making this testable in a few ways. The first way you can see in the second commit and is straightforward, but it can't actually observe most of the behaviour we hope to test.

The second way apes the `Policy` approach of gecko JSMs. (I had some struggles with the borrow checker on this approach, so it's very possible this is more ornate than it needs to be) I think this `Policy` approach is the better one because it allows us to inspect and test every operation of the MPS which, as history has taught us, is absolutely crucial for peace of mind.

(( I also attempted to automock Glean, but it would require some serious modifications to make the Glean struct's impl suit the mocking library (mockall), so I abandoned it. ))

I'm looking for some honest feedback about this, both for whether the approach itself is distasteful as well as what patterns I've missed in pursuing things in this manner. I've taken heart from the [Book suggesting RefCell-based interior mutability for test mocks itself](https://doc.rust-lang.org/book/ch15-05-interior-mutability.html) to mean I'm not completely out to lunch here, but I might be.

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -187,6 +187,8 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
                 language_binding_name: LANGUAGE_BINDING_NAME.into(),
                 max_events: cfg.max_events,
                 delay_ping_lifetime_io: cfg.delay_ping_lifetime_io,
+                app_build: client_info.app_build.clone(),
+                use_core_mps: true,
             };
 
             let glean = match Glean::new(core_cfg) {
@@ -361,6 +363,7 @@ pub fn shutdown() {
     }
 
     crate::launch_with_glean_mut(|glean| {
+        glean_core::cancel_metrics_ping_scheduler();
         glean.set_dirty_flag(false);
     });
 
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -28,6 +28,7 @@
 //!     channel: None,
 //!     server_endpoint: None,
 //!     uploader: None,
+//!     use_core_mps: false,
 //! };
 //! glean::initialize(cfg, ClientInfoMetrics::unknown());
 //!
@@ -188,7 +189,7 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
                 max_events: cfg.max_events,
                 delay_ping_lifetime_io: cfg.delay_ping_lifetime_io,
                 app_build: client_info.app_build.clone(),
-                use_core_mps: true,
+                use_core_mps: cfg.use_core_mps,
             };
 
             let glean = match Glean::new(core_cfg) {
@@ -304,7 +305,7 @@ pub fn initialize(cfg: Configuration, client_info: ClientInfoMetrics) {
                 // Set up information and scheduling for Glean owned pings. Ideally, the "metrics"
                 // ping startup check should be performed before any other ping, since it relies
                 // on being dispatched to the API context before any other metric.
-                // TODO: start the metrics ping scheduler, will happen in bug 1672951.
+                glean.start_metrics_ping_scheduler();
 
                 // Check if the "dirty flag" is set. That means the product was probably
                 // force-closed. If that's the case, submit a 'baseline' ping with the
@@ -363,7 +364,7 @@ pub fn shutdown() {
     }
 
     crate::launch_with_glean_mut(|glean| {
-        glean_core::cancel_metrics_ping_scheduler();
+        glean.cancel_metrics_ping_scheduler();
         glean.set_dirty_flag(false);
     });
 
@@ -464,16 +465,15 @@ pub fn set_upload_enabled(enabled: bool) {
         let old_enabled = glean.is_upload_enabled();
         glean.set_upload_enabled(enabled);
 
-        // TODO: Cancel upload and any outstanding metrics ping scheduler
-        // task. Will happen on bug 1672951.
-
         if !old_enabled && enabled {
+            glean.start_metrics_ping_scheduler();
             // If uploading is being re-enabled, we have to restore the
             // application-lifetime metrics.
             initialize_core_metrics(&glean, &state.client_info, state.channel.clone());
         }
 
         if old_enabled && !enabled {
+            glean.cancel_metrics_ping_scheduler();
             // If uploading is disabled, we need to send the deletion-request ping:
             // note that glean-core takes care of generating it.
             state.upload_manager.trigger_upload();
@@ -684,8 +684,6 @@ pub(crate) fn destroy_glean(clear_stores: bool) {
 pub fn test_reset_glean(cfg: Configuration, client_info: ClientInfoMetrics, clear_stores: bool) {
     destroy_glean(clear_stores);
 
-    // Always log pings for tests
-    //Glean.setLogPings(true)
     initialize(cfg, client_info);
 }
 
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -98,12 +98,7 @@ pub fn setup_glean(glean: Glean) -> Result<()> {
     // calling `initialize` on the global singleton and further operations check that it has been
     // initialized.
     if GLEAN.get().is_none() {
-        if GLEAN.set(Mutex::new(glean)).is_ok() {
-            let glean = &GLEAN.get().unwrap().lock().unwrap();
-            if glean.schedule_metrics_pings {
-                scheduler::schedule(&glean);
-            }
-        } else {
+        if GLEAN.set(Mutex::new(glean)).is_err() {
             log::warn!(
                 "Global Glean object is initialized already. This probably happened concurrently."
             )
@@ -1017,18 +1012,28 @@ impl Glean {
         // We don't care about this failing, maybe the data does just not exist.
         let _ = self.event_data_store.clear_all();
     }
+
+    /// Instructs the Metrics Ping Scheduler's thread to exit cleanly.
+    /// If Glean was configured with `use_core_mps: false`, this has no effect.
+    pub fn cancel_metrics_ping_scheduler(&self) {
+        if self.schedule_metrics_pings {
+            scheduler::cancel();
+        }
+    }
+
+    /// Instructs the Metrics Ping Scheduler to being scheduling metrics pings.
+    /// If Glean wsa configured with `use_core_mps: false`, this has no effect.
+    pub fn start_metrics_ping_scheduler(&self) {
+        if self.schedule_metrics_pings {
+            scheduler::schedule(&self);
+        }
+    }
 }
 
 /// Returns a timestamp corresponding to "now" with millisecond precision.
 pub fn get_timestamp_ms() -> u64 {
     const NANOS_PER_MILLI: u64 = 1_000_000;
     zeitstempel::now() / NANOS_PER_MILLI
-}
-
-/// Instructs the Metrics Ping Scheduler's thread to exit cleanly.
-/// If Glean was configured with `use_core_mps: false`, this has no effect.
-pub fn cancel_metrics_ping_scheduler() {
-    scheduler::cancel();
 }
 
 // Split unit tests to a separate file, to reduce the file of this one.
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -38,6 +38,7 @@ mod internal_metrics;
 mod internal_pings;
 pub mod metrics;
 pub mod ping;
+mod scheduler;
 pub mod storage;
 mod system;
 pub mod traits;
@@ -97,7 +98,12 @@ pub fn setup_glean(glean: Glean) -> Result<()> {
     // calling `initialize` on the global singleton and further operations check that it has been
     // initialized.
     if GLEAN.get().is_none() {
-        if GLEAN.set(Mutex::new(glean)).is_err() {
+        if GLEAN.set(Mutex::new(glean)).is_ok() {
+            let glean = &GLEAN.get().unwrap().lock().unwrap();
+            if glean.schedule_metrics_pings {
+                scheduler::schedule(&glean);
+            }
+        } else {
             log::warn!(
                 "Global Glean object is initialized already. This probably happened concurrently."
             )
@@ -129,6 +135,11 @@ pub struct Configuration {
     pub max_events: Option<usize>,
     /// Whether Glean should delay persistence of data from metrics with ping lifetime.
     pub delay_ping_lifetime_io: bool,
+    /// The application's build identifier. If this is different from the one provided for a previous init,
+    /// and use_core_mps is `true`, we will trigger a "metrics" ping.
+    pub app_build: String,
+    /// Whether Glean should schedule "metrics" pings.
+    pub use_core_mps: bool,
 }
 
 /// The object holding meta information about a Glean instance.
@@ -147,6 +158,8 @@ pub struct Configuration {
 ///     upload_enabled: true,
 ///     max_events: None,
 ///     delay_ping_lifetime_io: false,
+///     app_build: "".into(),
+///     use_core_mps: false,
 /// };
 /// let mut glean = Glean::new(cfg).unwrap();
 /// let ping = PingType::new("sample", true, false, vec![]);
@@ -185,6 +198,8 @@ pub struct Glean {
     is_first_run: bool,
     upload_manager: PingUploadManager,
     debug: DebugOptions,
+    app_build: String,
+    schedule_metrics_pings: bool,
 }
 
 impl Glean {
@@ -208,7 +223,7 @@ impl Glean {
             /* seconds per interval */ 60, /* max pings per interval */ 15,
         );
 
-        // We only scan the pending ping sdirectories when calling this from a subprocess,
+        // We only scan the pending ping directories when calling this from a subprocess,
         // when calling this from ::new we need to scan the directories after dealing with the upload state.
         if scan_directories {
             let _scanning_thread = upload_manager.scan_pending_pings_directories();
@@ -233,6 +248,9 @@ impl Glean {
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
             debug: DebugOptions::new(),
+            app_build: cfg.app_build.to_string(),
+            // Subprocess doesn't use "metrics" pings so has no need for a scheduler.
+            schedule_metrics_pings: false,
         };
 
         // Can't use `local_now_with_offset_and_record` above, because we needed a valid `Glean` first.
@@ -288,6 +306,9 @@ impl Glean {
             }
         }
 
+        // We set this only for non-subprocess situations.
+        glean.schedule_metrics_pings = cfg.use_core_mps;
+
         // We only scan the pendings pings directories **after** dealing with the upload state.
         // If upload is disabled, we delete all pending pings files
         // and we need to do that **before** scanning the pending pings folder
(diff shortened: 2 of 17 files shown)


glean-core/src/scheduler.rs
```diff
@@ -168,33 +248,67 @@ fn get_last_sent_build_metric() -> StringMetric {
...
         let (glean, _t) = new_glean(None);
-        let now = Local.ymd(2021, 4, 28).and_hms(14, 57, 15);
-        let last_sent = Local.ymd(2021, 4, 28).and_hms(4, 0, 0);
+        let mut policy = TestPolicy {
+            ..Default::default()
+        };
+        policy.app_build = "a build".to_string();
+        policy
+            .last_sent_build
+            .replace(Some("a different build".to_string()));
```
badboy, line 302: This can probably be simplified to:

```rust
        let mut policy = TestPolicy {
			app_buid: "a build".to_string(),
			last_sent_build: Some("a different build".to_string)),
            ..Default::default()
        };
```

glean-core/src/lib.rs
```diff
@@ -101,6 +102,10 @@ pub fn setup_glean(glean: Glean) -> Result<()> {
             log::warn!(
                 "Global Glean object is initialized already. This probably happened concurrently."
             )
+        } else {
+            // Can only kick off the "metrics" ping scheduler after we have a global Glean.
+            let glean = &GLEAN.get().unwrap().lock().unwrap();
+            scheduler::schedule(&glean);
```
badboy, line 108: I would probably swap around the `if` condition here:

```rust
if GLEAN.set(Mutex::new(glean)).is_ok() {
	// Can only kick off the "metrics" ping scheduler after we have a global Glean.
    let glean = &GLEAN.get().unwrap().lock().unwrap();
    scheduler::schedule(&glean);
} else {
    // ...
}
```

That is: handle the error case last

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,314 @@
...
+    }
+
+    fn schedule_on_thread(
+        &self,
+        sleep_duration: Duration,
+        generation: u32,
+        reason: Option<&'static str>,
+    ) {
+        let thread_gen = Arc::clone(&THREAD_GENERATION.get_or_init(Default::default));
+        std::thread::Builder::new()
```
badboy, line 95: So I'm a bit worried about this new thread. It's long running and sleeping all the time.
I _think_ we had the problem before that Firefox tooling wants us to ensure threads are stopped before the main thread. As it stands we don't have a way to inform this thread.

badboy: (Additionally I'd wish we would just have a way to hook into whatever-existing background-threaded timers the surrounding thing gives us :( )

badboy (commented):
I think the general approach is sound and indeed a 1:1 mapping of the "old" implementation.
I think if we treat this implementation as "experimental" (maybe even put it behind a feature with that name?) we can consider some (breaking) changes on it as followups.

I like the testing approach, though might bikeshed on the `Policy` name (but don't have a better one yet).
(and of course we need to add some more tests) 

I have a bunch of comments I left inline.

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,164 @@
...
+    // We aim to cover 3 cases here:
+    //
+    // 1. The ping was already collected on the current calendar day;
+    //    only schedule one for collection on the next calendar day at the due time.
+    // 2. The ping was NOT collected on the current calendar day AND we're later
+    //    than today's due time; collect the ping immediately.
+    // 3. The ping was NOT collected on the current calendar day BUT we still have
+    //    some time to the due time; schedule for sending the current calendar day.
+
+    let already_sent_today = last_sent_time.map_or(false, |d| d.ordinal() == now.ordinal());
```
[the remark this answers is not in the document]
chutten: `ordinal` gives day of year: https://docs.rs/chrono/0.4.19/chrono/trait.Datelike.html#tymethod.ordinal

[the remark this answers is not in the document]
badboy: Guess I looked at ... I don't know what I looked at.
Anyway, bug stands for a browser that's not used in a full year! (Edge case I guess...)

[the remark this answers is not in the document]
chutten: How 'bout I put in a comment about it then? I trust `chrono` to properly handle getting me calendar days since Jan 1 over me trying to figure out days since an epoch. I'd get the timezones wrong for sure.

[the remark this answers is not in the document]
badboy: 👍

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,164 @@
...
+
+const SCHEDULED_HOUR: u32 = 4;
+
+static THREAD_GENERATION: OnceCell<Arc<AtomicU32>> = OnceCell::new();
+
+/// Performs startup checks to decide when to schedule the next "metrics" ping collection.
+/// Must be called before draining the preinit queue.
+/// (We're at the Language Bindings' mercy for that)
+pub fn schedule(glean: &Glean) {
+    let now = Local::now();
```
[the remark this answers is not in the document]
chutten: ...but won't we have problems with any constructed value of `DateTime<Local>` no matter what? I mean, if `Local`'s timezone is problematic, there's not really much we can do with it.

I guess I'll rewrite scheduler to use `DateTime<FixedOffset>` and use `local_now_with_offset` to get it.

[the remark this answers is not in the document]
badboy: If we use the wrong timezone consistently then it's still fine.
But we shouldn't crash when encountering something wrong.

Using `DateTime<FixedOffset>` throughout should be still correct usage.

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,164 @@
...
+
+    schedule_internal(&glean, now)
+}
+
+fn schedule_internal(glean: &Glean, now: DateTime<Local>) {
+    let last_sent_build_metric = get_last_sent_build_metric();
+    if let Some(last_sent_build) = last_sent_build_metric.get_value(&glean, INTERNAL_STORAGE) {
+        // XXX: If `app_build` is longer than StringMetric's max length, we will always
```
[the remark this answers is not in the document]
chutten: Hm, that's a question: If the length is too long that it gets truncated, will the error we record (`invalid_overflow` methinks) show up in a place we'd see it? Or will it be hidden because I'm using `INTERNAL_STORAGE`?

Because if we report it through error reporting, then Woo to the Hoo, job done.

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,164 @@
...
+        // treat it as a changed build when really it isn't.
+        if &last_sent_build != glean.app_build.as_ref().unwrap() {
```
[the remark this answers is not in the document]
chutten: We have `app_build` iff `cfg.use_core_mps`. But I'm not happy with how it's only enforced by logic not language. `cfg.app_build` is always present (often "unknown"), I was using `Option` to signal more than the app_build... maybe I should add `use_core_mps` to `struct Glean`?

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,164 @@
...
+    };
+
+    // Other MPSes cancel outstanding tasks here. I'm not sure we need to, since schedule() is only
+    // called once and we're only called from there and the task. But let's follow the leader.
+    let thread_gen = Arc::clone(&THREAD_GENERATION.get_or_init(Default::default));
+    let my_gen = thread_gen.fetch_add(1, Ordering::Relaxed) + 1;
+    std::thread::Builder::new()
+        .name("glean.mps".into())
+        .spawn(move || {
+            // Waiting on `duration_zero #73544` to use Duration::ZERO directly.
```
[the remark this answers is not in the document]
chutten: Yes. 

Though I guess with our minimum Rust version as it stands we can't use it without upgrading. Should I omit the comment?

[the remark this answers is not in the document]
badboy: Nah, we can leave that in. Eventually we will update.

chutten:
Notable changes:
* Split ping submission away from scheduling for clarity
* Use of `Condvar` for scheduling, cancellation
* Lotsa tests

I should've caught all the previous comments. Lemme know if I missed any.


....and clippy immediately fails for something `make lint-rust` didn't catch locally? Weird. Maybe my clippy's out of date. Anyhoo. Nitfixes incoming.

chutten:
Had to add a lock for scheduler tests, which makes my nose twitch. Being able to cancel the scheduler from any thread without caring if the scheduler's got work to do is a feature, not a bug, but any time I have to do this sort of workaround it makes me wonder if I'm decreasing safety.

glean-core/src/lib.rs
```diff
@@ -233,6 +248,9 @@ impl Glean {
             max_events: cfg.max_events.unwrap_or(DEFAULT_MAX_EVENTS),
             is_first_run: false,
             debug: DebugOptions::new(),
+            // Subprocess doesn't use "metrics" pings so has no need for a scheduler.
```
badboy, line 251: This comment doesn't apply anymore I guess

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+// Clippy thinks an AtomicBool would be preferred, but Condvar requires a full Mutex.
+// See https://github.com/rust-lang/rust-clippy/issues/1516
+#[allow(clippy::mutex_atomic)]
+static TASK_CONDVAR: Lazy<Arc<(Mutex<bool>, Condvar)>> =
+    Lazy::new(|| Arc::new((Mutex::new(false), Condvar::new())));
+
+trait MetricsPingSubmitter {
+    fn submit_metrics_ping(&self, glean: &Glean, reason: Option<&str>, now: DateTime<FixedOffset>);
+}
+trait MetricsPingScheduler {
```
badboy, line 30: Newline to separate different traits

```suggestion

trait MetricsPingScheduler {
```

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+
+const SCHEDULED_HOUR: u32 = 4;
+
+// Clippy thinks an AtomicBool would be preferred, but Condvar requires a full Mutex.
+// See https://github.com/rust-lang/rust-clippy/issues/1516
+#[allow(clippy::mutex_atomic)]
+static TASK_CONDVAR: Lazy<Arc<(Mutex<bool>, Condvar)>> =
+    Lazy::new(|| Arc::new((Mutex::new(false), Condvar::new())));
+
+trait MetricsPingSubmitter {
```
badboy, line 27: Should this call out why we use this trait and the one below?

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+    fn submit_metrics_ping(&self, glean: &Glean, reason: Option<&str>, now: DateTime<FixedOffset>);
```
badboy, line 28: Should have some documentation.
E.g. it's not obvious why `now` is passed.

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+}
+trait MetricsPingScheduler {
+    fn start_scheduler(
```
badboy, line 31: Should have some documentation too

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+        &self,
+        submitter: impl MetricsPingSubmitter + Send + 'static,
+        now: DateTime<FixedOffset>,
+        when: When,
+    );
+}
+
+struct GleanMetricsPingSubmitter {}
```
badboy, line 39: Maybe explain that this models the default behavior?

```suggestion
/// Submit pings through Glean directly.
struct GleanMetricsPingSubmitter;
```

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+        submitter: impl MetricsPingSubmitter + Send + 'static,
+        now: DateTime<FixedOffset>,
+        when: When,
+    ) {
+        start_scheduler(submitter, now, when);
+    }
+}
+
+/// Performs startup checks to decide when to schedule the next "metrics" ping collection.
+/// Must be called before draining the preinit queue.
```
badboy, line 66: Maybe make this more obvious then?

```suggestion
/// **Must** be called before draining the preinit queue.
```

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+
+#[derive(Debug, PartialEq)]
+enum When {
+    Today,
+    Tomorrow,
+    Reschedule,
+}
+
+impl When {
+    fn until(&self, now: DateTime<FixedOffset>) -> std::time::Duration {
```
badboy, line 142: Probably useful to have docs here explaining what it tries to do:

> Return the duration from `now` until the specified deadline

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+                (now.date() + Duration::days(1)).and_hms(SCHEDULED_HOUR, 0, 0)
+            }
+        };
+        // After rust-lang/rust#73544 can use std::time::Duration::ZERO
+        (fire_date - now)
+            .to_std()
+            .unwrap_or_else(|_| std::time::Duration::from_millis(0))
+    }
+
+    fn reason(&self) -> &'static str {
```
badboy, line 157: I like this :)

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+            schedule_validator,
+            validator_run_count: Arc::clone(&scheduler_count),
+        };
+        (submitter, submitter_count, scheduler, scheduler_count)
+    }
+
+    // Ensure that if we have a different build, we immediately submit an "upgrade" ping
+    // and schedule a "reschedule" ping for tomorrow.
+    #[test]
+    fn test_different_app_builds() {
```
badboy, line 296: In Rust one usually doesn't use the `test_` prefix for this functions.
Otherwise all functions in the output just start the same, making it harder to parse at a glance.

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+        // If we bring it back before `SCHEDULED_HOUR` it should give us the duration:
+        let earlier = now.date().and_hms(SCHEDULED_HOUR - 1, 0, 0);
+        assert_eq!(
+            std::time::Duration::from_secs(3600),
+            When::Today.until(earlier)
+        );
+
+        // `Tomorrow` and `Reschedule` should differ only in their `reason()`
+        assert_eq!(
+            std::time::Duration::from_secs(46670),
```
badboy, line 390: I first was a bit confused by this "magic" duration of seconds. It's 12 hours, 57 minutes and 10 seconds. The exact time it takes from 15:02:10 to 00:04:00. Worth noting somewhere?

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+                },
+                |_, _| panic!("Not even using the scheduler this time."),
+            )
+        };
+
+        // Test Today.
+        let (submitter, submitter_count, _, _) = proxy_factory();
+        let handle = start_scheduler(submitter, now, When::Today);
+        *cancelled_lock.lock().unwrap() = true; // Cancel the scheduler thread.
+        condvar.notify_all(); // Notify any/all listening schedulers to check whether they were cancelled.
```
badboy, line 436: Shouldn't this just call `cancel()`? It's what the outside will call and does the same inside, so we don't need to repeat it over and over again in tests.

```suggestion
        super::cancel();
```

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,496 @@
...
+        handle.join().unwrap(); // Should complete immediately.
+        assert_eq!(0, submitter_count.swap(0, Ordering::Relaxed));
+
+        // Test Tomorrow.
+        let (submitter, submitter_count, _, _) = proxy_factory();
+        *cancelled_lock.lock().unwrap() = false; // Uncancel.
+        let handle = start_scheduler(submitter, now, When::Tomorrow);
+        *cancelled_lock.lock().unwrap() = true; // Cancel the scheduler thread.
+        condvar.notify_all(); // Notify any/all listening schedulers to check whether they were cancelled.
```
badboy, line 445: As above. Gonna stop calling it out.

glean-core/src/lib.rs
```diff
@@ -992,6 +1012,12 @@ pub fn get_timestamp_ms() -> u64 {
     zeitstempel::now() / NANOS_PER_MILLI
 }
 
+/// Instructs the Metrics Ping Scheduler's thread to exit cleanly.
+/// If Glean was configured with `use_core_mps: false`, this has no effect.
+pub fn cancel_metrics_ping_scheduler() {
```
badboy, line 1017: Right now this is not called anywhere.
Should this be called in RLB's `shutdown` function?

badboy (changes requested):
Bunch of questions inline.
Overall this code looks good though and we're close to landing this.

I wonder if you already managed to test this out in use in FOG?

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,515 @@
...
+        glean.app_build = "a build".into();
+        get_last_sent_build_metric().set(&glean, "a different build");
+
+        let (submitter, submitter_count, scheduler, scheduler_count) = new_proxies(
+            |_, reason| assert_eq!(reason, Some("upgrade")),
+            |_, when| assert_eq!(when, When::Reschedule),
+        );
+
+        schedule_internal(&glean, submitter, scheduler, local_now_with_offset().0);
+        assert_eq!(1, submitter_count.swap(0, Ordering::Relaxed));
```
badboy, line 330: Wondered why you use `swap(0)` here, when `load()` would work just as fine? (It's not wrong and can stay, just trying to understand your thinking)

glean-core/src/scheduler.rs
```diff
@@ -0,0 +1,515 @@
...
+        assert_eq!(std::time::Duration::from_secs(0), When::Today.until(now));
+        // If we bring it back before `SCHEDULED_HOUR` it should give us the duration:
+        let earlier = now.date().and_hms(SCHEDULED_HOUR - 1, 0, 0);
+        assert_eq!(
+            std::time::Duration::from_secs(3600),
+            When::Today.until(earlier)
+        );
+
+        // `Tomorrow` and `Reschedule` should differ only in their `reason()`
+        // 46670 is 12h57m10s (aka, the time from 15:02:10 to 04:00:00
```
badboy, line 411:
```suggestion
        // 46670s is 12h57m10s (aka, the time from 15:02:10 to 04:00:00
```

badboy (commented):
This looks good now! From the chat I take you took this to a test drive, did this require any last changes or was that purely FOG work now?

glean-core/src/scheduler.rs
chutten: I use `swap(0)` so that if someone extends the test the `submitter_count` has been reset.

glean-core/rlb/src/lib.rs
```diff
@@ -461,16 +465,15 @@ pub fn set_upload_enabled(enabled: bool) {
         let old_enabled = glean.is_upload_enabled();
         glean.set_upload_enabled(enabled);
 
-        // TODO: Cancel upload and any outstanding metrics ping scheduler
-        // task. Will happen on bug 1672951.
-
         if !old_enabled && enabled {
+            glean.start_metrics_ping_scheduler();
```
badboy, line 469: huh, interestingly we don't do this on Kotlin or Swift.

badboy: It's correct though, when re-enabled we should also let the MPS do its work again

badboy (approved):

PR #1601, opened 2021-04-29
Bug 1708204 - Apply new structure to book with no content changes / additions

brizental:
Hey @badboy it seems the table of contents stopped being created somehow. Is this normal? That is specifically for the "Adding new metrics" page.

I am confused because that pre-processor is not even defined in out book.toml 🤔 

Demo of the changes can be checked out at: https://innate-earth.surge.sh, redirections and all.


Dexterp37:
> Demo of the changes can be checked out at: https://innate-earth.surge.sh, redirections and all.

Hey @brizental the demo seems to break for me (it doesn't load). Can you double check?

brizental:
> Hey @brizental the demo seems to break for me (it doesn't load). Can you double check?

Hm, not sure. It loads for me. You can also download the branch and use `make docs`, if you can't load it. 

Note that it can be quite slow. The problem is not the demo itself, but the service I use to provide the demo.

bin/build-rust-docs.sh
```diff
@@ -32,6 +32,41 @@ rm -rf build/docs
 mkdir -p build/docs
 echo '<meta http-equiv=refresh content=0;url=book/index.html>' > build/docs/index.html
 
+# Add redirections for all pages moved on the restructuring done on Bug 1708204
+# Android Specific Information
```
Dexterp37, line 36: When are we planning on removing these? Are we even considering removing them? (just out of curiosity, no action required)

docs/user/README.md
```diff
@@ -38,25 +38,29 @@ file a bug in [Bugzilla in Data Platform & Tools :: Glean.js][gleanjs-bugs].
...
+### [API Reference](./reference/yaml/index.html)
+
+This section of the book contains reference pages for Glean’s user facing APIs.
+
+If you are looking for information a specific Glean API, this is the section you should check out.
 
-This sections lists all the metric types provided by Glean, with examples on how to define them
-and record data using them. Before diving into Glean's metric types details, don't forget to
-read the [Choosing a metric type](https://mozilla.github.io/glean/book/user/adding-new-metrics.html#choosing-a-metric-type) page.
+### [Implementation Specific Information](./impl/android/index.html)
```
Dexterp37, line 58: What about:

- Language binding-specific information?
- Platform-specific information? (might be a stretch, because this is not really a platform)

docs/user/user/metrics/adding-new-metrics.md
```diff
@@ -20,45 +20,45 @@ The following is a set of questions to ask about the data being collected to hel
...
+If you need to measure the total time spent doing a particular task, look to the [timespan metric](../../reference/metrics/timespan.html). Timespan metrics allow specifying the resolution they are collected at, and to stay [lean][lean-data], they should only be collected at the minimum resolution required to answer your question.
 Note that this metric should only be used to measure time on a single thread. If multiple overlapping timespans are measured for the same metric, an invalid state error is recorded.
 
-If you need to measure the relative occurrences of many timings, use a [timing distribution](metrics/timing_distribution.html). It builds a histogram of timing measurements, and is safe to record multiple concurrent timespans on different threads.
+If you need to measure the relative occurrences of many timings, use a [timing distribution]../..(/reference/metrics/timing_distribution.html). It builds a histogram of timing measurements, and is safe to record multiple concurrent timespans on different threads.
```
Dexterp37, line 55:
```suggestion
If you need to measure the relative occurrences of many timings, use a [timing distribution](../../reference/metrics/timing_distribution.html). It builds a histogram of timing measurements, and is safe to record multiple concurrent timespans on different threads.
```

docs/user/user/metrics/adding-new-metrics.md
```diff
@@ -20,45 +20,45 @@ The following is a set of questions to ask about the data being collected to hel
...
 
 For all of the metric types in this section that measure single values, it is especially important to consider how the lifetime of the value relates to the ping it is being sent in. Since these metrics don't perform any aggregation on the client side, when a ping containing the metric is submitted, it will contain only the "last known" value for the metric, potentially resulting in **data loss**.  There is further discussion of [metric lifetimes](#when-should-the-glean-sdk-automatically-clear-the-measurement) below.
 
 ### Are you counting things?
 
-If you want to know how many times something happened, use a [counter metric](metrics/counter.html).  If you are counting a group of related things, or you don't know what all of the things to count are at build time, use a [labeled counter metric](metrics/labeled_counters.html).
+If you want to know how many times something happened, use a [counter metric](../../reference/metrics/counter.html).  If you are counting a group of related things, or you don't know what all of the things to count are at build time, use a [labeled counter metric]../..(/reference/metrics/labeled_counters.html).
```
Dexterp37, line 39:
```suggestion
If you want to know how many times something happened, use a [counter metric](../../reference/metrics/counter.html).  If you are counting a group of related things, or you don't know what all of the things to count are at build time, use a [labeled counter metric](../../reference/metrics/labeled_counters.html).
```

docs/user/user/metrics/adding-new-metrics.md
```diff
@@ -20,45 +20,45 @@ The following is a set of questions to ask about the data being collected to hel
...
+> **Beware:** string metrics are exceedingly general, and you are probably best served by selecting the most specific metric for the job, since you'll get better error checking and richer analysis tools for free. For example, avoid storing a number in a string metric --- you probably want a [counter metric]../..(/reference/metrics/counter.html) instead.
 
-If you need to store multiple string values in a metric, use a [string list metric](metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
+If you need to store multiple string values in a metric, use a [string list metric]../..(/reference/metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
 
-<!-- TODO: If you do know the list of values you might store and you just want to record their presence, consider using a [labeled boolean](metrics/labeled_boolean.html) instead. -->
+<!-- TODO: If you do know the list of values you might store and you just want to record their presence, consider using a [labeled boolean]../..(/reference/metrics/labeled_boolean.html) instead. -->
```
Dexterp37, line 31:
```suggestion
<!-- TODO: If you do know the list of values you might store and you just want to record their presence, consider using a [labeled boolean](../../reference/metrics/labeled_boolean.html) instead. -->
```

docs/user/user/metrics/adding-new-metrics.md
```diff
@@ -20,45 +20,45 @@ The following is a set of questions to ask about the data being collected to hel
...
 
-> **Beware:** string metrics are exceedingly general, and you are probably best served by selecting the most specific metric for the job, since you'll get better error checking and richer analysis tools for free. For example, avoid storing a number in a string metric --- you probably want a [counter metric](metrics/counter.html) instead.
+> **Beware:** string metrics are exceedingly general, and you are probably best served by selecting the most specific metric for the job, since you'll get better error checking and richer analysis tools for free. For example, avoid storing a number in a string metric --- you probably want a [counter metric]../..(/reference/metrics/counter.html) instead.
 
-If you need to store multiple string values in a metric, use a [string list metric](metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
+If you need to store multiple string values in a metric, use a [string list metric]../..(/reference/metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
```
Dexterp37, line 29:
```suggestion
If you need to store multiple string values in a metric, use a [string list metric](../../reference/metrics/string_list.html). For example, you may want to record the list of other Mozilla products installed on the device.
```

Dexterp37 (changes requested):
Looks like a good start! I found a few broken things, see below

docs/user/user/metrics/adding-new-metrics.md
brizental: It's interesting that links checker doesn't break with this. Either mdbook is inteligent enough to compile correctly even when this is wrong or links checker is doing something different.

bin/build-rust-docs.sh
brizental: I would say we should remove them when this is at least a whole year in the past. This is completely a arbitrary timeline though.

Definitely nowhere in the near future.

docs/user/README.md
brizental: Hm, in retrospect I prefer Language Binding. For what it is worth, Glean.js is very much a language binding. That is a strech for Firefox Desktop, which should have it's own subsection here, but I think Firefox Desktop will always be stand out of pattern in all of the book, so that should be fine :)

Dexterp37 (approved):

docs/user/README.md
```diff
@@ -38,25 +38,29 @@ file a bug in [Bugzilla in Data Platform & Tools :: Glean.js][gleanjs-bugs].
 > if you want to use Glean.js and is missing some key Glean feature.
 ## Sections
 
-### [Using Glean](./user/index.html)
+### [User's Guides](./user/adding-glean-to-your-project.html)
```
chutten, line 41:
```suggestion
### [User Guides](./user/adding-glean-to-your-project.html)
```

docs/user/SUMMARY.md
```diff
@@ -1,35 +1,13 @@
...
-    - [Timespan](user/metrics/timespan.md)
-    - [Timing Distribution](user/metrics/timing_distribution.md)
-    - [Memory Distribution](user/metrics/memory_distribution.md)
-    - [UUID](user/metrics/uuid.md)
-    - [Datetime](user/metrics/datetime.md)
-    - [Event](user/metrics/event.md)
-    - [Custom Distribution](user/metrics/custom_distribution.md)
-    - [Quantity](user/metrics/quantity.md)
-    - [Rate](user/metrics/rate.md)
+# User's Guides
```
chutten, line 3:
```suggestion
# User Guides
```

bin/build-rust-docs.sh
```diff
@@ -32,6 +32,41 @@ rm -rf build/docs
...
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/quantity.html>' > build/docs/book/user/metrics/quantity.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/rate.html>' > build/docs/book/user/metrics/rate.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/string.html>' > build/docs/book/user/metrics/string.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/string_list.html>' > build/docs/book/user/metrics/string_list.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/timespan.html>' > build/docs/book/user/metrics/timespan.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/timing_distribution.html>' > build/docs/book/user/metrics/timing_distribution.html
+echo '<meta http-equiv=refresh content=0;url=../../reference/metrics/uuid.html>' > build/docs/book/user/metrics/uuid.html
+# YAML Format
+echo '<meta http-equiv=refresh content=0;url=../reference/yaml/index.html>' > build/docs/book/user/metric-parameters.html
+# Metrics User's Guides
```
chutten, line 65:
```suggestion
# Metrics User Guides
```

docs/user/README.md
```diff
@@ -38,25 +38,29 @@ file a bug in [Bugzilla in Data Platform & Tools :: Glean.js][gleanjs-bugs].
...
+If you are looking for information a specific Glean API, this is the section you should check out.
 
-This sections lists all the metric types provided by Glean, with examples on how to define them
-and record data using them. Before diving into Glean's metric types details, don't forget to
-read the [Choosing a metric type](https://mozilla.github.io/glean/book/user/adding-new-metrics.html#choosing-a-metric-type) page.
+### [Language Binding Specific Information](./language-bindings/android/index.html)
 
-### [Pings](./user/pings/index.html)
+This section contains guides and essays regarding specific usage's
+and possibilities in each of Glean's language bindings.
```
chutten, line 61:
```suggestion
This section contains guides and essays regarding specific usage information
and possibilities in each of Glean's language bindings.
```

docs/user/README.md
```diff
@@ -38,25 +38,29 @@ file a bug in [Bugzilla in Data Platform & Tools :: Glean.js][gleanjs-bugs].
...
 
-This section goes through what is a ping and how to define custom pings. A Glean client may provide
-off-the-shelf pings, such as the [`metrics`](https://mozilla.github.io/glean/book/user/pings/metrics.html)
-or [`baseline`](https://mozilla.github.io/glean/book/user/pings/baseline.html) pings. In this section,
-you will also find the descriptions and the schedules of each of these pings.
+Check out this section for important information on the language binding you are using.
```
chutten, line 63: This sounds as though this is a requirement ("important information"). Are we considering LB-specific knowledge required? (this is a philosophical question)

bin/build-rust-docs.sh
travis79, line 65: nit: "User's Guides" sounds a bit funny to me, maybe "Metrics User Guides" or rewording it somehow?

docs/user/README.md
travis79, line 41: Same nit as above: "User's Guides" just feels a little awkward due to the possesive form of user.

docs/user/SUMMARY.md
travis79, line 3: Same nit as above: possesive form of "User's" here seems awkward...

travis79 (commented):
From a first pass, I LOVE the new format.  I just had a nit about wording, but no strong opinions on that nit at all.

docs/user/README.md
```diff
@@ -38,25 +38,29 @@ file a bug in [Bugzilla in Data Platform & Tools :: Glean.js][gleanjs-bugs].
...
+### [API Reference](./reference/yaml/index.html)
+
+This section of the book contains reference pages for Glean’s user facing APIs.
+
+If you are looking for information a specific Glean API, this is the section you should check out.
 
-This sections lists all the metric types provided by Glean, with examples on how to define them
-and record data using them. Before diving into Glean's metric types details, don't forget to
-read the [Choosing a metric type](https://mozilla.github.io/glean/book/user/adding-new-metrics.html#choosing-a-metric-type) page.
+### [Language Binding Specific Information](./language-bindings/android/index.html)
```
chutten, line 58: In the ToC it says `Implementation Specific Information`. Which do we prefer?

I think `Language Binding Specific Information` is more correct (we use "implementation" to distinguish between SDK and .js above), but I'd be cool with using "Implementation" to refer to LBs across both clients and instead clarifying the above language in the front matter to be "client"

chutten (commented):
Love the structure, have some nitpicks.

docs/user/README.md
brizental: Does "User Guides" sound better? :)

docs/user/README.md
brizental: Hah, no strong opinion. I did not think long and hard before writing this page, I just wanted it to make sense with the new structure. I can remove "important", it only contains android information at this point so... :D

docs/user/README.md
brizental: Yeah, from the start I did always dislike the "Implementation Specific..." terminology. I made the change based on Alessio's review comment https://github.com/mozilla/glean/pull/1601#discussion_r623020279

Also the TOC was a mistake which I fixed already :D See 2960ff1

It is a fuzzy concept, what is a language binding and what is an implementation of Glean. Definitely something to add to the glossary... I would be fine with calling Glean.js a language binding. [Wikipedia tells me that is not quite correct though](https://en.wikipedia.org/wiki/Language_binding)... Why is terminology so hard?

badboy (approved):
The previous reviewers did most of the work already.
Kudos for all that work! 👍

PR #1603, opened 2021-04-29
Bug 1693486 - Implement new event extras API

badboy:
This still requires glean_parser changes before we can land it.
This is also a breaking change for the RLB:

  This removes support for the old API, passing a HashMap of ExtraKey -> String pairs.
  The new API will look like this for consumers:

    let mut extra = SomeExtra {
        key1: Some("1".into()),
        key2: Some("2".into)
    };
    metric.record(extra);

  glean_parser will generate for the new API only.

---

**Update 2021-05-05**

[API showcases for all languages](https://hackmd.io/5ABAR2FrTuGFoBmOYLH8vQ)

* I decided to use the `Extra` suffix for generated code, as written in the above language examples.
* I decided to use proper `CamelCasing` for the Kotlin classes (the previous `enums` used `camelCasing`)
* I consider the non-type variant deprecated in Kotlin, Swift & Python (should probably add that somewhere). The way forward is the new API. Rust will not even support the old API.
* The Python API can throw if used with wrong types.
* Documentation still (mostly) missing, needs #1625 to land

I think this is ready for a review, the [corresponding glean_parser PR](https://github.com/mozilla/glean_parser/pull/321) is also up to date.

---

Todo:

* [x] Decide on a better naming than the `Class` suffix
* [x] [glean_parser changes](https://github.com/mozilla/glean_parser/pull/321)
* [x] glean_parser release
* [x] Test this in FOG
* [x] Docs


badboy:
Now here's one interesting side note:

In Kotlin/Swift/Python we keep the old syntax working.
To activate the new syntax people would need to change their event definitions to the new field and list out the types.

For Rust (and FOG, thus C++/JS) however with the old definition syntax they only get the new API with only string fields.
They can switch to other fields using the new syntax.

badboy:
working through the changes in glean_parser I realized that we already nest below the `extra_keys`, so we don't need `event_properties` at all!
We can just add a new `type` attribute and default to string (and the old way) if it is missing:


```yaml
extra_keys:
  preference:
    type: string
    description: "This is key one"
  enabled:
    type: boolean
    description: "This is key two"
  swapped:
    type: quantity
    description: "This is key three"
```

glean-core/python/glean/_loader.py
```diff
@@ -90,6 +90,44 @@ def __getattr__(self, attr):
         )
 
 
+def event_extra_factory(name, argnames):
```
mdboom, line 93: Probably should be `_event_extra_factory` to make it "private" (by convention).  Also add type hints.
```suggestion
def _event_extra_factory(name: str, argnames: List[str]) -> EventExtraClass:
```

glean-core/python/glean/_loader.py
```diff
@@ -90,6 +90,44 @@ def __getattr__(self, attr):
...
+    Generate a new class, inheriting from `metrics.EventExtraClass`
+    and implementing the `to_ffi_extra` method,
+    which serializes expected attributes to pass over FFI.
+    """
+
+    def __init__(self, **kwargs):
+        for key, value in kwargs.items():
+            if key not in argnames:
+                raise TypeError(
+                    "Argument %s not valid for %s" % (key, self.__class__.__name__)
```
mdboom, line 104: Should probably use the more modern `f` strings:
```suggestion
                    f"Argument {key} not valid for {self.__class__.__name__}"
```

travis79 (approved):
Swift parts look good to me, as does the rest!

glean-core/python/glean/_loader.py
```diff
@@ -90,6 +90,44 @@ def __getattr__(self, attr):
...
+    def to_ffi_extra(self):
+        keys = []
+        values = []
+
+        for idx, name in enumerate(argnames):
+            attr = getattr(self, name, None)
+            if attr is not None:
+                keys.append(idx)
+                # Special-case needed for booleans to turn them lowercase (true/false)
+                if isinstance(attr, bool):
```
mdboom, line 117: Since this is Python, we probably want to be even more explicit about limiting the supported types.  For example, if the user sets one of these to a list, we'll get some Python-specific string representation of that out, which is probably not want we want.

We should *also* probably do that type checking in `__init__` so they are caught early.  And I say *also* because that still leaves open the opportunity to change the attributes after creation time, and those wouldn't be typechecked so we need to check them again here in `to_ffi_extra`.

Python dataclasses are sort of designed to address these problems, but they aren't available to us because they were introduced in Python 3.7 and we still support Python 3.6.

mdboom (dismissed):

glean-core/python/glean/_loader.py
badboy: Hah, that's on me for relying on the internet

glean-core/python/glean/_loader.py
badboy: Question is: what to do with wrong data types? Throw an exception?

badboy: I put this in an extra commit: [`5357018` (#1603)](https://github.com/mozilla/glean/pull/1603/commits/53570180c4882624c665802b54b51d3497adcd03)
I would probably squash that with the earlier commit, but for easier review iteration it's kept separate for now.

badboy:
```
Installed /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages/iso8601-0.1.14-py3.6.egg
Searching for glean_parser@ git+ssh://git@github.com/mozilla/glean_parser@new-event-api#egg=glean_parser
Reading https://pypi.org/simple/glean_parser/
Downloading https://files.pythonhosted.org/packages/f1/86/c507b33d8245797904c6b27dc3a89ed58e5544d57bee5b371aa186e2a2e4/glean_parser-3.2.0-py3-none-any.whl#sha256=f6eb2259707155e18ee97c5b9c868011c4eebfcfe2c98c596f3e82a9d02af4dd
Best match: glean-parser 3.2.0
Processing glean_parser-3.2.0-py3-none-any.whl
Installing glean_parser-3.2.0-py3-none-any.whl to /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages
writing requirements to /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages/glean_parser-3.2.0-py3.6.egg/EGG-INFO/requires.txt
Adding glean-parser 3.2.0 to easy-install.pth file
Installing glean_parser script to /home/circleci/project/glean-core/python/.venv3.6/bin
```

I actually don't know why it selects glean_parser from pypi when I clearly tell it to use the package from git.
Regardless, locally that works.

badboy:
> ```
> Installed /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages/iso8601-0.1.14-py3.6.egg
> Searching for glean_parser@ git+ssh://git@github.com/mozilla/glean_parser@new-event-api#egg=glean_parser
> Reading https://pypi.org/simple/glean_parser/
> Downloading https://files.pythonhosted.org/packages/f1/86/c507b33d8245797904c6b27dc3a89ed58e5544d57bee5b371aa186e2a2e4/glean_parser-3.2.0-py3-none-any.whl#sha256=f6eb2259707155e18ee97c5b9c868011c4eebfcfe2c98c596f3e82a9d02af4dd
> Best match: glean-parser 3.2.0
> Processing glean_parser-3.2.0-py3-none-any.whl
> Installing glean_parser-3.2.0-py3-none-any.whl to /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages
> writing requirements to /home/circleci/project/glean-core/python/.venv3.6/lib/python3.6/site-packages/glean_parser-3.2.0-py3.6.egg/EGG-INFO/requires.txt
> Adding glean-parser 3.2.0 to easy-install.pth file
> Installing glean_parser script to /home/circleci/project/glean-core/python/.venv3.6/bin
> ```
> 
> I actually don't know why it selects glean_parser from pypi when I clearly tell it to use the package from git.
> Regardless, locally that works.

Figured out most of it, but ugh, this is annoying. Let's just decide here and do  the glean_parser release.

badboy:
FOG WIP changeset: https://phabricator.services.mozilla.com/D114498

Dexterp37:
> 
> 
> working through the changes in glean_parser I realized that we already nest below the `extra_keys`, so we don't need `event_properties` at all!
> We can just add a new `type` attribute and default to string (and the old way) if it is missing:

And this is exactly the reason why we took that decision :)

glean-core/android/src/main/java/mozilla/telemetry/glean/private/EventMetricType.kt
```diff
@@ -54,6 +54,21 @@ enum class NoExtraKeys(
...
+     * 1. The list of extra key indices.
+     *    Unset keys will be skipped.
+     * 2. The list of extra values.
+     */
+    fun toFfiExtra(): Pair<IntArray, StringArray>
+}
+
 /**
  * This implements the developer facing API for recording events.
  *
```
Dexterp37, line 74: Leaving the comment here since I can't comment on the specific line. Can we add a [Deprecated](https://kotlinlang.org/api/latest/jvm/stdlib/kotlin/-deprecated/) annotation on line 122 ` fun record(extra: Map<ExtraKeysEnum, String>? = null) {`?

glean-core/python/setup.py
```diff
@@ -225,6 +225,9 @@ def run(self):
     ],
     description="Mozilla's Glean Telemetry SDK: The Machine that Goes 'Ping!'",
     install_requires=requirements,
+    dependency_links=[
+        "git+ssh://git@github.com/mozilla/glean_parser@new-event-api#egg=glean_parser-3.2.1"
```
Dexterp37, line 229: Just as a reminder to remove this and bump the version above

Dexterp37 (approved):
r+ on the Kotlin and Rust changes. I did not look to Python and Swift.

badboy:
I had to fix this up once more.
However thanks to our code generation nothing should change for consumers.
Please take another look at the last two commits (not the [TEST] one though).

* [`d17fcb1` (#1603)](https://github.com/mozilla/glean/pull/1603/commits/d17fcb1f70891ef059ba3fcf27a54902efbfc286)
* [`f418589` (#1603)](https://github.com/mozilla/glean/pull/1603/commits/f4185893876a94bc6afbb45eb81bc56ddb281442)

badboy:
The remaining glean_parser fixes are here: https://github.com/mozilla/glean_parser/pull/339

Dexterp37 (approved):
Kotlin changes look good to me. I also took a look at the related Kotlin glean_parser changes, and they are reasonable. Let's wait for the other reviews before merging.

glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
```diff
@@ -15,6 +15,27 @@ enum ClickKeys: Int32, ExtraKeys {
     }
 }
 
+struct ClickExtras: EventExtras {
```
travis79, line 18: nit: Can this have the same comment as the Kotlin definition? Even that brief explanation that this would usually be generated by glean_parser was useful to know that this was mainly a test construct.

glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
```diff
@@ -78,15 +99,20 @@ class EventMetricTypeTests: XCTestCase {
...
-        metric.record(extra: [.objectId: "buttonA", .other: "foo"])
+        // Newer API
+        metric.record(ClickExtras(objectId: "buttonA", other: "foo"))
+        // Some extra keys can be left undefined.
+        metric.record(ClickExtras(objectId: "buttonA"))
 
         /* SKIPPED: resetting system clock to return fixed time value */
 
-        metric.record(extra: [.objectId: "buttonB", .other: "bar"])
+        // Old API, this is available only because we manually implemented the enum.
```
travis79, line 109: _Is_ this the old API though?  It looks identical to the "new" API called just above on line 103...

glean-core/ios/sdk_generator.sh
```diff
@@ -142,7 +142,7 @@ VENVDIR="${SOURCE_ROOT}/.venv"
 # We need at least pip 20.3 for Big Sur support, see https://pip.pypa.io/en/stable/news/#id48
 # Latest pip is 21.0.1
 "${VENVDIR}"/bin/pip install "pip>=20.3"
-"${VENVDIR}"/bin/pip install --upgrade glean_parser==$GLEAN_PARSER_VERSION
+"${VENVDIR}"/bin/pip install --upgrade "git+ssh://git@github.com/mozilla/glean_parser@missing-event-imports#egg=glean_parser"
```
travis79, line 145: o_O  What's `@missing-event-imports`?

travis79 (approved):
I reviewed the Kotlin and Swift parts, just a couple of nits but nothing to block my approval on.

glean-core/ios/sdk_generator.sh
badboy: The branch! Will go away once I do another release and we use the version number again.

glean-core/ios/GleanTests/Metrics/EventMetricTests.swift
badboy: Sure, missed that when porting it over.

PR #1607, opened 2021-04-30
Bug 1698800 - Conform boolean and counter metric pages to new structure

brizental:
...And add Glean.js information to them.

[Rendered Boolean](https://innate-earth.surge.sh/reference/metrics/boolean.html)
[Rendered Counter](https://innate-earth.surge.sh/reference/metrics/counter.html)

Adding both metrics to the same PR, because they can then serve as a blueprint for all other metrics :)

### DIFF
--- a/docs/shared/tabs.js
+++ b/docs/shared/tabs.js
@@ -82,7 +82,12 @@ function switchAllTabs(language) {
             button.classList.add("tablinks");
             if (!tabcontent.innerHTML) {
                 button.classList.add("disabled");
-                button.title = `${tabcontent.dataset.lang} does not provide this API`;
+                if (!tabcontent.dataset.bug) {
+                    button.title = `${tabcontent.dataset.lang} does not provide this API`;
+                } else {
+                    button.title =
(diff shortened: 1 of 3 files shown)


docs/shared/tabs.js
```diff
@@ -82,7 +82,12 @@ function switchAllTabs(language) {
             button.classList.add("tablinks");
             if (!tabcontent.innerHTML) {
                 button.classList.add("disabled");
-                button.title = `${tabcontent.dataset.lang} does not provide this API`;
+                if (!tabcontent.dataset.bug) {
+                    button.title = `${tabcontent.dataset.lang} does not provide this API`;
+                } else {
+                    button.title =
+                        `${tabcontent.dataset.lang} does not provide this API yet. \nFollow https://bugzilla.mozilla.org/show_bug.cgi?id=${tabcontent.dataset.bug} for updates.`;
```
brizental, line 89: This is not great, because people can't blocki on the link... But I didn't want to rabbit hole on it too much.

Dexterp37: Yeha, maybe file a follow-up.

Dexterp37 (approved):

PR #1611, opened 2021-04-30
Bug 1698800 - Conform datetime metric reference page to new format

brizental:
...and Add Glean.js information to datetime metric page.

I also found a bug in the Swift example and fixed it (thanks @travis79 !).

[Rendered](https://innate-earth.surge.sh/reference/metrics/datetime.html).

Somethings I noticed:

- Java and Kotlin don't have the `testGetValue` API documented, was that on purpose? [That API seems to exist](https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/private/DatetimeMetricType.kt#L185). I can just include the missing examples in this PR, but wanted to double check first.
- Firefox Desktop `testGetValue` actually works like the `testGetValueAsString`. Worth filing a bug or was this on purpose?

### DIFF
--- a/docs/user/reference/metrics/datetime.md
+++ b/docs/user/reference/metrics/datetime.md
@@ -78,7 +78,15 @@ install::first_run.set(Some(custom_date)); // Records a custom datetime
 
 </div>
 
-<div data-lang="Javascript" class="tab"></div>
+<div data-lang="Javascript" class="tab">
+
+```js
+import * as install from "./path/to/generated/files/install.js";
+
+install.firstRun.set(); // Records "now"
+install.firstRun.set(new Date("March 25, 2019 00:00:00")); // Records a custom datetime
+```
+</div>
 
 <div data-lang="Firefox Desktop" class="tab">
 
@@ -146,7 +154,15 @@ assert_eq!(expected_date, metrics.install.first_run.test_get_value(None));
 
 </div>
 
-<div data-lang="Javascript" class="tab"></div>
+<div data-lang="Javascript" class="tab">
+
+```js
+import * as install from "./path/to/generated/files/install.js";
+
+const expectedDate = new Date("March 25, 2019 00:00:00");
+assert.deepStrictEqual(expectedDate, await install.firstRun.testGetValue());
+```
+</div>
 
 <div data-lang="Firefox Desktop" class="tab">
 
@@ -222,7 +238,14 @@ assert "2019-03-25-05:00" == metrics.install.first_run.test_get_value_as_str()
 
 <div data-lang="Rust" class="tab"></div>
 
-<div data-lang="Javascript" class="tab"></div>
+<div data-lang="Javascript" class="tab">
+
+```js
+import * as install from "./path/to/generated/files/install.js";
+
+assert.strictEqual("2019-03-25-05:00", await install.firstRun.testGetValueAsString());
+```
+</div>
 
 <div data-lang="Firefox Desktop" class="tab"></div>
 
--- a/docs/user/reference/metrics/datetime.md
+++ b/docs/user/reference/metrics/datetime.md
@@ -1,39 +1,18 @@
 # Datetime
 
-Datetimes are used to record an absolute date and time, for example the date and time that the application was first run.
+Datetimes are used to record an absolute date and time,
+for example the date and time that the application was first run.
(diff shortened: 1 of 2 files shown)


mdboom (approved):
This looks great.  It seems like the missing languages for testGetValue is just an oversight so we should put that in.

As for the Firefox Desktop question, I defer to @chutten .

chutten:
>     * Firefox Desktop `testGetValue` actually works like the `testGetValueAsString`. Worth filing a bug or was this on purpose?

Just because it was on purpose doesn't mean it doesn't need a bug : )

Hm, but to know for certain whether or not this was intentional we must ask the person responsible for implementing the API. Now [who was that](https://bugzilla.mozilla.org/show_bug.cgi?id=1673643)....  : D

At the very least we should consider this approach, and a bug could serve as documentation for why we choose to do it or not. I've filed https://bugzilla.mozilla.org/show_bug.cgi?id=1709147 for that purpose. Thank you for bringing this up!

docs/user/reference/metrics/datetime.md
```diff
@@ -124,149 +58,362 @@ import datetime
...
+
+<div data-lang="Firefox Desktop" class="tab"></div>
+
 {{#include ../../../shared/tab_footer.md}}
 
-## Limits
+### `testGetNumRecordedErrors`
 
-* None.
+Gets number of errors recorded for a given counter metric.
```
badboy, line 310:
```suggestion
Get number of errors recorded for a given datetime metric.
```

docs/user/reference/metrics/datetime.md
```diff
@@ -124,149 +58,362 @@ import datetime
...
 ```
 
 </div>
+{{#include ../../../shared/tab_footer.md}}
 
-<div data-lang="JS" class="tab">
+### `testGetValueAsString`
 
-> **Note**: JS APIs are only available in Firefox Desktop.
+Gets the recorded value for a given datetime metric an [ISO Date String](https://en.wikipedia.org/wiki/ISO_8601#Dates).
```
badboy, line 191:
```suggestion
Get the recorded value for a given datetime metric as an [ISO Date String](https://en.wikipedia.org/wiki/ISO_8601#Dates).
```

docs/user/reference/metrics/datetime.md
```diff
@@ -124,149 +58,362 @@ import datetime
...
-// Records "now"
-install::first_run.set(None);
-// Records a custom datetime
-let custom_date = FixedOffset::east(0).ymd(2019, 3, 25).and_hms(0, 0, 0);
-install::first_run.set(Some(custom_date));
+## Testing API
+
+### `testGetValue`
+
+Gets the recorded value for a given datetime metric as a language specific Date object.
```
badboy, line 121:
```suggestion
Get the recorded value for a given datetime metric as a language-specific Date object.
```

badboy (changes requested):

docs/user/reference/metrics/datetime.md
```diff
@@ -124,149 +58,362 @@ import datetime
...
+Gets number of errors recorded for a given counter metric.
 
-## Examples
+{{#include ../../../shared/tab_header.md}}
 
-* When did the user first run the application?
+<div data-lang="Kotlin" class="tab">
+
+```Kotlin
+import org.mozilla.yourApplication.GleanMetrics.Install
```
badboy, line 317:
```suggestion
import mozilla.telemetry.glean.testing.ErrorType
import org.mozilla.yourApplication.GleanMetrics.Install
```

docs/user/reference/metrics/datetime.md
```diff
@@ -124,149 +58,362 @@ import datetime
...
+assertEquals(0, Install.firstRun.testGetNumRecordedErrors(ErrorType.InvalidValue))
+```
 
-* `invalid_value`: Setting the date time to an invalid value.
+</div>
+
+<div data-lang="Java" class="tab">
+
+```Java
+import org.mozilla.yourApplication.GleanMetrics.Install;
```
badboy, line 327:
```suggestion
import mozilla.telemetry.glean.testing.ErrorType
import org.mozilla.yourApplication.GleanMetrics.Install;
```

PR #1613, opened 2021-05-04
Bug 1704606 - Rust: Don't return a result from `submit_ping`

badboy:
The `Result<bool>` value returned previously was not expressing what's
really happening anymore.
It was impossible to encounter an `Err` variant after recent changes.

Technically this is a breaking change for glean-core, but that library
is not consumed by any outsiders.


Dexterp37 (approved):
I think this is a sensible change. Even though this is not "really a breaking change", can we have a changelog entry? Just in case, to ease debugging in case future things go wrong somewhere! (no strong opinion, feel free to merge without)

Dexterp37:
(of course with CI passing)

CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v37.0.0...main)
 
+* Rust
+  * Don't return a result from `submit_ping`. The boolean result indicates whether a ping was submitted ([#1613](https://github.com/mozilla/glean/pull/1613))
```
chutten, line 6:
```suggestion
  * Don't return a result from `submit_ping`. The boolean return value indicates whether a ping was submitted ([#1613](https://github.com/mozilla/glean/pull/1613))
```

(the most minor of nits, and only because you're using "result" in the first sentence to mean `Result` and using the same word with a different meaning in the second sentence)

glean-core/ffi/src/lib.rs
```diff
@@ -301,9 +301,7 @@ pub extern "C" fn glean_set_upload_enabled(flag: u8) {
 #[no_mangle]
 pub extern "C" fn glean_submit_ping_by_name(ping_name: FfiStr, reason: FfiStr) -> u8 {
     with_glean(|glean| {
```
chutten, line 303: TIL we have a `with_glean` that unwraps `Result`.

glean-core/src/lib.rs
```diff
@@ -656,16 +656,24 @@ impl Glean {
                 if let Err(e) = ping_maker.store_ping(&self.get_data_path(), &ping) {
                     log::warn!("IO error while writing ping to file: {}. Enqueuing upload of what we have in memory.", e);
                     self.additional_metrics.io_errors.add(self, 1);
-                    let content = ::serde_json::to_string(&ping.content)?;
+                    let content = match ::serde_json::to_string(&ping.content) {
```
chutten, line 659: Why not `expect`?

glean-core/src/lib.rs
```diff
@@ -929,12 +938,12 @@ impl Glean {
...
-        if let Err(err) = self.internal_pings.baseline.submit(self, Some("inactive")) {
-            log::warn!("Failed to submit baseline ping on inactive: {}", err);
+        if !self.internal_pings.baseline.submit(self, Some("inactive")) {
+            log::info!("baseline ping not submitted on inactive");
         }
 
-        if let Err(err) = self.internal_pings.events.submit(self, Some("inactive")) {
-            log::warn!("Failed to submit events ping on inactive: {}", err);
+        if !self.internal_pings.events.submit(self, Some("inactive")) {
+            log::info!("events ping not submitted on inactive");
```
chutten, line 946: I wonder if it should still be louder than `info!`, though, since it should be impossible. Thoughts? Maybe it needs instrumentation/error stream.

glean-core/src/lib_unit_tests.rs
```diff
@@ -894,8 +894,7 @@ fn records_io_errors() {
     // Writing the ping file should fail.
     let submitted = glean.internal_pings.metrics.submit(&glean, None);
     // But the return value is still Ok(true) because we enqueue the ping anyway.
```
chutten, line 896: comment needs an update

chutten (approved):
Comments, but nothing that should block this from landing if you file follow-ups

glean-core/src/lib.rs
badboy: Not having an events ping on inactive is normal behavior though, if no events have been recorded nothing will be submitted.

PR #1614, opened 2021-05-04
Bug 1706729 - RLB: Dispatch setting the source tag.

badboy:
This avoids a race condition where there might not be a global Glean
yet.
But source tags are only needed for the ping assembly, and that's also
dispatched, so either happened before anyway or will happens afterwards.

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -743,22 +743,16 @@ pub fn set_log_pings(value: bool) {
 ///
 /// * `tags` - A vector of at most 5 valid HTTP header values. Individual
 ///   tags must match the regex: "[a-zA-Z0-9-]{1,20}".
-///
-/// # Returns
-///
-/// This will return `false` in case `value` contains invalid tags and `true`
-/// otherwise or if the tag is set before Glean is initialized.
-pub fn set_source_tags(tags: Vec<String>) -> bool {
+pub fn set_source_tags(tags: Vec<String>) {
     if was_initialize_called() {
-        with_glean_mut(|glean| glean.set_source_tags(tags))
+        crate::launch_with_glean_mut(|glean| {
+            glean.set_source_tags(tags);
+        });
     } else {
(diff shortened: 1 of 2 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -748,17 +748,16 @@ pub fn set_log_pings(value: bool) {
 ///
 /// This will return `false` in case `value` contains invalid tags and `true`
 /// otherwise or if the tag is set before Glean is initialized.
-pub fn set_source_tags(tags: Vec<String>) -> bool {
+pub fn set_source_tags(tags: Vec<String>) {
     if was_initialize_called() {
-        with_glean_mut(|glean| glean.set_source_tags(tags))
+        crate::launch_with_glean_mut(|glean| {
```
Dexterp37, line 753: Ah nice, that was an oversight. We already do this in Glean.js as well.

Dexterp37 (approved):

glean-core/rlb/src/lib.rs
```diff
@@ -748,17 +748,16 @@ pub fn set_log_pings(value: bool) {
 ///
 /// This will return `false` in case `value` contains invalid tags and `true`
```
chutten, line 749: Comment needs removal/update

glean-core/rlb/src/test.rs
```diff
@@ -849,6 +849,70 @@ fn setting_source_tags_before_initialization_should_not_crash() {
...
+        max_events: None,
+        delay_ping_lifetime_io: false,
+        channel: Some("testing".into()),
+        server_endpoint: Some("invalid-test-host".into()),
+        uploader: Some(Box::new(FakeUploader { sender: s })),
+    };
+
+    let _t = new_glean(Some(cfg), true);
+
+    // Attempt to set source tags before Glean is initialized.
```
chutten, line 896: ...but _after_ initialize was called, right? You should probably assert `was_initialize_called()`

chutten (changes requested):

chutten (approved):

PR #1616, opened 2021-05-04
Bug 1700942 - Gradle plugin: Provide forward-compatibility with Gradle 6.8

badboy:
In Gradle 6.x `getBuildConfigPackageName` was reaplced by `namespace`.
We want to be forward compatible, so we check that first or fallback to the old method.

See changes here:
https://cs.android.com/android-studio/platform/tools/base/+/c20e34494964552a5ce2787bb4c05016ac022378:build-system/gradle-core/src/main/java/com/android/build/gradle/tasks/GenerateBuildConfig.kt;dlc=fa4d27e36ad755e7f270d342c8cb6e367ea05149


mdboom (approved):
Nit: The rest of this file seems to have 4-space indents, but this uses 2.

r+wc

badboy:
> Nit: The rest of this file seems to have 4-space indents, but this uses 2.
> 
> r+wc

*shouts at editor config*

pocmo:
Thank you for fixing this :)

PR #1617, opened 2021-05-04
Bug 1708661 - Counter: Explicitly mention that `0` is not reported

badboy:
[doc only]


Dexterp37 (approved):
Good call in making this a bit more explicit.

PR #1618, opened 2021-05-04
Kotlin: Upgrade to Robolectric 4.5.1 to gain Java 11 support

badboy:
This requires disabling the GleanDebugActivity tests for now.
See https://bugzilla.mozilla.org/show_bug.cgi?id=1698471 for more.

---

Tests in #1613 failed again, this should finally solve that


Dexterp37 (approved):
Sad r+ (because we are rightfully ignoring the test)

badboy:
> Sad r+ (because we are rightfully ignoring the test)

I plan to have that re-enabled soon

PR #1629, opened 2021-05-10
Apply clippy suggestions (from Clippy beta)

badboy:


mdboom (approved):
Approving because clippy says so and the tests are passing :)

PR #1635, opened 2021-05-12
No bug - Remove unnecessary call to strip_label

brizental:
Nit I stumbled upon while implementing error recording for Glean.js.

### DIFF
--- a/glean-core/tests/labeled.rs
+++ b/glean-core/tests/labeled.rs
@@ -9,8 +9,8 @@ use serde_json::json;
 
 use glean_core::metrics::*;
 use glean_core::storage::StorageManager;
-use glean_core::{CommonMetricData, Lifetime};
 use glean_core::{test_get_num_recorded_errors, ErrorType};
+use glean_core::{CommonMetricData, Lifetime};
 
 #[test]
 fn can_create_labeled_counter_metric() {
--- a/glean-core/tests/labeled.rs
+++ b/glean-core/tests/labeled.rs
@@ -10,6 +10,7 @@ use serde_json::json;
 use glean_core::metrics::*;
 use glean_core::storage::StorageManager;
 use glean_core::{CommonMetricData, Lifetime};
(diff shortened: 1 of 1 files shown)


glean-core/src/error_recording.rs
```diff
@@ -69,7 +69,6 @@ fn get_error_metric_for_metric(meta: &CommonMetricData, error: ErrorType) -> Cou
     // Can't use meta.identifier here, since that might cause infinite recursion
     // if the label on this metric needs to report an error.
     let identifier = meta.base_identifier();
-    let name = strip_label(&identifier);
```
badboy, line 72: But isn't this potentially be called on a metric that was created as a labeled metric, and thus the name will include its label?

brizental: How is that possible? Do we use this pattern of creating a metric as labeled outside of the error_recording module? 

Users can't define metrics like that in the YAML files, glean_parser would complain about that.

I mean, I am ok with dropping this change. It's just confusing that we use base_indentifier, which gets the identifier without the label and immediatelly follow that with strip_label.

badboy: When getting a metric from a label on a labeled metric[^1], we use `name/label` as the internal name of that newly constructed metric. It's also how we store the data.
Errors on that (sub)metric are reported on the user-defined labeled metric, that's why we strip out the label during error recording.
See e.g. https://github.com/mozilla/glean/blob/dfd41e42c5d89d587d015cf6b328bcee00e900b5/glean-core/src/metrics/labeled.rs#L96-L99

Now the fact that no test failed means we don't actually have any test coverage for such errors.


[^1]: What a sentence ....

brizental: TIL! Thanks for the detailed explanation. This is great to know because I would have introduced a bug in Glean.js if I didn't add this, heh.

I dropped the first commit and added a test that fails with the previous change.

badboy (approved):

PR #1637, opened 2021-05-13
bug 1710990 - Don't take the Global Glean's lock while holding MPS's condvar lock

chutten:
The correct lock order is 1) Global Glean 2) MPS' condvar's cancelled lock.


chutten:
Try run: https://treeherder.mozilla.org/jobs?repo=try&tier=1%2C2%2C3&revision=7cc31b00b350d78599f767c8e852c730ef5f4812

May have to run it a bunch of times because it's an intermittent.

badboy (approved):
Looks good to me.
Gosh, locking code is hard.

PR #1647, opened 2021-05-21
Bug 1712322 - Add reference pages about debugging APIs

brizental:
Opening this as a draft before proceeding with the other pages. I want to get feedback on information layout and also some other things.

I only did the Debug View Tag page for now. Here is the rendered page for convenience: https://innate-earth.surge.sh/reference/debug/debugViewTag.html

Some thoughts:

1. I included the documentation for Kotlin and Java, but I am not sure the API really works there, because it is [marked as internal](https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L758).
2. I added a tooltip saying this API is not available for Firefox Desktop and environment variables should be preferred in that environment. I am not sure that is correct, though. I faintly remember @chutten saying environment variables work there, but I could be imagining.
3. Writing this page I noticed two things related to Glean.js:
   a. Glean.js provides [configuration options](https://github.com/mozilla/glean.js/blob/main/glean/src/core/debug_options.ts) to set the `debugViewTag` and other debug options. No other binding has that, I believe we should remove that from Glean.js for standardization sake.
   b. Glean.js provides [`unset`](https://github.com/mozilla/glean.js/blob/main/glean/src/core/glean.ts#L380) APIs for the debug APIs. At first I thought that made sense. After working on this I do not think so anymore, I think we should remove that API. WDYT @Dexterp37 ?


Dexterp37:
> 
> 
> Opening this as a draft before proceeding with the other pages. I want to get feedback on information layout and also some other things.
> 
> I only did the Debug View Tag page for now. Here is the rendered page for convenience: https://innate-earth.surge.sh/reference/debug/debugViewTag.html
> 
> Some thoughts:
> 
>     1. I included the documentation for Kotlin and Java, but I am not sure the API really works there, because it is [marked as internal](https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L758).

It shouldn't work and shouldn't probably be documented. Or it should be made public :-) The recommended way to use this on Android is by using the `GleanDebugActivity`.

>     3. Writing this page I noticed two things related to Glean.js:
>        a. Glean.js provides [configuration options](https://github.com/mozilla/glean.js/blob/main/glean/src/core/debug_options.ts) to set the `debugViewTag` and other debug options. No other binding has that, I believe we should remove that from Glean.js for standardization sake.

I remember that Python had something like that, but it doesn't anymore. I think that was removed when the environment variables were introduces. Unfortunately, we can't use environment variables with Glean.js. I see your point though, you're probably right in that the configuration options should be removed.

>        b. Glean.js provides [`unset`](https://github.com/mozilla/glean.js/blob/main/glean/src/core/glean.ts#L380) APIs for the debug APIs. At first I thought that made sense. After working on this I do not think so anymore, I think we should remove that API. WDYT @Dexterp37 ?

Yes, I think that should be removed.

chutten (commented):
(( Could we add an example to the Environment variable section? ))

Yes, in FOG environment variables work as described in the docs. We also have `about:glean` to set values at runtime (it is a hideous UI)

brizental:
> (( Could we add an example to the Environment variable section? ))

Done.

> Yes, in FOG environment variables work as described in the docs. We also have about:glean to set values at runtime (it is a hideous UI)

Added a little note about this in the Debugging page too, since I was there already :)

brizental:
This is ready for a review. I ended up also touching the "Debugging products using Glean" index page and since I was already there I also added a Javacript guide for debugging. The demo is updated with the latest code.

docs/user/SUMMARY.md
```diff
@@ -23,14 +23,18 @@
     - [Testing custom pings](user/pings/testing-custom-pings.md)
 - [Using the experiments API](user/experiments-api.md)
 - [Debugging products using Glean](user/debugging/index.md)
-    - [Glean Debug Ping View](user/debugging/debug-ping-view.md)
     - [Android](user/debugging/android.md)
```
brizental, line 26: I debated changing these to "Kotlin" and "Swift", for standardization sake. The guides are very Android and iOS specific, but then again so is everything else Kotlin / Swift related. Anyways, just a though, but it is weird to call the Kotlin bindings "Kotlin" / "Android" interchangeably and same for "Swift" / "iOS".

Dexterp37: I agree, it's indeed weird . I think that predates our idea / term of "language bindings"

docs/user/appendix/contribution-guidelines.md
```diff
@@ -170,6 +170,46 @@ And this is how those tabs will look like:
 
 {{#include ../../shared/tab_footer.md}}
 
+#### Tab tooltips
+
+When a tab is disabled i.e. does not have any content, it is possible to provide extra context
```
Dexterp37, line 175: When creating new content, what should we do? Just fill in for the existing tabs we know of or add them all and add a tooltip? May be worth making a call here (whatever that is) for consistency moving forward.

docs/user/reference/debug/index.md
```diff
@@ -0,0 +1,39 @@
...
+
+### [Source Tags](./logPings.md)
+
+This will tag all subsequent outgoing pings with a maximum of 5 comma-separated tags.
+
+### Send Pings
+
+_This feature is only available through the Android and iOS debug tools._
+
+This expects the name of a ping and forces its immediate collection and submission.
```
Dexterp37, line 39:
```suggestion
This expects the name of a ping and forces its immediate submission.
```

Because submission, according to [our glossary](https://mozilla.github.io/glean/book/appendix/glossary.html#submission) means collection + upload.

Dexterp37 (approved):
r+wc

docs/user/SUMMARY.md
brizental: I guess the naming proposal will help reach a consensus here, so I will ignore the weirdness for now.

docs/user/appendix/contribution-guidelines.md
brizental: There are guidelines about that in [this section](https://mozilla.github.io/glean/book/appendix/contribution-guidelines.html#tabs) of the contribution guidelines:

> Every tab section contains tabs for each Glean language binding, even that language binding does not provide the API in question. In this case, the tab div is still there without any inner HTML. When that is the case that tab will be rendered in a disabled state.

BUT I agree that this is not very clear 😄 

The idea is that tabs sections always have all the tabs, regardless if a given language binding doesn't provide the API being exemplified. When that is the case the tab should be added with no content inside and `tabs.js` will do the job of making sure emtpy tabs are shown in a disabled state.

By default, if you don't provide _any_ tooltip text, the default message is `Javascript does not provide this API`, if you provide a `data-bug` it shows `Javascript does not provide this API. Follow Bug XXXXX for updates.` and if you provide a `data-info` it will show whatever text you give it. The tooltip text itself depends on whether you think the default message is enough or not.

docs/shared/tabs.js
```diff
@@ -79,19 +79,24 @@ function switchAllTabs(language) {
...
                 button.classList.add("disabled");
-                if (!tabcontent.dataset.bug) {
-                    button.title = `${tabcontent.dataset.lang} does not provide this API`;
+                const tooltip = document.createElement("span");
+                tooltip.classList.add("tooltip");
+                if (tabcontent.dataset.bug) {
+                    tooltip.innerHTML +=
+                    `${tabcontent.dataset.lang} does not provide this API yet. \nFollow <a href="https://bugzilla.mozilla.org/show_bug.cgi?id=${tabcontent.dataset.bug}" target="_blank">Bug ${tabcontent.dataset.bug}</a> for updates.`;
+                } else if (tabcontent.dataset.info) {
+                    tooltip.innerHTML += `${tabcontent.dataset.info}`;
```
chutten, line 92: Ooooooh, these uses of `innerHTML` make my [webdev hackles rise with anxiety](https://developer.mozilla.org/en-US/docs/Web/API/Element/innerHTML#security_considerations). I do see how it's being used to provide rich output, but what if that output executes JS...

If we're going to continue using `innerHTML` please put a comment explaining why this is safe. Or an acceptable risk.

Otherwise I'd really prefer we come up with some other way to provide the richness these tooltips provide. I am really looking forward to being able to click links and select text in these new tooltips.

docs/user/appendix/changelog.md
```diff
@@ -0,0 +1,1092 @@
+# Unreleased changes
```
chutten, line 1: Is there a part of the release process that will update this file? Or are we expected to keep it up to date manually?

docs/user/reference/debug/index.md
```diff
@@ -0,0 +1,39 @@
...
+1. [Debugging Android applications using the Glean SDK](../../user/debugging/android.md)
+2. [Debugging iOS applications using the Glean SDK](../../user/debugging/ios.md)
+3. [Debugging Python applications using the Glean SDK](../../user/debugging/python.md)
+4. [Debugging Javascript applications using Glean.js](../../user/debugging/javascript.md)
+
+## Features
+
+The Glean SDK provides four debugging features.
+
+### [Log Pings](./logPings.md)
```
chutten, line 23: Links in headers: how discoverable are they? I know in the Firefox Source Docs they conflict with tables of contents and are really hard to discern since they are styled identically to unlinked headers.

docs/user/reference/debug/logPings.md
```diff
@@ -0,0 +1,99 @@
+# Log pings
+
+This flag causes all subsequent pings that are submitted, to also be echoed to the device's log.
```
chutten, line 3:
```suggestion
This flag causes all subsequent pings that are submitted to also be echoed to the product's log.
```

docs/user/user/debugging/index.md
```diff
@@ -1,51 +1,73 @@
 # Debugging products using the Glean SDK
 
-## Platform-specific debugging
+Glean provide a few debugging features to assist on debugging a product using Glean.
```
chutten, line 3:
```suggestion
Glean provides a few debugging features to assist with debugging a product using Glean.
```

chutten (commented):
I'm hesitant to approve this with the `innerHTML` stuff unresolved, sorry. I do think the structure and content is vastly improved and I reiterate that I'm looking forward to the new-style tooltips that I can click on and select text from.

docs/user/appendix/changelog.md
brizental: Hm, well actually this file should not even be here. This is me using `git add --all` again... Good catch though! (Changelog files are copied to the `changelog/` folder everytime we run `made docs`.)

docs/user/reference/debug/index.md
brizental: Links with headers turn blue in  `mdbook`. We use then quite a lot in this book https://mozilla.github.io/glean/book/index.html#sections.

docs/shared/tabs.js
brizental: It is possible for me to change this, but then I would need to parse any string passed to `data-info` and actually create the DOM elements which would be _very_ complex. Since the Glean book is a static website I don't see the issue in using `innerHTML` here. From the MDN link:

> there is still a security risk whenever you use innerHTML to set strings over which you have no control.

In this case we do have control as we are the ones defining which HTML string is passed to `innerHTML` and I don't see we ever changing this to actually receive dynamic user input. WDYT?

chutten: An inline comment to that effect would be very nice and would satisfy my anxiety, yes!

chutten (approved):

PR #1651, opened 2021-05-26
Bug 1683171 test_get_value assert num_errors == 0

chutten:
Presently left undone is the task of replacing using Language Binding-specific panics instead of `panic!`. Is there any place in specific I should perform that setup in the LBs?

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v38.0.1...main)
 
+* General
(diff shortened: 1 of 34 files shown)


chutten:
In trying to get this to actually prove out its usefulness on Python, I've found that exceptions raised in the platform panic callback do not propagate through or around Rust, making it impossible to use this to craft an ergonomic solution after all.

Therefore we either remove platform_panic and just `panic!` on these errors (with all the ergo failures that entails) or we explore changes to `test_get_value` on _every single metric type_.

PR #1661, opened 2021-06-02
Bug 1686689 - Add Glean.js changelog to changelog appendix

brizental:


mdboom (commented):
Side question for @badboy: Do you think we'll run into any issues in some context making a network request during the doc build like this?  I think because it's for *docs* and not code we should be ok, but want to confirm...

badboy:
> Side question for @badboy: Do you think we'll run into any issues in some context making a network request during the doc build like this? I think because it's for _docs_ and not code we should be ok, but want to confirm...

I don't think so, it might break if anyone ever works offline and tries to build the docs, but that's it.
On our CI we don't restrict access. It's HTTPS on a repo we control. Security wise I don't think we're worse off here.
GitHub _might_ ratelimit that endpoint, but I don't know if that happens.

mdboom (approved):

PR #1663, opened 2021-06-03
Bug 1711447 - Build for all iOS targets and bundle an xcframework

badboy:
Preparing for the xcframework build, first test run!

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -731,6 +731,19 @@ jobs:
             ./ghr -replace "${CIRCLE_TAG}" Glean.framework.zip
             ./ghr -replace "${CIRCLE_TAG}" Glean.xcframework.zip
 
+  glean-swift release:
+    docker:
+      - image: circleci/rust:latest
+    steps:
+      - attach_workspace:
+          at: .
+      - run:
+          name: Release glean-swift
+          command: |
+              git config --global user.email "jrediger@mozilla.com"
+              git config --global user.name "Glean automated release"
+              ./bin/publish-glean-swift.sh "${CIRCLE_TAG}"
+
   ###########################################################################
   # Python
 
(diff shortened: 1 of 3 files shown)


badboy:
Just under 3 minutes extra build time isn't too bad IMO, and it only happens on release.

badboy:
We can land this and I send a followup for the remaining work to push to [glean-swift](https://github.com/mozilla/glean-swift)

badboy:
Decided to include the glean-swift release part already

.circleci/config.yml
```diff
@@ -701,7 +717,7 @@ jobs:
           path: device_logs.zip
           destination: device_logs.zip
 
-  Carthage release:
+  iOS Framework release:
     macos:
       xcode: "12.4.0"
```
travis79, line 722: Can/should we update to 12.5?  Application Services just updated and Firefox-iOS is already on 12.5...

bin/publish-glean-swift.sh
```diff
@@ -0,0 +1,16 @@
...
+set -e
+
+# Strip any `v` prefix from `vX.Y.Z`
+NEW_VERSION=$(echo "$1" | sed 's/^v//')
+
+git clone https://github.com/mozilla/glean-swift glean-swift
+cd glean-swift
+bin/update.sh "$NEW_VERSION"
+git push -q https://${GLEAN_SWIFT_GITHUB_TOKEN}@github.com/mozilla/glean-swift.git main
+git push -q https://${GLEAN_SWIFT_GITHUB_TOKEN}@github.com/mozilla/glean-swift.git "$NEW_VERSION"
```
travis79, line 16: Nice!  Do you think we will need to keep a separate repo for publishing forever?

travis79 (approved):
Just a couple of question, nothing to block landing on.

.circleci/config.yml
badboy: Actually yeah, I thought I built the previous one on Xcode 12.4 for Focus, but actually it was 12.5 too

PR #1668, opened 2021-06-04
Fix remaining issues for release

badboy:
I manually fixed up the v39.0.1 release.
The steps I took went into this patch now.
Yes, it was that easy.


Dexterp37 (approved):
🤦

PR #1670, opened 2021-06-07
Bug 1714932 - Document how to add Glean test dependencies to Android projects

badboy:
[doc only]


Dexterp37 (approved):
Take that, review time!

PR #1673, opened 2021-06-08
Bug 1710424 - Document more precisely how the UUID type behaves

badboy:
Generate only v4 UUIDs.
Accepts any UUID in the typical 8-4-4-4-12 format.

[doc only]

### DIFF
--- a/docs/user/reference/metrics/uuid.md
+++ b/docs/user/reference/metrics/uuid.md
@@ -6,7 +6,7 @@ UUIDs metrics are used to record values that uniquely identify some entity, such
 
 ### `generateAndSet`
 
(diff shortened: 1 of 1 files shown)


docs/user/reference/metrics/uuid.md
```diff
@@ -100,6 +100,7 @@ Glean.user.clientId.generateAndSet();
 ### `set`
 
 Sets a UUID metric to a specific value.
+Accepts any UUID version.
```
brizental, line 103: We accept more than any UUID version... But I agree that we should not advertise that too much :)

brizental (approved):

docs/user/reference/metrics/uuid.md
badboy: well, technically anything not RFC'd is still _a version_

PR #1683, opened 2021-06-11
Bug 1712071 - Update structure of Rate metric reference page 

badboy:
[doc only]


Dexterp37 (approved):
r+ with the links fixed

PR #1684, opened 2021-06-11
Bug 1713574 - Add documentation on JavaScript plugins and the PingEncryptionPlugin

brizental:


docs/user/language-bindings/javascript/plugins/index.md
```diff
@@ -0,0 +1,37 @@
...
+  uploadStatus,
+  {
+    plugins: [
+      new HypotheticalPlugin("with", "hypothetical", "arguments")
+    ]
+  }
+);
+```
+
+Plugins attach to specific internal events on the SDK and can modify its behavior.
```
Dexterp37, line 22: Can you list out the currently supported events?

Dexterp37 (approved):
r+wc

docs/user/language-bindings/javascript/plugins/index.md
brizental: I though about that, but that is internal detail for now. Since we don't let people write their own plugins, what is the point in listing events?

Dexterp37: Ok, let's move on for now

brizental: (FWIW, there is a bug to document events on the Glean.js developer docs. It's [Bug 1698089](https://bugzilla.mozilla.org/show_bug.cgi?id=1698089) :D)

PR #1685, opened 2021-06-11
Fix wording

badboy:
[doc only]


brizental (approved):
Thanks for caching this :)

PR #1687, opened 2021-06-14
Update to latest mdbook-dtmo

badboy:
[doc only]


badboy:
Merging without review. Only a CI tool update.

PR #1691, opened 2021-06-14
Bug 1714578 - Pings testing api docs

Dexterp37:
[docs only]


docs/user/user/pings/testing-custom-pings.md
```diff
@@ -1,300 +1,11 @@
-# Unit testing Glean custom pings for Android
+# Testing custom pings
 
-Applications defining [custom pings](custom.md) can use use the strategy defined in this document to test these pings in unit tests.
+Applications defining [custom pings](custom.md) can use use the [ping testing API](../../reference/pings/index.md#testing-api) test these pings in unit tests.
```
brizental, line 3:
```suggestion
Applications defining [custom pings](custom.md) can use the [ping testing API](../../reference/pings/index.md#testing-api) to test these pings in unit tests.
```

docs/user/user/pings/testing-custom-pings.md
```diff
@@ -1,300 +1,11 @@
-# Unit testing Glean custom pings for Android
+# Testing custom pings
```
brizental, line 1: It could be interesting to add a paragraph or two on the importance of testing custom pings. That can be done as a follow up though.

docs/user/user/pings/testing-custom-pings.md
```diff
@@ -1,300 +1,11 @@
...
-    scheduler.add_some_data()
-    scheduler.schedule_ping()
-```
-
-</div>
-
-{{#include ../../../shared/tab_footer.md}}
+1. Triggering the code path that accumulates/records the data.
+2. Defining a callback validation function using the [ping testing API](../../reference/pings/index.md#testbeforenextsubmit).
+3. Finally submitting the ping using the [`submit` API](../../reference/pings/index.md#submit).
```
brizental, line 11:
```suggestion
3. Finally triggering the code path that submits the custom ping or submitting the ping using the [`submit` API](../../reference/pings/index.md#submit).
```
?

docs/user/reference/pings/index.md
```diff
@@ -1,6 +1,9 @@
-# Submitting a custom ping
+# Pings
 
-To collect and queue a custom ping for eventual uploading, call the `submit` method on the `PingType` object that the Glean SDK generated for your ping.
+## Submission API
+
+### `submit`
+Collect and queue a custom ping for eventual uploading.
```
brizental, line 6: For the sake of redundancy, can we add an info or warning box saying that internal pings don't need to be submitted? Something like "HEY, THIS REALLY ONLY MATTERS IF YOU HAVE CUSTOM PINGS!!!! GO AWAY OTHERWISE!!11!!!", but nicer and then link to the internal pings page, ofc.

docs/user/reference/pings/index.md
```diff
@@ -83,3 +86,178 @@ GleanPings.search.submit("performed");
...
+let validatorRun = false;
+const p = pings.search.testBeforeNextSubmit(async reason => {
+  assert.strictEqual(reason, "performed");
+  assert.strictEqual(await search.defaultEngine.testGetValue(), 5);
+  validatorRun = true;
+});
+
+// Submit the ping.
+pings.search.submit("performed");
+// Wait for the validation to finish.
```
brizental, line 252: I could be asking for this due to being traumatized by this API, but I think it might be worth clearly stating the limitations of this API on JavaScript. That can be a follow up no problem. This could even be a whole page on the JavaScript specific section honestly........................................ Yeah, definitely I am asking out of _trauma_.

docs/user/reference/pings/index.md
```diff
@@ -83,3 +86,178 @@ GleanPings.search.submit("performed");
...
+await p;
```
brizental, line 253: Maybe wrap this is a `assert.doesNotThrow(await p)` so it's clear that IT'S ULTRA IMPORTANT

brizental (changes requested):
I noticed the `submit` section is lacking Java examples, would you mind filing a follow-up for that or just doing it in this PR?

On the examples you create this `validatorRun` boolean that is not really necessary it seems, or am I missing something? If not, can we remove that from the examples so as to not add more to it than necessary?

docs/user/user/pings/testing-custom-pings.md
Dexterp37: I think that's better material for https://mozilla.github.io/glean/book/user/pings/index.html

docs/user/reference/pings/index.md
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1716540

docs/user/user/pings/testing-custom-pings.md
brizental: Agree!

brizental:
> The boolean is used to make sure that the validator function actually runs. We have no other way to be completely sure of that in our APIs. We can't distinguish between "the callback run and detected no error" with the "callback did not run at all" (we do that in our tests, too)

In this case, why does the Rust example not have that?

docs/user/user/pings/testing-custom-pings.md
brizental: Would you mind filing a bug for it? (Sorry if you already did it)

docs/user/user/pings/testing-custom-pings.md
```diff
@@ -1,300 +1,11 @@
...
-    scheduler.add_some_data()
-    scheduler.schedule_ping()
-```
-
-</div>
-
-{{#include ../../../shared/tab_footer.md}}
+1. Triggering the code path that accumulates/records the data.
+2. Defining a callback validation function using the [ping testing API](../../reference/pings/index.md#testbeforenextsubmit).
+3. Finally triggering the code path that submits the custom ping or submitting the ping using the [`submit` API](../../reference/pings/index.md#submit).
```
brizental, line 11: IMHO, we could go further and maybe go through a _real world_ case of ping testing in this page. It is kinda dry at the moment i.e. does not add too much to the docs. Good enough for this PR though, would you mind filing a follow-up bug though?

brizental (approved):
r+wc

docs/user/user/pings/testing-custom-pings.md
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1716800

docs/user/user/pings/testing-custom-pings.md
Dexterp37: https://bugzilla.mozilla.org/show_bug.cgi?id=1716801

badboy (commented):
> > The boolean is used to make sure that the validator function actually runs. We have no other way to be completely sure of that in our APIs. We can't distinguish between "the callback run and detected no error" with the "callback did not run at all" (we do that in our tests, too)
> 
> In this case, why does the Rust example not have that?

Probably because it's so much more annoying to do in Rust:

```rust
let callback_was_called = Arc::new(AtomicBool::new(false));
let callback_was_called_2 = Arc::clone(&callback_was_called);

ping.test_before_next_submit(move |reason| {
    assert_eq!(None, reason);
    assert_eq!(1, metric.test_get_value(None).unwrap());
    callback_was_called_2.store(true, Ordering::SeqCst);
});

ping.submit(None);
assert!(callback_was_called.load(Ordering::SeqCst));
```

I'd say if you directly call `ping.submit()` in your test there's no need for `callback_was_called` (because users should trust us that we call it from `submit`).
It's only when you invoke `submit` behind the scenes and custom logic it might be a good idea I guess.

PR #1693, opened 2021-06-17
Use the correct type for the `seq` field

Dexterp37:
This definition is only useful for documentation purposes, because [we are not using it internally](https://searchfox.org/glean/rev/42e5778f4aaf240330eae1ce5bebfa35e19642a0/glean-core/src/ping/mod.rs#71). We are not even using a field with the same name :-) We're using `sequence` internally, instead of `seq`.

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -243,12 +243,14 @@ glean.internal.metrics:
     expires: never
 
   seq:
-    type: string
+    type: counter
     lifetime: user
     send_in_pings:
       - glean_internal_info
     description: |
-      A running counter of the number of times pings of this type have been sent
+      A running counter of the number of times pings of this type have been sent.
+      This metric definition is only used for documentation purposed: internally,
+      Glean instantiates the metric manually and calls it `sequence`.
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1556964
     data_reviews:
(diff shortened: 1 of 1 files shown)


Dexterp37:
@fbertsch the type for this metric was mistakenly marked as `string`, but it really was a counter. This should have no impact on the pipeline (because internally we don't use this definition in the Glean SDKs), but this is being used downstream for at least the Glean Dictionary (and maybe the schema generator)? Would fixing this break things downstream?

fbertsch (approved):
This should be fine. [Ping info](https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/main/schemas/glean/glean/glean.1.schema.json#L536-L538) and Client Info are both static entries from MPS, so this won't affect the downstream schema generation.

glean-core/metrics.yaml
```diff
@@ -243,12 +243,16 @@ glean.internal.metrics:
...
-    type: string
+    type: counter
     lifetime: user
     send_in_pings:
       - glean_internal_info
     description: |
-      A running counter of the number of times pings of this type have been sent
+      A running counter of the number of times pings of this type have been
+      sent.
+      This metric definition is only used for documentation purposed:
```
mdboom, line 253:
```suggestion
      This metric definition is only used for documentation purposes:
```

mdboom (approved):
r+ (with minor change)

PR #1697, opened 2021-06-21
Bug 1717402 - Extend `invalid_timezone_offset` metric until the end of the year

badboy:
This is an internal metric and thus the code already uses a never-ending
metric. All prior versions will continue to collect this metric too,
even beyond the initial date.
This is therefore mostly a documentation update.

[doc only]

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -634,11 +634,13 @@ glean.time:
       - metrics
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1611770
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1717402#c1
     data_reviews:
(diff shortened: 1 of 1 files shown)


glean-core/metrics.yaml
```diff
@@ -634,11 +634,13 @@ glean.time:
       - metrics
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1611770
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1717402#c1
     data_reviews:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1611770#c9
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1717402#c3
```
badboy, line 640: Should link to the proper response once I have it.

PR #1702, opened 2021-06-25
Bug 1717043 - Provide a way to iterate all error types

badboy:


glean-core/src/error_recording.rs
```diff
@@ -48,6 +48,25 @@ impl ErrorType {
...
+    ///
+    /// ```
+    /// # use glean_core::ErrorType;
+    /// let errors = ErrorType::iter();
+    /// let all_errors = errors.collect::<Vec<_>>();
+    /// assert_eq!(4, all_errors.len());
+    /// ```
+    pub fn iter() -> impl Iterator<Item=Self> {
+        // N.B.: This has no compile-time guarantees that it is complete.
+        // New `ErrorType` variants will need to be added manually.
```
chutten, line 62: Should we also comment on the enum to ensure they stay synced?

chutten (approved):
LGTM, one optional note

PR #1705, opened 2021-06-25
Bug 1714365 - Update JS docs for testGetNumRecordedErrors to use ErrorType enum

brizental:
[doc only]


badboy (approved):
Beta failures unrelated, fixed in main already.

PR #1714, opened 2021-07-13
Bug 1718040 - Document database format internals

badboy:
[doc only]

### DIFF
--- a/docs/dev/core/internal/database.md
+++ b/docs/dev/core/internal/database.md
@@ -1,0 +1,58 @@
+# Database format
+
+The Glean SDK stores all recorded data in a database for persistence.
+This data is read, written and transformed by the core implementation.
+
+Some internal metrics are stored similar to user-defined metrics,
+but encode additional implementation-defined information in the key or value of the entry.
+
+We guarantee backwards-compatibility of already stored data.
+If necessary an old database will be converted to the new format.
+
+## Database stores
+
+The Glean SDK will use one store per metric lifetime:
+`user`, `application` and `ping`.
+This allows to separately read and clear metrics based on their respective lifetimes.
+
+## Key
+
+The key of a database entry uniquely identifies the stored metric data.
+It encodes additional information about the stored data in the key name using special characters.
+The full list of special characters in use is:
+
+`. # / +`
+
+These characters cannot be used in a user-defined ping name, metric category,  metric name or label.
+
+A key will usually look like:
+
+```
+ping#category.name[/label]
+```
+
+where:
+
+| Field | Description | Allowed characters | Maximum length\* | Note |
+| ----- | ----------- | ------ | ----- | ----- |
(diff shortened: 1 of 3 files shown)


docs/dev/core/internal/database.md
```diff
@@ -0,0 +1,51 @@
...
+The full list of special characters in use is:
+
+`. # / +`
+
+These characters cannot be used in a user-defined ping name, metric category,  metric name or label.
+
+A key will usually look like:
+
+```
+[type+]ping#category.name[/label]
```
brizental, line 31: The `+` character used in Qt/QML is not to separate type information. It it not tied to any Glean specific context. Glean.js saves data in JSON format and it saves it more or less like it would be in the ping payload. In Qt/QML we use the `+` character to signal that we are going one step further in a (non-existent) JSON object.

An example will make this clearer I think, so take this hypothetical data stored in JSON format. 
(This is exactly how we would store a user lifetime datetime metric in the webext storage.)

```json
{
  "userLifetimeMetrics": {
    "my_ping": {
      "datetime": {
        "glean.start_time": {
          "timeUnit": "minute",
          "timezone": -60,
          "date": "2021-07-14T13:07:10.723Z"
        }
      }
    }
  }
}
```

In Qt/QML that will be stored its own table, so the `userLifetimeMetrics` index is ignored. Otherwise, this is how that same data would look on SQLite:

```
| key                                          | value                      |
| -------------------------------------------- | -------------------------- |
| "my_ping+datetime+glean.start_time+timeUnit" | "minute"                   |
| "my_ping+datetime+glean.start_time+timezone" | -60                        |
| "my_ping+datetime+glean.start_time+date"     | "2021-07-14T13:07:10.723Z" |
```

The Qt/QML store then transforms this into the same object as the JSON storage would, before returning. To make it opaque.

brizental: Since database is very different in Glean.js, I think it would make more sense to leave that information out of this page. Maybe, add a note saying that `+` is a character used in the Glean.js storage implementation for Qt/QML and thus should never be allowed for user defined metrics or something. 

Otherwise, it makes more sense IMO to actually add a database internal doc over on the Glean.js developer documentation that goes deeper on the subject.

Filed a bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1720476

brizental (changes requested):

docs/dev/core/internal/database.md
badboy: Ah, then I misunderstood the explanation yesterday. I had a more simpl note before that `+` is only used in Glean.js, so I'll bring that back in favor of the more detailed, but completely different notes.

brizental: Yeah, this is very hard to explain with plain text :D That sounds good.

brizental (approved):

PR #1715, opened 2021-07-14
Bug 1719315 - Add documentation about the URL metric type

brizental:
Blocked until a release of Glean.js comes out with https://github.com/mozilla/glean.js/pull/529.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 237 utf-8
+personal_ws-1.1 en 238 utf-8
 AAR
 AARs
 ABI
@@ -226,6 +226,7 @@ tooltips
 transpiler
 travis
 und
+unencoded
 unhandled
 uniffi
 uploader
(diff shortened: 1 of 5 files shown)


brizental:
[v0.17.0](https://github.com/mozilla/glean.js/releases/tag/v0.17.0) is out with the new type :)

travis79 (commented):
I'm good with everything here, but I'm withholding approval in order to put on my data-steward hat for a moment and ask about the data sensitivity category for this particular metric type.  I feel that calling out more explicitly here in our docs that URL metrics will almost always be Category 3 and above would be a good way to reduce overuse and ensure that it gets data-reviewed properly.  What do you think?

brizental:
> I'm good with everything here, but I'm withholding approval in order to put on my data-steward hat for a moment and ask about the data sensitivity category for this particular metric type. I feel that calling out more explicitly here in our docs that URL metrics will almost always be Category 3 and above would be a good way to reduce overuse and ensure that it gets data-reviewed properly. What do you think?

Agreed 100%. Let me make the change :)

brizental:
I added a note, let me know if you think it is too hidden. I also filed a bug to maybe reject that on glean_parser directly https://bugzilla.mozilla.org/show_bug.cgi?id=1721358.

travis79 (approved):
Thanks for adding the data sensitivity note, this LGTM!

PR #1716, opened 2021-07-15
bug 1694496 - Add FOG docs for rate [doc only]

chutten:


docs/user/reference/metrics/rate.md
```diff
@@ -90,7 +111,28 @@ assert_eq!((1, 1), network::http_connection_error.test_get_value(None).unwrap())
...
 <div data-lang="Javascript" class="tab"></div>
-<div data-lang="Firefox Desktop" class="tab"></div>
+<div data-lang="Firefox Desktop" class="tab">
+
+  **C++**
+
+  ```cpp
+  #include "mozilla/glean/GleanMetrics.h"
+
+  auto pair = mozilla::glean::network::http_connection_error.TestGetValue().unwrap();
```
brizental, line 121: What is the error thrown when an error was recorded in C++? Worth adding a comment like you do [here](https://github.com/mozilla/glean/pull/1716/files#diff-801e14f2e0aa0683adbf04426b847d42a9a632356fd74f4a341fd21fb6783411R129) even in the case that no error is ever thrown.

brizental (approved):
r+wc

docs/user/reference/metrics/rate.md
chutten: It'll crash/explode on the `unwrap()` like Rust does. I wasn't sure whether to say anything given that `unwrap` is so idiomatic.

brizental: Hm, yeah I missed that `unwrap` there. Yeah I agree, ignore my previous comment in this case :)

Issue #1717, opened 2021-07-15
Run taskgraph CI with Python 3


ahal:
We'd like to drop Python 2 support in the [taskgraph](https://hg.mozilla.org/ci/taskgraph) module. So we need to make sure consumers are all running it with Python 3.

badboy:
Fixed with ebfd84372b52f704deba205002619dc3eba31a17

PR #1718, opened 2021-07-15
Run taskgraph generation with Python 3

ahal:
Fixes issue #1717, see also RELENG-617

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -243,8 +243,8 @@ tasks:
                                     - bash
                                     - -cx
                                     - >
-                                      PIP_IGNORE_INSTALLED=0 pip install --user /builds/worker/checkouts/taskgraph &&
-                                      PIP_IGNORE_INSTALLED=0 pip install --user arrow taskcluster pyyaml &&
+                                      PIP_IGNORE_INSTALLED=0 pip3 install --user /builds/worker/checkouts/taskgraph &&
+                                      PIP_IGNORE_INSTALLED=0 pip3 install --user arrow taskcluster pyyaml &&
                                       ln -s /builds/worker/artifacts artifacts &&
                                       ~/.local/bin/taskgraph decision
                                       --pushlog-id='0'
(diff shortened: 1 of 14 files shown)


ahal:
This still needs some work, I don't see an option to convert it to a draft PR.

ahal:
This repo installs the `arrow` and `pyyaml` packages, and I guess one of them started pulling in some dependencies that need to be compiled after we switched to Python 3. So I think we'll need to release new `decision` and `decision-mobile` images here which include `python3-dev`.

badboy:
Let me know if there's something we can help with.
Otherwise leave a comment when this is ready for review & merge, I'll get to it then. (and don't worry about this not being a Draft, it's fine)

ahal:
Tracking image updates here:
https://bugzilla.mozilla.org/show_bug.cgi?id=1721187

ahal:
Made progress here, decision task is passing. Though it doesn't appear to be generating the docker contexts required by the `build-docker-image-linux` task. So far this has always "just worked", so not sure what's going on. Further investigation required.

ahal:
Figured out the problem. The decision task was properly generating the docker-context artifacts under `/builds/worker/checkouts/src/docker-contexts`. However glean's `.taskcluster.yml` was setting `MOZ_FETCHES_DIR=/builds/worker/checkouts/src`, and `run-task` has a cleanup at the end that removes `MOZ_FETCHES_DIR` entirely.

Afaict, the decision task doesn't use any fetches.. so it should be safe to remove this line. There is a commented out line in `glean_taskgraph` so maybe the intent was to use a fetch so maybe that's why it got added? But if so it was still added in error as we need `MOZ_FETCHES_DIR` to be set in the env of the task that will download the fetch, not the env of the Decision task itself.

ahal:
Hi @badboy , this should be ready for review now. I don't have permissions, but if you like you can add the `mozilla-releng/releng` as an additional reviewer if you aren't comfortable with these changes.

badboy (approved):
Looks good to me. I'll double-check the task that ran just to make sure.
I need to do a release later today anyway, so we can test that that part still works ok as well.

PR #1743, opened 2021-08-06
Bug 1716957 - Link to Glean.js collected metrics docs on the Glean book [doc only]

brizental:
This feels like a hack, but it was the way I found to add this information to the book. Final render does not look too bad:

Screenshot 2021-08-06 at 12 57 16

Open to suggestions here to make this nices though.


travis79 (approved):
LGTM!

PR #1747, opened 2021-08-10
Bug 1722271 - Update to Kotlin 1.5

badboy:
Also upgrading Gradle and some dependencies.

Draft as I also want to update some stuff on CI

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -93,6 +93,15 @@ commands:
             rustup target add i686-linux-android
             rustup target add x86_64-linux-android
 
+  install-android-ndk:
+    steps:
+      - run:
+          name: Install missing Android SDK & NDK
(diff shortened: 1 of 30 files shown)


travis79:
This builds and runs all the Glean kotlin tests, but when I try and build and run the glean-sample-app, it fails in `SharedTestUtils.kt` with an unresolved reference to `setDispatcher`, and it looks like something is returning a `String?` instead of a `String` for the `request.path` in the `waitForPingContent` function, which is triggering a `Only safe (?.) or non-null asserted (!!.) calls are allowed ...`

badboy:
The `Generate Kotlin documentation` was removed in this PR, I'm going to remove it from our github enforcement rules.

travis79 (approved):

PR #1752, opened 2021-08-16
Bug 1691953 - Android: Remove support for reading app version info from package manager

badboy:
This removes the previously deprecated functionatilty.
Passing in the build-time generated `BuildInfo` struct is required.

---

I'll wait with landing this until I have the geckoview work done, so that this can go in a later major release as needed (because technicall a breaking change)


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -105,42 +103,7 @@ open class GleanInternalAPI internal constructor () {
...
-     * @param uploadEnabled A [Boolean] that determines whether telemetry is enabled.
-     *     If disabled, all persisted metrics, events and queued pings (except
-     *     first_run_date and first_run_hour) are cleared.
-     * @param configuration A Glean [Configuration] object with global settings.
-     */
-    @JvmOverloads
-    @Synchronized
-    @MainThread
-    @Deprecated("The buildInfo parameter will be required in the future. See bug 1691953")
-    fun initialize(
```
Dexterp37, line 137: Not sure we need to worry right away, but did you test how this affects (if at all) Java consumers?

Dexterp37 (approved):
r+ with a comment and much joy.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Java consumers can't leave out the config parameter anymore: https://github.com/mozilla/glean/pull/1752/files#diff-55350221eb40c8d32d4c30f8392220d105c0a44fa9ccbbd338b975f09acdb6cdR42

PR #1754, opened 2021-08-17
Bug 1725943 - Use a consistent Unknown value for the app build

badboy:
I choose `Unknown` over `unknown` for `app_build` and `app_display_version` because it was used more already.

`os` (and `os_version`) on the other hand use lowercase `unknown`.

Is it worth being fully consistent everywhere? Are we breaking anyone by changing this? (I hope not)
Looping in @brizental for docs review and @Dexterp37 for an opinion.


Dexterp37:
> 
> 
> I choose `Unknown` over `unknown` for `app_build` and `app_display_version` because it was used more already.
> 
> `os` (and `os_version`) on the other hand use lowercase `unknown`.

🤦 

> Is it worth being fully consistent everywhere?

I would always value consistency over anything else, if possible. I don't think this value is being used too much, so I would not be too concern. Worst case, we could always write an announcement to the mailing list.

> Are we breaking anyone by changing this? (I hope not)

I hope not!

badboy:
> > I choose `Unknown` over `unknown` for `app_build` and `app_display_version` because it was used more already.
> > `os` (and `os_version`) on the other hand use lowercase `unknown`.
> 
> 🤦
> 
> > Is it worth being fully consistent everywhere?
> 
> I would always value consistency over anything else, if possible. I don't think this value is being used too much, so I would not be too concern. Worst case, we could always write an announcement to the mailing list.
> 
> > Are we breaking anyone by changing this? (I hope not)
> 
> I hope not!

So... `unknown`, `Unknown`, `UnKnOwN` or `ünkn0wn` everywhere?

brizental:
> So... unknown, Unknown, UnKnOwN or ünkn0wn everywhere?

Although I am a big fan of `ünkn0wn`, I'd prefer `Unknown`.

(As a bonus that value is the one used on Glean.js, so it would require no changes over there).

Dexterp37:
> So... `unknown`, `Unknown`, `UnKnOwN` or `ünkn0wn` everywhere?

 `ünkn0wn` is the clear winner.
 
 But, if we want to go with a least perfect choice, I'd say either `unknown` or `Unknown` is fine: maybe the latter, since it was used more? But I have no strong opinion.

docs/user/user/pings/index.md
```diff
@@ -175,6 +172,7 @@ _Type: [String](../../reference/metrics/counter.md),
 Lifetime: [Application](../../reference/yaml/metrics.md#application)_
 
 The user-visible version of the operating system (e.g. "1.2.3").
+If the version detection fails, this metric gets set to `unknown`.
```
brizental, line 175:
```suggestion
If the version detection fails, this metric gets set to `Unknown`.
```
?

brizental (approved):
r+, only related to the doc changes :)

docs/user/user/pings/index.md
badboy: only once I change the code too.

PR #1756, opened 2021-08-20
Bug 1726765 - Force a compile failure when multiple Glean versions are detected in the build

badboy:
Because of the native code shipped it's essential that all downstream
users rely on the same Glean version.
Gradle can be overeager in chosing a single version when multiple are in
the dependency tree, always picking the latest, even across potentially
breaking-change (that is: major) releases.

Gradle has a way to influence dependency resolution and can even
disallow differing versions using `failOnVersionConflict`.
But that applies to all dependencies and there's no native way to
enforce that for a single dependency.
So the recommended solution[1] is to "manually" check if a dependency was
resolved from multiple versions and raise an issue.

For the ease of consumption by users, such as Fenix, we can apply this
in the Gradle plugin, because that's guaranteed to execute as part of
the build (because if not how are they even using Glean?)

[1]: https://github.com/gradle/gradle/issues/8813

---

Once landed this will go into a release.

### DIFF
--- a/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
+++ b/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
@@ -567,6 +567,16 @@ except:
                 }
                 because 'use GeckoView Glean instead of standalone Glean'
             }
+
+            incoming.afterResolve {
+                resolutionResult.allComponents {
+                    if (selectionReason.conflictResolution && moduleVersion != null) {
+                        if (moduleVersion.group.contains("org.mozilla.telemetry") && moduleVersion.name.contains("glean")) {
+                            throw AssertionError("Cannot have a conflict on Glean ${selectionReason}")
+                        }
+                    }
+                }
+            }
         }
 
         if (project.android.hasProperty('applicationVariants')) {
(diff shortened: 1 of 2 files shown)


gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -567,6 +567,16 @@ except:
                 }
                 because 'use GeckoView Glean instead of standalone Glean'
             }
+
+            incoming.afterResolve {
```
Dexterp37, line 571: Can you add a changelog entry for this?

Dexterp37 (commented):
Did you run manual QA on this? How can this be tested?

gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
badboy: Yup, change the version a-c depends on to something different from what m-c uses, and BOOM, it breaks. See also [this Fenix PR](https://github.com/mozilla-mobile/fenix/pull/20889) that does the same. If we get this released and into m-c first, then we don't need that in Fenix.

badboy:
> Did you run manual QA on this? How can this be tested?

Yup, change the version a-c depends on to something different from what m-c uses, and BOOM, it breaks. See also [this Fenix PR](https://github.com/mozilla-mobile/fenix/pull/20889) that does the same. If we get this released and into m-c first, then we don't need that in Fenix.

Dexterp37 (approved):
:shipit:

PR #1759, opened 2021-08-25
Capability is now added by GeckoView directly 

badboy:


badboy:
exercising admin rights.

PR #1761, opened 2021-08-25
Bug 1727299 - Document new text metric

badboy:
[doc only]

### DIFF
--- a/docs/user/reference/metrics/text.md
+++ b/docs/user/reference/metrics/text.md
@@ -1,0 +1,151 @@
+# Text
+
+Records a single long Unicode text, used when the limits on `String` are too low.
+
+{{#include ../../../shared/blockquote-warning.html}}
+
+## Important
+
+> This type should only be used in special cases when other metrics don't fit.
+> See [limitations](#limits) below.  
+> Reach out to the Glean team before using this.
+
+## Recording API
+
+### `set`
+
+Sets a text metric to a specific value.
+
+{{#include ../../../shared/tab_header.md}}
+
+<div data-lang="Kotlin" class="tab"></div>
+
+<div data-lang="Java" class="tab"></div>
+
+<div data-lang="Swift" class="tab"></div>
+
+<div data-lang="Python" class="tab"></div>
+
+<div data-lang="JavaScript" class="tab">
+
+```js
+import * as article from "./path/to/generated/files/article.js";
+
+article.content.set(extractedText);
+```
+</div>
+
+<div data-lang="Rust" class="tab"></div>
+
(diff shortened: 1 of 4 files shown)


docs/user/reference/metrics/text.md
```diff
@@ -0,0 +1,150 @@
...
+<div data-lang="Java" class="tab"></div>
+
+<div data-lang="Swift" class="tab"></div>
+
+<div data-lang="Python" class="tab"></div>
+
+<div data-lang="JavaScript" class="tab">
+
+```js
+import * as display from "./path/to/generated/files/display.js";
```
brizental, line 31: This example is a bit confusing as `display` is imported but the name of the metric used is `article`.

docs/user/reference/metrics/text.md
```diff
@@ -0,0 +1,150 @@
...
+
+assert.strictEqual("some content", await article.content.testGetValue());
```
brizental, line 76: Some comment about the confusing import.

docs/user/reference/metrics/text.md
```diff
@@ -0,0 +1,150 @@
...
+
+<div data-lang="Rust" class="tab"></div>
+
+<div data-lang="Firefox Desktop" class="tab"></div>
+
+{{#include ../../../shared/tab_footer.md}}
+
+#### Limits
+
+* Text metrics can only be sent in custom pings.
```
brizental, line 45: Please linkinfy "custom pings".

docs/user/reference/metrics/text.md
```diff
@@ -0,0 +1,150 @@
...
+* Text metrics are always of data collection category 3 (`web_activity`) or category 4 (`highly_sensitive`).
+* Only `ping` and `application` lifetime are allowed.
```
brizental, line 47:
```suggestion
* Only `ping` and `application` lifetimes are allowed.
```

docs/user/reference/metrics/text.md
```diff
@@ -0,0 +1,150 @@
+# Text
+
+Records a single long Unicode text, used when the limits on `String` are too low.
+
+{{#include ../../../shared/blockquote-warning.html}}
+
+## Important
+
+> This type should only be used in special cases. See [limitations](#limits) below.
+> It is best to talk to the Glean team before using this.
```
brizental, line 10: Kinda weird for me that we suggest people ask us before using a specific metric type in their own apps.

badboy: Yeah, I wasn't sure about this. I'm fine with removing that sentence, the limitations should be enough of a "warning"

PR #1782, opened 2021-09-10
bug 1729723 - Clear only the specified storage of delayed ping data

chutten:


travis79 (approved):
Nice 😎 !

chutten:
@badboy I don't suppose we could increase our min rust version to 1.53.0 so I can use `BTreeMap::retain`? m-c's on 1.54.0 for [at least a month](https://bugzilla.mozilla.org/show_bug.cgi?id=1723016).

travis79:
If we bump the minimum version, we may need to coordinate with A-S since [they appear to be using a minimum rust version of 1.50.0](https://github.com/mozilla/application-services/blob/e1abe5ad61983f00a8abf1b477afa5470876e484/.circleci/config.yml#L16)

brizental (approved):
Nice catch!

badboy:
> @badboy I don't suppose we could increase our min rust version to 1.53.0 so I can use `BTreeMap::retain`? m-c's on 1.54.0 for [at least a month](https://bugzilla.mozilla.org/show_bug.cgi?id=1723016).

I'm not too opposed to that. We can coordinate with a-s afterwards as it's not too critical (they don't use Glean Rust yet directly)

PR #1790, opened 2021-09-14
Add note that C# is deprecated

badboy:
[doc only]

---

Noticed it's nowhere written that this lacks functionality and is deprecated. Shall we add this note?


Dexterp37 (approved):
:'(

badboy:
> :'(

I know, but I'm only stating the facts.

PR #1792, opened 2021-09-15
Bug 1727589 - Upgrade to Gradle 7, rust-android-gradle 0.9 and other updates

badboy:
Glean now requires Java 11 due to the dependency updates.

---

* [ ] Changelog entry
* [ ] Check that consumers are fine with this (e.g. m-c)


badboy:
Closing to figure out Docker issues first.

PR #1793, opened 2021-09-15
Bug1729723 expose persist ping lifetime

chutten:
We'll need a minor (new API) release after.

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -357,10 +357,7 @@ fn initialize_internal(
     Some(init_handle)
 }
 
-/// Shuts down Glean.
-///
-/// This currently only attempts to shut down the
-/// internal dispatcher.
+/// Shuts down Glean in an orderly fashion.
 pub fn shutdown() {
     if global_glean().is_none() {
         log::warn!("Shutdown called before Glean is initialized");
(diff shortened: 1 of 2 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -379,6 +376,13 @@ pub fn shutdown() {
     if let Err(e) = dispatcher::shutdown() {
         log::error!("Can't shutdown dispatcher thread: {:?}", e);
     }
+
+    // Be sure to call this _after_ draining the dispatcher
+    crate::launch_with_glean(|glean| {
```
badboy, line 381: `launch_with_glean` uses the dispatcher. By the time this is called the dispatcher is shutdown and the code won't run.
I might have run over by [implementing a test for this](https://github.com/mozilla/glean/commit/300c286c8deb8e4e9ea4e010cf4fd1f3ace6d693).
It [fails on the current codebase](https://app.circleci.com/pipelines/github/mozilla/glean/8337/workflows/73efaadf-3f11-4a0b-a668-1736c3cc7cfb/jobs/150554) and does fail on this PR as is, it succeeds after fixing this.
(I think it would make sense to pull in the test).

```suggestion
    crate::with_glean(|glean| {
```

badboy (changes requested):

glean-core/rlb/src/lib.rs
chutten: Ah, of course. I knew that, of course I knew that. :facepalm: 

As for the test, we can probably write this as an integration test in Rust. Lemme try.

*some time later, after having re-remembered `-- --nocapture`*

Yup, it works.

badboy (approved):

badboy:
I'll merge & release and will add a changelog entry when doing so.

PR #1798, opened 2021-09-17
TC: Use Java 11 in CI and update Android tooling [ci full]

badboy:
Glean will require Java 11 due to upcoming dependency updates.
Dockerfile changes are copied from application-services.
They are also using Java 11, so we should be just as fine.

[1]: https://github.com/mozilla/application-services/commit/07a7f542c37828247965046bf47dd58b8e371f65

---

Let's see what happens if we only try to land this.


travis79 (approved):
LGTM, I see Docker is being a bit of a pain for the Android tests...

badboy:
Yeah, blocked on https://bugzilla.mozilla.org/show_bug.cgi?id=1731329

badboy:
Close/Re-open to re-trigger TaskCluster with a full build

PR #1805, opened 2021-09-27
Bug 1731963 - Document that it is unnecessary to check if Glean is enabled before submitting pings

brizental:
I also removed a redundant paragraph from the bottom of the "Submission API" section and moved the info box related to built-in pings to the top of the page.


travis79 (approved):
LGTM!

PR #1808, opened 2021-09-29
Bug 1733152 - Explicitly only allow glean logs on Android

badboy:
Previously we enabled debug logging for all Rust modules within the same build.
This causes far too many logs in geckoview builds.
This stops that and only enables Glean logging.


chutten (approved):
Ah, we talked about this, I remember.

PR #1812, opened 2021-09-30
bug 1732988 - Make glean::persist_ping_lifetime_data async

chutten:
By its nature, persist_ping_lifetime_data performs I/O.
To be consistent across our public API surface, we should dispatch the call.

### DIFF
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -788,25 +788,12 @@ pub fn get_timestamp_ms() -> u64 {
     glean_core::get_timestamp_ms()
 }
 
-/// Asks the database to persist ping-lifetime data to disk. Probably expensive to call.
-/// Only has effect when Glean is configured with `delay_ping_lifetime_io: true`.
-/// If Glean hasn't been initialized this will dispatch and return Ok(()),
-/// otherwise it will block until the persist is done and return its Result.
-pub fn persist_ping_lifetime_data() -> Result<()> {
-    if !was_initialize_called() {
-        crate::launch_with_glean(|glean| {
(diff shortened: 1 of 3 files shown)


glean-core/rlb/src/lib.rs
```diff
@@ -788,25 +788,12 @@ pub fn get_timestamp_ms() -> u64 {
...
-        // (In production Glean can't be re-initialized).
-        dispatcher::block_on_queue().map_err(|_| Error::not_initialized())?;
-        with_glean(|glean| glean.persist_ping_lifetime_data())
-    }
+/// Dispatches a request to the database to persist ping-lifetime data to disk.
+pub fn persist_ping_lifetime_data() {
+    crate::launch_with_glean(|glean| {
+        // This is async, we can't get the Error back to the caller.
+        let _ = glean.persist_ping_lifetime_data();
+    });
```
badboy, line 796: `shutdown` calls `persist_ping_lifetime_data` after `dispatcher::shutdown`, which means it will never actually run at that point.

badboy (changes requested):
I agree that this can be async, but then we need to restructure `shutdown` as well.

glean-core/rlb/src/lib.rs
chutten: Actually, it calls `glean.persist_ping_lifetime_data()` which goes straight to the db synchronously. It's only the static version of `persist_ping_lifetime_data` on the public API that dispatches. Or am I misunderstanding?

badboy: D'oh! No, you might be correct and I'm bad at reading code.

badboy (approved):

PR #1816, opened 2021-10-05
Extend general API with logging and setUploadEnabled

badboy:
This creates a new crate `glean-uniffi` that will be consumed by the
different foreign-language SDKs.
Eventually `glean-core` and `glean` should go away and be merged into
that `glean-uniffi`

---

This is the first PR against the `uniffi` branch.
It doesn't do much, except

* Fixing up the general API just slightly
* Extending the counter metric and implementing it (I accidentally committed the minimal counter API definition before).

Might not fully compile just yet.


travis79 (approved):
uploadEnabled is enabled!

badboy:
Ok, let's merge this, will make my future commits easier to handle.

PR #1817, opened 2021-10-06
Bring back the database & a dispatcher

badboy:
This brings back the database implementation on top of Rkv, unmodified.
It then introduces the dispatcher, taken from the RLB and uses that for the counter.
Rust code builds, Kotlin code builds. Tests don't.

---

The way I see this working is we now start moving over the pieces we need (like here: database, dispatcher) nearly unmodified and iterate on it to get it working (I have some more patches to get some initial counter tests back).
Once we're there I'd like to move some of the changes into a sort of ADR/proposal/doc just to call them out, e.g. initialize is now changed due to the back and forth between language SDKs and the Rust core.
After that we can work on metrics.

When we feel all confident about it we need to check against the codebase in `main` and see how to best merge that back.

what do you think about that plan, @travis79?

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -362,13 +362,16 @@ name = "glean-uniffi"
 version = "0.1.0"
 dependencies = [
  "android_logger 0.9.2",
+ "bincode",
  "chrono",
  "crossbeam-channel",
  "env_logger 0.7.1",
+ "ffi-support",
  "inherent",
  "log",
  "once_cell",
  "oslog",
+ "rkv",
  "serde",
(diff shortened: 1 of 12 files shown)


travis79:
> what do you think about that plan, @travis79?

Sounds good to me!

mdboom:
Yeah, :+1: to the doc about these changes (not as upfront design, but more as a massive CHANGELOG of the differences).  So that when we deploy and we see data go differently we'll all have context to understand why.  That could just be in the developer appendix in this repo or something.

PR #1818, opened 2021-10-06
Support Python 3.10

mdboom:
This also uses the "next gen" Python Docker images from CircleCI.

Also removes some `persist_to_workspace` that is no longer needed.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -280,7 +280,7 @@ jobs:
       - image: cimg/python:3.8
     steps:
       - checkout
-      - run: sudo pip install yamllint
+      - run: pip install yamllint
       - run: make lint-yaml
 
   ###########################################################################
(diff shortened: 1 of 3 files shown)


.circleci/config.yml
```diff
@@ -858,6 +860,7 @@ jobs:
 
   Python Windows i686 tests:
     docker:
+      # Windows tests require "legacy" Docker images
```
Dexterp37, line 863: Do you know why that is?

Dexterp37 (approved):

.circleci/config.yml
mdboom: I don't know why this is, but I would guess some kind of difference in the mingw compiler, causing one of the tests to fail.  Good point that I should probably file a new bug to address this later (I don't think it's a high enough priority to do this now).

PR #1822, opened 2021-10-07
Use uniffi for the python bindings

badboy:
The smoke test works, but only if you copy the libglean_uniff.dylib around.
I'll fix that next.


travis79 (approved):
Woosh! That's really not so bad..

PR #1823, opened 2021-10-08
Bug 1733757 - Store created label metrics to avoid rapidly reconstructing them all …

badboy:
…the time

This fixes a crashing bug, that probably started being an issue since
March 2020 / Glean v26, when we removed finalizers from the Kotlin
library[1]. Depending on how GC runs it was theoretically possible to
hit that even with finalizers in the SDK.

Contrary to the commit message in [1] we do have one more place where we
dynamically construct metrics: Labeled Metrics!
It used to be the case that whenever you access the submetric of a
labeled metric, that is `Metrics.Category.Name["label"]`, we created a
new Rust object, put that into the internal ConcurrentHandleMap and
returned a handle to it.

A ConcurrentHandleMap does a lot of work to ensure that you can't misuse
handles against the wrong map, when the item was already replaced, etc.
It also sets an upper limit of entries it can handle[2].
That limit is high enough for the static use of Rust (you would have to
use 32767+ metrics of a single type), but quickly[3] breaks down

The solution is to not re-create labeled submetrics on every access.
We cache those ourselves now, identified by the parent's metric handle
and the label.
Now accessing the same label 32768+ times is fine, you get the same
Rust object every time (as long as it exists).

There's one last piece: We do expose a way to destroy metrics, even if
that shouldn't happen for any statically defined ones.
This is not happening on Kotlin, where we removed finalizers some time
ago [1].
We're now removing those finalizers on Swift and Python as well.
We could try to get smart and do some double-checks that the handle is
not stale when the submetric is accessed.
But that breaks down quickly, because the metric could be destroyed
between the access and the recording, still leading to issues.
The only way to handle that would be some reference counting on the
object itself, but the current implementation doens't easily allow that.
So for now we just don't destroy boolean/counter/string metrics, ever.

The included tests fail without the corresponding library changes,
giving us a bit more confidence that the fix works.

Now there's still one way to trigger a crash, in theory:
Keep accessing previously-unused labels 32768+ times.
If you do that you're really on your own.

[1]: https://github.com/mozilla/glean/commit/25bf9ea0b26791f80235e2a421884ee63aae2326
[2]: https://github.com/mozilla/ffi-support/blob/0fdc22a8dfe3731be5fd39b311e4e4885219e26c/src/handle_map.rs#L160-L166
[3]: Well, "quickly". Reproducing that on get.webgl.org takes 8+ minutes.

### DIFF
--- a/glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
+++ b/glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
@@ -149,7 +149,7 @@ class LabeledMetricTypeTests: XCTestCase {
     }
 
     func testLabeledStringType() {
-        let counterMetric = StringMetricType(
+        let stringMetric = StringMetricType(
             category: "telemetry",
             name: "labeled_counter_metric",
             sendInPings: ["metrics"],
@@ -163,7 +163,7 @@ class LabeledMetricTypeTests: XCTestCase {
             sendInPings: ["metrics"],
             lifetime: .application,
             disabled: false,
-            subMetric: counterMetric
+            subMetric: stringMetric
         )
 
         labeledStringMetric["label1"].set("foo")
@@ -217,5 +217,39 @@ class LabeledMetricTypeTests: XCTestCase {
         )) { error in
             XCTAssertEqual(error as! String, "Can not create a labeled version of this metric type")
         }
+    }
+
+    func testRapidlyRecreatingLabeledMetricsDoesNotCrash() {
+        // Regression test for bug 1733757.
+        // The underlying map implementation has an upper limit of entries it can handle,
+        // currently set to (1<<15)-1 = 32767.
+        // We used to create a new object every time a label was referenced,
+        // leading to exhausting the available storage in that map, which finally results in a panic.
+
+        let counterMetric = CounterMetricType(
+            category: "telemetry",
+            name: "labeled_nocrash_counter",
+            sendInPings: ["metrics"],
+            lifetime: .application,
+            disabled: false
+        )
+
+        let labeledCounterMetric = try! LabeledMetricType<CounterMetricType>(
+            category: "telemetry",
+            name: "labeled_nocrash",
+            sendInPings: ["metrics"],
+            lifetime: .application,
+            disabled: false,
+            subMetric: counterMetric,
+            labels: ["foo"]
+        )
+
+        // We go higher than the maximum of `(1<<15)-1 = 32767`.
+        let maxAttempts = Int32(1 << 16)
+        for _ in 1 ... maxAttempts {
+            labeledCounterMetric["foo"].add(1)
+        }
+
+        XCTAssertEqual(maxAttempts, try labeledCounterMetric["foo"].testGetValue())
     }
 }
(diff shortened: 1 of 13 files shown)


glean-core/python/tests/metrics/test_labeled.py
```diff
@@ -251,3 +251,30 @@ def test_invalid_labels_go_to_other():
...
+        labels=["foo"]
+    )
+
+    # We go higher than the maximum we know of.
+    # Python is slow, so we only go a tiny bit higher.
+    max_attempts = (1 << 15) + 1
+    for _ in range(max_attempts):
+        labeled_counter_metric["foo"].add(1)
+
+    assert max_attempts, labeled_counter_metric["foo"].test_get_value()
```
mdboom, line 280: First clause is the test, second is a message.  So I think you meant: 
```suggestion
    assert max_attempts == labeled_counter_metric["foo"].test_get_value()
```

glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
```diff
@@ -149,7 +149,7 @@ class LabeledMetricTypeTests: XCTestCase {
     }
 
     func testLabeledStringType() {
-        let counterMetric = StringMetricType(
+        let stringMetric = StringMetricType(
```
travis79, line 152: 😎

glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
```diff
@@ -218,4 +218,38 @@ class LabeledMetricTypeTests: XCTestCase {
...
+            name: "labeled_nocrash",
+            sendInPings: ["metrics"],
+            lifetime: .application,
+            disabled: false,
+            subMetric: counterMetric,
+            labels: ["foo"]
+        )
+
+        // We go higher than the maximum we know of.
+        let maxAttempts = Int32(1 << 16)
```
travis79, line 248: nit: After some extra thought on this, bit shifting isn't the best at readability and understanding the code. I mean, I understand what you are doing here but someone else might come along and wonder "why use shift here?". Did you do this simply because it is explicitly defined as `(1<<15)-1` in Rust or something? I guess I'm just wondering why not do something like `Int32(Int16.max +1)` or something.

glean-core/python/tests/metrics/test_labeled.py
```diff
@@ -251,3 +251,30 @@ def test_invalid_labels_go_to_other():
...
+        name="labeled_nocrash",
+        send_in_pings=["metrics"],
+        lifetime=Lifetime.APPLICATION,
+        disabled=False,
+        labels=["foo"]
+    )
+
+    # We go higher than the maximum we know of.
+    # Python is slow, so we only go a tiny bit higher.
+    max_attempts = (1 << 15) + 1
```
travis79, line 276: So, Kotlin and Swift you used 1<<16 for...  I guess this is what you meant by Python being slow.

badboy: yes, it double the runtime in Python to somewhere between 13 and 20 seconds :/

glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
badboy: I can change the comment to call out the maximum.

PR #1831, opened 2021-10-13
Bug 1734408 - Add docs on how to avoid Glean.js + iOS + Qt issue

brizental:
[doc only]

### DIFF
--- a/docs/user/user/adding-glean-to-your-project/qt.md
+++ b/docs/user/user/adding-glean-to-your-project/qt.md
@@ -140,6 +140,4 @@ This bug is only reproduced in iOS devices, it does not happen in emulators. It 
 **only happens when using the Qt debug library for iOS**.
 
 There is no way around this bug other than avoiding the Qt debug library for iOS altogether until
-this bug is fixed. In order to do that, we suggest overwriting the `QT_LIBRARY_SUFFIX` environment
-variable with `""` when building for this environment. Other techniques for achieving this same goal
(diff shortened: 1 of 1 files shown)


docs/user/user/adding-glean-to-your-project/qt.md
```diff
@@ -127,3 +128,18 @@ to
...
+
+The [`submitPing`](../../reference/pings/index.md) function hits a
+[known bug](https://bugreports.qt.io/browse/QTBUG-96788) in the Qt JavaScript interpreter.
+
+This bug is only reproduced in iOS devices, it does not happen in emulators. It also
+**only happens when using the Qt debug library for iOS**.
+
+There is no way around this bug other than avoiding the Qt debug library for iOS altogether until
+this bug is fixed. In order to do that, we suggest overwriting the `QT_LIBRARY_SUFFIX` environment
+variable with `""` when building for this environment. Other techniques for achieving this same goal
```
badboy, line 144: Does this variable need to be set as part of the build process?
So something like `QT_LIBRARY_SUFFIX="" bash scripts/qt5_compile.sh qt qt` following the instructions from the VPN client readme.

Also I think this can be phrased stronger. We're not only suggesting it.
Maybe:

```suggestion
this bug is fixed. 
To build a non-debug version of Qt set the environment variable `QT_LIBRARY_SUFFIX` to `""` for the build, e.g. `QT_LIBRARY_SUFFIX="" make`
Other techniques for achieving this same goal
```

brizental: Well, ended up removing all that and just linking ot the Qt docs, let me know if you think that is a bit better.

PR #1833, opened 2021-10-15
Get current Rust test passing 

badboy:
This creates a new crate `glean-uniffi` that will be consumed by the
different foreign-language SDKs.
Eventually `glean-core` and `glean` should go away and be merged into
that `glean-uniffi`

--- 

mostly I want to get `cargo test` green so that I can rely on that going forward.
Kotlin & Python tests don't work yet.


glean-core/uniffi/src/lib.rs
```diff
@@ -113,3 +113,8 @@ pub fn enable_logging() {
 pub fn set_upload_enabled(enabled: bool) {
     core::with_glean_mut(|glean| glean.set_upload_enabled(enabled))
 }
+
+// Split unit tests to a separate file, to reduce the file of this one.
```
travis79, line 117:
```suggestion
// Split unit tests to a separate file, to reduce the length of this one.
```

glean-core/uniffi/src/core/mod.rs
```diff
@@ -110,6 +110,28 @@ impl Glean {
         Ok(this)
     }
 
+    /// For tests make it easy to create a Glean object using only the required configuration.
+    #[cfg(test)]
+    pub(crate) fn with_options(
```
travis79, line 115: 🤩

PR #1835, opened 2021-10-18
Enable uniffi in the Xcode build

badboy:
This doesn't build yet due to all the missing/replaced functionality, but it's probably best to iterate on this in small chunks.


build-scripts/xc-cargo.sh
```diff
@@ -0,0 +1,10 @@
+#!/usr/bin/env bash
+#
+# This is a small wrapper for running `cargo` inside of an XCode build,
+# which unfortunately doesn't seem to work quite right out-of-the-box.
+set -eEuvx
+
+# Xcode tries to be helpful and overwrites the PATH. Reset that.
```
travis79, line 7: Xcode is always so helpful

PR #1838, opened 2021-10-20
Define & implement the general experiment API

badboy:
This does not yet store or return any data, but only defines the API
functions and adds a (failing) smoke test.
It also only covers Kotlin.

---

Additional notes inline.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -17,14 +17,6 @@ import kotlinx.coroutines.Job
 import kotlinx.coroutines.MainScope
 import kotlinx.coroutines.launch
 import mozilla.telemetry.glean.GleanMetrics.GleanBaseline
-import mozilla.telemetry.glean.internal.InternalConfiguration
-import mozilla.telemetry.glean.internal.initialize as gleanInitialize
-import mozilla.telemetry.glean.internal.enableLogging
-import mozilla.telemetry.glean.internal.finishInitialize
-import mozilla.telemetry.glean.internal.setExperimentActive as gleanSetExperimentActive
-import mozilla.telemetry.glean.internal.setExperimentInactive as gleanSetExperimentInactive
-import mozilla.telemetry.glean.internal.testIsExperimentActive as gleanTestIsExperimentActive
-import mozilla.telemetry.glean.internal.testGetExperimentData as gleanTestGetExperimentData
 import mozilla.telemetry.glean.internal.*
 import mozilla.telemetry.glean.config.Configuration
 import mozilla.telemetry.glean.config.FfiConfiguration
@@ -173,7 +165,7 @@ open class GleanInternalAPI internal constructor () {
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
 
         Log.e(LOG_TAG, "Glean. enable logging.")
-        enableLogging()
+        gleanEnableLogging()
         // Execute startup off the main thread.
         @Suppress("EXPERIMENTAL_API_USAGE")
         Dispatchers.API.executeTask {
@@ -198,7 +190,7 @@ open class GleanInternalAPI internal constructor () {
 
             // Signal Dispatcher that init is complete
             Dispatchers.API.flushQueuedInitialTasks()
-            finishInitialize()
+            gleanFinishInitialize()
 
             // At this point, all metrics and events can be recorded.
             // This should only be called from the main thread. This is enforced by
--- a/glean-core/uniffi/src/glean.udl
+++ b/glean-core/uniffi/src/glean.udl
@@ -1,15 +1,15 @@
 namespace glean {
-    boolean initialize(InternalConfiguration cfg);
-    boolean finish_initialize();
+    boolean glean_initialize(InternalConfiguration cfg);
+    boolean glean_finish_initialize();
 
-    void enable_logging();
+    void glean_enable_logging();
 
-    void set_upload_enabled(boolean enabled);
+    void glean_set_upload_enabled(boolean enabled);
 
     // Experiment reporting API
-    void set_experiment_active(string experiment_id, string branch, record<DOMString, string> extra);
-    void set_experiment_inactive(string experiment_id);
-    RecordedExperiment? test_get_experiment_data(string experiment_id);
+    void glean_set_experiment_active(string experiment_id, string branch, record<DOMString, string> extra);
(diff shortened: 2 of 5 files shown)


glean-core/uniffi/src/glean.udl
```diff
@@ -5,6 +5,17 @@ namespace glean {
     void enable_logging();
 
     void set_upload_enabled(boolean enabled);
+
+    // Experiment reporting API
+    void set_experiment_active(string experiment_id, string branch, record<DOMString, string> extra);
+    void set_experiment_inactive(string experiment_id);
+    boolean test_is_experiment_active(string experiment_id);
+    RecordedExperiment test_get_experiment_data(string experiment_id);
```
badboy, line 13: I actually realised that we can now merge these together into this:

```suggestion
    RecordedExperiment? test_get_experiment_data(string experiment_id);
```

Then the SDK wrapper can just call that single function in both `testIs` and `testGet` and do the right thing depending on whether it got null or not.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -21,6 +21,11 @@ import mozilla.telemetry.glean.internal.InternalConfiguration
 import mozilla.telemetry.glean.internal.initialize as gleanInitialize
 import mozilla.telemetry.glean.internal.enableLogging
 import mozilla.telemetry.glean.internal.finishInitialize
+import mozilla.telemetry.glean.internal.setExperimentActive as gleanSetExperimentActive
+import mozilla.telemetry.glean.internal.setExperimentInactive as gleanSetExperimentInactive
+import mozilla.telemetry.glean.internal.testIsExperimentActive as gleanTestIsExperimentActive
+import mozilla.telemetry.glean.internal.testGetExperimentData as gleanTestGetExperimentData
+import mozilla.telemetry.glean.internal.*
```
badboy, line 28: I am considering just naming the top-level functions with a `glean_` prefix in the UDL. 
That avoids the renaming here in Kotlin and the wrapping we will need to do in Swift.

@travis, wdyt?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -327,34 +331,8 @@ open class GleanInternalAPI internal constructor () {
...
-            LibGleanFFI.INSTANCE.glean_set_experiment_active(
-                experimentId,
-                branch,
-                keys,
-                values,
-                numKeys
-            )
-        }
+        var map = extra ?: mapOf()
+        gleanSetExperimentActive(experimentId, branch, map)
```
badboy, line 335: Decided to just take a non-nullable value here and let the SDK handle the "old" API.
We can just as easily take an optional value in the UDL though.

glean-core/android/src/test/java/mozilla/telemetry/glean/UniffiTest.kt
```diff
@@ -81,4 +81,16 @@ class UniffiTest {
...
+    @Test
+    fun smokeExperimentAPI() {
+        val buildInfo = BuildInfo("0.0.1", "0.0.1")
+        Glean.initialize(context, uploadEnabled = true, Configuration(), buildInfo)
+
+        Glean.setExperimentActive("my-experiment", "control")
+        Glean.setExperimentActive("my-experiment", "control", mapOf("report" to "nothing"))
+        Glean.setExperimentInactive("my-experiment")
+        Glean.testIsExperimentActive("my-experiment")
+        val experiment = Glean.testGetExperimentData("my-experiment")
```
badboy, line 94: will all panic because the implementation is stubbed out.

glean-core/uniffi/src/lib.rs
```diff
@@ -114,6 +116,39 @@ pub fn set_upload_enabled(enabled: bool) {
...
+/// experiment annotation to the environment which is sent with pings. This
+/// infomration is not persisted between runs.
+///
+/// See [`glean_core::Glean::set_experiment_active`].
+pub fn set_experiment_active(
+    _experiment_id: String,
+    _branch: String,
+    _extra: HashMap<String, String>,
+) {
+    todo!()
```
badboy, line 129: These will need to run through the dispatcher.

glean-core/uniffi/src/private/recorded_experiment.rs
```diff
@@ -0,0 +1,16 @@
...
+
+use serde::Deserialize;
+
+/// Deserialized experiment data.
+#[derive(Clone, Deserialize, Debug)]
+pub struct RecordedExperiment {
+    /// The experiment's branch as set through [`set_experiment_active`](crate::set_experiment_active).
+    pub branch: String,
+    /// Any extra data associated with this experiment through [`set_experiment_active`](crate::set_experiment_active).
+    pub extra: HashMap<String, String>,
```
badboy, line 15: No more `Option`. Breaks the API on the SDK side technically.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -363,12 +341,7 @@ open class GleanInternalAPI internal constructor () {
      * @param experimentId The id of the experiment to deactivate.
      */
     fun setExperimentInactive(experimentId: String) {
```
travis79, line 343: These experiment API functions are razor thin now...  any reason to have the wrapper around the `gleanSetExperiment*` functions at all (aside from consistency with the other "glean" prefixed UDL functions)?

glean-core/uniffi/src/glean.udl
travis79: Nice! I don't think anyone is using the experiment test api aside from our internal tests right now, so I'd be okay with making this change even without the SDK wrapper, maybe with a comment that explains `null` = no experiment with that id.

glean-core/uniffi/src/private/recorded_experiment.rs
travis79: With Nimbus as the only customer, I'm less concerned about breaking this particular API. I can deal with the changes downstream, if necessary.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Because API consistency. We need those exposed on the global Glean object. We can only create that Glean object on the Kotlin side, needing to wrap UDL-exported functions.

We _could_ have an `Glean` interface in the UDL which has those as methods, but we most likely still need to wrap that interface on the Kotlin side, because of the additional functionality that still exists.
Plain old functions seemed more reasonable to me.

PR #1840, opened 2021-10-21
Use ios-sim target on Rust stable

badboy:
[Rust 1.56.0](https://blog.rust-lang.org/2021/10/21/Rust-1.56.0.html) is here and with it the ios-sim target on stable.


travis79 (approved):
Woot! ❤️ `stable`!

PR #1847, opened 2021-11-01
[doc only] Add in labeling and doc only info

Marlene-M-Hirose:
These two parts of the bug have not been addressed:

- Only explains how to run the Kotlin and Switf tests
- Mentions a CODEOWNERS file which does not exist


travis79 (approved):
Nice catch!

PR #1848, opened 2021-11-01
Use in-tree references for the YAML format documentation

wlach:
The glean_parser documentation just redirects back to this, so let's
save our users some trouble. We still link to the glean_parser
documentation for the actual schema definition.


brizental (approved):
Thanks, this is a great change. 🎉

PR #1850, opened 2021-11-02
Implement debug view/source tags & log pings API

badboy:
This is internal API, but called from the SDK wrappers, thus we need to
define it in the UDL, then expose it to users.

---

This includes 2 cleanup commits.
I have one more thing and then I can probably start working on making this all compile again properly.

### DIFF
--- a/glean-core/uniffi/src/core/mod.rs
+++ b/glean-core/uniffi/src/core/mod.rs
@@ -4,6 +4,7 @@ use std::sync::Mutex;
 
 use once_cell::sync::OnceCell;
 
+use crate::debug::DebugOptions;
 use crate::database::Database;
 use crate::private::{ExperimentMetric, RecordedExperiment};
 use crate::InternalConfiguration;
@@ -67,6 +68,7 @@ pub struct Glean {
     data_path: PathBuf,
     application_id: String,
     max_events: u32,
+    debug: DebugOptions,
     schedule_metrics_pings: bool,
 }
 
@@ -106,6 +108,7 @@ impl Glean {
             data_path,
             application_id,
             max_events: cfg.max_events.unwrap_or(500),
+            debug: DebugOptions::new(),
             schedule_metrics_pings: cfg.use_core_mps,
         };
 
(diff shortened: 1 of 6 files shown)


glean-core/uniffi/src/core/mod.rs
```diff
@@ -222,4 +225,71 @@ impl Glean {
...
+        self.debug.debug_view_tag.set(value.into())
+    }
+
+    /// Return the value for the debug view tag or [`None`] if it hasn't been set.
+    ///
+    /// The `debug_view_tag` may be set from an environment variable
+    /// (`GLEAN_DEBUG_VIEW_TAG`) or through the [`set_debug_view_tag`] function.
+    pub(crate) fn debug_view_tag(&self) -> Option<&String> {
+        self.debug.debug_view_tag.get()
+    }
```
travis79, line 249: I assume the need for the "property accessor" pattern here is to control access and only expose the `set` to UniFFI?

glean-core/uniffi/src/debug.rs
```diff
@@ -0,0 +1,319 @@
...
+
+impl std::fmt::Debug for DebugOptions {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        fmt.debug_struct("DebugOptions")
+            .field("log_pings", &self.log_pings.get())
+            .field("debug_view_tag", &self.debug_view_tag.get())
+            .field("source_tags", &self.source_tags.get())
+            .finish()
+    }
+}
```
travis79, line 54: This is nifty!

glean-core/uniffi/src/glean.udl
```diff
@@ -10,6 +10,10 @@ namespace glean {
     void glean_set_experiment_active(string experiment_id, string branch, record<DOMString, string> extra);
     void glean_set_experiment_inactive(string experiment_id);
     RecordedExperiment? glean_test_get_experiment_data(string experiment_id);
+
+    boolean glean_set_debug_view_tag(string tag);
+    boolean glean_set_source_tags(sequence<string> tags);
+    void glean_set_log_pings(boolean value);
```
travis79, line 16: I guess it's been this way for a while, but it seems weird that both of the other `set` functions return the boolean and the `log_pings` doesn't.

glean-core/uniffi/src/lib.rs
```diff
@@ -166,6 +218,74 @@ pub fn glean_test_get_experiment_data(experiment_id: String) -> Option<RecordedE
...
+///
+/// When the log pings debug option is `true`,
+/// we log the payload of all succesfully assembled pings.
+///
+/// # Arguments
+///
+/// * `value` - The value of the log pings option
+pub fn glean_set_log_pings(value: bool) {
+    if was_initialize_called() {
+        core::with_glean_mut(|glean| glean.set_log_pings(value));
```
travis79, line 283: Since `set_log_pings` returns a boolean, I'm curious what does Rust do here since this wrapper function returns void?

travis79 (approved):
Nothing worth blocking on, but a few questions inline.

glean-core/uniffi/src/lib.rs
travis79: Ahh, nevermind, I missed the `;` at the end

glean-core/uniffi/src/core/mod.rs
badboy: It's only used by the (not-yet-included) ping maker: https://github.com/mozilla/glean/blob/uniffi-settags-api/glean-core/src/ping/mod.rs#L194-L200

glean-core/uniffi/src/debug.rs
badboy: old code, copied from glean-core, but yes.

glean-core/uniffi/src/glean.udl
badboy: both the others validate the tags, at which point we bail out further processing if they are wrong.
`log_pings` has nothing to validate.

PR #1851, opened 2021-11-02
Implement client activity API

badboy:
This now makes me believe I _should_ store a general callback to trigger the uploader, just to avoid passing a new callback all the time

### DIFF
--- a/glean-core/uniffi/src/core/mod.rs
+++ b/glean-core/uniffi/src/core/mod.rs
@@ -292,4 +292,32 @@ impl Glean {
     pub(crate) fn log_pings(&self) -> bool {
         self.debug.log_pings.get().copied().unwrap_or(false)
     }
+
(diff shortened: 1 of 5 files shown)


glean-core/uniffi/src/core/mod.rs
```diff
@@ -292,4 +292,32 @@ impl Glean {
...
+    pub fn handle_client_inactive(&mut self) {
+        //if !self.internal_pings.baseline.submit(self, Some("inactive")) {
+        //    log::info!("baseline ping not submitted on inactive");
+        //}
+
+        //if !self.internal_pings.events.submit(self, Some("inactive")) {
+        //    log::info!("events ping not submitted on inactive");
+        //}
+
+        //self.set_dirty_flag(false);
```
travis79, line 321: Is this just cruft left behind from moving this functionality?

travis79 (approved):

PR #1852, opened 2021-11-03
Bug 1738878 - Glean parser with tags

badboy:
Requires https://github.com/mozilla/glean_parser/pull/409 and a followup release.
We should probably document that.

cc @wlach


Dexterp37 (approved):
Seems sane to me!

badboy:
Test is green now (but running with SSH enabled, I'll cancel that, then still merge this)

PR #1859, opened 2021-11-03
Bug 1686675 - Update the Glean book with the "SDKs" terminology

brizental:
This is a best effort, I am sure I missed some occurances.

### DIFF
--- a/docs/user/README.md
+++ b/docs/user/README.md
@@ -1,47 +1,38 @@
 # Introduction
 
-Glean is a modern approach for a telemetry library
-and is part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
+The Glean SDKs are modern cross-platform telemetry client libraries
+and are a part of the [Glean project](https://docs.telemetry.mozilla.org/concepts/glean/glean.html).
 
 ![Glean logo](glean.jpeg)
 
-There are two implementations of Glean, with support for 5 different programming languages in total.
-Both implementations strive to contain the same features with similar, but idiomatic APIs.
+The Glean SDKs are available for several programming languages and development environments.
+Each SDK aims to contain the same group of features with similar, but idiomatic APIs.
 
-Unless clearly stated otherwise, regard the text in this book as valid for both clients
-and all the supported programming languages and environments.
+To learn more about each SDK, refer to the [SDKs overview](./language-bindings/index.md) page.
 
-### [The Glean SDK](https://github.com/mozilla/glean)
+To get started adding Glean to your project, choose one of the following guides:
 
-The Glean SDK is an implementation of Glean in Rust, with language bindings for **Kotlin**,
-**Python**, **Rust** and **Swift**.
+- **[Kotlin](./user/adding-glean-to-your-project/kotlin.md)**
+  - Get started adding Glean to an Android application or library.
+- **[Swift](./user/adding-glean-to-your-project/swift.md)**
+  - Get started adding Glean to an iOS application or library.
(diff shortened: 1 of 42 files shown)


docs/user/README.md
```diff
@@ -89,11 +80,9 @@ To contact the Glean team you can:
 
 - Find us in the [#glean channel on chat.mozilla.org](https://chat.mozilla.org/#/room/#glean:mozilla.org).
 - Send an email to *glean-team@mozilla.com*.
-- The Glean SDK team is: *:janerik*, *:dexter*, *:travis*, *:mdroettboom*, *:gfritzsche*, *:chutten*, *:brizental*.
+- The Glean team is: *:janerik*, *:dexter*, *:travis*, *:mdroettboom*, *:gfritzsche*, *:chutten*, *:brizental*.
```
Dexterp37, line 83: I think we could safely remove Georg from this list.

Please note that I believe this is the "Glean SDKs" team still. The Glean team also include others (wlach and relud for example)

docs/user/appendix/changelog/index.md
```diff
@@ -1,6 +1,4 @@
 # Changelog
 
-This section contains the changelog for both Glean clients.
-
-- [The Glean SDK](./sdk.md)
-- [Glean.js](./js.md)
+- [`mozilla/glean` (glean-core)](./sdk.md)
```
Dexterp37, line 3: I'm not sure that's "glean-core". Isn't glean-core the Rust core alone? Isn't this just Glean, given that you're referring to the other SDK as Glean.js?

Dexterp37 (changes requested):

docs/user/README.md
brizental: I agree, thanks for catching this. 

I'll remove Georg too.

docs/user/appendix/changelog/index.md
brizental: Good question, not really sure here. Maybe I could remove this completely and just leave the repository name to avoid the rabbit hole. This is kinda old, but I was clinging to [this](https://mozilla.github.io/glean/book/appendix/changelog/sdk.html#v1900-2019-10-22) changelog entry that referred to the whole thing as "glean-core".

brizental: > Isn't this just Glean

Isn't Glean the collection of SDKs + pipeline + tools?

Dexterp37: > 
> 
> > Isn't this just Glean
> 
> Isn't Glean the collection of SDKs + pipeline + tools?

I've been Dexter'd. You are correct. Then Glean.js is probably Glean JS SDK and Glean is "Other Glean SDKs"?

brizental: So, there is also the horrible solution of calling them, "Rust core-based SDKs" and "JavaScript core-based SDKs". Which I started with and then removed. I did use it in the SDKs overview page I edited in this PR, see: https://github.com/mozilla/glean/pull/1859/files#diff-5358def7708fb975f1114844a9a38f2f5c0972f6262ec00962303e6029e1d605R18-R23 and https://github.com/mozilla/glean/pull/1859/files#diff-5358def7708fb975f1114844a9a38f2f5c0972f6262ec00962303e6029e1d605R84-R89.

brizental: Talked to @badboy about this just now and leaving just the repository name seems like a reasonable solution here.

Dexterp37 (approved):

PR #1862, opened 2021-11-05
Bug 1737656 - Document that url and text metric types are not usable inside Mozilla

wlach:
I feel a bit guilty putting this very Mozilla-data-pipeline-specific
text in there, but it does seem important to warn people.

### DIFF
--- a/docs/user/reference/metrics/text.md
+++ b/docs/user/reference/metrics/text.md
@@ -2,6 +2,13 @@
 
 Records a single long Unicode text, used when the limits on `String` are too low.
 
+{{#include ../../shared/blockquote-stop.html}}
+
+## Warning
+
+> This metric is not yet usable for most Mozilla products due to a data pipeline issue.
+> See [bug 1737656](https://bugzilla.mozilla.org/show_bug.cgi?id=1737656) for more information.
+
 {{#include ../../../shared/blockquote-warning.html}}
 
 ## Important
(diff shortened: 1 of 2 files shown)


docs/user/reference/metrics/text.md
```diff
@@ -2,6 +2,13 @@
 
 Records a single long Unicode text, used when the limits on `String` are too low.
 
+{{#include ../../shared/blockquote-stop.html}}
```
brizental, line 5: This blockquote type is a _very_ strong visual warning (example: https://mozilla.github.io/glean/book/appendix/contribution-guidelines.html#stop-quote). From the bug it seems like this issue is under control i.e. we have a plan on how to deal with it. I wonder if it's even necessary to add this warning here at all. Do you know a rough timeline for the issue to be fixed and for this warning to be stale? 

If we keep it, let's please rephrase it and also not use such a strong visual cue :) I worry that a warning phrased and presented like this could cause a long lasting trust issue related to these metric types or even to Glean itself. We have had such problems with less strong warnings on the book e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1734454.

badboy:
With the move of the schema generator to handle this as differently-named columns and renaming them back in views, I feel like we don't need this.
I'm ok with waiting for the implementation of those changes instead.

wlach:
Agreed, let's close this for now.

PR #1866, opened 2021-11-15
Bug 1741229 - Make `RecordedEventData` fields public

travis79:
This makes the fields in `RecordedEventData` for iOS public so that they are accessible in tests.


travis79:
Closed in favor of #1867

PR #1867, opened 2021-11-15
Bug 1741229 - Public access for struct RecordedEventData fields

lmarceau:
Label each field public on the RecordedEventData struct for EventMetric - otherwise internal by default.
Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1741229


travis79 (approved):
Thanks!

PR #1868, opened 2021-11-16
Bug 1734608 - Merge with glean-core and enable full general API and some tests

badboy:
This Pull Request does a whole lot of things all at once:

* It gets rid of the glean-uniffi crate. That was an experiment. Going forward we'll use glean-core again
* It implements all pieces necessary for Glean.kt to work
* It is able to pass a whole bunch of tests again.
* It uses a fork of glean-parser

~~It's far from ready-for-review. I will re-arrange and re-write commits to make it easier.~~
Commits are reviewable. I tried to structure everything so that individual commits make sense, and sometimes include a bit more information.


badboy:
Current state: Rust tests on glean-core pass, most Kotlin tests around the general API pass. What's left is all the metric types that are now obviously broken.

<details>

image

</details>

travis79:
> @travis79 Do you think you will be able to review it in this state? I can also spend more time on splitting this up into individual PRs instead.

I think I can manage it, if it becomes too difficult I will talk with you and we can compromise

glean-core/ios/Glean/Glean.swift
```diff
@@ -132,135 +117,33 @@ public class Glean {
...
+            languageBindingName: Constants.languageBindingName,
+            uploadEnabled: uploadEnabled,
+            maxEvents: configuration.maxEvents.map { UInt32($0) },
+            delayPingLifetimeIo: false,
+            appBuild: "0.0.0",
+            useCoreMps: false
+        )
+        let clientInfo = getClientInfo(configuration)
+        let callbacks = OnGleanEventsImpl(glean: self)
+        gleanInitialize(cfg: cfg, clientInfo: clientInfo, callbacks: callbacks)
```
travis79, line 132: This condensed down nicely!

build.gradle
```diff
@@ -147,6 +147,11 @@ subprojects {
     tasks.withType(org.jetbrains.kotlin.gradle.tasks.KotlinCompile).all {
         // TODO: Re-enable when we fixed all those warnings.
         //kotlinOptions.allWarningsAsErrors = true
+        kotlinOptions {
+            freeCompilerArgs += [
+                "-Xuse-experimental=kotlin.ExperimentalUnsignedTypes",
```
travis79, line 152: I don't suppose we can use the version of Kotlin where these are no longer experimental?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -60,9 +60,9 @@ internal class OnGleanEventsImpl(
         PingUploadWorker.enqueueWorker(glean.applicationContext)
     }
 
-    override fun startMetricsPingScheduler() {
+    override fun startMetricsPingScheduler(): Boolean {
         glean.metricsPingScheduler = MetricsPingScheduler(glean.applicationContext, glean.buildInfo)
-        glean.metricsPingScheduler?.schedule()
+        return glean.metricsPingScheduler!!.schedule()
```
travis79, line 65: I guess it is better to explode here than to fail silently with the `?`

travis79 (approved):
WHEW!  This is amazing and I am very happy you broke it down into so many commits.  I don't have any feedback worth blocking on and I definitely feel like my understanding of the new architecture is better after this review.  🎉 🍻

build.gradle
badboy: Ah good point, This is needed for the UIntArray stuff used somewhere. I'm not sure why the in-code annotations didn't work. I'll add another comment so we remember to double-check this.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: We set the value in the line above, so `!!` can't fail (unless the java VM breaks)

PR #1870, opened 2021-11-17
Fix path to metric-lifetime-timeline.svg

mhammond:
https://mozilla.github.io/glean/book/user/metrics/adding-new-metrics.html is supposed to show a link to metric-lifetime-timeline.svg, but it doesn't get the path quite correct - so it does show the alt-text ("Metric lifetime timeline") but no image.


brizental (approved):
Thanks a lot!

PR #1887, opened 2021-12-02
BUGFIX: Swap category and name

badboy:
Otherwise this records the metric into the wrong full identifier


Dexterp37 (approved):
Can you add a changelog entry?

PR #1889, opened 2021-12-06
Bug 1744504 - iOS: Skip code generation in `indexbuild` build

badboy:


Dexterp37 (approved):
Given what's in the bug, this makes sense to me. Sadly, I don't have a Mac to test. i trust you @badboy on this.

PR #1893, opened 2021-12-07
Bug 1741578 - Document that metric types must not change after release

badboy:


docs/user/reference/yaml/metrics.md
```diff
@@ -97,13 +97,12 @@ See the list of [supported metric types](../metrics/index.md).
...
 
-> Once a metric is released in a product, its `type` should not be changed.
-> If any data was collected locally with the older `type`, and hasn't yet been sent in a ping,
-> recording data with the new `type` may cause any old persisted data to be lost for that metric.
-> See [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1621757#c1) for an extended
-> explanation of the different scenarios.
+> Once a metric is defined in a product, its `type` must not be changed.
+> The ingestion pipeline will not be able to handle such a change.
+> If a type change is required a new metric must be added with a new name and the new type.
+> This will require an additional data review, in which you can also reference the old data review.
```
brizental, line 105: This could become more problematic the longer an application is using Glean. Is it something we should consider filing a bug in the pipeline about?

brizental (approved):

docs/user/reference/yaml/metrics.md
badboy: We already have that with legacy as well. People will come up with better names or `_v2` suffixes.
I don't think the pipeline can do much about it at this point.

PR #1899, opened 2021-12-09
Migrate timespan metric implementation

badboy:
Based on #1898.

PR #1900, opened 2021-12-09
Migrate boolean metric implementation

badboy:
Based on #1899

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
@@ -5,12 +5,6 @@
 package mozilla.telemetry.glean.private
 
 import androidx.annotation.VisibleForTesting
-import com.sun.jna.StringArray
-import mozilla.telemetry.glean.rust.LibGleanFFI
-import mozilla.telemetry.glean.rust.toByte
-
-import mozilla.telemetry.glean.Dispatchers
-import mozilla.telemetry.glean.rust.toBoolean
 
 /**
  * This implements the developer facing API for recording boolean metrics.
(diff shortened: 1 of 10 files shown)


glean-core/src/glean.udl
```diff
@@ -195,3 +195,11 @@ interface TimespanMetric {
 
     i32 test_get_num_recorded_errors(ErrorType error, optional string? ping_name = null);
 };
+
+interface BooleanMetric {
```
badboy, line 199: Documentation here will follow in a later commit.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/BooleanMetricType.kt
```diff
@@ -19,106 +13,11 @@ import mozilla.telemetry.glean.rust.toBoolean
...
-     *                 Defaults to the first value in `sendInPings`.
-     * @return value of the stored metric
-     * @throws [NullPointerException] if no value is stored
-     */
-    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
-    @JvmOverloads
-    fun testGetValue(pingName: String = sendInPings.first()): Boolean {
-        @Suppress("EXPERIMENTAL_API_USAGE")
-        Dispatchers.API.assertInTestingMode()
+typealias BooleanMetricType = mozilla.telemetry.glean.internal.BooleanMetric
```
travis79, line 17: I love how thin the metric type definitions are with this!

glean-core/src/glean.udl
```diff
@@ -195,3 +195,11 @@ interface TimespanMetric {
...
+    constructor(CommonMetricData meta);
+
+    void set(boolean value);
+
+    boolean? test_get_value(optional string? ping_name = null);
```
travis79, line 204: Does this need a function to return the number of recorded errors also?

travis79 (approved):
r+wc

glean-core/src/glean.udl
badboy: There are no errors for boolean metrics!
(well, technically there are for labeled booleans and because of how this is build it's required later, but that comes with the labeled ones)

PR #1908, opened 2021-12-15
Bug 1745908 - Python: Copy the environment when launching the upload process

badboy:
The previous fix set the `PYTHONPATH` for the subprocess.
That overwrites the complete environment, clearing out any other
variables.
That can cause problems when other parts rely on something, such as the
RNG in Python relying on `SYSTEMROOT`


Dexterp37 (approved):
Ha! Whoops.

Dexterp37:
@badboy would you kindly add a changelog entry?

badboy:
> @badboy would you kindly add a changelog entry?

I'll follow this up with a release, I'll do it there.

PR #1915, opened 2021-12-28
Upgrade cc crate

glandium:
When building the wheel on platforms that don't come with a pre-built
wheel, it may fail to build because of issues that were fixed in recent
versions of the cc crate.


badboy:
@glandium Which issue was that?

badboy:
(close-reopen to trigger CI))

badboy (approved):

glandium:
The one I hit personally was https://github.com/alexcrichton/cc-rs/pull/627

PR #1918, opened 2022-01-06
Bug 1748858: Log with Python successfully updates a ping

mdboom:


badboy:
In theory there is logging [internally of glean core](https://github.com/mozilla/glean/blob/3e839a38f11525064e432b39000804bed19e6985/glean-core/src/upload/mod.rs#L652).
It doesn't have the same details though.
Is the logging not setup correctly to enable Glean logs in that subprocess?

mdboom:
So, it seems the root of the issue is that we don't get log messages from the subprocess on which pings are sent, because the log configuration isn't copied over to the subprocess.  Normally, this would be handled by `fork`, but we don't `fork` because, Windows.  Even if we don't try to get the log messages from Rust and use this PR as-is, we would have this problem  Unfortunately, it's doesn't seem the Python `logging` module has a way to copy a configuration to another process, which isn't surprising that it's sort of a deeply stateful, highly flexible sort of API.  We could probably create an API to just "turn on all Glean logging" and forgo the complexity of the Python logging API as a stopgap (even just following the `GLEAN_LOG_PINGS` environment variable we already have).  But the better solution is probably to shuttle *all* log messages from the subprocess to the main process and handle them there -- maybe there's something off-the-shelf to handle that general problem.

mdboom:
I don't think there's any way to shuttle the log messages from the subprocess to the main process that wouldn't keep the main process alive the entire time (which is the main point of the subprocess).

So, added a simple API to set the logging level that we can pass to the subprocess, which I think will be sufficient for the Burnham use case.

badboy (approved):
code wise this looks ok.
I'd like @hackebrot's look on that as a python pro (and glean consumer)

docs/user/user/debugging/python.md
```diff
@@ -13,12 +13,25 @@ such as `pings.custom_ping.submit()`, will send it.
...
+
+### Simple logging API
+
+You can set the logging level for Glean log messages to `DEBUG` as follows:
+
+```python
+import logging
+from glean import Glean
+
+Glean.set_log_level(logging.DEBUG)
```
hackebrot, line 29: As a consumer of the Glean Python package, I would find it useful to set the logging level for the simple logging API on Glean initialization. Either as a keyword argument to `initialize` or on the `Configuration` object. This way we would only have 1 call for Glean configuration instead of 2.

```python
Glean.initialize(
    application_id=__title__,
    application_version=__version__,
    upload_enabled=enable_telemetry is True,
    data_dir=Path(TemporaryDirectory().name),
    logging_level=logging.DEBUG,
    configuration=Configuration(server_endpoint=platform),
)
```

Or

```python
Glean.initialize(
    application_id=__title__,
    application_version=__version__,
    upload_enabled=enable_telemetry is True,
    data_dir=Path(TemporaryDirectory().name),
    configuration=Configuration(
        server_endpoint=platform, logging_level=logging.DEBUG
    ),
)
```

As opposed to:

```python
Glean.initialize(
    application_id=__title__,
    application_version=__version__,
    upload_enabled=enable_telemetry is True,
    data_dir=Path(TemporaryDirectory().name),
    configuration=Configuration(server_endpoint=platform),
)
Glean.set_log_level(logging.DEBUG)
```

hackebrot (changes requested):
Thank you for tagging me!

I added my comment on the API to the docs inline. 📋

PR #1928, opened 2022-01-14
Apply clippy beta suggestions

badboy:


badboy:
Beta tests are fixed. Merging without review.

PR #1936, opened 2022-01-18
[UniFFI] Migrate metrics: string, string list, url, uuid, remove jwe

badboy:
Finally getting around to split up #1903 into smaller chunks.

These commits can be reviewed individually.


travis79 (approved):
Goodbye JWE...

PR #1940, opened 2022-01-21
[UniFFI] Re-enable most tests

badboy:
This is the second-to-last batch split out of the mega PR.
(The last one will be docs)


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -798,8 +793,8 @@ class GleanTest {
         // real-world async behaviour of this.
         // We don't need to care about clearing it,
         // the test-unit hooks will call `resetGlean` anyway.
-        Dispatchers.API.setTaskQueueing(true)
-        Dispatchers.API.setTestingMode(false)
+        //Dispatchers.API.setTaskQueueing(true)
+        //Dispatchers.API.setTestingMode(false)
```
travis79, line 797: Commented out and leaving it in?

travis79 (approved):
Glad to see all the `@Suppress("EXPERIMENTAL_API_USAGE")` going away!

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
badboy: this can go.

badboy: Ah! That's in the disabled test anyway. I left it in because if we ever re-enable that test it is good to know that we _used_ to require task qeueuing but not test mode.

PR #1942, opened 2022-01-25
No bug - Change path to deletion-request ping page

brizental:
To keep it the same as the default ping urls generated by
glean_parser. Ping names don't have underscores, but hyphens.

Reported in: https://github.com/mozilla-rally/facebook-pixel-hunt/pull/96/files


badboy (approved):
It once had an underscore! *runs away*

PR #1943, opened 2022-01-25
Don't generate Carthage archive anymore

badboy:


badboy:
> (Carthage delenda est - sorry, couldn't resist)

ehehe, we're not there yet, but once Firefox iOS switches we can drop it and not test that code path anymore.
(In a second step we can change our own Carthage deps as well)

badboy:
Firefox iOS merged SPM support: https://github.com/mozilla-mobile/firefox-ios/pull/9507
That means there's no more Carthage consumer of Glean.

travis79 (approved):
Carthago delenda est!

PR #1954, opened 2022-02-08
Bug 1753194 - Enable support for expiration by version on Android

Dexterp37:
This bumps the version of glean_parser currently being used by Glean and enables Android products to use expiration by version. To make sure things actually work, we change the Android sample to use expiration by version instead of a weird date.

### DIFF
--- a/docs/user/language-bindings/android/android-build-configuration-options.md
+++ b/docs/user/language-bindings/android/android-build-configuration-options.md
@@ -66,3 +66,15 @@ To override this, `ext.gleanYamlFiles` may be set to a list of explicit paths.
 ```groovy
 ext.gleanYamlFiles = ["$rootDir/glean-core/metrics.yaml", "$rootDir/glean-core/pings.yaml"]
 ```
+
+## `gleanExpireByVersion`
+
+Expire the metrics and pings by version, using the provided major version.
+
+If enabled, expiring metrics or pings by date will produce an error.
+
+```groovy
+ext.gleanExpireByVersion = 25
+```
+
+
--- a/docs/user/language-bindings/android/android-build-configuration-options.md
+++ b/docs/user/language-bindings/android/android-build-configuration-options.md
@@ -77,4 +77,6 @@ If enabled, expiring metrics or pings by date will produce an error.
 ext.gleanExpireByVersion = 25
 ```
 
-
+Different products have different ways to compute the product version at build-time.
+For this reason the Glean Gradle plugin cannot provide an automated way to detect the product major version at build time.
+When using the expiration by version feature in Android, products must provide the major version by themselves.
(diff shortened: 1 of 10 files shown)


Dexterp37:
This is the error that shows up when using dates instead of versions in the registry files if expiration by version is turned on:

immagine

docs/user/language-bindings/android/android-build-configuration-options.md
```diff
@@ -66,3 +66,15 @@ To override this, `ext.gleanYamlFiles` may be set to a list of explicit paths.
...
+
+Expire the metrics and pings by version, using the provided major version.
+
+If enabled, expiring metrics or pings by date will produce an error.
+
+```groovy
+ext.gleanExpireByVersion = 25
+```
+
+
```
badboy, line 80:
```suggestion
```

too many newlines

samples/android/app/build.gradle
```diff
@@ -17,6 +17,8 @@ android {
         applicationId "org.mozilla.samples.gleancore"
         minSdkVersion rootProject.ext.build['minSdkVersion']
         targetSdkVersion rootProject.ext.build['targetSdkVersion']
+        // Keep the version in sync with gleanExpireByVersion which is
+        // set below, near the end of the file.
         versionCode 1
```
badboy, line 22: I hoped you pick something "fun" :(

samples/android/app/build.gradle
```diff
@@ -57,6 +59,9 @@ dependencies {
 ext.gleanNamespace = "mozilla.telemetry.glean"
 // Fixed build date so we can test for it
 ext.gleanBuildDate = "2020-11-06T11:30:50+00:00"
+// Enable expiration by major version. The sample is at version 1,
+// see versionName.
+ext.gleanExpireByVersion = 1
```
badboy, line 64: Will we have some guidance how this value should be fetched from build variables at build time?

samples/android/app/build.gradle
Dexterp37: Yeah, my day turned out to be "fun fun fun" :-P

samples/android/app/build.gradle
badboy: well, if we want Fenix to use that we probably have to figure out how they can set it, don't we?

build.gradle
```diff
@@ -135,7 +135,7 @@ switch (DefaultPlatform.RESOURCE_PREFIX) {
         ext.nativeRustTarget = 'linux-x86-64'
         break
     case 'win32-x86-64':
-        ext.nativeRustTarget = 'win32-x86-64-gnu'
+        ext.nativeRustTarget = 'win32-x86-64-msvc'
```
badboy, line 138: this breaks CI

badboy: Let's revert that and maybe open a bug so we can handle that afterwards?

Dexterp37: Eugh, sorry, this was not intentional. Let me revert changes to this file.

PR #1961, opened 2022-02-15
On shutdown kill the dispatcher if Glean failed to initialize

badboy:


Dexterp37 (dismissed):
Good catch. Please add a changelog entry.

badboy:
> Good catch. Please add a changelog entry.

This is UniFFI. `main` is doing the right thing.
CI will continue to be broken for a bit for UniFFI.

Dexterp37:
Gosh, sorry :\

badboy:
Merging without review. It unblocks the test!

PR #1963, opened 2022-02-16
Document that Python can send the baseline ping

badboy:
The events pings is sent on startup and on the API call.

[doc only]

---

I think marking "not sent" explicitly with ❌ makes things more obvious.


brizental (approved):
Ah, so this is why I was under the imporession Python did not send these pings.

badboy:
> Ah, so this is why I was under the imporession Python did not send these pings.

That's what made me look what's actually documented! :D

PR #1964, opened 2022-02-16
 UniFFI & Python - Part 2 of 3

badboy:
Batch 2, now with all metric types migrated.
Some require additional wrapping, others are just using the UniFFI-exported types now.

Side note: this breaks documentation. :/ 
Tests will not yet pass because further cleanup is required, that's coming in Part 3.
Commits are split by metric type.


travis79 (approved):
🎉

PR #1967, opened 2022-02-18
Bug 1755651 - Enable native building & testing on Apple M1 machines

badboy:


travis79 (approved):
Should we also mention the `rustup target add aarch64-apple-darwin` command in the dev docs (on setting up your environment)?

badboy:
> Should we also mention the `rustup target add aarch64-apple-darwin` command in the dev docs (on setting up your environment)?

That's only necessary for cross-compilation. If you're on an M1 that's your host target already so just setting up Rust will have that installed.
On x86_64 macos there's next to no reason to build for aarch64-apple-darwin.

PR #1973, opened 2022-03-04
Bug 1733365 - Document testResetGlean on JavaScript SDK + initialize page restructure

brizental:
[Rendered](https://delirious-match.surge.sh/reference/general/initializing.html)


docs/user/reference/general/initializing.md
```diff
@@ -223,22 +247,136 @@ Glean.initialize(
...
+describe("myTestSuite", () => {
+    beforeEach(async () => {
+        await testResetGlean("my-test-id");
+    });
+});
+
+```
+
+</div>
+<div data-lang="Firefox Desktop" class="tab"></div>
```
brizental, line 373: @chutten I am assuming testing Glean metrics in Firefox Desktop means Glean resetting is handled internally, is that correct? Would a message such as "On Firefox Desktop Glean resetting is handled internally." be correct here?

chutten: Accccctttually, we just implemented `Services.fog.testResetFog()` for test reset capability (cf [bug 1741674](https://bugzilla.mozilla.org/show_bug.cgi?id=1741674)). You can read all about it in [our brand new instrumentation testing docs](https://firefox-source-docs.mozilla.org/toolkit/components/glean/user/instrumentation_tests.html).

What I missed was linking heavily to that doc from the `Firefox Desktop` tabs in all the metric types. Want me to file a bug for that?

brizental: Nah, we do not mention resetting for any of the SDKs on the metric types pages. Actually this PR is the first time resetting is documented in the reference part of the book. Before it was mentioned in the "testing metric" guide only.

brizental: Back to "what I should add here", maybe the tooltip can say "Resetting in FOG is documented on <this link you just sent me>"? WDYT?

chutten: Sounds good to me.

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
 # Initializing
 
-The following steps are required for applications using the Glean SDK, but not libraries.
+Glean needs to be initialized in order to be able to send pings,
+record metrics and perform maintenance tasks. Thus is is advised that Glean be initialized as
```
badboy, line 4:
```suggestion
record metrics and perform maintenance tasks. Thus it is advised that Glean be initialized as
```

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
+The number of calls dropped, if any,
+is recorded in the `glean.error.preinit_tasks_overflow` metric.
+
+## Behavior once initialized
+
+### When upload is enabled
+
+Once initialized, if upload is enabled,
+Glean will automatically start applying enqueued tasks, such as
+collection of [baseline metrics](../../user/pings/metrics.md).
```
badboy, line 31: We shouldn't call them "baseline" metrics or people will confuse that with the baseline ping.
I think it's a bit confusing to keep talking about "tasks" here, that's not a concept exposed to users.
Additionally this links to one specific ping. Maybe we can rephrase that a bit.

```suggestion
Once initialized, if upload is enabled,
Glean applies all metric recordings, for both user-defined metrics and builtin metrics.
This always happens off the main thread.
```

(not sure about the last sentencen, but I guess it's good to call it out once?)

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
+
+### When upload is disabled
+
+If upload is disabled, any persisted metrics, events and pings (other than `first_run_date` and `first_run_hour`) are cleared, any pending deletion-request is sent and subsequent calls to record metrics and submit pings will be no-ops. Because Glean does that as part of its initialization, users are _required_ to always initialize Glean.
```
badboy, line 35: Add some newlines for easier editing and a bit of restructuring.

```suggestion
If upload is disabled, any persisted metrics, events and pings (other than `first_run_date` and `first_run_hour`) are cleared.
Pending `deletion-request` pings are sent.
Subsequent calls to record metrics and submit pings will be no-ops.
Because Glean does that as part of its initialization, users are _required_ to always initialize Glean.
```

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
+
+**Glean must be initialized even if upload is disabled**.
```
badboy, line 37:
```suggestion
**Glean must be initialized even if upload is disabled.**
```

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
+May only be called once. Subsequent calls to `initialize` are no-op.
+
+#### Configuration
+
+The available initialize configuration options may vary depending on the SDK.
+Below are listed the configuration options available on most SDKs.
+
+- `applicationId`: Application identifier. For Android and iOS applications, this is the id used on the platform's respective app store and is extracted automatically from the application context.
+- `uploadEnabled`: The user preference on whether or not data upload is enabled.
+- `appChannel`: The application's release channel. When present, the `app_channel` will be reported on all pings [`client_info`](../../user/pings/index.html#the-client_info-section) section.
```
badboy, line 57:
```suggestion
- `appChannel`: The application's release channel. When present, the `app_channel` will be reported in all ping's [`client_info` section](../../user/pings/index.html#the-client_info-section).
```

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
+- `appBuild`: A build identifier e.g. the build identifier generated by a CI system (e.g. "1234/A"). If not present, `app_build` will be reported as "Unknown" on all pings [`client_info`](../../user/pings/index.html#the-client_info-section) section.
+- `appDisplayVersion`: The user visible version string for the application running Glean. If not present, `app_display_version` will be reported as "Unknown" on all pings [`client_info`](../../user/pings/index.html#the-client_info-section) section.
+- `serverEndpoint`: The server pings are sent to. Defaults to `https://incoming.telemetry.mozilla.org`.
+
+To learn about SDK specific configuration options available, refer to the [Reference](#reference) section.
 
 {{#include ../../../shared/blockquote-stop.html}}
 
-##### Initialize Glean with the correct value for `uploadEnabled`!
+##### Always initialize Glean with the correct value upload preference
```
badboy, line 66:
```suggestion
##### Always initialize Glean with the correct upload preference
```

docs/user/reference/general/initializing.md
```diff
@@ -1,39 +1,80 @@
...
 
-> `Glean.initialize` must **always** be called with real values.
-> Always pass the user preference, e.g. `Glean.initialize(upload=userSettings.telemetry_enabled)` or the equivalent for your application.
-> Never call `Glean.initialize(upload=true)` if `true` is a placeholder value that later gets reset by `Glean.setUploadEnabled(false)`.
+> Glean must **always** be initialized with real values.
+>
+> Always pass the user preference, e.g. `Glean.initialize(upload=userSettings.telemetry_enabled)` or the equivalent for your application. Never call `Glean.initialize(upload=true)` if `true` is a placeholder value that later gets reset by `Glean.setUploadEnabled(false)` and vice-versa.
```
badboy, line 70:
```suggestion
> Always pass the user preference, e.g. `Glean.initialize(upload=userSettings.telemetry_enabled)` or the equivalent for your application.
> Never call `Glean.initialize(upload=true)` if `true` is a placeholder value that later gets reset by `Glean.setUploadEnabled(false)` and vice-versa.
```

badboy (changes requested):
Couple of inline notes.
All in all this is looking good though.

docs/user/reference/general/initializing.md
brizental: This is actually just copy pasted from what is there already, but I agree with the changes. Just need to change the "off the main thread" call, because that does not apply to Javascript. I'll send these changes in a follow-up commit.

badboy (approved):

PR #1975, opened 2022-03-10
Rename 'src -> vcs' in hardcoded paths in CI tasks [ci full]

ahal:
The previous commit updated Taskgraph to the latest version, which
contained a backwards incompatible change (introduced last summer) that
renamed the default checkout path from 'src' -> 'vcs'. This is causing
the build to fail.

Since this backwards incompatible change was introduced, we've since
started releasing to Pypi with proper version numbers and are now
following SemVer. So hopefully cases like this won't happen again.

(Ironically preventing cases like this is what the PR was trying to
solve).

### DIFF
--- a/.taskcluster.yml
+++ b/.taskcluster.yml
@@ -234,8 +234,8 @@ tasks:
 
                             command:
                                 - /usr/local/bin/run-task
-                                - '--glean-checkout=/builds/worker/checkouts/src'
-                                - '--task-cwd=/builds/worker/checkouts/src'
+                                - '--glean-checkout=/builds/worker/checkouts/vcs'
+                                - '--task-cwd=/builds/worker/checkouts/vcs'
                                 - '--'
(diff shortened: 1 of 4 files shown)


ahal:
I'm hoping this will fix the bustage currently on main, but we should test out the failing task before landing. I'm also ok with backing out the PR if that's easier.

badboy:
and here I wondered if I should force a full CI run on that other PR, and then didn't. Let's do it for this.

badboy:
close & reopen should trigger the decision task again.

ahal:
Yeah thanks, a full CI run is definitely valuable here. Apologies, yours is the first repo I've hit this with, though it will happen for any repo who hasn't updated their `taskgraph` dependency in a long enough time. I'll be on the lookout going forward :)

badboy (approved):

PR #1981, opened 2022-03-14
Bug 1746183 - Swift: Migrate general API & setup to UniFFI

badboy:
This is part 1 of the big UniFFI migration for Swift.
It should be reviewable commit-by-commit.

The tests don't run yet, that will require changed from part 2 as well.
There's some little things that will be cleaned up in followups (e.g. re-enabling or finally removing tests)


glean-core/ios/GleanTests/GleanTests.swift
```diff
@@ -260,6 +263,8 @@ class GleanTests: XCTestCase {
         }
     }
 
+    /*
+    FIXME: todo
```
travis79, line 267: 🎗️

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
```diff
@@ -2,10 +2,10 @@
  * License, v. 2.0. If a copy of the MPL was not distributed with this
  * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
 
+/*
```
travis79, line 5: And with one fell swoop, @badboy fixes the MPS tests for good!

glean-core/ios/sdk_generator.sh
```diff
@@ -142,7 +142,7 @@ VENVDIR="${SOURCE_ROOT}/.venv"
 # We need at least pip 20.3 for Big Sur support, see https://pip.pypa.io/en/stable/news/#id48
 # Latest pip is 21.0.1
 "${VENVDIR}"/bin/pip install "pip>=20.3"
-"${VENVDIR}"/bin/pip install --upgrade glean_parser==$GLEAN_PARSER_VERSION
+"${VENVDIR}"/bin/pip install --upgrade "git+ssh://git@github.com/badboy/glean_parser@uniffi-preparation#glean-parser"
```
travis79, line 145: 🎗️ to change this back when it's time

travis79 (approved):
Just a couple of reminders (🎗️ ) left in the review, but LGTM!

glean-core/ios/GleanTests/Scheduler/MetricsPingSchedulerTests.swift
badboy: image

PR #1983, opened 2022-03-15
Bug 1711928 - Stop overwriting User-Agent header in favor of X-Telemetry-Agent header

perrymcmanis144:


travis79 (approved):
Looks good to me!

CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * Android
   * The `glean-native-forUnitTests` now ships with separate libraries for macOS x86_64 and macOS aarch64 ([#1967](https://github.com/mozilla/glean/pull/1967))
+* Rust
+  * Glean-Core will no longer overwrite User-Agent, but instead send that information as X-Telemetry-Agent ([bug 1711928](https://bugzilla.mozilla.org/show_bug.cgi?id=1711928))
```
badboy, line 8:
```suggestion
* General
  * Glean will no longer overwrite the `User-Agent` header, but instead send that information as `X-Telemetry-Agent` ([bug 1711928](https://bugzilla.mozilla.org/show_bug.cgi?id=1711928))
```

docs/user/user/pings/index.md
```diff
@@ -261,7 +261,7 @@ Describes the data sent to the server. Value is always `application/json; charse
 
 Submission date/time in GMT/UTC+0 offset, e.g. `Mon, 23 Jan 2019 10:10:10 GMT+00:00`.
 
-#### `User-Agent`
+#### `User-Agent` (deprecated)
```
badboy, line 264: Let's change the text below:

```
Up to Glean v44.0.0 and Glean.js v0.13.0 this contained the Glean SDK version and platform information.
Newer Glean SDKs do not overwrite this header.
See [`X-Telemetry-Agent`](#x-telemetry-agent) for details.

Clients might still send it, for example, when sending pings from browsers
it will contain the characteristic [browser UA string](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/User-Agent#syntax).

This header is parsed [...]
```

And then also remove the footnote at the end of this doc.

badboy (changes requested):

docs/user/user/pings/index.md
perrymcmanis144: 👍 I believe this should be resolved now, assuming I understood your request on the footnote correctly (thinking it was the "this header is only sent by the Glean JS SDK..." one)

badboy (approved):

PR #1986, opened 2022-03-16
Migrate all Swift metrics to UniFFI

badboy:
This migrates every single metric implemenation to UniFFI.
It was supposed to also fix the tests, but they remain broken and I'll take care of that in followups (hopefully).


travis79 (approved):
Nice!!

PR #1989, opened 2022-03-17
Glean Swift on UniFFI - More test fixes

badboy:
Sometimes all tests pass, sometimes they don't.
It's coming down to race conditions of the uploader and when we shutdown the Glean object.

This will need another proper look to ensure our tests are good enough, but for now this can get some review.

Based on #1986, will auto-update the branch when that is merged.


glean-core/ios/Glean/Glean.swift
```diff
@@ -355,6 +355,7 @@ public class Glean {
         gleanSetTestMode(false)
         self.testingMode.value = false
         self.initialized = false
+        self.metricsPingScheduler = nil
```
travis79, line 358: Aha!

travis79 (approved):

PR #2007, opened 2022-04-06
Change glean-sample-app to use Swift Packages

travis79:
I did this to help get this working on my M1


badboy:
I want this change!

This has the unfortunate downside that it makes local testing much harder and testing on CI near impossible until we figure out how to replace the package there.

travis79:
> I want this change!
> 
> This has the unfortunate downside that it makes local testing much harder and testing on CI near impossible until we figure out how to replace the package there.

:D I meant to make this a draft, I was sharing this with @perrymcmanis144 to try and figure out why he cannot build on his M1

PR #2013, opened 2022-04-12
Bug 1753975 - Support expiration by major version on iOS

badboy:
It's up to the embedding application to provide the correct version
number.


chutten (approved):
LGTM

PR #2014, opened 2022-04-13
CI: Migrate away from CircleCI legacy convenience images

badboy:
See https://discuss.circleci.com/t/legacy-convenience-image-deprecation for details


badboy:
Timings look pretty similar to the old images.

PR #2015, opened 2022-04-13
bug 1764549 - Bump the RLB preinit queue from 10^2 to 10^3

chutten:
FOG's on Desktop and is init whenever BrowserGlue gets around to it.
We have the memory to play with.

This is a temporary measure until we make this configurable or limitless.

### DIFF
--- a/glean-core/rlb/src/dispatcher/global.rs
+++ b/glean-core/rlb/src/dispatcher/global.rs
@@ -7,7 +7,7 @@ use std::sync::RwLock;
 
 use super::{DispatchError, DispatchGuard, Dispatcher};
 
-pub const GLOBAL_DISPATCHER_LIMIT: usize = 100;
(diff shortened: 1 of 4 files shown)


glean-core/rlb/src/dispatcher/global.rs
```diff
@@ -7,7 +7,7 @@ use std::sync::RwLock;
 
 use super::{DispatchError, DispatchGuard, Dispatcher};
 
-pub const GLOBAL_DISPATCHER_LIMIT: usize = 100;
+pub const GLOBAL_DISPATCHER_LIMIT: usize = 1000;
```
badboy, line 10: There's a test for this: `glean-core/rlb/tests/overflowing_preinit.rs`

chutten: Well, poo. Shows me for just redressing a quick'n'dirty patch as a proper change.

badboy (approved):

PR #2020, opened 2022-04-25
Use integer keys in distribution data

badboy:


glean-core/src/lib_unit_tests.rs
```diff
@@ -751,12 +751,11 @@ fn timing_distribution_truncation() {
...
         let mut recorded_values = 0;
 
-        for (key, &value) in &snapshot.values {
-            let key = key.parse().unwrap();
+        for (&key, &value) in &snapshot.values {
             // A snapshot potentially includes buckets with a 0 count.
             // We can ignore them here.
             if value > 0 {
-                assert!(key < max_sample_time * unit.as_nanos(1));
+                assert!((key as u64) < max_sample_time * unit.as_nanos(1));
```
travis79, line 758: We are using signed integers in other places, is the `u64` intentional here?

travis79 (approved):

glean-core/src/lib_unit_tests.rs
badboy: So yes. It's a bit weird, but it is correct.
We use u64 internally for e.g. the time difference in a timing dist. But for compatibility we use `i64` to the outside (`Long` in Kotlin).
We loose a bit of precision, but we already don't have that currently.

PR #2025, opened 2022-04-26
Set user-agent again.

badboy:
We still want to send _some_ User-Agent.
Uploader implementations can in theory overwrite this without problems.


chutten (changes requested):
Tests? Docs? Changelog?

badboy:
> Tests? Docs? Changelog?

We're not even landing this!

PR #2036, opened 2022-05-03
Android Sample: Put metrics in the custom ping and add a send button

badboy:


samples/android/app/src/main/java/org/mozilla/samples/glean/MainActivity.kt
```diff
@@ -18,7 +20,7 @@ open class MainActivity : AppCompatActivity() {
         super.onCreate(savedInstanceState)
         binding = ActivityMainBinding.inflate(layoutInflater)
 
-        setContentView(R.layout.activity_main)
+        setContentView(binding.root)
```
chutten, line 23: I don't know what this change does.

badboy: Neither do I! But that's what [a-c does and that one runs on CI](https://github.com/mozilla-mobile/android-components/blob/89c54fb2c9e0f35c996cc29a2599e5c770247cb0/samples/glean/src/main/java/org/mozilla/samples/glean/MainActivity.kt#L28).

To explain a bit more: [4 weeks ago I removed an outdated android dependency](https://github.com/mozilla/glean/commit/0250cecc903b050d2efbf82028cf3dfeca29249d) and adopted the code that a-c also has, but missed that one line. Now testing locally pressing any button didn't do anything. Only after I changed that line the button handler started working again.

chutten: :shipit:

PR #2044, opened 2022-05-05
Use a tooling crate to run uniffi-bindgen and builtin bindgen in build.rs

badboy:
That's adopted from application-services once more.


travis79 (approved):
The builtin-bindgen made things better on the appservices build, if I recall.

travis79:
:D `TimerIds` everywhere

PR #2048, opened 2022-05-10
Implement Rate, Denominator and Numerator using UniFFI

badboy:
…Swift


glean-core/src/glean.udl
```diff
@@ -498,3 +498,40 @@ interface EventMetric {
 
     i32 test_get_num_recorded_errors(ErrorType error, optional string? ping_name = null);
 };
+
+dictionary Rate {
+    i32 numerator;
+    i32 denominator;
```
chutten, line 504: How I long for the day we can use unsigned types

glean-core/src/metrics/denominator.rs
```diff
@@ -81,15 +90,51 @@ impl DenominatorMetric {
     /// Gets the currently stored value as an integer.
     ///
     /// This doesn't clear the stored value.
-    pub fn test_get_value(&self, glean: &Glean, storage_name: &str) -> Option<i32> {
+    pub fn test_get_value(&self, ping_name: Option<String>) -> Option<i32> {
+        crate::block_on_dispatcher();
```
chutten, line 94: Wait, uniffi brings the dispatcher to glean-core? Nice.

glean-core/src/metrics/denominator.rs
```diff
@@ -81,15 +90,51 @@ impl DenominatorMetric {
...
+        crate::core::with_glean(|glean| self.get_value(glean, ping_name.as_deref()))
+    }
+
+    #[doc(hidden)]
+    pub fn get_value<'a, S: Into<Option<&'a str>>>(
```
chutten, line 99: Is there a reason `test_get_value` doesn't follow the `_sync` suffix pattern of the other methods? Also, I'm a tad worried about someone seeing a method without a `test_` prefix and deciding to use it in non-test code.

glean-core/src/metrics/numerator.rs
```diff
@@ -0,0 +1,98 @@
...
+use crate::metrics::MetricType;
+use crate::metrics::Rate;
+use crate::metrics::RateMetric;
+use crate::CommonMetricData;
+use crate::Glean;
+
+// We need to wrap the glean-core type: otherwise if we try to implement
+// the trait for the metric in `glean_core::metrics` we hit error[E0117]:
+// only traits defined in the current crate can be implemented for arbitrary
+// types.
```
chutten, line 17: ...but we're _in_ glean_core::metrics....

badboy: Copy'n'pasta. Good point, gonna fix that.

glean-core/src/metrics/denominator.rs
badboy: Yeah, so this is all .. not perfect right now. I have this pattern across several places right now because of our current test layout.
Let me create a followup bug to address this.

glean-core/src/metrics/denominator.rs
badboy: Yes!

glean-core/src/glean.udl
badboy: *side-eyes Kotlin*

glean-core/src/metrics/denominator.rs
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1768832

glean-core/src/glean.udl
chutten: [a-c's up to Kotlin 1.6.10](https://github.com/mozilla-mobile/android-components/pull/12000), unsigned types are stable [as of 1.5](https://kotlinlang.org/docs/whatsnew15.html#standard-library) ......

badboy: unsigned types are. literals are signed though, and the compiler will complain if you pass it `1` where it expects something unsigned.

BUT ... [turns out literals can now have a `u` suffix](https://kotlinlang.org/docs/basic-types.html#literals).

Time to revisit our APIs, but maybe _after_ this.

chutten (approved):

PR #2054, opened 2022-05-18
Kotlin: Use the library name `xul` for the core parts

badboy:
This is a hack:
Glean is build into libxul from mozilla-central and users consume it as part of GeckoView.
Because UniFFI does not have the option to switch the library to consume,
except at build time, we use this hack for now.


travis79 (approved):
Interesting!

PR #2064, opened 2022-05-23
Bug 1770791 - Remove duplicate initialization in python glean package

arai-a:
for [bug 1770791](https://bugzilla.mozilla.org/show_bug.cgi?id=1770791)

On Windows, Creating `FdLogger` twice causes the subsequent `subprocess` operation somehow fail.
And apparently the duplicate initialization is completely unnecessary.
So, applied 2 fixes, both on Rust side and Python side.

Rust side's `glean_enable_logging_to_fd` uses `OnceCell` to prevent subsequent call overwriting the value,
but `fd_logger::FdLogger::new(fd)` is not guarded.
So, it should check `FD_LOGGER.get().is_some()` before `fd_logger::FdLogger::new(fd)`.

Python side calls `setup_logging` twice, first time from `_ffi.py` top-level, and second time from `glean.py` top-level.
Either of them should be removed, and I've removed the former one.

The attached testcase fails on windows, with `py.test -s tests/test_subprocess.py`.
the `-s` option seems to be necessary to hit the problematic case.

### DIFF
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -889,13 +889,12 @@ pub fn glean_enable_logging_to_fd(fd: u64) {
         // language binding should setup a pipe and pass in the descriptor to
         // the writing side of the pipe as the `fd` parameter. Log messages are
         // written as JSON to the file descriptor.
-        if FD_LOGGER.set(fd_logger::FdLogger::new(fd)).is_ok() {
-            // Set the level so everything goes through to the language
-            // binding side where it will be filtered by the language
-            // binding's logging system.
-            if log::set_logger(FD_LOGGER.get().unwrap()).is_ok() {
-                log::set_max_level(log::LevelFilter::Debug);
-            }
+        let logger = FD_LOGGER.get_or_init(|| fd_logger::FdLogger::new(fd));
+        // Set the level so everything goes through to the language
+        // binding side where it will be filtered by the language
+        // binding's logging system.
+        if log::set_logger(logger).is_ok() {
+            log::set_max_level(log::LevelFilter::Debug);
         }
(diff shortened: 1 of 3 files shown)


glean-core/src/lib.rs
```diff
@@ -885,6 +885,10 @@ static FD_LOGGER: OnceCell<fd_logger::FdLogger> = OnceCell::new();
 pub fn glean_enable_logging_to_fd(fd: u64) {
     // SAFETY: TODO.
     unsafe {
+        // If the logger is already set, do not create new logger.
+        if FD_LOGGER.get().is_some() {
+            return;
+        }
```
badboy, line 891: Seems like we should just make use of `OnceCell`'s `get_or_init` for exactly that purpose.

arai-a: `get_or_init` returns the reference to the cell's content, regardless of whether the cell was empty or not.
https://docs.rs/once_cell/latest/once_cell/sync/struct.OnceCell.html#method.get_or_init
So, the remaining operation (`log::set_logger` etc) cannot be written conditional, unless we somehow return the state.

I think the early return is simpler.

arai-a: ah, good to know :)
I'll fix.

badboy:
Turns out our Python Windows tests have been broken for a while. Grmpf.
(See https://app.circleci.com/pipelines/github/mozilla/glean/9332/workflows/86672023-5c22-450d-b756-7904a2b3a1a5/jobs/173298 -- they crash, but don't report that as an error value).

Gonna go with your fix though, and fix CI in a followup.

badboy (approved):

PR #2071, opened 2022-05-27
Bug 1770149 - UniFFI: Adjust docs for all metric types

badboy:
[doc only]

### DIFF
--- a/docs/user/reference/metrics/boolean.md
+++ b/docs/user/reference/metrics/boolean.md
@@ -53,7 +53,7 @@ metrics.flags.a11y_enabled.set(is_accessibility_enabled())
 <div data-lang="Rust" class="tab">
 
 ```Rust
-use glean_metrics;
+use glean_metrics::flags;
 
 flags::a11y_enabled.set(system.is_accessibility_enabled());
 ```
@@ -91,13 +91,15 @@ Glean.flags.a11yEnabled.set(false);
 
 #### Recorded errors
 
-* [`invalid_type`](../../user/metrics/error-reporting.md): if a non-boolean value is given.
+* [`invalid_type`](../../user/metrics/error-reporting.md): if a non-boolean value is given (JavaScript only).
 
 ## Testing API
 
 ### `testGetValue`
 
-Gets the recorded value for a given boolean metric.
+Gets the recorded value for a given boolean metric.  
+Returns `true` or `false` if data is stored.  
+Returns a language-specific empty/null value if no data is stored.
 
 {{#include ../../../shared/tab_header.md}}
 
@@ -125,7 +127,7 @@ assertTrue(Flags.INSTANCE.a11yEnabled.testGetValue());
 <div data-lang="Swift" class="tab">
 
 ```Swift
-XCTAssertTrue(try Flags.a11yEnabled.testGetValue())
+XCTAssertTrue(Flags.a11yEnabled.testGetValue())
 ```
 
 </div>
@@ -144,7 +146,7 @@ assert True is metrics.flags.a11y_enabled.test_get_value()
 <div data-lang="Rust" class="tab">
 
 ```Rust
-use glean_metrics;
+use glean_metrics::flags;
 
 assert!(flags::a11y_enabled.test_get_value(None).unwrap());
 ```
@@ -180,9 +182,9 @@ Assert.equal(false, Glean.flags.a11yEnabled.testGetValue());
 
 {{#include ../../../shared/tab_footer.md}}
 
-### `testHasValue`
+### `testGetNumRecordedErrors`
 
-Whether or not **any** value was recorded for a given boolean metric.
+Gets the number of errors recorded for a given boolean metric.
 
 {{#include ../../../shared/tab_header.md}}
 
@@ -191,9 +193,10 @@ Whether or not **any** value was recorded for a given boolean metric.
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Flags
 
-assertTrue(Flags.a11yEnabled.testHasValue())
+assertEquals(
+    0, Flags.a11yEnabled.testGetNumRecordedErrors(ErrorType.INVALID_VALUE)
+)
 ```
-
 </div>
 
 <div data-lang="Java" class="tab">
@@ -201,16 +204,17 @@ assertTrue(Flags.a11yEnabled.testHasValue())
 ```Java
 import org.mozilla.yourApplication.GleanMetrics.Flags;
 
-assertTrue(Flags.INSTANCE.a11yEnabled.testHasValue());
+assertEquals(
+    0, Flags.INSTANCE.a11yEnabled.testGetNumRecordedErrors(ErrorType.INVALID_VALUE)
+);
 ```
 
 </div>
 
-
 <div data-lang="Swift" class="tab">
 
 ```Swift
-XCTAssertTrue(try Flags.a11yEnabled.testHasValue())
+XCTAssertEqual(0, Flags.a11yEnabled.testGetNumRecordedErrors(.invalidValue))
 ```
 
 </div>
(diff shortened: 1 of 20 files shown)


Dexterp37 (approved):

docs/user/reference/metrics/boolean.md
```diff
@@ -180,9 +182,9 @@ Assert.equal(false, Glean.flags.a11yEnabled.testGetValue());
 
 {{#include ../../../shared/tab_footer.md}}
 
-### `testHasValue`
+### `testGetNumRecordedErrors`
 
-Whether or not **any** value was recorded for a given boolean metric.
+Gets number of errors recorded for a given boolean metric.
```
chutten, line 187:
```suggestion
Gets the number of errors recorded for a given boolean metric.
```

docs/user/reference/metrics/counter.md
```diff
@@ -112,7 +123,9 @@ Increases the counter by a certain amount. If no amount is passed it defaults to
 
 ### `testGetValue`
 
-Gets the recorded value for a given counter metric.
+Gets the recorded value for a given counter metric.  
+Returns the count if data is stored.  
+Returns `null` if no data is stored.
```
chutten, line 128: Hmmmm... what to do about languages where we don't return `null`. Like Rust where we return `None` or Firefox Desktop C++ where we return `mozilla::Nothing()` or Firefox Desktop JS where we return `undefined`...

I mean, is saying "Returns `null`" close enough for folks to get the gist? Should we pedantically list everything it could be? Will it ruin comprehension if we say "Returns `null` or some other empty/nothing value native to the specific platform"?

Hrm.

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -74,7 +97,9 @@ Glean.graphics.checkerboardPeak.accumulateSamples([23])
 
 ### `testGetValue`
 
-Gets the recorded value for a given counter metric.
+Gets the recorded value for a given counter metric.
```
chutten, line 100:
```suggestion
Gets the recorded value for the custom distribution metric.  
```

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -87,21 +112,57 @@ import org.mozilla.yourApplication.GleanMetrics.Graphics
...
+
+```Swift
+// Get snapshot.
+let snapshot = graphics.checkerboardPeak.testGetValue()
 
-<div data-lang="Java" class="tab"></div>
-<div data-lang="Swift" class="tab"></div>
-<div data-lang="Python" class="tab"></div>
+// Usually you don't know the exact timing values,
+// but how many should have been recorded.
```
chutten, line 143: In which case shouldn't we include an example of checking that?

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -87,21 +112,57 @@ import org.mozilla.yourApplication.GleanMetrics.Graphics
...
+<div data-lang="Swift" class="tab">
+
+```Swift
+// Get snapshot.
+let snapshot = graphics.checkerboardPeak.testGetValue()
 
-<div data-lang="Java" class="tab"></div>
-<div data-lang="Swift" class="tab"></div>
-<div data-lang="Python" class="tab"></div>
+// Usually you don't know the exact timing values,
```
chutten, line 142: timing? This is a custom distribution, sir.

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -87,21 +112,57 @@ import org.mozilla.yourApplication.GleanMetrics.Graphics
 val snapshot = Graphics.checkerboardPeak.testGetValue()
 
 // Does the sum have the expected value?
-assertEquals(11, snapshot.sum)
+assertEquals(23, snapshot.sum)
 
 // Usually you don't know the exact timing values, but how many should have been recorded.
```
chutten, line 117:
```suggestion
// Usually you don't know the exact values, but how many should have been recorded.
```

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -87,21 +112,57 @@ import org.mozilla.yourApplication.GleanMetrics.Graphics
...
+<div data-lang="Python" class="tab">
+
+```Python
+from glean import load_metrics
+metrics = load_metrics("metrics.yaml")
+
+# Get snapshot.
+snapshot = metrics.graphics.checkerboard_peak.test_get_value()
+
+# Usually you don't know the exact timing values,
```
chutten, line 157:
```suggestion
# Usually you don't know the exact values,
```

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -132,9 +193,9 @@ Assert.equal(23, data.sum);
 
 {{#include ../../../shared/tab_footer.md}}
 
-### `testHasValue`
+### `testGetNumRecordedErrors`
 
-Whether or not **any** value was recorded for a given counter metric.
+Gets number of errors recorded for a given counter metric.
```
chutten, line 198:
```suggestion
Gets the number of errors recorded for the custom distribution metric.
```

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -187,7 +258,11 @@ use glean_metrics::graphics;
 // Were any of the values negative and thus caused an error to be recorded?
 assert_eq!(
     0,
-    graphics::checkerboard_peak.test_get_num_recorded_errors(ErrorType::InvalidValue));
+    graphics::checkerboard_peak.test_get_num_recorded_errors(
+        ErrorType::InvalidValue,
+        None
```
chutten, line 263: What's this second param for? counter and boolean don't have it in Rust (that I can see)

docs/user/reference/metrics/event.md
```diff
@@ -423,7 +377,10 @@ use metrics::views;
 
 assert_eq!(
     0,
-    views::login_opened.test_get_num_recorded_errors(ErrorType::InvalidOverflow, None)
+    views::login_opened.test_get_num_recorded_errors(
+        ErrorType::InvalidOverflow,
+        None
```
chutten, line 382: Huh, here too with the second argument. Though this one seems to have come from the previous version of the doc.

docs/user/reference/metrics/quantity.md
```diff
@@ -78,13 +78,13 @@ display::width.set(width);
...
 
-mozilla::glean::display::width.Set(innerHeight);
+mozilla::glean::display::width.Set(innerWidth);
 ```
 
 **JavaScript**
 
 ```js
-Glean.display.width.set(innerHeight);
+Glean.display.width.set(innerWidth);
```
chutten, line 87: Nice catches.

docs/user/reference/metrics/quantity.md
```diff
@@ -290,12 +236,28 @@ metrics = load_metrics("metrics.yaml")
...
+
+```Rust
+use glean::ErrorType;
+use glean_metrics::display;
+
+assert_eq!(
+  0,
+  display::width.test_get_num_recorded_errors(
+    ErrorType::InvalidValue,
+    None
```
chutten, line 255: Another second param

docs/user/reference/metrics/timespan.md
```diff
@@ -174,11 +174,11 @@ def on_login():
 </div>
 <div data-lang="Rust" class="tab">
 
-```rust
-use fog::metrics;
+```Rust
+use glean_metrics::auth;;
```
chutten, line 178: Hrm. Does this mean we should rewrite FOG's metrics API generation to appear in a `glean_metrics` crate instead of in the `metrics` mod of the `fog` crate? I wouldn't mind it being consistent to support e.g. standalone WR builds linking to a full Glean SDK outside of gecko build.

chutten (approved):
r+wc

docs/user/reference/metrics/counter.md
badboy: I did briefly think about that and didn't come up with a good solution either.
I went with just `null` for brevity, but I could go with `a language-specific empty/null value`

docs/user/reference/metrics/custom_distribution.md
badboy: coooopy-pasta.

docs/user/reference/metrics/custom_distribution.md
badboy: haha, see. I made a note in my `notes.txt` to go back and fix that everywhere:
`test_get_num_recorded_errors` takes as the second parameter the (optional) ping to look into because ... well, technically you can look in other pings (and on Rust we can't leave it out).

Now the thing is: we don't actually need that: errors _always_ also go to the `metrics` ping on top of whatever ping this metric normally goes in. So we could be safe just removing that and always assuming `metrics`.
That's a bigger change though, so I'm filling a bug for that.
And because of that I'm less motivated to fix it everywhere just to revert it soon after.

docs/user/reference/metrics/timespan.md
badboy: I mostly did _not_ want to make a decision about the precise naming here, but instead go with what we used more already, which was `glean_metrics`.
Now neither of these is really more correct than the other, because in the end it depends on the application.
And heck, because it's a module in every other potential rust consumer other than FOG, this line wouldn't even work (it would need to be `use crate::glean_metrics::...`).

I'd be ok to revisit this in a followup, maybe just fully removing it? or adding comments about it.

docs/user/reference/metrics/custom_distribution.md
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1771705

docs/user/reference/metrics/timespan.md
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=1771707

PR #2073, opened 2022-05-27
Bug 1746181 - Implement the text metric for Kotlin and Swift

badboy:
Note: this is not required by anything really.
But it was so simple to implement I just went ahead.


docs/user/reference/metrics/text.md
```diff
@@ -18,13 +18,54 @@ Sets a text metric to a specific value.
...
+
+</div>
+
+<div data-lang="Rust" class="tab">
 
-<div data-lang="Python" class="tab"></div>
+```Rust
+use glean_metrics::article;
+
+article::conent.set(extracted_text);
```
chutten, line 65:
```suggestion
article::content.set(extracted_text);
```

glean-core/src/metrics/text.rs
```diff
@@ -0,0 +1,178 @@
...
+
+    /// Sets to the specified value.
+    ///
+    /// # Arguments
+    ///
+    /// * `value` - The text to set the metric to.
+    ///
+    /// ## Notes
+    ///
+    /// Truncates the value if it is longer than `MAX_LENGTH_VALUE` bytes and logs an error.
```
chutten, line 70: How? Do we truncate even in the middle of a multi-byte sequence? Or only on character boundaries?

chutten: (yes I know `truncate_string_at_boundary_with_error` rather conveniently answers my questions, but that's not in the docs, that's in the code)

glean-core/src/metrics/text.rs
```diff
@@ -0,0 +1,178 @@
...
+    pub fn set(&self, value: String) {
+        let metric = self.clone();
+        crate::launch_with_glean(move |glean| metric.set_sync(glean, &value))
+    }
+
+    /// Sets to the specified value synchronously.
```
chutten, line 76:
```suggestion
    /// Sets to the specified value synchronously, truncating and recording an error if longer than `MAX_LENGTH_VALUE`.
```

glean-core/src/metrics/text.rs
```diff
@@ -0,0 +1,178 @@
...
+            return;
+        }
+
+        let s = truncate_string_at_boundary_with_error(glean, &self.meta, value, MAX_LENGTH_VALUE);
+
+        let value = Metric::Text(s);
+        glean.storage().record(glean, &self.meta, &value)
+    }
+
+    /// Gets the current-stored value as a string, or None if there is no value.
```
chutten, line 89:
```suggestion
    /// Gets the currently-stored value as a string, or None if there is no value.
```

glean-core/tests/text.rs
```diff
@@ -0,0 +1,121 @@
...
+
+use serde_json::json;
+
+use glean_core::metrics::*;
+use glean_core::storage::StorageManager;
+use glean_core::{test_get_num_recorded_errors, ErrorType};
+use glean_core::{CommonMetricData, Lifetime};
+
+// SKIPPED from glean-ac: string deserializer should correctly parse integers
+// This test doesn't really apply to rkv
```
chutten, line 16: I don't understand this comment.

glean-core/tests/text.rs
```diff
@@ -0,0 +1,121 @@
...
+
+        assert_eq!(
+            json!({"text": {"telemetry.text_metric": "test_text_value"}}),
+            snapshot
+        );
+    }
+}
+
+// SKIPPED from glean-ac: strings are serialized in the correct JSON format
+// Completely redundant with other tests.
```
chutten, line 92: This too

chutten (approved):
r+wc

PR #2076, opened 2022-05-30
CI: Use Rust 1.61.0 everywhere

badboy:
Let's try this again.


badboy:
Now that timing is back to normal I'm merging this.

PR #2079, opened 2022-06-02
Update docs on how to add a new metric type to Glean

badboy:
[doc only]


docs/dev/core/new-metric-type.md
```diff
@@ -165,4 +186,6 @@ Tests are written in the Language Bindings and tend to just cover basic function
 
 ---
 
-In the next step we will create the FFI wrapper and platform-specific wrappers.
+At this point the metric type will have an auto-generated API in all target languages.
+This needs to be re-exported in the target language.
+For additional modifications to the API see the following chapters.
```
Dexterp37, line 191: I believe we should mention somewhere what things we watch our for, e.g. the generated API must work well with the recommended IDE for the target language (autocompletion must work, docs should be discoverable, etc.)

Dexterp37 (approved):
r+wc

docs/dev/core/new-metric-type.md
badboy: fwiw docs currently ... suck. That's a thing we're hoping to fix in UniFFI eventually (docs written in the UDL are never exposed anywhere).

PR #2096, opened 2022-06-15
Update env_logger to 0.9 and android_logger to 0.11

glandium:


badboy:
No idea why tests weren't started. TC ran and it has Rust tests, so that's good enough to merge.

PR #2097, opened 2022-06-16
Bug 1774550 - Use single quotes for strings in the Python inline script

badboy:


Dexterp37 (approved):
r+wc. 

I tested this locally in Fenix and it works as expected. Please add a changelog entry before merging.

PR #2102, opened 2022-06-20
Upgrade to uniffi 0.19

badboy:


chutten (approved):
I'd love if the commit message or PR explained why (even if it's just "because keeping up to date is good hygiene")

badboy:
> I'd love if the commit message or PR explained why (even if it's just "because keeping up to date is good hygiene")

Will add it. Right now the 0.19.1 release is broken for us anyway and I need to wait for a new one,.

travis79:
UniFFI version 19.2 seems to work according to my local build and CI. The rust-beta-tests are a [separate problem](https://github.com/mozilla/glean/pull/2107)

badboy:
> UniFFI version 19.2 seems to work according to my local build and CI. The rust-beta-tests are a [separate problem](https://github.com/mozilla/glean/pull/2107). @badboy I can't flag you for review since this is your PR, but if you don't mind taking a look before this is landed (when you are back, of course)

yup, landing this, then we can pull in #2107 as well and get to a release soon.

PR #2103, opened 2022-06-21
Bug 1772566 - Implement a glean-build crate for use in Rust consumers

badboy:
This crate should be used as a build dependency in Rust consumers.
In their `build.rs` they can call the builder to auto-generate code from
`metrics.yaml` and `pings.yaml` definitions:

    glean_build::Builder::default()
        .metrics("metrics.yaml")
        .pings("pings.yaml")
        .generate()
        .expect("Error generating Glean Rust bindings");

This requires a working Python3 & pip installation.
Windows untested.

Fixes https://bugzilla.mozilla.org/show_bug.cgi?id=1713573 and https://bugzilla.mozilla.org/show_bug.cgi?id=1772566

---

This is my first stab at bug 1772566. It seems to work.
Todo:

* [ ] Automate the release

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -62,17 +62,28 @@ commands:
           name: Test
           command: |
             export GLEAN_TEST_COVERAGE=$(realpath glean_coverage.txt)
-            cargo test --all --verbose --jobs 6 -- --nocapture
+            cargo test --verbose --jobs 6 -- --nocapture
       - run:
           name: Test Glean with rkv safe-mode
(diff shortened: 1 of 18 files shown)


travis79 (commented):
"Seems to work" sounds good enough for me to try this out with Nimbus in A-S, I'm having some problems with getting CI to run my Python script that tries to do this there.

badboy:
@travis79 will need another r+ after taking it out of draft mode

PR #2105, opened 2022-06-24
Bug 1756153 - Put Glean initialization on a async Kotlin task

badboy:
This should delay loading the bundled library, which can be quite time
consuming itself.

---

This restores the minimal functionality from our previous Dispatcher.
I opted for that way because we know it works.

I have _not_ tested this in an app yet. In our tests we rely on test mode -> synchronous, so we're not exercising that code path. The sample app though should use that code path. We don't run sample app tests in CI though.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -175,19 +175,23 @@ open class GleanInternalAPI internal constructor() {
         this.httpClient = BaseUploader(configuration.httpClient)
         this.gleanDataDir = File(applicationContext.applicationInfo.dataDir, GLEAN_DATA_DIR)
 
-        val cfg = InternalConfiguration(
-            dataPath = gleanDataDir.path,
-            applicationId = applicationContext.packageName,
-            languageBindingName = LANGUAGE_BINDING_NAME,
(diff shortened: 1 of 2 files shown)


perrymcmanis144 (approved):
LGTM as changes reflect stated intent 👍

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -411,6 +415,8 @@ open class GleanInternalAPI internal constructor() {
     internal fun setTestingMode(enabled: Boolean) {
         this.testingMode = enabled
         gleanSetTestMode(enabled)
+        @Suppress("EXPERIMENTAL_API_USAGE")
```
travis79, line 418: Ugh, why is this still experimental?

travis79 (approved):
I'm a bit sad that we had to put a dispatcher back in a language binding, but better than impacting performance.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: Because ... reasons! (No idea)

badboy:
blocked on me actually testing this in Fenix and see if it has an effect.

badboy:
Not really a change on its own, but we're still landing this.

PR #2107, opened 2022-06-30
Fixes clippy lints and warnings for Rust 1.63.0-beta.2

travis79:
This fixes some clippy warnings that have popped up with the latest rust-beta.

In the places where I moved a method call outside of a loop or match and replaced it with a variable holding the result, it was due to this lint: https://rust-lang.github.io/rust-clippy/master/index.html#significant_drop_in_scrutinee

In the places where I added the derive for `Eq`, it was to satisfy this lint: https://rust-lang.github.io/rust-clippy/master/index.html#derive_partial_eq_without_eq

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -96,7 +96,7 @@ bench-rust: ## Run Rust benchmarks
 # Linting
 
 lint-rust: ## Run cargo-clippy to lint Rust code
-	cargo clippy --all --all-targets --all-features -- -D warnings
+	cargo clippy --all --all-targets --all-features -- -D warnings -A unknown-lints
 
 lint-kotlin: ## Run ktlint to lint Kotlin code
 	./gradlew lint ktlint detekt
--- a/glean-core/src/dispatcher/mod.rs
+++ b/glean-core/src/dispatcher/mod.rs
@@ -63,7 +63,7 @@ enum Command {
 }
 
 /// The error returned from operations on the dispatcher
-#[derive(Error, Debug, PartialEq)]
+#[derive(Error, Debug, PartialEq, Eq)]
 pub enum DispatchError {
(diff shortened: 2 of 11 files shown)


chutten (commented):
`for (key, value) in data.iter() {` and friends... are they really subject to significant drop? (Maybe we're running afoul of rust-lang/rust-clippy#8963 or [other issues with this lint](https://github.com/rust-lang/rust-clippy/issues?q=is%3Aissue+significant_drop_in_scrutinee))

I'd be happier with lint suppression for most of these cases.

(The "Be nice and derive `Eq` whenever you derive `PartialEq`" seems alright (though I wonder if it increases binary size))

travis79:
I don't think we were using the referenced values in a way that subjected us to the significant drop issue, and I'm completely okay with silencing the lints for these rather than extracting the assignment out of the loop/match.

travis79:
Ugh, disabling the lints doesn't work because stable rust doesn't know what these lints are

travis79:
I think I will wait for @badboy's input on this to see what he thinks about the best way to deal with this.

badboy:
> (The "Be nice and derive `Eq` whenever you derive `PartialEq`" seems alright (though I wonder if it increases binary size))

I don't think it does. [`Eq`](https://doc.rust-lang.org/std/cmp/trait.Eq.html) doesn't have any methods.

travis79:
@badboy I think it fails because of this: `note: '-D unknown-lints' implied by '-D warnings'`

badboy:
> @badboy I think it fails because of this: `note: '-D unknown-lints' implied by '-D warnings'`

I see that now locally as well.
Change the makefile to have `-A unknown-lints` for now after the `-D warnings`

PR #2111, opened 2022-07-08
Python: drop dependency on cffi

badboy:
UniFFI uses ctypes which comes with Python.
No need for anything else anymore.


chutten (approved):
Love me a good all-red patch

PR #2114, opened 2022-07-11
Bug 1776445 - Bump the Android Gradle Plugin to v7.2.1

Dexterp37:
This additionally bumps the Gradle version and fixes the breakage due to querying the `namespace` property at configuration time. We instead query it at evaluation time now.

This unblocks mozilla-mobile/android-components#12386

**Note**: I'm making this a draft until I test it in A-C.

### DIFF
--- a/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
+++ b/gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
@@ -139,22 +139,7 @@ except:
     def setupTasks(Project project, File envDir, boolean isApplication, String parserVersion) {
         return { variant ->
             def sourceOutputDir = "${project.buildDir}/generated/source/glean/${variant.dirName}/kotlin"
-            // Get the name of the package as if it were to be used in the R or BuildConfig
-            // files. This is required since applications can define different application ids
-            // depending on the variant type: the generated API definitions don't need to be
-            // different due to that.
-            TaskProvider buildConfigProvider = variant.getGenerateBuildConfigProvider()
-            def configProvider = buildConfigProvider.get()
-            def originalPackageName
-            // In Gradle 6.x `getBuildConfigPackageName` was reaplced by `namespace`.
-            // We want to be forward compatible, so we check that first or fallback to the old method.
-            if (configProvider.hasProperty("namespace")) {
-                originalPackageName = configProvider.namespace.get()
-            } else {
-                originalPackageName = configProvider.getBuildConfigPackageName().get()
-            }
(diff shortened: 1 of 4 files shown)


gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -229,6 +212,26 @@ except:
...
+                    // Get the name of the package as if it were to be used in the R or BuildConfig
+                    // files. This is required since applications can define different application ids
+                    // depending on the variant type: the generated API definitions don't need to be
+                    // different due to that.
+                    // Note that this needs to be done at evaluation time rather than configuration
+                    // time, otherwise the `namespace` property won't be available.
+                    TaskProvider buildConfigProvider = variant.getGenerateBuildConfigProvider()
+                    def configProvider = buildConfigProvider.get()
+                    def originalPackageName
+                    // In Gradle 6.x `getBuildConfigPackageName` was reaplced by `namespace`.
```
badboy, line 224: While we're here:
```suggestion
                    // In Gradle 6.x `getBuildConfigPackageName` was replaced by `namespace`.
```

badboy (approved):
Wait, just moving the code into the `doFirst` block is enough? Well, if CI agrees and the test in a-c works I have nothing against it.

badboy:
CI failure for the clippy lints, that is fixed in `main`, so either rebase here or just ignore it and it will be green when merged.

Dexterp37:
> Wait, just moving the code into the `doFirst` block is enough? Well, if CI agrees and the test in a-c works I have nothing against it.

Yes, because `doFirst` makes the code in its closure be executed as the [first thing when the task executes](https://docs.gradle.org/current/dsl/org.gradle.api.Task.html#org.gradle.api.Task:doFirst(org.gradle.api.Action) (vs configuration time)

Dexterp37:
I tested this in Android Components and I was able to correctly build and run the Glean Sample app  (all the other components built just fine).

mcarare:
@Dexterp37 Is there any timeline set for releasing the next version that contains this fix? TY!

badboy:
> @Dexterp37 Is there any timeline set for releasing the next version that contains this fix? TY!

We have a couple of things I'd like to bundle into a release, so we probably do a release end of next week, which means it will take until the week after that (July 25th) before it reaches A-C.

mcarare:
@badboy I guess app services should also be updated with the new Glean version and bumped before we can also update Glean in AC?

badboy:
> @badboy I guess app services should also be updated with the new Glean version and bumped before we can also update Glean in AC?

Yes, I'm working on that today.

PR #2116, opened 2022-07-15
Bug 1779638 - Kotlin: Add a hook to run tasks after init

badboy:
We saw some intermittents on Android Components failing in the `assert(isInitialized())` line.
This leads me to believe that it runs before Glean.initialize is
finished off-thread.
To fix this we introduce a new task queue: run things right after
Glean.initialize calls back into the wrapper.
We need to be careful though: This is triggered from the Rust side so we
can't call back and forth between Kotlin and Rust inside that task or we
risk deadlocks.
But in this instance we're safe:
1. We only allow tasks in test mode
2. We only call it in a single place

In theory there's a race condition still there, because we use
unsynchronized access to the `afterInitQueue`, but really ... it's test
mode and might just be fine enough for all we care? Let's hope so.

I am not able to reproduce the same error as in the intermittent locally
at all (neither in the simulator nor on an actual device), so really I
don't know if this "fixes" it.
The only way to test this out is to go through a release, get it into
A-C and then re-enable the tests and hope for the best.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -44,6 +44,12 @@ internal class OnGleanEventsImpl(val glean: GleanInternalAPI) : OnGleanEvents {
             ProcessLifecycleOwner.get().lifecycle.addObserver(glean.gleanLifecycleObserver)
         }
         glean.initialized = true
+
+        if (glean.testingMode) {
+            glean.afterInitQueue.forEach { block ->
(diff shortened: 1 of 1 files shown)


travis79 (approved):
Looks good to me, the only slight nit I might have is that seeing `block()` made me think that we were going to block and wait on something rather than what it is actually doing in executing a block of code. Just semantic really, not worth "blocking" on ;)

badboy:
Gonna land this so we get that to a-c. Still can't reproduce the initial error, but at the very least this doesn't make things worse.

PR #2117, opened 2022-07-15
bug 1767037 - RLB: Add a new ctor to Event that takes a list of allowed event extra keys.

chutten:
This enables runtime-defined extra keys for runtime-defined event metrics.

Alternatively to this we could give RLB a RuntimeExtraKeys like NoExtraKeys and
specialize the impl of EventMetric to not use ALLOWED_EXTRA or into_ffi_extra,
but given that runtime-defined metrics are a dev-only feature (at present)
maybe this is enough?

### DIFF
--- a/glean-core/rlb/src/private/event.rs
+++ b/glean-core/rlb/src/private/event.rs
@@ -38,6 +38,18 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
         }
     }
 
+    /// The public constructor used by runtime-defined metrics.
+    pub fn with_runtime_extra_keys(
+        meta: glean_core::CommonMetricData,
+        allowed_extra_keys: Vec<String>,
+    ) -> Self {
+        let inner = glean_core::metrics::EventMetric::new(meta, allowed_extra_keys);
+        Self {
+            inner,
+            extra_keys: PhantomData,
+        }
+    }
+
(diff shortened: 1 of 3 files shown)


chutten:
Interesting note on the rust formatting. The [contributing docs](https://mozilla.github.io/glean/dev/contributing.html) no longer tell people changing rust code to run `make fmt-rust` (only `make lint-rust`). I'll add a patch to this PR to add it back to the instructions.

glean-core/rlb/src/private/event.rs
```diff
@@ -38,6 +38,18 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
         }
     }
 
+    /// The public constructor used by runtime-defined metrics.
+    pub fn new_runtime(
+        meta: glean_core::CommonMetricData,
+        allowed_extra_keys: Vec<String>,
+    ) -> Self {
```
badboy, line 45: For some reason I thought we had additional checks _somehwere_.
But of course `﻿ALLOWED_KEYS` is the static thing we generate for the RLB and the internal implementation uses only runtime values anyway.

I think this way is okay then.
In Rust the convention for alternative constructors is usually `with_*`.
What do you think about `with_extra_keys` or `with_runtime_extra_keys`?

badboy (approved):

PR #2119, opened 2022-07-15
Adding a uniffi.toml that is compatible with building the AppServices…

travis79:
… iOS megazord for Focus/Firefox iOS

This will be used as a replacement for the existing uniffi.toml, to be swapped out as part of the megazord build process when the AppServices swift packages are bundled in rust-components-swift

### DIFF
--- a/glean-core/megazord.uniffi.toml
+++ b/glean-core/megazord.uniffi.toml
@@ -1,0 +1,5 @@
+[bindings.swift]
+ffi_module_name = "MozillaRustComponents"
+ffi_module_filename = "gleanFFI"
+generate_module_map = false
+omit_argument_labels = true


glean-core/megazord.uniffi.toml
```diff
@@ -0,0 +1,6 @@
+[bindings.swift]
+cdylib_name = "glean_ffi"
```
badboy, line 2: Is this actually still correct?

badboy (approved):
one question, but otherwise good to go.

glean-core/megazord.uniffi.toml
travis79: Oops, no...   while it doesn't make a difference whether this works or not this line isn't necessary, it works with or without it.  I intended to remove it but forgot.

PR #2120, opened 2022-07-18
Remove ffi-support crate

badboy:
Everything is handled by UniFFI now


badboy:
Removing effectively unused code, skipping review here.

PR #2131, opened 2022-07-25
Bug 1781085 - Add back missing client_info metrics

badboy:


chutten (changes requested):
Needs changelog, test.

chutten (approved):
LGTM 🚢

PR #2136, opened 2022-07-29
Bug 1780806 - Allow the uploader to signal it is done handling uploads.

badboy:
Putting this up so it's up.
I want to discuss this with chutten when he's back.

I'm not super happy with it because the `UploadResult::Done` needs to be handled on a different layer than all the other `UploadResult` variants.
(It's also only implemented for the RLB right now)

### DIFF
--- a/.detekt.yml
+++ b/.detekt.yml
@@ -9,6 +9,14 @@ style:
   # Multiple returns are ok
   ReturnCount:
     active: false
+
+  # Go ahead and jump (Jump!)
+  # Go ahead and jump
+  #
+  # break/continue in loops are ok.
+  LoopWithTooManyJumpStatements:
+    active: false
+
 build:
   maxIssues: 0
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
@@ -17,6 +17,7 @@ import androidx.work.Worker
 import androidx.work.WorkerParameters
 import mozilla.telemetry.glean.Glean
 import mozilla.telemetry.glean.internal.PingUploadTask
+import mozilla.telemetry.glean.internal.UploadTaskAction
 import mozilla.telemetry.glean.internal.gleanGetUploadTask
 import mozilla.telemetry.glean.internal.gleanProcessPingUploadResponse
 import mozilla.telemetry.glean.utils.testFlushWorkManagerJob
@@ -112,13 +113,19 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                     )
 
                     // Process the upload response
-                    gleanProcessPingUploadResponse(action.request.documentId, result)
+                    val action = gleanProcessPingUploadResponse(action.request.documentId, result)
+                    when (action) {
+                        UploadTaskAction.NEXT -> continue
+                        UploadTaskAction.END -> break
+                    }
                 }
                 is PingUploadTask.Wait -> SystemClock.sleep(action.time.toLong())
-                is PingUploadTask.Done -> return Result.success()
+                is PingUploadTask.Done -> break
             }
         } while (true)
         // Limits are enforced by glean-core to avoid an inifinite loop here.
         // Whenever a limit is reached, this binding will receive `PingUploadTask.Done` and step out.
+
+        return Result.success()
     }
 }
--- a/glean-core/rlb/src/net/mod.rs
+++ b/glean-core/rlb/src/net/mod.rs
@@ -15,7 +15,7 @@ use std::thread;
 use std::time::Duration;
 
 use glean_core::upload::PingUploadTask;
-pub use glean_core::upload::{PingRequest, UploadResult};
+pub use glean_core::upload::{PingRequest, UploadResult, UploadTaskAction};
 
 pub use http_uploader::*;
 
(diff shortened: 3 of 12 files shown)


glean-core/rlb/src/test.rs
```diff
@@ -1131,3 +1135,73 @@ fn test_boolean_get_num_errors() {
...
+    // Give it time to spin down.
+    thread::sleep(Duration::from_millis(UPLOAD_WAIT_TIME * 2));
+
+    // Submit another ping and wait for it to do work.
+    custom_ping.submit(None);
+    thread::sleep(Duration::from_millis(UPLOAD_WAIT_TIME * 2));
+
+    let map = call_count.lock().unwrap();
+    assert_eq!(2, map.len(), "should have launched 2 uploader threads");
+    for (_tid, &count) in &*map {
```
chutten, line 1204: How many threads do we use to call out to the uploader, anyway? If it's one per submission, then this won't properly ensure we've handled `Done` properly

glean-core/rlb/src/net/mod.rs
```diff
@@ -97,21 +98,35 @@ impl UploadManager {
                             let upload_url = format!("{}{}", inner.server_endpoint, request.path);
                             let headers: Vec<(String, String)> =
                                 request.headers.into_iter().collect();
-                            let result = inner.uploader.upload(upload_url, request.body, headers);
+                            let result =
+                                match inner.uploader.upload(upload_url, request.body, headers) {
```
chutten, line 102: If you're concerned about maintaining the abstraction, you could drill through a return value from `PingUploadManager::process_ping_upload_response`. Then you could add this `UploadResult::Done` case to its `match`, returning `PingUploadTask::Done` or something. Then handle the correct `Done` here.

badboy: I'm not sure I understand what you mean.
Right now `glean_process_ping_upload_response` doesn't return anything. Are you suggesting it _should_ return something that this uploading loop handles to determine the next step?
(That again doesn't seem _too_ bad I think)

badboy:
Ok, I reworked this a bit now.
The `Done` is passed through and then `glean_process_ping_upload_response` returns something that instructs the upload loop whether to continue or end.  I also implemented that for Kotlin now to show that it works.

Additionally I reworked the test. Instead of arbitrary sleep times it now uses a barrier to sync up the 2 threads. That should be much more reliable.

glean-core/rlb/src/net/mod.rs
chutten: That's precisely what I meant. Have the thing that handles `UploadResult::*` handle this new one, then have it return the result all the way back out here.

glean-core/android/src/main/java/mozilla/telemetry/glean/scheduler/PingUploadWorker.kt
```diff
@@ -112,13 +113,19 @@ class PingUploadWorker(context: Context, params: WorkerParameters) : Worker(cont
                     )
 
                     // Process the upload response
-                    gleanProcessPingUploadResponse(action.request.documentId, result)
+                    val action = gleanProcessPingUploadResponse(action.request.documentId, result)
+                    when (action) {
+                        UploadTaskAction.CONTINUE -> continue
+                        UploadTaskAction.END -> break
+                    }
```
chutten, line 120: No longer RLB-only, I see.

glean-core/rlb/src/net/mod.rs
```diff
@@ -99,19 +100,25 @@ impl UploadManager {
...
                             let result = inner.uploader.upload(upload_url, request.body, headers);
                             // Process the upload response.
-                            glean_core::glean_process_ping_upload_response(doc_id, result);
+                            match glean_core::glean_process_ping_upload_response(doc_id, result) {
+                                UploadTaskAction::Continue => continue,
+                                UploadTaskAction::End => break,
+                            }
                         }
                         PingUploadTask::Wait { time } => {
+                            log::trace!("Instructed to wait for {:?}ms", time);
```
chutten, line 109: If we're tracing the other `PingUploadTask::*` we should trace `PingUploadTask::Upload`, no?

.detekt.yml
```diff
@@ -9,6 +9,14 @@ style:
   # Multiple returns are ok
   ReturnCount:
     active: false
+
+  # Go ahead and jump (Jump!)
+  # Go ahead and jump
```
chutten, line 14: 👏

glean-core/rlb/src/test.rs
```diff
@@ -1131,3 +1134,80 @@ fn test_boolean_get_num_errors() {
...
+    let _t = new_glean(Some(cfg), true);
+
+    // Define a new ping and submit it.
+    const PING_NAME: &str = "test-ping";
+    let custom_ping = private::PingType::new(PING_NAME, true, true, vec![]);
+    custom_ping.submit(None);
+    custom_ping.submit(None);
+
+    // Sync up with the upload thread.
+    barrier.wait();
```
chutten, line 1199: Ah, and if the threading model changes (ie, if each `submit` starts getting its own thread) this test will fail by locking itself?

glean-core/src/upload/result.rs
```diff
@@ -32,6 +32,14 @@ pub enum UploadResult {
         /// The HTTP status code
         code: i32,
     },
+
+    /// Signal that this uploader is done with work
+    /// and doesn't accept new work.
```
chutten, line 37:
```suggestion
    /// and won't accept new work.
```

chutten (commented):
Needs CHANGELOG

glean-core/rlb/src/test.rs
badboy: yep

badboy: well, in theory. *goes back to the code* why doesn't this fail now? oO

badboy: ugh, turns out I'm still running into timing issues here. Let me once again rethink the test.

chutten (approved):

PR #2146, opened 2022-08-03
Bug 1754396 - Python: Verbose logging to maybe catch the intermittent.

badboy:


Dexterp37 (approved):
🤞

badboy:
I ran the test 100 times on CI now and of course it doesn't fail this time. Gonna merge this and maybe we have some more logging next time this fails.

PR #2150, opened 2022-08-08
Bug 1783219 - Upgrade to Android API level 32, upgraded tools and upgraded ndk

badboy:


.circleci/config.yml
```diff
@@ -99,8 +99,8 @@ commands:
           name: Install missing Android SDK & NDK
           command: |
               sdkmanager \
-                "build-tools;28.0.3" \
-                "ndk;21.3.6528147"
+                "build-tools;33.0.0" \
+                "ndk;25.0.8775105"
```
Dexterp37, line 103: I wonder if this could create problems to Focus: are they using the same NDK as A-C?

badboy: That doesn't matter really.
First of all they are not using any NDK (they don't have native code of their own).
Secondly NDKs support a huge range of SDK api levels. Google only provides support for the latest NDK release, but that covers (currently) everything back to Android 5.

PR #2152, opened 2022-08-08
1783739 sleep causing intermittent

perrymcmanis144:
Hmm, not sure about the readme. Will fix tomorrow.

Anyway, here is the strategy I've cooked up (thanks stackoverflow!). It uses a guarded while loop (so, max number of iterations)

Here are the test results I did up quickly. Average over 100 loops for how much deviation there is from the desired time. While this doesn't perfectly mirror the test env or guarantee the problem is fixed, it should make that error far less likely.

I also tried using a small value for sleep and looping and checking but this was... not good lol

```
(base) pmcmanis@MacBook-Pro scratch % python time_loop_tests.py
sleep test average deviation 3.47556ms
sleep loop test average deviation 28.14823ms
perf counter test average deviation 0.0032599999999979447ms
guarded perf counter test average deviation 0.003129999999996755ms
```

### DIFF
--- a/glean-core/python/tests/metrics/test_timing_distribution.py
+++ b/glean-core/python/tests/metrics/test_timing_distribution.py
@@ -162,13 +162,20 @@ def test_measure():
     )
 
     with metric.measure():
-        time.sleep(0.1)
+        iteration_guard = 10**8  # this guard should be
+        # about 100 times as long as the loop
+        iterations = 0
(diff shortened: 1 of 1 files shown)


Dexterp37:
@perrymcmanis144 I think you need to rebase this :E

badboy (approved):
Looks good to me.

travis79 (approved):
Nice!

PR #2158, opened 2022-08-12
TC: Limit to 6 test jobs on the Rust Beta tests

badboy:
Let's see if that fixes the issues we have.


perrymcmanis144 (approved):
CL/config change, tests pass, 👍

PR #2159, opened 2022-08-12
Bug 1784249 - Upgrade to Android NDK r25

badboy:
This reverts commit dee4e7a.

The updated rust-android-gradle plugin takes care of fixing up the build and using the right tools


chutten (changes requested):
Needs `CHANGELOG`

PR #2164, opened 2022-08-19
Update Rust glean::initialize documentation

brizental:
Was trying something out and found this 👀


travis79 (approved):
Nice catch, thanks @brizental

PR #2165, opened 2022-08-19
Bug 1780410 - TC: Use only 2 tasks.

badboy:
This runs on a 2 vCPU/4G memory machine.
Maybe we're running into limits here?


badboy:
Well, that quite obviously didn't work, so we're back to the drawing board.

PR #2168, opened 2022-08-22
Bug 1780410 - RLB: Ignore two tests by default, run them on CI.

badboy:
These tests cause test timeouts on TaskCluster with Rust Beta.
We don't know yet why, they run fine on CircleCI/Rust Stable.

We ignore them for the time being, but explicitly run them on CI.


chutten (approved):
It is more traditional to ref the bug that's open to track fixing the underlying problems than the bug that will be closed (since that's already implicit from the git log), but our BMO doctrine's pretty solid at linking related bugs together so I guess this'll do.

PR #2171, opened 2022-08-24
Bug 1780410 - TC: Don't run Rust tests

badboy:
This reverts commit 01b1770

We still haven't figured out why some tests keep stalling.
This only happens on the TaskCluster instances and is so far not
reproducible in any other environment, including a local Docker run.

I'm out of options here.

---

I've now run the tests hundred of times in my local Docker on a Linux machine, limiting it to 2 vCPUs and 4 G of RAM, same as the taskcluster instance has, with no success to make them fail in the same way.


chutten (approved):
That's fair. And https://bugzilla.mozilla.org/show_bug.cgi?id=1786469 can be for anyone who wants to look into this further if more information comes to light.

PR #2182, opened 2022-09-07
Bug 1789615 - Always store a UUID hyphenated

badboy:
This fixes a regression in the Glean Python SDK.
We used to do this, but failed to correctly encode UUIDs since v50.

The Glean ping schema expects UUIDs in a hyphenated form a la `060f830d-3bcc-4a6a-b4b1-7336271a2b34`. However we passed the user data through without re-encoding, thus leading to non-hyphenated strings landing in the JSON payload of a ping and then getting rejected by the pipeline.

This is only an issue for Python, because for Kotlin and Swift we don't accept plain strings, but only `UUID` types, which serialize to hyphenated strings.


perrymcmanis144 (approved):
👍 code matches my expectation from description; lgtm

PR #2186, opened 2022-09-12
Clarify that intermediary timestamp values do not persist [doc only]

chutten:


docs/user/reference/metrics/timespan.md
```diff
@@ -218,6 +219,7 @@ function onLogin() {
 
 * [`invalid_state`](../../user/metrics/error-reporting.md):
   Calling `stop` without calling `start` first.
+  e.g. if the `start` happened on a previous application run.
```
badboy, line 222:
```suggestion
  Calling `stop` without calling `start` first,
  e.g. if the `start` happened on a previous application run.
```

badboy (approved):
minimal nit.

PR #2187, opened 2022-09-12
Bug 1716954 - Update timing_distribution metric docs

rosahbruno:
Now that timing_distribution is being implemented in Glean.js, we no longer need to say that the API is not available.

Glean.js PR - https://github.com/mozilla/glean.js/pull/1475

### DIFF
--- a/docs/user/reference/metrics/timing_distribution.md
+++ b/docs/user/reference/metrics/timing_distribution.md
@@ -111,7 +111,18 @@ fn on_page_start() {
 ```
 
 </div>
-<div data-lang="JavaScript" class="tab" data-bug="1716954"></div>
+<div data-lang="JavaScript" class="tab">
+
+```Javascript
+import * as timing from "./path/to/generated/files/timing.js";
+
+function onPageStart() {
+    // store this ID, you will need it later to stop or cancel your timer
+    const timerId = timing.start();
+}
+```
+
+</div>
 <div data-lang="Firefox Desktop" class="tab">
 
 **C++**
@@ -197,7 +208,17 @@ fn on_page_loaded() {
 ```
 
 </div>
-<div data-lang="JavaScript" class="tab" data-bug="1716954"></div>
+<div data-lang="JavaScript" class="tab">
+
+```Javascript
+import * as timing from "./path/to/generated/files/timing.js";
+
+function onPageLoaded() {
+    timing.stopAndAccumulate(timerId);
+}
+```
+
+</div>
 <div data-lang="Firefox Desktop" class="tab">
 
 **C++**
@@ -265,7 +286,7 @@ with metrics.pages.page_load.measure():
 
 </div>
 <div data-lang="Rust" class="tab"></div>
-<div data-lang="JavaScript" class="tab" data-bug="1716954"></div>
+<div data-lang="JavaScript" class="tab"></div>
 <div data-lang="Firefox Desktop" class="tab"></div>
 
 {{#include ../../../shared/tab_footer.md}}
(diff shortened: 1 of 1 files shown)


docs/user/reference/metrics/timing_distribution.md
```diff
@@ -435,7 +466,19 @@ assert_eq!(1, snapshot.values.len());
 
 
 </div>
-<div data-lang="JavaScript" class="tab" data-bug="1716954"></div>
+<div data-lang="JavaScript" class="tab">
+
+```Javascript
+import * as timing from "./path/to/generated/files/timing.js";
+
+const snapshot = await timing.testGetValue("<yourPingName>");
```
travis79, line 474: This example seems a bit off. First of all, there should be a metric defined under the timing module, right? And passing the ping name explicitly in the parameter is not the normal use-case, because most metrics are only sent in a single ping passing nothing here should default to the "default" ping for this metric.

I suggest making this look a little more similar to the other examples by adding a metric under the `timing` category and removing the `"<yourPingName>"`:
```suggestion
const snapshot = await timing.serverRequest.testGetValue();
```

docs/user/reference/metrics/timing_distribution.md
```diff
@@ -524,7 +567,16 @@ assert_eq!(
...
 
 </div>
-<div data-lang="JavaScript" class="tab" data-bug="1716954"></div>
+<div data-lang="JavaScript" class="tab">
+
+```Javascript
+import * as timing from "./path/to/generated/files/timing.js";
+import { ErrorType } from "@mozilla/glean/<platform>";
+
+assert.equal(1, await metric.testGetNumRecordedErrors(ErrorType.InvalidValue));
```
travis79, line 576: While I can make the leap to see where `metric` comes from here, perhaps calling it as a member of the `timing` module would be clearer (or even using the same example metric as used in the `testGetValue` example):

```suggestion
assert.equal(1, await timing.metric.testGetNumRecordedErrors(ErrorType.InvalidValue));
```

travis79 (approved):
Looks good to me, just a couple of suggestions to make the test API examples clearer.

rosahbruno:
@travis79 
Thanks for the feedback. I agree, when I was writing these I was thinking in terms of how I used the APIs when working inside of Glean.js, not using the generated types. Let me take another run at the examples, I want to make them consistent using what you suggested.

rosahbruno:
@travis79 
Looking at those examples now, with all the Glean knowledge I've acquired in the last week, I can see those weren't very clear.

I updated the JS examples to follow the same pattern that the other languages use of interacting with a page load metric. Let me know if these look better or if anything else still needs addressed!

https://github.com/mozilla/glean/pull/2187/commits/f240b4a8055146e69b5360e8544297af9b0dc5df

docs/user/reference/metrics/timing_distribution.md
rosahbruno: Should be fixed in https://github.com/mozilla/glean/commit/f240b4a8055146e69b5360e8544297af9b0dc5df

docs/user/reference/metrics/timing_distribution.md
rosahbruno: Should be fixed in https://github.com/mozilla/glean/commit/f240b4a8055146e69b5360e8544297af9b0dc5df

badboy (approved):

travis79 (approved):
🎉

rosahbruno:
For reference:

Chatted about this offline. These changes can land once we are closer to a new release for `Glean.js` since these doc updates are a part of the Timing Distribution implementation in `Glean.js`.

PR #2196, opened 2022-09-19
Bug 1790501 - Add count to DistributionData payload

rosahbruno:
Update `DistributionData` object to include `count` so that a dev can easily check how many samples end up in the distribution. This change effects timing, memory, and custom distributions.

### DIFF
--- a/glean-core/tests/timing_distribution.rs
+++ b/glean-core/tests/timing_distribution.rs
@@ -252,7 +252,7 @@ fn the_accumulate_samples_api_correctly_handles_overflowing_values() {
     // Overflowing values are truncated to MAX_SAMPLE_TIME and recorded.
     assert_eq!(snapshot.sum as u64, MAX_SAMPLE_TIME + 6);
 
-    // Check that we got the right number of values.
+    // Check that we got the right number of samples.
     assert_eq!(snapshot.count, 4);
 
     // We should get a sample in each of the first 3 buckets.


glean-core/src/metrics/mod.rs
```diff
@@ -77,6 +77,9 @@ pub struct DistributionData {
 
     /// The accumulated sum of all the samples in the distribution.
     pub sum: i64,
+
+    /// The total number of entries in the distribution.
+    pub count: u64,
```
rosahbruno, line 82: Trying to set this as an `i64` caused errors in the Kotlin tests. I am not exactly sure why that might be, but I would prefer this work the same as `sum`.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt
```diff
@@ -53,6 +53,7 @@ class TimingDistributionMetricTypeTest {
         val snapshot = metric.testGetValue()!!
         // Check the sum
         assertTrue(snapshot.sum > 0L)
+        assertEquals(snapshot.count, 3UL)
```
rosahbruno, line 56: These have to be `UL` since I can't set the count as an `i64`.

docs/user/reference/metrics/memory_distribution.md
```diff
@@ -128,7 +128,7 @@ assertEquals(11, snapshot.sum)
 
 // Usually you don't know the exact memory values,
 // but how many should have been recorded.
-assertEquals(2L, snapshot.count)
+assertEquals(2UL, snapshot.count)
```
rosahbruno, line 131: Need to use `UL` since `count` is a `u64` at the moment.

rosahbruno (commented):
Seeing some weirdness that I don't understand with `i64` & `u64` in the Kotlin tests. If anyone more knowledgeable about Rust or Kotlin has any ideas, it would be much appreciated!

rosahbruno:
@travis79 
No rush on this, but do you have any ideas about the issue I described above? The Rust tests don't seem to care if I use an `i64` or a `u64`, but Kotlin seems to not like the type conversion and throws weird errors when trying to run the tests.

travis79:
> @travis79 No rush on this, but do you have any ideas about the issue I described above? The Rust tests don't seem to care if I use an `i64` or a `u64`, but Kotlin seems to not like the type conversion and throws weird errors when trying to run the tests.

So, for the longest time unsigned types were a "prototype" feature in kotlin, so we used signed types even though we never had a use case to record negative values.  So I think any of the values coming across the FFI should be signed types because of this.

What failures were you seeing on the Kotlin side, maybe that will help me to point you in the right direction to get this working with an `i64`

rosahbruno:
Talked with @travis79 and looks like my issues were caused (potentially, unconfirmed) by gradle caching. I ran the Kotlin tests via `make test-kotlin` instead of `./gradlew` and everything worked as expected - the CI tests also passed!

So looks like the issue I was seeing was just a local thing, thanks @travis79 !

docs/user/reference/metrics/custom_distribution.md
```diff
@@ -171,10 +181,12 @@ assertEquals(1L, snapshot.values[19])
...
-// Usually you don't know the exact values,
-// but how many should have been recorded.
+// Does the sum have the expected value?
 assert_eq!(23, graphics::checkerboard_peak.test_get_value(None).unwrap().sum);
 
+// Does the count have the expected value?
+assert_eq!(1, graphics::checkerboard_peak.test_get_value(None).unwrap().count);
+
 // Buckets are indexed by their lower bound.
 assert_eq(1, snapshot.values[19])
```
travis79, line 191: This isn't a nit on you, but could you fix this while you are here?
```suggestion
assert_eq!(1, snapshot.values[19])
```

glean-core/tests/timing_distribution.rs
```diff
@@ -245,6 +252,9 @@ fn the_accumulate_samples_api_correctly_handles_overflowing_values() {
     // Overflowing values are truncated to MAX_SAMPLE_TIME and recorded.
     assert_eq!(snapshot.sum as u64, MAX_SAMPLE_TIME + 6);
 
+    // Check that we got the right number of values.
```
travis79, line 255: Uber-nit: you used "samples" instead of "values" just about everywhere else this comment appears

travis79 (approved):
This is looking great!  I had just a couple of non-blocking nits, and I don't see a need to re-review this.  🎉

glean-core/tests/timing_distribution.rs
rosahbruno: Good call, fixed in [`04fd7db` (#2196)](https://github.com/mozilla/glean/pull/2196/commits/04fd7db311d49b0ee25253d872a1397b87eca0d1)

PR #2198, opened 2022-09-20
Apply beta clippy suggestions [ci full]

badboy:
Clippy got smarter again.

PR #2199, opened 2022-09-20
Bug 1790949 - Bump URL metric limit to 8k

rosahbruno:
Update the URL metric to have an 8k limit. Also updated the docs to reflect the new URL length.

Glean.js PR - https://github.com/mozilla/glean.js/pull/1513

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v51.2.0...main)
 
 * General
-  * Update URL metric character length to 8192 ([#2199](https://github.com/mozilla/glean/pull/2199))
+  * Update URL metric character limit to 8k to support longer URLs. URLs that are too long now are truncated to `MAX_URL_LENGTH` and still recorded along with an Overflow error. ([#2199](https://github.com/mozilla/glean/pull/2199))
 * Kotlin
   * Gradle plugin: Fix quoting issue in Python wrapper code ([#2193](https://github.com/mozilla/glean/pull/2193))
   * Bumped the required Android NDK to version 25.1.8937393 ([#2195](https://github.com/mozilla/glean/pull/2195))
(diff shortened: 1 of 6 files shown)


rosahbruno:
I believe the tests are failing because of the issues that should be getting fixed in #2198

travis79 (approved):
Yep, the beta tests are failing because of an unrelated Rust change. This looks great!

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v51.2.0...main)
 
+* General
+  * Update URL metric character length to 8192 ([#2199](https://github.com/mozilla/glean/pull/2199))
```
badboy, line 6: and truncate!

glean-core/src/metrics/url.rs
```diff
@@ -80,15 +80,15 @@ impl UrlMetric {
...
         if s.len() > MAX_URL_LENGTH {
             let msg = format!(
                 "Value length {} exceeds maximum of {}",
                 s.len(),
                 MAX_URL_LENGTH
             );
             record_error(glean, &self.meta, ErrorType::InvalidOverflow, msg, None);
-            return;
+            s = s[..MAX_URL_LENGTH].to_string();
         }
```
badboy, line 92: We have a whole function to make this easier and safer!

```suggestion
        let s = truncate_string_at_boundary_with_error(glean, &self.meta, value, MAX_URL_LENGTH);
```

badboy (changes requested):

CHANGELOG.md
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2199/commits/591e57f6bb01c435d1e08e5291d78bcdfd44f815

glean-core/src/metrics/url.rs
rosahbruno: Fixed this in https://github.com/mozilla/glean/pull/2199/commits/e0eede33369df85e8a767ccb75a50ad3b513432d

Had to add import as well

badboy (approved):

PR #2206, opened 2022-09-22
Bug 1791789 - Fix for iOS startup crash caused by Glean

travis79:
The tl;dr of the fix: we had recursion wrapped in a loop which I _think_ could have caused two `process()` operations to be active at the same time in certain circumstances. This removes the outer while loop and just uses recursion to spawn a new `process()` operation when the previous upload operation completes.

Note: This also shuffles some iOS tests around to help reduce the intermittent test failures.


travis79:
The "iOS build and test" task has a single failing test, which is already a known intermittent.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -150,36 +150,33 @@ public class HttpPingUploader {
...
-                // we don't want to launch multiple uploads at once.
-                // if the upload finishes, we going to launch the next one.
-                return
-            case .wait(let time):
-                sleep(UInt32(time) / 1000)
-                continue
-            case .done:
-                return
             }
+        case .wait(let time):
```
rosahbruno, line 172: The change makes sense and it makes sense that how it was set up before could've caused multiple processes to run at once.

I don't know a lot about this, so this is more of a question for me. Is the only instance when we need to spawn a new process when we hit `.wait`? Just based on the verbiage and the cases I was curious why we wouldn't create a new process on `.done`. Is `.done` just more of a teardown state?

travis79: `.done` in this case means there are no more pings to send according to glean-core so no need to spawn another `process()`.  The only times this should spawn a new `process()` to handle the next ping is either when the previous ping response is complete or after glean-core has told us to back off (`.wait`).

rosahbruno: Ok cool, thanks! The changes make sense, but I don't think I have enough context/expertise to give it a proper sign-off.

badboy (approved):

PR #2211, opened 2022-10-03
Update kotlin and gradle plugin to latest versions

rvandermeulen:


build.gradle
```diff
@@ -11,14 +11,14 @@ buildscript {
     // changing them. Please note that, for using in Android-Components, the
     // versions below must match the ones in that repository.
     ext.versions = [
-        android_gradle_plugin: '7.2.2',
+        android_gradle_plugin: '7.3.0',
```
Dexterp37, line 14: @rvandermeulen would you kindly add a changelog entry for the version bumps?

I'm curious a bit about these updates: is our policy now to be aligned with the most recent version of these?

Dexterp37 (changes requested):

build.gradle
rvandermeulen: We don't really have a strong policy around these deps unless they affect downstream AC/Fenix/Focus (at which point, more care and coordination is needed). That said, these particular ones don't impact downstream and have already been accepted by the mobile team and bumped in m-c, so that seems like a reasonable-enough reason to update elsewhere IMO 😄.

PR #2221, opened 2022-10-14
Bump jna to 5.12.1

rvandermeulen:
AC and A-S have already picked up this change, let's keep Glean in sync as well.


Dexterp37 (approved):
Wow, the changelog is [quite impressive](https://github.com/java-native-access/jna/blob/master/CHANGES.md). There's a bunch of fixes that could be helpful for us.

PR #2228, opened 2022-10-18
Bug 1791842 - Revert "Revert "Kotlin: Upgrade to NDK r25"" [ci full]

badboy:
This reverts commit c69ed251ec4436eb72c147cd84b88eeb5a0f496a.

This might not have been the issue afterall.

---

This is the revert of the revert. Again. I'd rather have us ship with the latest NDK, but I'm holding back until we get a final say about it in the bug.


perrymcmanis144 (approved):
lgtm 👍

PR #2229, opened 2022-10-18
Upgrade to UniFFI 0.21.0

badboy:
Also updated the changelog while I was at it.


rosahbruno (approved):
Looks good to me, assuming tests pass

PR #2233, opened 2022-10-19
Bug 1795925 - Document version expiry in the yaml reference

badboy:
[doc only]


Dexterp37 (approved):
r+ with the linkchecker fixed :)

badboy:
> r+ with the linkchecker fixed :)

"works locally" doesn't count?

PR #2234, opened 2022-10-19
use an unbounded queue, but emulate a bounded one

perrymcmanis144:
Initial pass at doing an unbounded queue.

I added the limit to the dispatch guard and check before sending if the unbounded queue is already "full". It struck me that we probably want this act operationally identical to the bounded version, so I return the same error as if a bounded queue had been exceeded. If this isn't the best way to do this, lmk.

CC @chutten


chutten (approved):

glean-core/src/dispatcher/mod.rs
```diff
@@ -135,20 +138,16 @@ impl DispatchGuard {
...
-                err => err,
+            if self.preinit_sender.len() < self.max_queue_size {
+                self.preinit_sender
+                    .try_send(task)
+                    .map_err(DispatchError::from)
+            } else {
+                self.overflow_count.fetch_add(1, Ordering::SeqCst);
+                // Instead of using a bounded queue, we are handling the bounds
+                // checking ourselves. A bounded queue would return a QueueFull
+                // DispatchError if full, so we do the same.
```
badboy, line 149: Actually we translate the send error to a `QueueFull` ourselves here: https://github.com/mozilla/glean/blob/main/glean-core/src/dispatcher/mod.rs#L92
Which means we could now remove that `from` implementation and above use `send`, not `try_send` (as we can't hit a full queue there anyway)

badboy:
First review to quick, see comment inline, let's switch from `try_send` to `send`

glean-core/src/dispatcher/mod.rs
perrymcmanis144: 👍 ok I switched it over and updated the comment to be more accurate, lmk if any further changes are necessary

badboy: you can now remove these lines: https://github.com/mozilla/glean/blob/main/glean-core/src/dispatcher/mod.rs#L89-L96

perrymcmanis144: updated, also removed the now unused import.

PR #2235, opened 2022-10-20
Bug 1791789 - Fix number 2 for iOS crash due to invalidated session

travis79:
This is the second in a series of fixes to address a crash on iOS due to Glean attempting to interact with an invalidated session.


OrlaM:
Do you have a theory on how something is attempting to use the session object after it's been invalidated? From what I can see in the code a new session object is always created for each request, and then it's immediately used. Invalidating it at some point after it's done is good in order to clean up but from what I can see nothing else can possibly attempt to use it so the timing of this call really shouldn't matter.

travis79:
> Do you have a theory on how something is attempting to use the session object after it's been invalidated? From what I can see in the code a new session object is always created for each request, and then it's immediately used. Invalidating it at some point after it's done is good in order to clean up but from what I can see nothing else can possibly attempt to use it so the timing of this call really shouldn't matter.

This is a speculative change based on the theory that the background update task isn't able to complete the work in the delegate before being cancelled (by either the OS suspending the process or the user force-stopping the app), and that can leave behind a file which would trigger creating a session with the same identifier. I think that captures what @badboy and I discussed last week, but I will allow him to elaborate if I misunderstood. 

The thinking behind this change is that waiting to invalidate the session until the delegate has completed would allow us to reuse/rebuild the session in this case. Basically the session will never be invalidated before the work is complete to the point where we _shouldn't_ be creating a new session with this identifier.

badboy (approved):

PR #2241, opened 2022-10-25
Bug 1797262 - Handle upload trigger/cancel failure by logging, but not panicking

badboy:
I'm not 100% happy with just logging this. Maybe we should record an error? hmmm.


perrymcmanis144 (approved):
This new error makes sense to me so LGTM 👍

travis79 (approved):
Looks good to me. This feels like it is awfully close to where the crash in iOS is occurring, too 🤔

PR #2250, opened 2022-11-01
Bug 1797934 - Pull up label limits to the recording API docs

badboy:
These docs are relevant for the set/add methods, so better to call them out there and then.
To reduce copy-pasta I pulled that info out into includes.

[doc only]


perrymcmanis144 (approved):
👍 this language is clear to me and matches my understanding so lgtm

PR #2253, opened 2022-11-01
Bug 1821437 - Upgrade to Xcode 14.0

badboy:
This is essentially a revert of the revert-commit f941d76 again.

---

We can hold off until our consumers are good to go with this.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -481,7 +481,7 @@ jobs:
 
   Check Swift formatting:
     macos:
-      xcode: "14.0"
+      xcode: "14.3"
     resource_class: "macos.x86.medium.gen2"
     steps:
       - checkout
@@ -500,7 +500,7 @@ jobs:
 
   iOS build and test:
     macos:
-      xcode: "13.4.1"
+      xcode: "14.3"
     resource_class: "macos.x86.medium.gen2"
     steps:
       - checkout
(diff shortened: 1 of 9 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,8 @@
 
 * Kotlin
   * Increase to Android target/compile SDK version 33 ([#2246](https://github.com/mozilla/glean/pull/2246))
+* iOS
+  * Glean for iOS is now being built with Xcode 14.0 ([#](https://github.com/mozilla/glean/pull/TODO))
```
Dexterp37, line 8: Please update this:)

Dexterp37 (approved):
r+wc

rvandermeulen:
@badboy IIUC, Firefox and Focus iOS are both on XCode 14.3 now. Are there other consumers that need to update or can this land now? I guess AppServices is still on 13.4.1, though. Not sure if that blocks or not.

.circleci/config.yml
```diff
@@ -513,14 +513,14 @@ jobs:
       - setup-rust-toolchain
       - restore_cache:
           name: Restore rubygems cache
-          key: swift-docs-gems-v13
+          key: swift-docs-gems-v15
```
badboy, line 516: We used `-v14` in the reverted commit before. Thus skipping that and going for v15 right away.

rosahbruno (approved):

PR #2255, opened 2022-11-03
Bug 1798713 - Events: Increase event extras limit to 15 and event extra value limit to 500 bytes

badboy:


Dexterp37 (approved):

badboy:
test failure intermittent, will rerun when merged

PR #2267, opened 2022-11-15
Bug 1799509 - Do not serialize a histogram's count into the payload

badboy:


perrymcmanis144 (approved):
This matches my expectation for solving the issue pointed out here: https://bugzilla.mozilla.org/show_bug.cgi?id=1799509#c5

I like the new test object as well, very clear what is being tested. As such, 👍  and LGTM

PR #2272, opened 2022-11-15
Bug 1800640 - Use a Cow<'static, str> for labels to reduce heap allocation when use…

badboy:
…d with static labels in Rust

---

This will require changes to the code generator in FOG (and probably glean_parser).


chutten (commented):
Seems legit. The lower-case `string` in uniffi is already owned? So the to/from the builtin type is just type coercion, no actual allocation?

badboy:
> Seems legit. The lower-case `string` in uniffi is already owned? So the to/from the builtin type is just type coercion, no actual allocation?

`string` in udl is `String in Rust, yes (https://mozilla.github.io/uniffi-rs/udl/builtin_types.html).
Nothing really changes for UDL consumers, they pass a `String` buffer in memory. The `into_custom` turns that into `Cow::Owned` on the Rust side. `from_custom` shouldn't actually be used because we're not passing back a `CowString` to the other side.

chutten (approved):

chutten:
Needs `CHANGELOG`, though

PR #2276, opened 2022-11-17
Disable chrono default features

glandium:


badboy (changes requested):
This is incorrect as we depend on the `clock` feature.
It's just enabled because of transitive dependencies.
I updated `jsonschema-valid` to use `default-features = false` and then this here breaks.

I wonder though: What made you do this change? Given [that is it based on a checkout older than July](https://github.com/mozilla/glean/commit/16cae15c2a957221fbf53dd81ddbf27f17605317).

badboy:
We have https://bugzilla.mozilla.org/show_bug.cgi?id=1780401 open. chrono recently changed a lot and we have to verify it still works as expected on all platforms for our release builds.
We just didn't get to it yet, I'll wrap this together.

glandium:
> This is incorrect as we depend on the clock feature.

Are you? cargo check is happy with default-features=false.

badboy:
> > This is incorrect as we depend on the clock feature.
> 
> Are you? cargo check is happy with default-features=false.

yeah, because chrono is a dependency with `default-features = true` in some other crate, so `clock` is ultimately activated in the build.

PR #2282, opened 2022-11-22
Bug 1798919 - Server Knobs implementation work for glean-core

travis79:
I haven't added any test coverage for this yet, but this is the current wip implementation with an internal `CommonMetricData`

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -367,7 +367,7 @@ dependencies = [
 
 [[package]]
 name = "glean-build"
-version = "6.3.0"
+version = "6.4.0"
 dependencies = [
  "xshell-venv",
 ]
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -3,6 +3,7 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::convert::TryFrom;
+use std::sync::atomic::{AtomicU8, Ordering};
 
 use crate::error::{Error, ErrorKind};
 use crate::metrics::labeled::validate_dynamic_label;
@@ -78,18 +79,45 @@ pub struct CommonMetricData {
     pub dynamic_label: Option<String>,
 }
 
-impl CommonMetricData {
+#[derive(Default, Debug, Deserialize, Serialize)]
+pub struct CommonMetricDataInternal {
+    pub inner: CommonMetricData,
+    pub disabled: AtomicU8,
+}
+
+impl Clone for CommonMetricDataInternal {
+    fn clone(&self) -> Self {
+        Self {
+            inner: self.inner.clone(),
+            disabled: AtomicU8::new(self.disabled.load(Ordering::Relaxed)),
+        }
+    }
+}
+
+impl From<CommonMetricData> for CommonMetricDataInternal {
+    fn from(input_data: CommonMetricData) -> Self {
+        Self {
+            inner: input_data.clone(),
+            disabled: AtomicU8::new(u8::from(input_data.disabled)),
+        }
+    }
+}
+
+impl CommonMetricDataInternal {
     /// Creates a new metadata object.
     pub fn new<A: Into<String>, B: Into<String>, C: Into<String>>(
         category: A,
         name: B,
         ping_name: C,
-    ) -> CommonMetricData {
-        CommonMetricData {
-            name: name.into(),
-            category: category.into(),
-            send_in_pings: vec![ping_name.into()],
-            ..Default::default()
+    ) -> CommonMetricDataInternal {
+        CommonMetricDataInternal {
+            inner: CommonMetricData {
+                name: name.into(),
+                category: category.into(),
+                send_in_pings: vec![ping_name.into()],
+                ..Default::default()
+            },
+            disabled: AtomicU8::new(0),
         }
     }
 
@@ -98,10 +126,10 @@ impl CommonMetricData {
     /// If `category` is empty, it's ommitted.
     /// Otherwise, it's the combination of the metric's `category` and `name`.
     pub(crate) fn base_identifier(&self) -> String {
-        if self.category.is_empty() {
-            self.name.clone()
+        if self.inner.category.is_empty() {
+            self.inner.name.clone()
         } else {
-            format!("{}.{}", self.category, self.name)
+            format!("{}.{}", self.inner.category, self.inner.name)
         }
     }
 
@@ -112,20 +140,15 @@ impl CommonMetricData {
     pub(crate) fn identifier(&self, glean: &Glean) -> String {
         let base_identifier = self.base_identifier();
 
-        if let Some(label) = &self.dynamic_label {
+        if let Some(label) = &self.inner.dynamic_label {
             validate_dynamic_label(glean, self, &base_identifier, label)
         } else {
             base_identifier
         }
     }
 
-    /// Whether this metric should be recorded.
-    pub fn should_record(&self) -> bool {
-        !self.disabled
-    }
-
     /// The list of storages this metric should be recorded into.
     pub fn storage_names(&self) -> &[String] {
-        &self.send_in_pings
+        &self.inner.send_in_pings
     }
 }
--- a/glean-core/src/core/mod.rs
+++ b/glean-core/src/core/mod.rs
@@ -1,6 +1,7 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
-use std::sync::Mutex;
+use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::{Arc, Mutex};
 
 use chrono::{DateTime, FixedOffset};
 use once_cell::sync::OnceCell;
@@ -10,7 +11,9 @@ use crate::debug::DebugOptions;
 use crate::event_database::EventDatabase;
 use crate::internal_metrics::{AdditionalMetrics, CoreMetrics, DatabaseMetrics};
 use crate::internal_pings::InternalPings;
-use crate::metrics::{self, ExperimentMetric, Metric, MetricType, PingType, RecordedExperiment};
+use crate::metrics::{
+    self, ExperimentMetric, Metric, MetricType, MetricsDisabledConfig, PingType, RecordedExperiment,
+};
 use crate::ping::PingMaker;
 use crate::storage::{StorageManager, INTERNAL_STORAGE};
 use crate::upload::{PingUploadManager, PingUploadTask, UploadResult, UploadTaskAction};
@@ -148,6 +151,8 @@ pub struct Glean {
     debug: DebugOptions,
     pub(crate) app_build: String,
     pub(crate) schedule_metrics_pings: bool,
+    pub(crate) remote_settings_epoch: AtomicU8,
+    pub(crate) remote_settings_metrics_config: Arc<Mutex<MetricsDisabledConfig>>,
 }
 
 impl Glean {
@@ -200,6 +205,8 @@ impl Glean {
             app_build: cfg.app_build.to_string(),
             // Subprocess doesn't use "metrics" pings so has no need for a scheduler.
             schedule_metrics_pings: false,
+            remote_settings_epoch: AtomicU8::new(0),
+            remote_settings_metrics_config: Arc::new(Mutex::new(MetricsDisabledConfig::new())),
         };
 
         // Ensuring these pings are registered.
@@ -687,6 +694,22 @@ impl Glean {
         metric.test_get_value(self)
     }
 
+    /// Set configuration for metrics' disabled property, typically from a remote_settings experiment
+    /// or rollout
+    ///
+    /// # Arguments
+    ///
+    /// * `json` - The stringified JSON representation of a `MetricsDisabledConfig` object
+    pub fn set_metrics_disabled_config(&self, cfg: MetricsDisabledConfig) {
+        // Set the current MetricsDisabledConfig, keeping the lock until the epoch is
+        // updated to prevent against reading a "new" config but an "old" epoch
+        let mut lock = self.remote_settings_metrics_config.lock().unwrap();
+        *lock = cfg;
+
+        // Update remote_settings epoch
+        self.remote_settings_epoch.fetch_add(1, Ordering::SeqCst);
+    }
+
     /// Persists [`Lifetime::Ping`] data that might be in memory in case
     /// [`delay_ping_lifetime_io`](InternalConfiguration::delay_ping_lifetime_io) is set
     /// or was set at a previous time.
@@ -829,7 +852,7 @@ impl Glean {
             self.storage(),
             INTERNAL_STORAGE,
             &dirty_bit_metric.meta().identifier(self),
-            dirty_bit_metric.meta().lifetime,
+            dirty_bit_metric.meta().inner.lifetime,
         ) {
             Some(Metric::Boolean(b)) => b,
             _ => false,
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -16,10 +16,12 @@
 
 use std::borrow::Cow;
 use std::collections::HashMap;
+use std::convert::TryFrom;
 use std::fmt;
 use std::sync::atomic::{AtomicBool, Ordering};
 use std::sync::{Arc, Mutex};
 
+use metrics::MetricsDisabledConfig;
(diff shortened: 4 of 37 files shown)


glean-core/src/metrics/metrics_disabled_config.rs
```diff
@@ -0,0 +1,44 @@
...
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::{collections::HashMap, convert::TryFrom};
+
+use serde::{Deserialize, Serialize};
+
+/// Represents a list of metrics and their associated `disabled` property from the
+/// remote-settings configuration store
+#[derive(Serialize, Deserialize, Debug, Clone)]
+#[serde(rename_all = "camelCase")]
```
travis79, line 12: May not need this, I'm checking now to see what the default will be coming from experimenter based on what I defined in m-c

glean-core/src/metrics/mod.rs
```diff
@@ -165,7 +169,43 @@ pub trait MetricType {
...
     /// and whether upload is enabled on the Glean object.
     fn should_record(&self, glean: &Glean) -> bool {
-        glean.is_upload_enabled() && self.meta().should_record()
+        if !glean.is_upload_enabled() {
+            return false;
+        }
+
+        // Get the current disabled field from the metric metadata, including
+        // the encoded Nimbus epoch
+        let disabled_field = self.meta().disabled.load(Ordering::Relaxed);
```
travis79, line 178: Something I wasn't entirely sure about, do I need `Ordering::Acquire` here instead of `::Relaxed`?

glean-core/src/lib.rs
```diff
@@ -694,6 +694,15 @@ pub fn glean_test_get_experiment_data(experiment_id: String) -> Option<RecordedE
     core::with_glean(|glean| glean.test_get_experiment_data(experiment_id.to_owned()))
 }
 
+/// Sets a remote configuration for the metrics' disabled property
+///
+/// See [`core::Glean::set_metrics_disabled_config`].
+pub fn glean_set_metrics_disabled_config(json: String) {
+    launch_with_glean(|glean| {
+        _ = glean.set_metrics_disabled_config(json);
```
travis79, line 702: Not sure what to do with the error here, probably should allow it to bubble up across the FFI

glean-core/src/core/mod.rs
```diff
@@ -697,6 +705,22 @@ impl Glean {
         metric.test_get_value(self)
     }
 
+    /// Set configuration for metrics' disabled property, typically from a Nimbus experiment
+    /// or rollout
+    ///
+    /// # Arguments
+    ///
+    /// * `json` - The stringified JSON representation of a MetricsDisabled object
```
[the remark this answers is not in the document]
badboy:
```suggestion
    /// * `json` - The stringified JSON representation of a `MetricsDisabledConfig` object
```

glean-core/src/metrics/metrics_disabled_config.rs
```diff
@@ -0,0 +1,43 @@
...
+    /// property of the metric
+    pub metrics_disabled: HashMap<String, bool>,
+}
+
+impl MetricsDisabledConfig {
+    /// Creates a new MetricsDisabledConfig
+    pub fn new() -> Self {
+        Self {
+            metrics_disabled: HashMap::new(),
+        }
```
badboy, line 24: Call `Default::default()` here:

```suggestion
        Default::default()
```

glean-core/src/metrics/metrics_disabled_config.rs
```diff
@@ -0,0 +1,43 @@
...
+
+impl TryFrom<String> for MetricsDisabledConfig {
+    fn try_from(json: String) -> Result<Self, Self::Error> {
+        match serde_json::from_str(json.as_str()) {
+            Ok(config) => Ok(config),
+            Err(e) => Err(crate::ErrorKind::Json(e)),
+        }
+    }
+
+    type Error = crate::ErrorKind;
```
badboy, line 42: minimal stylistic nit: I prefer the `type Error` on top

```suggestion
    type Error = crate::ErrorKind;
    
    fn try_from(json: String) -> Result<Self, Self::Error> {
        match serde_json::from_str(json.as_str()) {
            Ok(config) => Ok(config),
            Err(e) => Err(crate::ErrorKind::Json(e)),
        }
    }
```

glean-core/src/metrics/mod.rs
badboy: I never really know. I'm usually conservative and go with `Acquire`.
But let's think here for a moment:

Most metrics recording happens on the dispatcher. That runs one task at a time. Thus `Relaxed` would be fine enough. There's no concurrent reading going on.

_Some_ metrics are record synchronously.
* This happens during startup -> No dispatcher tasks running yet.
* This happens during ping submission -> Itself runs synchronously during startup OR _also_ runs on the dispatcher, thus no other task concurrently.

Therefore I think `Relaxed` is good enough.

glean-core/src/metrics/mod.rs
```diff
@@ -165,7 +169,40 @@ pub trait MetricType {
...
+        let current_disabled =
+            if let Some(is_disabled) = metrics_disabled.get(&self.meta().base_identifier()) {
+                u8::from(*is_disabled)
+            } else {
+                u8::from(self.meta().defaults.disabled)
+            };
+
+        // Re-encode the epoch and enabled status and update the metadata
+        let new_disabled = (nimbus_epoch << 4) | (current_disabled & 0xF);
+        self.meta().disabled.store(new_disabled, Ordering::Relaxed);
```
badboy, line 202: So _technically_ nothing stops multiple `should_record()` to run in parallel, and both reading `self.meta().disabled`, then later store to it again.
In between it can read `nimbus_metrics_config`, which itself _could_ be modified in between those 2 reads.
Now we could write the _wrong_ `nimbus_epoch | current_disabled` value until at least the next time this metric is written.
All in all at worst we would see that metric enabled/disabled wrongly once. 

But also from the outside everything is tunneled through the dispatcher, so this should never ever happen to begin with.
I therefore say this is all fine, but maybe we can leave a comment on top calling out that it is fine that this reading/calculating/writing is not atomic?

glean-core/src/metrics/mod.rs
```diff
@@ -165,7 +169,40 @@ pub trait MetricType {
...
+
+        // Return the bool form of whether or not the metric is disabled
```
badboy, line 204: Heh, reading this I got confused again. The method is `should_record`, so exactly the opposite of `is disabled?`.

```suggestion
        // Return the bool form of whether or not the metric should be recorded
```

glean-core/src/core/mod.rs
```diff
@@ -697,6 +705,22 @@ impl Glean {
...
+    pub fn set_metrics_disabled_config(&self, json: String) -> Result<()> {
```
badboy, line 714: Maybe this should take `cfg: MetricsDisabledConfig` instead and the outer function do the parsing?

badboy (changes requested):

glean-core/src/common_metric_data.rs
```diff
@@ -78,18 +79,58 @@ pub struct CommonMetricData {
...
+impl From<CommonMetricData> for CommonMetricDataInternal {
+    fn from(input_data: CommonMetricData) -> Self {
+        Self {
+            defaults: input_data.clone(),
+            disabled: AtomicU8::new(u8::from(input_data.disabled)),
+        }
+    }
+}
+
+impl From<CommonMetricDataInternal> for CommonMetricData {
```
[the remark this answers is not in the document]
travis79: We don't. I think I originally put this there thinking we may need it for the test API functions but we don't.

glean-core/src/common_metric_data.rs
```diff
@@ -78,18 +79,58 @@ pub struct CommonMetricData {
     pub dynamic_label: Option<String>,
 }
 
-impl CommonMetricData {
+#[derive(Default, Debug, Deserialize, Serialize)]
+pub struct CommonMetricDataInternal {
+    pub defaults: CommonMetricData,
```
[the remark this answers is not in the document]
travis79: I'm okay with that, I was naming them by how I was treating them but `inner` wfm.

glean-core/src/core/mod.rs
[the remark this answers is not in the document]
travis79: Not yet, I was only holding off on that until I was sure that the format wouldn't need to change. Ultimately I'd like to abstract this away from a human needing to think about it, but that's further down the road. I'll be sure the "schema" for this is well documented, and eventually having a Nimbus FML definition living in tree can help document it too (we'll need one of those for mobile support)

glean-core/src/lib_unit_tests.rs
```diff
@@ -860,10 +862,20 @@ fn test_set_metrics_disabled() {
...
-        Ok(_) => log::info!("Successfuly set metrics disabled"),
-        Err(e) => log::error!("{}", e.to_string()),
+    match MetricsDisabledConfig::try_from(
+        json!(
+        {
+            "metrics_disabled": {
+                "category.string_metric": true
+            }
+        })
+        .to_string(),
```
badboy, line 872: can we pull this out into a variable? I dislike when the `match`  header is so long

glean-core/src/core/mod.rs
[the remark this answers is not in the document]
travis79: Documentation added for the format in a couple of places now.

glean-core/src/lib_unit_tests.rs
```diff
@@ -811,6 +814,119 @@ fn test_setting_log_pings() {
...
+
+    use crate::metrics::MetricType;
+    // The "epoch" resides in the upper nibble of the `inner.disabled` field
+    let epoch = metric.meta().disabled.load(Ordering::Acquire) >> 4;
+    assert_eq!(
+        current_epoch, epoch,
+        "Epoch must match between metric and Glean core"
+    );
+}
+
```
badboy, line 929: should we have a test that first disables, then later re-enables a metric?

badboy (changes requested):
Calling out an edge case and a test I'd like to see

glean-core/src/lib.rs
```diff
@@ -694,6 +696,20 @@ pub fn glean_test_get_experiment_data(experiment_id: String) -> Option<RecordedE
...
 }
 
+/// Sets a remote configuration for the metrics' disabled property
+///
+/// See [`core::Glean::set_metrics_disabled_config`].
+pub fn glean_set_metrics_disabled_config(json: String) {
+    match MetricsDisabledConfig::try_from(json) {
+        Ok(cfg) => launch_with_glean(|glean| {
+            glean.set_metrics_disabled_config(cfg);
+        }),
```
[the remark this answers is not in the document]
travis79: Yes, in Desktop this _only_ happens after Glean is initialized, specifically right after FOG init is called in BrowserGlue.jsm.  After that, it could be called via the callback registered with Nimbus, but Glean should be initialized at that point.

glean-core/src/lib_unit_tests.rs
travis79: I can add tests for both cases

glean-core/src/core/mod.rs
```diff
@@ -687,6 +694,20 @@ impl Glean {
...
+    ///
+    /// # Arguments
+    ///
+    /// * `json` - The stringified JSON representation of a `MetricsDisabledConfig` object
+    pub fn set_metrics_disabled_config(&self, cfg: MetricsDisabledConfig) {
+        // Set the current MetricsDisabled object
+        *self.remote_settings_metrics_config.lock().unwrap() = cfg;
+
+        // Update remote_settings epoch
+        self.remote_settings_epoch.fetch_add(1, Ordering::SeqCst);
```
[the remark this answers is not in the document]
travis79: Would it be better to release the lock immediately after setting the `cfg`?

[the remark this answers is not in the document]
badboy: That's what the current code does in that one-liner.
I argue it's more correct to keep it until the epoch is also incremented.
Otherwise _in theory_ someone can read the new config but the old epoch.

Cargo.lock
```diff
@@ -367,7 +367,7 @@ dependencies = [
 
 [[package]]
 name = "glean-build"
-version = "6.3.0"
+version = "6.4.0"
```
badboy, line 370: huh, was this missed somewhere? Interesting.

glean-core/src/metrics/mod.rs
```diff
@@ -165,7 +169,51 @@ pub trait MetricType {
...
+            .lock()
+            .unwrap()
+            .metrics_disabled;
+        // Get the value from the remote configuration if it is there, otherwise return the default value.
+        let current_disabled =
+            if let Some(is_disabled) = metrics_disabled.get(&self.meta().base_identifier()) {
+                u8::from(*is_disabled)
+            } else {
+                u8::from(self.meta().inner.disabled)
+            };
```
badboy, line 209: This is still missing the part to handle labeled metrics correctly.

This should do it:

```suggestion
        let current_disabled = {
            let base_id = self.meta().base_identifier();
            let identifier = base_id.split_once('/').map(|split| split.0).unwrap_or(&base_id);
            if let Some(is_disabled) = metrics_disabled.get(identifier) {
                u8::from(*is_disabled)
            } else {
                u8::from(self.meta().inner.disabled)
            }
        };
```

glean-core/src/lib_unit_tests.rs
```diff
@@ -812,6 +815,158 @@ fn test_setting_log_pings() {
...
+        name: "string_metric".to_string(),
+        send_in_pings: vec!["baseline".to_string()],
+        ..Default::default()
+    });
+    let another_metric = StringMetric::new(CommonMetricData {
+        category: "category".to_string(),
+        name: "another_string_metric".to_string(),
+        send_in_pings: vec!["baseline".to_string()],
+        ..Default::default()
+    });
```
badboy, line 832: Let's turn this into a labeled metric to test with:

```suggestion
    let another_metric = LabeledString::new(
        CommonMetricData {
            category: "category".to_string(),
            name: "another_string_metric".to_string(),
            send_in_pings: vec!["baseline".to_string()],
            ..Default::default()
        },
        Some(vec!["label1".into()]),
    );
```

Then use this below:

```rust
    another_metric.get("label1").set_sync(&glean, "TEST_VALUE");
    assert_eq!(
        "TEST_VALUE",
        another_metric
            .get("label1")
            .get_value(&glean, "baseline")
            .unwrap(),
        "Initial value must match"
    );
```

badboy (changes requested):

Cargo.lock
travis79: Probably from the 5 times I rebased on commits that landed yesterday 😭

badboy: yeah, no, that was me. I'll fix this next week and make sure upgrading glean-parser correctly changes this!

badboy (approved):
one last nit! but then good to go

glean-core/src/lib_unit_tests.rs
```diff
@@ -812,6 +815,172 @@ fn test_setting_log_pings() {
...
+    );
+
+    // 3. Since the metrics were disabled, setting a new value will be ignored
+    metric.set_sync(&glean, "VALUE_AFTER_DISABLED");
+    assert_eq!(
+        "TEST_VALUE",
+        metric.get_value(&glean, "baseline").unwrap(),
+        "Shouldn't set when disabled"
+    );
+    another_metric.get("label1").set_sync(&glean, "TEST_VALUE");
```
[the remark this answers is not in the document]
travis79: Whoops, copy pasta in my haste to get it landed :D

glean-core/src/lib_unit_tests.rs
```diff
@@ -812,6 +815,174 @@ fn test_setting_log_pings() {
...
+            .get_value(&glean, "baseline")
+            .unwrap(),
+        "Initial value must match"
+    );
+
+    // 2. Set a configuration to disable the metrics
+    let mut metrics_disabled_config = json!(
+        {
+            "category.string_metric": true,
+            "category.another_string_metric": true,
```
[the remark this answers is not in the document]
travis79: Distracted and pushed before I read my local test output :/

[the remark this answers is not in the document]
travis79: Okay, fmt/clippy/test all happy now

PR #2283, opened 2022-11-22
Allow ErrorType to be passed thorugh an FFI boundary

brizental:
No bug. See discussion: https://github.com/mozilla-mobile/mozilla-vpn-client/pull/4919#discussion_r1021961072


travis79 (approved):
I don't see any potential harm in this change, if this is all it takes to get the ErrorType correctly on the other side of the FFI.

PR #2285, opened 2022-11-23
iOS: Build _only_ a staticlib.

badboy:
`--crate-type` was stabilized in cargo 1.64.0 (cargo#10838)

### DIFF
--- a/samples/ios/app/glean-sample-app.xcodeproj/project.pbxproj
+++ b/samples/ios/app/glean-sample-app.xcodeproj/project.pbxproj
@@ -19,9 +19,10 @@
 		BFDA7AD02371CE4900575C7B /* ViewControllerTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFDA7ACF2371CE4900575C7B /* ViewControllerTest.swift */; };
 		BFDA7AD22371D2BA00575C7B /* MockServer.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFDA7AD12371D2BA00575C7B /* MockServer.swift */; };
 		BFED6B2C238C1638006E2BC4 /* DeletionRequestPingTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFED6B2B238C1638006E2BC4 /* DeletionRequestPingTest.swift */; };
-		CD70CF8C2850D3F400FC2014 /* Glean.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = CD70CF8B2850D3F400FC2014 /* Glean.xcframework */; };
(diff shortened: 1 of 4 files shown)


samples/ios/app/glean-sample-app.xcodeproj/project.pbxproj
```diff
@@ -19,9 +19,10 @@
...
 		BFDA7AD22371D2BA00575C7B /* MockServer.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFDA7AD12371D2BA00575C7B /* MockServer.swift */; };
 		BFED6B2C238C1638006E2BC4 /* DeletionRequestPingTest.swift in Sources */ = {isa = PBXBuildFile; fileRef = BFED6B2B238C1638006E2BC4 /* DeletionRequestPingTest.swift */; };
-		CD70CF8C2850D3F400FC2014 /* Glean.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = CD70CF8B2850D3F400FC2014 /* Glean.xcframework */; };
 		CD70CF8E2850D4B700FC2014 /* Swifter in Frameworks */ = {isa = PBXBuildFile; productRef = CD70CF8D2850D4B700FC2014 /* Swifter */; };
 		CD70CF902850D4BA00FC2014 /* Gzip in Frameworks */ = {isa = PBXBuildFile; productRef = CD70CF8F2850D4BA00FC2014 /* Gzip */; };
+		CD802A47292E5AC800D46BE9 /* Glean.xcframework in Frameworks */ = {isa = PBXBuildFile; fileRef = CD70CF8B2850D3F400FC2014 /* Glean.xcframework */; };
+		CD802A48292E5AC800D46BE9 /* Glean.xcframework in Embed Frameworks */ = {isa = PBXBuildFile; fileRef = CD70CF8B2850D3F400FC2014 /* Glean.xcframework */; settings = {ATTRIBUTES = (CodeSignOnCopy, RemoveHeadersOnCopy, ); }; };
```
travis79, line 25: Any reason we need to embed this framework in the app now?  Is this something downstream consumers will need to do as well?

travis79 (approved):
Just a single question about embedding the framework, no need to block on the answer.

samples/ios/app/glean-sample-app.xcodeproj/project.pbxproj
badboy: This is the sample app, without embedding it it wouldn't work on my test device.
Other users consume it as a Swift package through glean-swift, which does the right thing already.
Only our sample app uses the bare xcframework, and those need to be embedded to run on actual hardware.

PR #2290, opened 2022-11-25
Bug 1716725 Events in Custom Pings shouldn't force submit

chutten:


chutten:
I am not sure what Android tests is complaining about, but it's consistent. It might take me a while to refresh my android environment enough to be able to run tests and figure out why it's failing (beyond "`glean_parser` failed to validate ping against schema" which, though helpful, isn't helpful enough).

travis79:
> I am not sure what Android tests is complaining about, but it's consistent. It might take me a while to refresh my android environment enough to be able to run tests and figure out why it's failing (beyond "`glean_parser` failed to validate ping against schema" which, though helpful, isn't helpful enough).

Let me know if I can assist in working out the issue with the Android testing. I'll review what's here in the interim.

glean-core/src/event_database/mod.rs
```diff
@@ -100,26 +113,66 @@ impl EventDatabase {
...
+                if !glean_restarted_stores.is_empty() {
+                    let glean_restarted = CommonMetricData {
+                        name: "restarted".into(),
+                        category: "glean".into(),
+                        send_in_pings: glean_restarted_stores,
+                        lifetime: Lifetime::Ping,
+                        ..Default::default()
+                    };
+                    let startup = get_iso_time_string(glean.start_time(), TimeUnit::Minute);
+                    let extra = [("glean.startup.date".into(), startup)].into();
```
travis79, line 166: Is this why it's failing the schema checks on Android with Glean parser? ref: https://github.com/mozilla/glean_parser/blob/0960293637d40af31e3db356f91c8686a37801bf/glean_parser/schemas/metrics.2-0-0.schema.yaml#L28

`dotted snake case` looks like it may only expect a single dot? (or am I misreading the regex?)

This also begs the question, why is Kotlin the only language we are checking the schema in 🤔

travis79 (commented):
First pass, looks okay. I had one suggestion that may be why the Kotlin tests are failing.

glean-core/metrics.yaml
```diff
@@ -662,3 +662,25 @@ glean.validation:
...
+  restarted:
+    type: event
+    description: |
+      Recorded when the Glean SDK is restarted.
+
+      Only included in custom pings that record events.
+
+      For more information, please consult the Custom Ping documentation.
+    send_in_pings:
+      - all-pings
```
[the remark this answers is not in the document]
chutten: glean_parser doesn't actually run on Glean SDK builds except for `metrics-docs`. We can put whatever we want in here, really. The only way I write/read from this event is by manually creating the `CommonMetricData` and/or `EventMetric` instances myself, which have appropriate `send_in_pings` arguments to only target the relevant stores.

The purpose of having this definition, then, is for pipeline consumption (though it's an event, so no biggie), for human consumption as docs via Glean Dictionary (what I'm mostly thinking of here, which is why I'm going to disable yamllint so I can link to the Custom Ping docs directly), and to make `python -m glean_parser data-review 1716725 glean-core/metrics.yaml` work (nice benefit).

travis79 (approved):

docs/user/reference/metrics/event.md
```diff
@@ -458,11 +457,13 @@ Each extra key contains additional metadata:
 
 ## Limits
 
-* When 500 events are queued on the client an events ping is immediately sent.
-* The `extra_keys` allows for a maximum of 10 keys.
+* When 500 events are queued on the client an "events" ping is immediately sent.
+* When 500 events are queues on the client for a Custom Ping,
+  subsequent events are dropped and errors of type `invalid_overflow` are recorded.
```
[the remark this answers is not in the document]
chutten: It's [the designed behaviour](https://docs.google.com/document/d/1R-tIZYz4Izx8rDB_YUiJXU9Xtk8VcJoF8Pmhggps_GU/edit#heading=h.2wb70llunlqy). Let me check the code... You know [I think you're right](https://github.com/mozilla/glean.js/blob/main/glean/src/core/metrics/events_database/index.ts#L170-L206).

So, uh, which would you prefer? I like "don't enforce the upper limit" as it has less of a  "Surprise! Data Loss!" vibe.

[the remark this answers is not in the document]
badboy: Huh.
I'm kinda surprised it's the designed behavior. 
The Glean.js implementation is thus basically a spec violation.

But IMO it's the better decision. The whole "rethink threshold" thing was to explicitly have custom pings not need to adhere to that, because they ultimately should not track how many events they added anyway.
I don't remember why we specified the overflow behavior.

glean-core/src/event_database/mod.rs
```diff
@@ -238,34 +311,216 @@ impl EventDatabase {
...
-                    let first_timestamp = store[0].timestamp;
-                    let snapshot = store
-                        .iter()
-                        .map(|e| e.serialize_relative(first_timestamp))
-                        .collect();
-                    Some(snapshot)
+                    // Normalization happens in-place, so if we're not clearing,
+                    // operate on a clone.
+                    if !clear_store {
+                        *store = store.clone();
```
badboy, line 519: Doesn't this overwrite the data we have in memory and operates on that?
You're cloning the data, but then write it in the same place it came from.

badboy (changes requested):

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -162,7 +162,7 @@ class CustomPingTest {
             allowedExtraKeys = emptyList()
         )
         // and record it in the currently initialized Glean instance.
-        click.record()
+        click.record(NoExtras())
```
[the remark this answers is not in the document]
chutten: This is indeed to avoid the deprecated way.

glean-core/src/event_database/mod.rs
```diff
@@ -189,34 +229,67 @@ impl EventDatabase {
...
+                };
+                let execution_counter = match StorageManager.snapshot_metric(
+                    glean.storage(),
+                    INTERNAL_STORAGE,
+                    &execution_cmd.identifier(glean),
+                    execution_cmd.lifetime,
+                ) {
+                    Some(Metric::Counter(i)) => Some(i),
+                    _ => None,
+                };
```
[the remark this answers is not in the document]
chutten: I was matching `seq`'s impl [over here](https://searchfox.org/glean/source/glean-core/src/ping/mod.rs#66), but I'll happily wrap this in `CounterMetric::new` to DRY.

glean-core/src/glean.udl
```diff
@@ -512,6 +512,12 @@ dictionary RecordedEvent {
     //
     // The set of allowed extra keys is defined by users in the metrics file.
     record<DOMString, string>? extra;
+
+    /// The monotonically-increasing execution counter.
```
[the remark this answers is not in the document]
chutten: Oof. Right.

glean-core/src/event_database/mod.rs
```diff
@@ -11,16 +13,23 @@ use std::io::Write;
...
+use crate::metrics::{DatetimeMetric, Metric, TimeUnit};
+use crate::storage::{StorageManager, INTERNAL_STORAGE};
+use crate::util::get_iso_time_string;
 use crate::Glean;
 use crate::Result;
+use crate::{CommonMetricData, Lifetime};
 
 /// Represents the recorded data for a single event.
 #[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+#[cfg_attr(test, derive(Default))]
```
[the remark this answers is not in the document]
chutten: Premature optimization, I suppose. Only `test` needs it (for now).

glean-core/src/event_database/mod.rs
```diff
@@ -238,34 +311,216 @@ impl EventDatabase {
...
+    ///
+    /// In the event that something goes awry, this will record an invalid_state on
+    /// glean.restarted if it is due to internal inconsistencies, or invalid_value
+    /// on client clock weirdness.
+    ///
+    /// # Arguments
+    ///
+    /// * `glean` - Used to report errors
+    /// * `store_name` - The name of the store we're normalizing.
+    /// * `store` - The store we're to normalize. Must not be empty.
```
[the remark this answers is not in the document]
chutten: At the time the comment was written it was true. Not sure if it is any more... lemme check.

You're right, this is no longer the case. (It was previously the case due to the index math)

glean-core/src/event_database/mod.rs
chutten: You're right. I think I need a `let store =` in here.

glean-core/src/event_database/mod.rs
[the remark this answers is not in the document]
chutten: There is a _slight_ difference in that `get_value` can record coverage info.

chutten:
Alrighty, the whole "don't put internal stuff on a public API" required wrapping `RecordedEvent` in a transparent-to-serde `StoredEvent`, which caused a lot of churn up and down the patch. The "don't overflow and don't send a ping when hitting the limit" part required a test fix, which you'll see inline. 

And I think that handles all of the reviewers' requests. Ready for another look.

CHANGELOG.md
```diff
@@ -2,6 +2,12 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v51.8.3...main)
 
+* General
+  * BEHAVIOUR CHANGE: Events in Custom Pings no longer trigger their submission. ([bug 1716725](https://bugzilla.mozilla.org/show_bug.cgi?id=1716725))
+    * Events recorded to custom pings beyond the Glean's `max_events` (default: 500) will be dropped and record `InvalidOverflow` errors.
```
badboy, line 7: don't forget to update this

glean-core/src/event_database/mod.rs
```diff
@@ -44,16 +53,19 @@ pub struct RecordedEvent {
...
-            timestamp: self.timestamp - timestamp_offset,
-            category: self.category.clone(),
-            name: self.name.clone(),
-            extra: self.extra.clone(),
-        })
-    }
+/// Represents the stored data for a single event.
+#[derive(Debug, Clone, Deserialize, Serialize, PartialEq, Eq)]
+struct StoredEvent {
+    #[serde(flatten)]
```
badboy, line 59: A load-bearing flatten!

badboy: Do we have a test to ensure the serialized form doesn't change?

badboy: 🤔 https://github.com/mozilla/glean/blob/adef1eca3fb1c73a1cb6cf0acc6d6dce63741df0/glean-core/tests/event.rs#L144-L149

glean-core/src/event_database/mod.rs
```diff
@@ -189,34 +236,39 @@ impl EventDatabase {
...
             for store_name in meta.send_in_pings.iter() {
                 let store = db.entry(store_name.to_string()).or_insert_with(Vec::new);
-                store.push(event.clone());
+                let execution_counter = CounterMetric::new(CommonMetricData {
+                    name: "execution_counter".into(),
+                    category: store_name.into(),
+                    send_in_pings: vec![INTERNAL_STORAGE.into()],
+                    lifetime: Lifetime::Ping,
+                    ..Default::default()
+                })
```
badboy, line 250: The `execution_counter` is never actually set/incremented anywhere. Therefore this will always be `None`.
That means:

* We're missing an `execution_counter.add_sync(&glean, 1)` somewhere
* We don't have a test that captures this behavior properly (might be easiest to "reset" Glean in Kotlin to try out this behavior

I added an event to the RLB prototype example (`glean-core/rlb/examples/prototype.rs`), removed the ping submission there, and then ran it several times.

badboy (changes requested):

glean-core/src/event_database/mod.rs
chutten: See `deserialize_existing_data` in event_database/mod.rs

CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v51.8.3...main)
 
+* General
```
badboy, line 5: Will need a rebase against `main` before landing

badboy (approved):

PR #2296, opened 2022-12-07
Update uuid to 1.0

glandium:


glandium:
Note that with an updated rkv with https://github.com/mozilla/rkv/pull/232, uuid 0.8.2 would disappear from Cargo.lock.

badboy (approved):

badboy:
We have a Glean release planned for end of this week, getting it into the next m-c nightly cycle

glandium:
If you don't mind I'll do the glean update after its release in bug 1804359.

PR #2298, opened 2022-12-08
Disable env_logger features in glean-core

glandium:
glean-core shouldn't be making this decision for its users.


glandium:
(honestly, a library shouldn't depend on env_logger in the first place)

badboy:
> (honestly, a library shouldn't depend on env_logger in the first place)

That is true and comes from a time when Glean was not really _used_ as a Rust library, and so `env_logger` was just convenience outside of Android and iOS.

badboy (approved):

glandium:
I suppose that's true of oslog and android_logger?

PR #2305, opened 2022-12-12
Consistently named tasks in the Makefile

badboy:


perrymcmanis144 (approved):
CI failure seems odd, formatting shouldn't break a working test.

Otherwise, it looks more consistent to me 👍

PR #2312, opened 2022-12-15
Bug 1804632 - Include `env_logger` only in debug builds

badboy:
This has no immediate effect on any other usage.
It's still in use as a dev-dependency in Rust (and thus can be enabled for tests).
It also is active for local builds of Glean for Android for testing.


rosahbruno (approved):
Looks good besides the test that failed, but looks like it is something likely not caused by your change?

badboy:
> Looks good besides the test that failed, but looks like it is something likely not caused by your change?

yup, addressed in #2314

PR #2315, opened 2022-12-15
CI: Update tools in use (mdbook, ghr, cargo-deny)

badboy:


Dexterp37 (approved):
r+ since the linting on beta failure seems to be unrelated.

PR #2316, opened 2022-12-15
Bug 1802503 - Rust: Implement a rust dispatcher connected to GCD

badboy:
GCD (Grand Central Dispatch) is a runtime system for executing code concurrently.
It's available on iOS and macOS by default as provided by the system, and available as libdispatch on other *nix systems.

We used that before when our dispatcher was purely in Swift. We now go back to use it, but from Rust.


badboy:
This is in draft mode because it's missing some obvious things:

* Working CI
* Passing tests on CI
* changelog
* extensive testing on iOS

PR #2325, opened 2022-12-19
Bug 1802094 - Windows os buildid

perrymcmanis144:
https://bugzilla.mozilla.org/show_bug.cgi?id=1802094

Draft PR while I manage data review and review everything


glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,18 @@ pub fn get_os_version() -> String {
...
+#[cfg(target_os = "windows")]
+/// Returns the Build ID for Windows kernels versions >= 10.0
+pub fn get_windows_build_number() -> String {
+    whatsys::windows::get_windows_build_number().unwrap_or_else(|| "Unknown".to_string())
+}
+
+#[cfg(not(target_os = "windows"))]
+/// Returns the Build ID for Windows kernels versions >= 10.0
+pub fn get_windows_build_number() -> String {
+    "Unknown".to_string()
```
perrymcmanis144, line 85: hey @badboy I am not sure this is correct. I would expect this to more appropriately be None, but I need the non-windows on here right? However, when I tried to wrap these function as an option, the compiler fought me a bit.

Is my instinct right that this should be Option<String>, and None in the `not(target_os="windows")` case? and I just need to figure out how to make it work?

badboy: Yup, should be `fn get_windows_build_number() -> Option<String>` in both cases.
Wait, why string? didn't we say we go with a number?

glean-core/rlb/src/core_metrics.rs
```diff
@@ -33,6 +33,7 @@ impl From<ClientInfoMetrics> for glean_core::ClientInfoMetrics {
             app_display_version: metrics.app_display_version,
             channel: metrics.channel,
             os_version: system::get_os_version(),
+            windows_build_number: Some(system::get_windows_build_number()),
```
badboy, line 36: When the function returns `Option<T>` you will only need this:

```suggestion
            windows_build_number: system::get_windows_build_number(),
```

glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,18 @@ pub fn get_os_version() -> String {
     parse_linux_os_string(whatsys::kernel_version().unwrap_or_else(|| "Unknown".to_owned()))
 }
 
+#[cfg(target_os = "windows")]
+/// Returns the Build ID for Windows kernels versions >= 10.0
+pub fn get_windows_build_number() -> String {
+    whatsys::windows::get_windows_build_number().unwrap_or_else(|| "Unknown".to_string())
```
badboy, line 79:
```suggestion
pub fn get_windows_build_number() -> Option<i32> {
    whatsys::windows_build_number()
```

glean-core/src/core_metrics.rs
```diff
@@ -26,6 +26,8 @@ pub struct ClientInfoMetrics {
     pub channel: Option<String>,
     /// The Android specific SDK version of the software running on this hardware device (e.g. "23").
     pub android_sdk_version: Option<String>,
+    /// The Windows specific OS build version (e.g. "19043")
+    pub windows_build_number: Option<String>,
```
badboy, line 30:
```suggestion
    pub windows_build_number: Option<i32>,
```

double-check how we send that on legacy telemetry.

glean.1.schema.json
```diff
@@ -1,596 +1,600 @@
 {
-  "$id": "moz://mozilla.org/schemas/glean/ping/1",
```
badboy, line 2: That's an awful lot of changes in here for what should be a small change.
Please update it upstream first: https://github.com/mozilla-services/mozilla-pipeline-schemas/
We have a script to update it from there: `bin/update-schema.sh`

glean-core/metrics.yaml
```diff
@@ -75,6 +75,26 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+
```
badboy, line 78:
```suggestion
```

glean-core/metrics.yaml
```diff
@@ -75,6 +75,26 @@ glean.internal.metrics:
...
+  windows_build_number:
+    type: string
+    lifetime: application
+    send_in_pings:
+      - glean_client_info
+    description: |
+      The optional user-visible version of the Windows Build Number
+      (e.g. "1.2.3"). If the version detection fails, this metric gets set
+      to `Unknown`.
```
badboy, line 87: windows build numbers are _not_ "1.2.3".
I argue we do not set it to `Unknown` but don't set it at all if we can't detect it (as we obviously can't on non-windows)

glean-core/metrics.yaml
```diff
@@ -75,6 +75,26 @@ glean.internal.metrics:
...
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1802094
+    data_reviews:
+      - todo
```
badboy, line 91: don't forget this.

glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,22 @@ pub fn get_os_version() -> String {
     parse_linux_os_string(whatsys::kernel_version().unwrap_or_else(|| "Unknown".to_owned()))
 }
 
+#[cfg(target_os = "windows")]
+/// Returns the Build ID for Windows kernels versions >= 10.0
```
badboy, line 77:
```suggestion
/// Returns the Windows Build Number.
```

The build number exists on Windows < 10 as well.

glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,22 @@ pub fn get_os_version() -> String {
...
+/// Casts to i64 to work with QuantityMetric type
```
badboy, line 78: That doesn't really have to be a doc comment, that can go inline below.

glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,22 @@ pub fn get_os_version() -> String {
...
+pub fn get_windows_build_number() -> Option<i64> {
+    match whatsys::windows::get_windows_build_number() {
```
badboy, line 80: Should be
```suggestion
    match whatsys::windows_build_number {
```

glean-core/rlb/src/system.rs
```diff
@@ -73,6 +73,22 @@ pub fn get_os_version() -> String {
...
+        Some(i) => Some(i as i64),
+        _ => None,
+    }
+}
+
+#[cfg(not(target_os = "windows"))]
+/// Returns the Build ID for Windows kernels versions >= 10.0
```
badboy, line 87: Not true. This returns `None` always.

glean-core/metrics.yaml
```diff
@@ -75,6 +75,26 @@ glean.internal.metrics:
...
 
+  windows_build_number:
+    type: quantity
+    lifetime: application
+    send_in_pings:
+      - glean_client_info
+    description: |
+      The optional user-visible version of the Windows Build Number
+      (e.g. "22000"). If the version detection fails, this metric gets set
+      to `Unknown`.
```
chutten, line 86: `quantity` doesn't support a value of `Unknown`, though?

glean-core/src/core_metrics.rs
```diff
@@ -26,6 +28,8 @@ pub struct ClientInfoMetrics {
     pub channel: Option<String>,
     /// The Android specific SDK version of the software running on this hardware device (e.g. "23").
     pub android_sdk_version: Option<String>,
+    /// The Windows specific OS build version (e.g. "19043")
```
chutten, line 31: (since it isn't a string)
```suggestion
    /// The Windows specific OS build version (e.g. 19043)
```

glean-core/metrics.yaml
```diff
@@ -75,6 +75,26 @@ glean.internal.metrics:
...
     expires: never
 
+  windows_build_number:
+    type: quantity
+    lifetime: application
+    send_in_pings:
+      - glean_client_info
+    description: |
+      The optional user-visible version of the Windows Build Number
+      (e.g. "22000"). If the version detection fails, this metric gets set
```
chutten, line 85:
```suggestion
      (e.g. 22000). If the version detection fails, this metric gets set
```

chutten (changes requested):

glean-core/metrics.yaml
perrymcmanis144: Ah, good point, I had written this patterned off of `os_version`, but this raises an interesting question to me: if this is optional, and this value is not always present, what should it be in those cases?

Does it make more sense to cast this value to a string so that we can support Unknown? 

It also strikes me that windows build numbers are not _really_ interpretable as numbers, i.e. they are more ordinal and as such string comparison is probably a better fit.

cc @badboy curious what you think as well

badboy: You are right. Quantity is an integer ultimately and `"Unknown"` is not used anywhere in the code.
According to https://firefox-source-docs.mozilla.org/toolkit/components/telemetry/data/environment.html it's a number in legacy, too.

There's relatively few cases where on Windows we would NOT be able to determine the value, right?
What does "Unknown" give us over just not setting the value (as the code does right now)?

perrymcmanis144: > What does "Unknown" give us over just not setting the value (as the code does right now)?

Slightly better conceptual alignment. Practically, nothing I suppose. Ok updating

badboy (approved):

badboy:
I went ahead, manually squashed this and also added a changelog entry. Landing it then.

badboy:
gnaaah, now that I added docs I of course also broke them...

PR #2332, opened 2023-01-10
Bug 1749510 - RLB: Wait for uploader on shutdown using new shutdown mechanism

badboy:
RLB: Wait for uploader on shutdown using new `shutdown` mechanism

On shutdown we now wait up to 30s for the upload thread to finish.
This gets done just after we drain and shut down the dispatcher,
but before Glean is considered fully shutdown.

The callback MUST NOT put any new tasks on the dispatcher.
The callback SHOULD NOT block arbitrarily long.

The RLB implementation of that callback instructs a potential uploader
thread to end operation after it finished one upload.
It also blocks any new upload thread from getting started.
This means an in-flight upload can finish as usual, but no new one
is launched.
If the upload thread is currently sleeping, it gets one more upload.

This also adds a new test.
The only way to reliably test this is from the outside,
so the test invokes an example implementation and checks that the
`pending_pings` directory is empty after the program ends.
The uploader inside sleeps for a short period of time.
Without waiting for this uploader the main thread would end and kill the
uploader thread before it wakes up.
With waiting the uploader finishes correctly before the main thread
exits.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -75,6 +75,10 @@ commands:
           command: |
             cargo run -p sample
       - run:
+          name: Run Rust RLB test
+          command: |
+            glean-core/rlb/tests/test-shutdown-blocking.sh
+      - run:
           name: Upload coverage report
           command: |
             pip3 install glean_parser
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -70,7 +70,7 @@ struct GleanEvents {
 }
 
 impl glean_core::OnGleanEvents for GleanEvents {
-    fn on_initialize_finished(&self) {
+    fn initialize_finished(&self) {
         // intentionally left empty
     }
 
--- a/glean-core/rlb/src/lib.rs
+++ b/glean-core/rlb/src/lib.rs
@@ -89,6 +89,10 @@ impl glean_core::OnGleanEvents for GleanEvents {
         // intentionally left empty
         Ok(())
     }
+
+    fn shutdown(&self) {
+        self.upload_manager.shutdown();
+    }
 }
 
 fn initialize_internal(cfg: Configuration, client_info: ClientInfoMetrics) -> Option<()> {
--- a/glean-core/rlb/src/net/mod.rs
+++ b/glean-core/rlb/src/net/mod.rs
@@ -7,17 +7,15 @@
 //! This doesn't perform the actual upload but rather handles
 //! retries, upload limitations and error tracking.
 
-use std::sync::{
-    atomic::{AtomicBool, Ordering},
-    Arc,
-};
-use std::thread;
+use std::sync::{atomic::Ordering, Arc, Mutex};
+use std::thread::{self, JoinHandle};
 use std::time::Duration;
 
 use glean_core::upload::PingUploadTask;
 pub use glean_core::upload::{PingRequest, UploadResult, UploadTaskAction};
 
 pub use http_uploader::*;
+use thread_state::{AtomicState, State};
 
 mod http_uploader;
 
@@ -45,7 +43,8 @@ pub(crate) struct UploadManager {
 struct Inner {
     server_endpoint: String,
     uploader: Box<dyn PingUploader + 'static>,
-    thread_running: AtomicBool,
+    thread_running: AtomicState,
+    handle: Mutex<Option<JoinHandle<()>>>,
 }
 
 impl UploadManager {
@@ -63,7 +62,8 @@ impl UploadManager {
             inner: Arc::new(Inner {
                 server_endpoint,
                 uploader: new_uploader,
-                thread_running: AtomicBool::new(false),
+                thread_running: AtomicState::new(State::Stopped),
+                handle: Mutex::new(None),
             }),
         }
     }
@@ -76,7 +76,12 @@ impl UploadManager {
         if self
             .inner
             .thread_running
-            .compare_exchange(false, true, Ordering::SeqCst, Ordering::SeqCst)
+            .compare_exchange(
+                State::Stopped,
+                State::Running,
+                Ordering::SeqCst,
+                Ordering::SeqCst,
+            )
             .is_err()
         {
             return;
@@ -84,7 +89,9 @@ impl UploadManager {
 
         let inner = Arc::clone(&self.inner);
 
-        thread::Builder::new()
+        // Need to lock before we start so that noone thinks we're not running.
+        let mut handle = self.inner.handle.lock().unwrap();
+        let thread = thread::Builder::new()
             .name("glean.upload".into())
             .spawn(move || {
                 log::trace!("Started glean.upload thread");
@@ -101,8 +108,14 @@ impl UploadManager {
                             let result = inner.uploader.upload(upload_url, request.body, headers);
                             // Process the upload response.
                             match glean_core::glean_process_ping_upload_response(doc_id, result) {
-                                UploadTaskAction::Next => continue,
+                                UploadTaskAction::Next => (),
                                 UploadTaskAction::End => break,
+                            }
+
+                            let status = inner.thread_running.load(Ordering::SeqCst);
+                            // asked to shut down. let's do it.
+                            if status == State::ShuttingDown {
+                                break;
                             }
                         }
                         PingUploadTask::Wait { time } => {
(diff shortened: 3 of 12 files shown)


glean-core/src/lib.rs
```diff
@@ -500,6 +508,13 @@ pub fn shutdown() {
         log::error!("Can't shutdown dispatcher thread: {:?}", e);
     }
 
+    // Call on_shutdown.
+    // Need to be done without holding a lock on the global Glean object.
+    {
+        let state = global_state().lock().unwrap();
+        state.callbacks.on_shutdown();
```
chutten, line 515: If we're worried about timeouts, we could spawn a thread for `on_shutdown` and give it one half of a Condvar that this thread'll be `wait_timeout()`-ing on?

chutten: That'd solve the arbitrarily long part, and if we named the thread (or used some other thread-local storage) we could have the dispatcher forbid adding of tasks from that thread, which'd solve the "do not add tasks" part.

glean-core/rlb/src/net/mod.rs
```diff
@@ -41,11 +41,16 @@ pub(crate) struct UploadManager {
...
+const STATE_STOPPED: u8 = 0;
+const STATE_RUNNING: u8 = 1;
+const STATE_SHUTTING_DOWN: u8 = 2;
+
 #[derive(Debug)]
 struct Inner {
     server_endpoint: String,
     uploader: Box<dyn PingUploader + 'static>,
-    thread_running: AtomicBool,
+    thread_running: AtomicU8,
```
chutten, line 52: Yeah, I'm definitely gonna want to make this look more `AtomicEnum`-ish before landing.

glean-core/rlb/src/net/mod.rs
```diff
@@ -115,11 +128,41 @@ impl UploadManager {
...
+            .thread_running
+            .store(STATE_SHUTTING_DOWN, Ordering::SeqCst);
+
+        // take the thread handle out.
+        let mut handle = self.inner.handle.lock().unwrap();
+        let thread = handle.take();
+
+        if let Some(thread) = thread {
+            thread
+                .join()
```
chutten, line 164: So this will work if our "we're killing it too early" is interrupting us in the middle of a `PingTask::Upload`, but if we're in a `PingTask::Wait`, then not only will we wait, we'll then not perform an upload attempt?

I'm tempted to put the `if SHUTTING_DOWN: break;` inside `PingTask::Upload` to make it more likely that we actually give an upload a try here. But then we might want to set the rate limiter to one that doesn't wait so we don't cycle from Wait to Wait for the full `3*interval` (3min?). Thoughts?

chutten (commented):
Nothing that's not surmountable, I think.

glean-core/rlb/src/lib.rs
```diff
@@ -89,6 +89,10 @@ impl glean_core::OnGleanEvents for GleanEvents {
         // intentionally left empty
         Ok(())
     }
+
+    fn on_shutdown(&self) {
```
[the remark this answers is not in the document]
badboy: Right now:
2 with `on_`: `on_initialize_finished`, `on_shutdown`,
3 without: `trigger_upload`, `start_metrics_ping_scheduler`, `cancel_uploads`

Without `on_` definitely sounds right for the 3 that don't have it. `shutdown` instead of `on_shutdown` sounds ok to me as well.
`initialize_finished` works as well.

So ... going with no `on_` (Heh, it's in the trait name already anyway)

glean-core/rlb/src/net/mod.rs
badboy: With the current throttle we're not able to run into it 3 times though, because we do sleep the appropriate amount of time and so the next call after a sleep should be allowed.
Rust's `thread::sleep` handles spurious wakeups and will go back to sleep to reach the appropriate `Duration`.

glean-core/src/lib.rs
badboy: Both implemented now.

.circleci/config.yml
```diff
@@ -74,6 +74,10 @@ commands:
           name: Run Rust sample
           command: |
             cargo run -p sample
+      - run:
+          name: Run Rust RLB test
+          command: |
+            bin/rust-external-test.sh
```
chutten, line 80: Nit: maybe we name this something more tightly connected to what it's testing? It isn't testing RLB wholesale, it's "just" running the `long-running` example and checking if the pending ping dir is empty.

Or do you have plans for this becoming a test suite of its own, so the generic naming is future-looking?

glean-core/rlb/src/net/mod.rs
```diff
@@ -117,9 +130,92 @@ impl UploadManager {
...
+                .expect("couldn't join on the uploader thread.");
+        }
+    }
+}
+
+mod thread_state {
```
chutten, line 164: Most excellent, thank you.

glean-core/src/lib.rs
```diff
@@ -21,11 +21,15 @@ use std::convert::TryFrom;
...
 use std::sync::{Arc, Mutex};
+use std::thread;
+use std::time::Duration;
 
-use metrics::MetricsDisabledConfig;
+use crossbeam_channel::unbounded;
 use once_cell::sync::{Lazy, OnceCell};
 use uuid::Uuid;
 
+use metrics::MetricsDisabledConfig;
```
chutten, line 31: rebase garbage?

glean-core/rlb/src/net/mod.rs
chutten: Thank you for moving the `if ShuttingDown: break` inside `Upload`. Even if we're stuck waiting, at least we'll give one upload a try after we're done.

glean-core/src/lib.rs
```diff
@@ -500,6 +514,30 @@ pub fn shutdown() {
...
+            .spawn(move || {
+                let state = global_state().lock().unwrap();
+                state.callbacks.shutdown();
+
+                // Best-effort sending. The other side might have timed out already.
+                let _ = tx.send(()).ok();
+            })
+            .expect("Unable to spawn thread to wait on shutdown");
+
+        // TODO: 30 seconds? What's a good default here? Should this be configurable?
```
chutten, line 533: For the current situation the shutdown processing duration is dominated by two things: network conditions for the single upload we might try, and whether we're waiting.

Network conditions around upload success and failure are something we can (and maybe should) instrument. We have `glean.upload.ping_upload_failure`, maybe we need a `timing_distribution` `glean.upload.ping_upload_success`? If we're willing to presume that Glean upload is at worst similarly performant to Legacy Firefox Desktop Telemetry upload, [on Beta we're looking at p95 around 4.5s](https://telemetry.mozilla.org/new-pipeline/dist.html#!cumulative=0&end_date=2023-01-29&include_spill=0&keys=__none__!__none__!__none__&max_channel_version=beta%252F110&measure=TELEMETRY_SEND_SUCCESS&min_channel_version=null&processType=*&product=Firefox&sanitize=1&sort_by_value=0&sort_keys=submissions&start_date=2023-01-16&table=0&trim=1&use_submission_date=0) (0.58% of successful uploads are in buckets above 33s). (send failures take longer, but that's likely dominated by timeout statuses).

As for `Wait`, the way I understand it we only `Wait` in two situations: 
1) we're not done processing pending pings, 
2) we're rate-limiting.

In 1 we might `Wait` -> `Wait` -> `Wait` because we don't know how long it takes to process pending pings. There's nothing that'll interrupt the loading of the pending pings, so we might be stuck in `ThreadJoinHandle('glean.upload').join()` for `3 * WAIT_TIME_FOR_PING_PROCESSING = 3s` (I originally worried this'd be `3 * Rate Limiter Interval = 3min`, but processing wisely has its own timing).

In 2 we don't `Wait` -> `Wait` because the rate limiter tells us to wait the full amount of time in one go, and as you say Rust should be trustworthy to get us to the full duration needed for the rate limiter to reset itself and start processing again. In this case we'll go `Wait` -> `Upload` (if there are tasks and we were throttled) meaning we can only get stuck for one full interval (1min). This is suboptimal because the 30s timeout here ensures we'll always wait the full 30s and still not perform any upload at the end of it.

---

In short, I think the 30s is (until we find out differently) a fine interval to choose. I'd like some instrumentation about what situations this new "wait for shutdown" module finds itself in so we can decide to tune it later, but I'm not sure we'll need it to be configurable. Tuning it release-by-release feels good enough.

...and I am a touch bothered by the rate limiter case. If we're in rate limiting then we've processed a bunch of pings which means we're out of danger of the problem this patch aims to solve... but it means we'll be sitting at shutdown for 30s for no benefit whatsoever.

A way out of this could be having us check before we `join` whether we're in `Wait` for longer than the timeout. (We could pass the shutdown timeout like `callbacks.shutdown(timeout)` which could pass it to the uploader which could compare it against the duration we gave to sleep. If it's too long we could opt to not join and allow `glean.shutdown` to fast-exit). It's adding more complication to cover what might be (without instrumentation we'd never know whether it is) a rare occurrence, so I'll not be blocking my review on it.

chutten (approved):
r+ conditional on adding a CHANGELOG entry and some instrumentation.

.circleci/config.yml
badboy: I ... initially thought about making this more generic, but given it's only one test right now naming it more appropriately is in order.

glean-core/src/lib.rs
badboy: how dare you call our code garbage?

nah, just reordering `use` statements and separating into blocks: std, external crates, internal modules

badboy:
> r+ conditional on adding a CHANGELOG entry and some instrumentation.

I would prefer to do the instrumentation in a followup. Because of how we upload this is ... not as easy as I would like and so to keep review on point doing that in a followup seems easier.

travis79 (approved):
👍   Nothing to really add, this looks good! (And +1 on removing the "on" from the function names)

PR #2337, opened 2023-01-16
ci: Append action to assumed role for Github releases

ahal:
Bug: 1810118


badboy (approved):

badboy:
ignoring test failure, this is release CI only anyway.

PR #2344, opened 2023-01-23
Rust: Mention the correct option in the context

badboy:
[doc only]


perrymcmanis144 (approved):
👍 very reasonable lgtm

PR #2345, opened 2023-01-23
Python: Switch to maturin for all Python building/development

badboy:
~~(based on #2344)~~

After the [recent universal2 wheel work](https://bugzilla.mozilla.org/show_bug.cgi?id=1810737) I realized how messy our Python setup is:
A big home-grown solution inside a `setup.py` reinventing half of Python package management all over again.

There must be something better for it.
There is: [maturin](https://github.com/PyO3/maturin/).

Tried and trusted [in other projects](https://github.com/mozilla/mozfun-local), this tool can take care of most of the build and development hassles of a Python package with UniFFI bindings.

---

This PR switches our development tooling over to maturin, for both local development and on CI.
And while I was at it I also moved all Python tooling configuration into the one and only `pyproject.toml`.

This is not yet ready for merge, because:

* [x] Waiting on a release that will include https://github.com/PyO3/maturin/pull/1416
  * CI is probably broken at the moment
* [x] UniFFI 0.23 (upcoming) will move things around, we probably want to have that first, then see how maturin needs to adopt that
* [x] Ensure the release tasks on CI work
  * Might need some test runs pushing to https://test.pypi.org/
* [ ] Remove check requirements on GitHub for `Python 3_6 tests`, `Python 3_6 tests minimum dependencies` and `Python 3_9 tests minimum dependencies` as those tests are gone.


perrymcmanis144 (approved):
👍 yep, I like it a lot.

I especially like having a pyproject.toml

badboy:
fwiw I pushed a locally-built package to pypi test here: https://test.pypi.org/project/glean-sdk/52.0.0/
then installed it manually (fetching dependencies from normal pypi though) and that all worked out just fine.

IMO this is good to go.

PR #2348, opened 2023-01-24
Documentation update to indicate time for distribution 

perrymcmanis144:
Update to the book to indicate that distributions take a few days to settle. I also made a change to temper language in one spot that I noticed right afterwards.


travis79 (approved):
LGTM!

PR #2352, opened 2023-01-26
properly invoke the windows buildid function from whatsys

perrymcmanis144:
Due to this code being off the execution path in Mac/Linux, the prior PR managed to pass tests even though it was not invoking this function correctly.

This rectifies that.


travis79 (approved):
😅

PR #2358, opened 2023-01-30
Rust: Update env_logger dependency

badboy:
m-c wants to update for sure: https://bugzilla.mozilla.org/show_bug.cgi?id=1813411


rosahbruno (approved):
Seems good, no tests failing. Though it does appear in the ticket they are backing out the changes?

badboy:
> Seems good, no tests failing. Though it does appear in the ticket they are backing out the changes?

It did land now and also it takes some time until we get a Glean release with this into m-c

PR #2363, opened 2023-02-02
Swift: Remove super long-running test

badboy:
This test takes up to 40s on a dev machine and closer to 70s on CI.
The underlying implementation has changed since the regression.
We also still have tests for this in Kotlin and Python (Python takes ~7s),
so we can keep those only.

Second commit reduces timeouts just a tiny bit.
Both commits can be reviewed on each own and also separate landed.


travis79 (approved):
⏩

PR #2364, opened 2023-02-02
Remove label regex in favour of just the 60-byte limit

chutten:
(and ASCII)


docs/user/reference/metrics/index.md
```diff
@@ -66,17 +66,7 @@ Labeled metrics come in two forms:
...
-Specifically, they must conform to this regular expression:
-
-```
-^[a-z_][a-z0-9_-]{0,29}(\\.[a-z_][a-z0-9_-]{0,29})*$
-```
-
-Check your label:
-<input type="text" label="Label" id="label">
-<span id="result">unchecked</span>
+Labels must not exceed 60 characters in length, and may comprise any ASCII characters.
```
badboy, line 69: *me starts putting null bytes in there*

badboy: Your proposal calls out "Or at least everything printable." Maybe we should go for that.

glean-core/src/metrics/labeled.rs
```diff
@@ -338,8 +278,8 @@ pub fn validate_dynamic_label(
         );
         record_error(glean, meta, ErrorType::InvalidLabel, msg, None);
         true
-    } else if !matches_label_regex(label) {
-        let msg = format!("label must be snake_case, got '{}'", label);
+    } else if !label.is_ascii() {
```
badboy, line 281: I was about to suggest something like `.is_printable` but apparently that doesn't exists.
There's only [`char::is_ascii_control`](https://doc.rust-lang.org/std/primitive.char.html#method.is_ascii_control),
so this would need to be:

```suggestion
    } else if label.chars().any(|c| c.is_ascii_control()) {
```

Note-worth: this allows spaces, but I guess that's fine.

glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
```diff
@@ -130,17 +130,29 @@ class LabeledMetricTypeTests: XCTestCase {
...
         XCTAssertEqual(
-            4,
+            0,
             labeledCounterMetric.testGetNumRecordedErrors(.invalidLabel)
         )
 
-        XCTAssertEqual(4, labeledCounterMetric["__other__"].testGetValue())
+        XCTAssertEqual(nil, labeledCounterMetric["__other__"].testGetValue())
+
+        // More than 61 characters? Not okay.
```
badboy, line 146:
```suggestion
        // More than 60 characters? Not okay.
```

That's the limit (using `61` is not _wrong_ here, but inaccurate I guess)

badboy (approved):
Minimal comment, otherwise good to go.

glean-core/ios/GleanTests/Metrics/LabeledMetricTests.swift
chutten: Ah ha, but excitingly you're wrong! If you put in a 61-character label it will be accepted.

badboy: then it's wrong in other places!
At least the Python tests say "60 characters" and use 70 and the kotlin tests say "60 characters" and use 62.

chutten:
Interestingly, the limit should actually be 71 since category's max length is 40, not 30. Let's standardize on 71B of printable ASCII (` ` to `~`, inclusive)

badboy:
> Interestingly, the limit should actually be 71 since category's max length is 40, not 30. Let's standardize on 71B of printable ASCII (` ` to `~`, inclusive)

Thanks for double-checking and adjusting.

docs/user/_includes/label-limits.md
```diff
@@ -1,5 +1,6 @@
-* Labels must conform to the [label formatting regular expression](index.md#label-format).
-* Each label must have a maximum of 60 bytes, when encoded as UTF-8.
+* Labels must conform to the [label format](index.md#label-format).
+* Each label must have a maximum of 71 bytes, when encoded as UTF-8.
+* Each label must only contain printable ASCII characters.
```
badboy, line 3: Well, given it's ASCII and for that "characters" is bascially intuitive what do you think about just phrasing that as such? Might be easier for people to read and understand.

```suggestion
* Each label must have a maximum of 71 characters.
* Each label must only contain printable ASCII characters.
```

(though I know we use the "encoded as UTF-8" part in other places, so maybe keeping it consistent is the better call here. I let you decide)

glean-core/src/metrics/labeled.rs
```diff
@@ -338,8 +278,8 @@ pub fn validate_dynamic_label(
         );
         record_error(glean, meta, ErrorType::InvalidLabel, msg, None);
         true
-    } else if !matches_label_regex(label) {
-        let msg = format!("label must be snake_case, got '{}'", label);
+    } else if !label.is_ascii() || label.chars().any(|c| c.is_ascii_control()) {
```
badboy, line 281: performance nit: this does two loops: one to check everything is ascii, one to check for control characters.
In theory this should be "faster":

```suggestion
    } else label.chars().any(|c| !c.is_ascii() || c.is_ascii_control()) {
```

but honestly for strings <= 71 chars it probably doesn't matter whatsoever.

badboy (approved):
needs a rebase because of the changelog. see inline nits, good to go then!

PR #2368, opened 2023-02-06
Bug 1716847 - Fix for Glean internal metric `baseline.duration` InvalidState errors

travis79:
This puts an interlock on calling `glean_handle_active` to ensure it is only called once until `glean_handle_inactive` is called.

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -88,6 +88,9 @@ public class Glean {
     // order to simulate not running in the main process.  DO NOT SET EXCEPT IN TESTS!
     var isMainProcess: Bool?
 
+    // Tracks the active/inactive state to prevent calling `handleClientActive` multiple times.
+    var isActive: Bool = false
+
     private init() {
         // intentionally left private, no external user can instantiate a new global object.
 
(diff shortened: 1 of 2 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -231,14 +234,20 @@ public class Glean {
 
     /// Handle foreground event and submit appropriate pings
     func handleForegroundEvent() {
-        gleanHandleClientActive()
+        if !isActive {
+            gleanHandleClientActive()
+            isActive = true
+        }
 
         GleanValidation.foregroundCount.add(1)
```
badboy, line 242: This can then be off-by-one, right?
Guess that's fine. Maybe we should just get rid of this metric anyway.

badboy (approved):

glean-core/ios/Glean/Glean.swift
travis79: Yeah, it could potentially be compared to the number of baseline pings with reason=active to see how often this happens in the wild, but that would better be instrumented as a dedicated metric in an `else` branch here. As far as I know we aren't even using this for anything currently, so it could be removed.

PR #2372, opened 2023-02-07
Update Android commandlinetools to version 9.0

rvandermeulen:


chutten:
I'll rerun the failed test to double-check it's an intermittent.

@rvandermeulen , is this for a specific bug?

rvandermeulen:
@chutten I'm going to be filing a bug for the m-c bump that we could attach it to if you wanted. Otherwise, no.

chutten (approved):

PR #2376, opened 2023-02-10
Bug 1814592 - Instrument the time needed for a successful/failure upload of a ping & upload shutdown in testing

badboy:
This PR does multiple things:

* Bring back #2365 -- that's the first 3 commits, cherry-picked unchanged (#2374 reverted them)
* Refactors the uploader shutdown to be used in test reset 
* Implement uploader shutdown in Swift
  * This is only used in test mode, because `glean::shutdown` is not called by the iOS bindings
  * Tries to be a bit more defensive for the newly introduced timing distributions done in the uploader.

Out of 20+ runs locally I had only 3 test failures ("No database found"). I don't fully understand why that still happens.
But hey, with only #2365 we had 7 tests fail basically every time. Improvement!

Especially the last part though bugs me. The iOS uploader still has some runs where it does some work after we told it to shut ~~up~~ down. And yet it happens in some tests.
Because Glean Swift _does not_ shut down in production use (it either runs or the embedding app is stopped/killed), this should not occur.
If it _does_ occur though this is strictly better by NOT crashing, but just logging an error now (but also: we only cover the known cases of metric types Glean collects there itself).

Time-wise: I'm still around on Monday, so we can have that reviewed and merged before.
I'd rather not have #2332 shipping without instrumentation, but it's also essentially desktop only for the time being in `main`, so it could.

I haven't (yet) tested this exact code on Desktop, but a prior version at least compiled in m-c just fine and ran tests.

### DIFF
--- a/glean-core/ios/Glean/Net/HttpPingUploader.swift
+++ b/glean-core/ios/Glean/Net/HttpPingUploader.swift
@@ -4,6 +4,30 @@
 
 import Foundation
 
+/// `true` to allow the uploader to process pings.
+/// Note that this does not mean that an uploader is actually running.
+/// It will be invoked when a ping is submitted.
+///
+/// `false` to stop the uploader from starting new uploads.
+/// An already running uploader will finish work and then stop.
+var stateRunAllowed: AtomicBoolean = AtomicBoolean(false)
+
+/// The uploader will enter and leave this group as it starts and finishes work.
+let uploadGroup = DispatchGroup()
+
+// TODO(bug 1816403): Move this and the associated global state
+// into a singleton instance of `HttpPingUploader`.
+func shutdownUploader() {
+    stateRunAllowed.value = false
+    // Wait for an uploader to finish it's current upload.
+    // If no uploader is running, this returns immediately.
+    uploadGroup.wait()
+}
+
+func startUploader() {
+    stateRunAllowed.value = true
+}
+
 /// This class represents a ping uploader via HTTP.
 ///
 /// This will typically be invoked by the appropriate scheduling mechanism to upload a ping to the server.
@@ -143,6 +167,12 @@ public class HttpPingUploader {
     /// It will report back the task status to Glean, which will take care of deleting pending ping files.
     /// It will continue upload as long as it can fetch new tasks.
     func process() {
+        if !stateRunAllowed.value {
+            self.logger.info("Not allowed to continue running. Bye!")
+        }
+
+        uploadGroup.enter()
+
         // Limits are enforced by glean-core to avoid an inifinite loop here.
         // Whenever a limit is reached, this binding will receive `.done` and step out.
         let task = gleanGetUploadTask()
@@ -152,6 +182,10 @@ public class HttpPingUploader {
             var body = Data(capacity: request.body.count)
             body.append(contentsOf: request.body)
             self.upload(path: request.path, data: body, headers: request.headers) { result in
+                defer {
+                    uploadGroup.leave()
+                }
+
                 let action = gleanProcessPingUploadResponse(request.documentId, result)
                 switch action {
                 case .next:
@@ -162,7 +196,6 @@ public class HttpPingUploader {
                 case .end:
                     return
                 }
-
             }
         case .wait(let time):
             sleep(UInt32(time) / 1000)
@@ -170,8 +203,9 @@ public class HttpPingUploader {
             Dispatchers.shared.launchAsync {
                 HttpPingUploader(configuration: self.config, testingMode: self.testingMode).process()
             }
+            uploadGroup.leave()
         case .done:
-            return
+            uploadGroup.leave()
         }
     }
 }
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -248,8 +248,9 @@ pub trait OnGleanEvents: Send {
     ///   * New tasks will be ignored.
     /// * This SHOULD NOT block arbitrarily long.
     ///   * Shutdown waits for a maximum of 30 seconds.
-    fn shutdown(&self) {
+    fn shutdown(&self) -> Result<(), CallbackError> {
         // empty by default
+        Ok(())
     }
 }
 
@@ -502,7 +503,9 @@ fn uploader_shutdown() {
         .name("glean.shutdown".to_string())
         .spawn(move || {
             let state = global_state().lock().unwrap();
-            state.callbacks.shutdown();
+            if let Err(e) = state.callbacks.shutdown() {
+                log::error!("Shutdown callback failed: {e:?}");
+            }
 
             // Best-effort sending. The other side might have timed out already.
             let _ = tx.send(()).ok();
--- a/glean-core/src/lib.rs
+++ b/glean-core/src/lib.rs
@@ -517,6 +517,8 @@ pub fn shutdown() {
     // Call on_shutdown.
     // Need to be done without holding a lock on the global Glean object.
     {
+        let timer_id =
+            core::with_glean(|glean| glean.additional_metrics.shutdown_wait.start_sync());
         let (tx, rx) = unbounded();
 
         let handle = thread::Builder::new()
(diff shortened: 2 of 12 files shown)


glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -4,6 +4,27 @@
...
+func shutdownUploader() {
+    stateRunning.value = false
+    // Wait for an uploader to finish it's current upload.
+    // If no uploader is running, this returns immediately.
+    uploadGroup.wait()
+}
+
+func startUploader() {
+    stateRunning.value = true
+}
```
badboy, line 26: oh this whole thing? more mutable global state!
I'm considering moving this into `HttpPingUploader` below and have it be a singleton like `Glean`

badboy:
> Base: **28.57**% // Head: **26.31**% // Decreases project coverage by **`-2.26%`** ⚠️

Ah yeah, new metrics, but no tests for those in Glean. Didn't I add one or did that get lost in my rebasing? Hmmm

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -216,9 +224,14 @@ impl TimingDistributionMetric {
...
 
-        glean
-            .storage()
-            .record_with(glean, &self.meta, |old_value| match old_value {
+        // Let's be defensive here:
+        // The uploader tries to store some timing distribution metrics,
+        // but in tests that storage might be gone already.
+        // Let's just ignore those.
+        // We do the same for counters.
+        // This should never happen in real app usage.
```
travis79, line 232: 😬

travis79 (approved):
I'd not seen the inline Codecov comments before, kinda nifty.  This looks good to me, especially after making Codecov happy.

glean-core/src/lib.rs
```diff
@@ -514,37 +563,7 @@ pub fn shutdown() {
...
-        //   * We don't know how long uploads take until we get data from bug 1814592.
-        let result = rx.recv_timeout(Duration::from_secs(30));
-
-        if result.is_err() {
-            log::warn!("Waiting for upload failed. We're shutting down.");
-        } else {
-            let _ = handle.join().ok();
-        }
-    }
+    uploader_shutdown();
```
chutten, line 566: Musing: I was reading through `uploader_shutdown` wondering if we could indeed call `with_glean` safely, and then get out here and read up-method to find that we do guard against init being fully complete, so that's helpful... but oof, this shutdown stuff is getting more and more complex and I'm not sure where the ideal place is to document or manage this complexity.

glean-core/src/lib.rs
```diff
@@ -941,6 +960,13 @@ pub fn glean_test_destroy_glean(clear_stores: bool, data_path: Option<String>) {
 
         dispatcher::reset_dispatcher();
 
+        // Only useful if Glean was fully initialized.
```
chutten, line 963: Possible inconsistency: Is there a reason "Glean was fully initd" is determined differently here than at the top of shutdown()?

glean-core/src/metrics/timing_distribution.rs
chutten, line 232: Suggestion: ...which means we can log it so we'll have a chance at being told?

glean-core/src/upload/mod.rs
```diff
@@ -360,6 +365,18 @@ impl PingUploadManager {
...
 
+        {
+            let in_flight = self.in_flight.read().unwrap();
+            if in_flight.contains_key(document_id) {
+                log::warn!(
+                    "Attempted to enqueue an in-flight ping {} at {}.",
+                    document_id,
+                    path
+                );
+                return;
```
chutten, line 376: Suggestion: This `return;` could be a big deal. We should have a health metric for it and a place on the monitoring dashboards for it.

glean-core/src/upload/mod.rs
```diff
@@ -575,6 +592,13 @@ impl PingUploadManager {
                     }
                 }
 
+                {
+                    let mut in_flight = self.in_flight.write().unwrap();
+                    let success_id = self.upload_metrics.sending_success.start_sync();
+                    let failure_id = self.upload_metrics.sending_failure.start_sync();
```
chutten, line 598: Musing: I wonder if we should have safety notices for synchronous I/O. Like "Sync I/O: Acceptable because it's on the uploader thread which itself does waiting I/O on the network" or something. Might not be a problem until someone finds us on a perf profile.

glean-core/src/upload/mod.rs
```diff
@@ -656,14 +680,27 @@ impl PingUploadManager {
...
+
         match status {
             HttpStatus { code } if (200..=299).contains(&code) => {
                 log::info!("Ping {} successfully sent {}.", document_id, code);
+                if let Some((success_id, failure_id)) = send_ids {
+                    self.upload_metrics
+                        .sending_success
+                        .set_stop_and_accumulate(glean, success_id, stop_time);
+                    self.upload_metrics.sending_failure.cancel_sync(failure_id);
+                }
```
chutten, line 703: Suggestion: If `send_ids.is_none()` then that means 🤔  we've removed this doc_id before we got here, or we never inserted it in the first place... either way, that's an inconsistency. Perhaps we should instrument this?

chutten (approved):
Obviously also want Travis on here, but broadly I'm in favour of the approach and implementation.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -159,19 +186,20 @@ public class HttpPingUploader {
                     Dispatchers.shared.launchAsync {
                         HttpPingUploader(configuration: self.config, testingMode: self.testingMode).process()
                     }
+                    uploadGroup.leave()
```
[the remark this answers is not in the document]
travis79: I always forget `defer` even exists...   I mostly see it used more like a `try/finally` but yeah, it would work here.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -143,6 +164,12 @@ public class HttpPingUploader {
     /// It will report back the task status to Glean, which will take care of deleting pending ping files.
     /// It will continue upload as long as it can fetch new tasks.
     func process() {
+        if !stateRunning.value {
+            self.logger.info("Not allowed to continue running. Bye!")
```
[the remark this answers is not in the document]
travis79: It does, the logger tags this with `glean/HttpPingUploader`

glean-core/src/upload/mod.rs
travis79: I second this. But I think if we file a bug and drop a TODO here we don't need to block on this

glean-core/src/upload/mod.rs
travis79: +1 from me on instrumenting this (or filing a bug to instrument it)

travis79 (approved):

glean-core/src/upload/mod.rs
badboy: Instrument the instrumentation 🤔 I'm all for it.

glean-core/ios/Glean/Net/HttpPingUploader.swift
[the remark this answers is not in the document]
badboy: This only solves half the problem though.
In this `case` we're within the closure used by `self.upload`.
In the `.wait` and `.done` further below (lines 194ff) we're outside of it. 

So regardless we wil need one defer in there, but still the individual calls further below (because we don't want to double-fire it)

glean-core/src/lib.rs
badboy: Inconsistency indeed. `was_initialize_called` vs. "was called, did work and actually finish, including setting up a database"

glean-core/src/lib.rs
badboy: ugh, yeah, it's ... a lot and spread in several places. This could definitely use some refactoring to streamline it again.

badboy:
I addressed all comments and filed bugs.
I also found one more thing: The metrics were named differently in `metrics.yaml` then they were in code. I fixed that ([bug 1815213 when?](https://bugzilla.mozilla.org/show_bug.cgi?id=1815213))
I also re-wrote the `upload_timing` test to use the metric test APIs instead of parsing the JSON (this should also satisfy codecov, because we had tests, it just didn't know about it)

badboy:
image

yey

PR #2389, opened 2023-02-25
Upgrade Taskgraph + add missing Decision task route

### LINKED ISSUE #2388: Chain of Trust errors on signing tasks
RyanVM noticed some chain of trust errors:
https://firefox-ci-tc.services.mozilla.com/tasks/groups/CrIshxNpQPGHuAMsnsOjEA

This is due to a key rotation we did last week and the fact that the cached docker-image task was run on a worker that had the old key. To fix it, we'll need to cause a rebuild.

Unfortunately glean is using a non-standard index route which is causing the `add-new-jobs` action to fail, and it is also using an older version of Taskgraph that is missing the `rebuild-cached-tasks` action. Both of these things should be simple to fix.

ahal:
Fixes #2388


ahal:
Hello, could someone add the `releng` team as a reviewer here? I'll also need help getting it landed after review. Thanks in advance!

Dexterp37:
@JohanLorenzo would you kindly take a review pass at this? Can't flag you as reviewer, sorry :)

bhearsum (approved):

Dexterp37:
Cool, I'm re-runnng the tests just to be safe, but it shouldn't be impacted :)

PR #2399, opened 2023-03-05
Various Android build updates

rvandermeulen:


Dexterp37:
(Fly-by, because we've been bitten by this before: @rvandermeulen does A-C/Fenix/Focus already use this? Is this confirmed to work fine on Windows?)

rvandermeulen:
AC/Fenix/Focus builds fine with the latest Android Studio release. They don't use the NDK toolchain directly, but do compile with the 33 SDK already. Regardless, I'm not super worried about issues here since the toolchain updates are just minor point releases to what we're already using anyway.

Dexterp37 (approved):

PR #2402, opened 2023-03-07
Bug 1820792 - handle rkv DatabaseCorrupted error by creating a new rkv

perrymcmanis144:
As discussed, this change only covers having this log in case of a DatabaseError.

I could find no tests around InvalidFile so I have not included any new ones.

This change is small enough that I wasn't sure a changelog entry was justified, especially given this isn't a change external folks should ever notice.

### DIFF
--- a/glean-core/src/database/mod.rs
+++ b/glean-core/src/database/mod.rs
@@ -52,6 +52,12 @@ pub fn rkv_new(path: &Path) -> std::result::Result<Rkv, rkv::StoreError> {
             // Now try again, we only handle that error once.
             Rkv::new::<rkv::backend::SafeMode>(path)
         }
+        Err(rkv::StoreError::DatabaseCorrupted) => {
+            let safebin = path.join("data.safe.bin");
+            fs::remove_file(safebin).map_err(|_| rkv::StoreError::DatabaseCorrupted)?;
+            // Try again, only allowing the error once.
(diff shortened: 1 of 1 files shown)


travis79 (approved):
LGTM! Are there any more errors that might be returned here that would be worth logging?

perrymcmanis144:
[There are other errors](https://docs.rs/rkv/latest/rkv/enum.StoreError.html#variant.ManagerPoisonError)

butttt in terms of which ones are worth actually removing the existing, that's potentially a larger discussion. In this case, I think *this* error fits in the spirit of what I believe rkv_new is desired to do. The rest should at least get logged.

If you're good with that thinking, I will hit the button.

travis79:
> [There are other errors](https://docs.rs/rkv/latest/rkv/enum.StoreError.html#variant.ManagerPoisonError)
> 
> butttt in terms of which ones are worth actually removing the existing rkv over, that's potentially a larger discussion. In this case, I think _this_ error fits in the spirit of what I believe rkv_new is desired to do. The rest should at least should get logged down the line as the error gets passed.
> 
> Edit, ah ok in this case I think what I'm actually doing is more accurately described as _handle DatabaseCorrupted errors by creating a new RKV_, so I think what I should do is update the PR name and commit message and then merge.
> 
> Any opposition @travis79 ?

Nope...  thanks for updating the title and commit message for documentation purposes :)

PR #2404, opened 2023-03-08
Bug 1820927 - Update remove_dir_all to v0.8.1

travis79:


glean-core/Cargo.toml
```diff
@@ -42,7 +42,7 @@ crossbeam-channel = "0.5"
 thiserror = "1.0.4"
 uniffi = { version = "0.23.0", default-features = false }
 time = "0.1.40"
-remove_dir_all = "0.5.3"
+remove_dir_all = "0.8.1"
```
Dexterp37, line 45: The current failing on CI seems to indicate that this version bump requires a specific version bump of rustc. Before merging this, we should make sure what's the minimum required version on m-c. Leaving r- for now until we figure this out.

Dexterp37 (changes requested):

glean-core/Cargo.toml
travis79: Looks like `rust-version` is pinned on lots of different versions in m-c.  From very early versions up to a few newer versions tha 1.62 that we would need.

Application-Services doesn't pin any versions that I can see.

travis79:
This is still blocked due to [an upstream bug](https://github.com/rbtcollins/fs_at/issues/55) causing bustage in our iOS builds.  Can't land this yet.

travis79:
Upstream bug resolved and a release is already out!  Bumping the dependency in Cargo.lock and kicking the CI again.

PR #2408, opened 2023-03-09
Bug 1821437 - Update iOS simulator from iPhone 11 to 14

travis79:


rosahbruno (commented):
Looks like CircleCI failed because it can't find the iPhone 14 sim. Also, locally I see two other files that have references to iPhone 11 (for CircleCI). Do we want to update those as well?
- `.circleci/config.yml`
- `.circleci/jazzy.yml`

travis79:
> Looks like CircleCI failed because it can't find the iPhone 14 sim. Also, locally I see two other files that have references to iPhone 11 (for CircleCI). Do we want to update those as well?
> 
>     * `.circleci/config.yml`
> 
>     * `.circleci/jazzy.yml`

Probably...   let me do a more thorough search for places to update this, and maybe I can add it as a make command or something? 🤔

travis79:
Hmm....  this may require me to update the version of Xcode we are using also.  I'm going to let this sit for a bit while I think about whether or not it's worth it right now.

badboy:
#2253 updates Xcode and switches to an iPhone 14 simulator.

travis79:
> #2253 updates Xcode and switches to an iPhone 14 simulator.

Ahh, I probably should have looked for this first. I was just annoyed that I don't have iPhone 11 simulator installed locally while trying to run tests. It would be really nifty if GitHub told me "hey, there's this other PR that's doing something like this too...."

Closing this in favor of the other PR.

PR #2415, opened 2023-03-14
Revert to libstd's remove_dir_all instead of external crate.

badboy:
This avoids the Windows dependencies.
The bug that caused us to switch to the crate in the first place was addressed in Rust 1.66+ (https://github.com/rust-lang/rust/pull/104558)


travis79 (approved):
This is better this way, thanks!

PR #2416, opened 2023-03-14
Use the sparse protocol everywhere

badboy:
It became available in Rust 1.68.0
Additionally lets avoid incremental builds on CI.
That's just extra work which we don't need there.


badboy:
Rust test stable old: 4m 41s
Rust tets stable this PR: 3m 8s

across pretty much all of the tasks we shave of 30-60s it seems, as they are all building Rust.

perrymcmanis144 (commented):
looks like it does what it sets out to do, lgtm 👍

perrymcmanis144 (approved):
ok, approved on this comment

PR #2425, opened 2023-03-20
Bug 1820792 - Create new RKV on DatabaseCorrupted Error and log error

perrymcmanis144:
Putting this up so we can discuss/not forget about it

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v52.4.2...main)
 
 * Rust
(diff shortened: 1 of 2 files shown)


badboy (changes requested):
👍 This seems better than the status quo because at least we would get back to a working state again.

It would be great if we start recording when these errors happen so we actually know how frequent that is.
That will require some plumbing to pass back the error and record it once we have a database though. Best done in a followup, could you file a bug?

We should also rebase this once the latest release has landed in `main` so that the changelog ends up in the right place.

badboy (approved):

PR #2427, opened 2023-03-21
Adjust docs to call out the right maximum number of extra keys

badboy:
The limit was raised a while ago and changed in the next section of the docs, but missed here.

[docs only]


perrymcmanis144 (approved):
hmm, could have sworn I clicked approve...

PR #2428, opened 2023-03-21
Bug 1823676 - expose Server Knobs APIs for mobile

rosahbruno:
Enable server knobs for mobile by exposing API to Kotlin and Swift bindings.

### Testing with the sample Glean projects
Tested using the debug ping viewer.

#### iOS
At the end of `func application()` in `AppDelegate.swift` add
```
glean.setMetricsDisabledConfig("{\"test.string_list\":true}")
```

`true` if you want to DISABLE the metric, `false` if you want to ENABLE the metric

You also need to add the `sample` ping in `send_in_pings` for the `test.string_list` metric

#### Android
In `GleanApplication.kt`

At the end of `onCreate()` add

```
val disabledMetrics = mapOf("test.string_list" to true)
Glean.setMetricsDisabledConfig(JSONObject(disabledMetrics).toString())
```

`true` if you want to DISABLE the metric, `false` if you want to ENABLE the metric


glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -394,6 +394,16 @@ open class GleanInternalAPI internal constructor() {
...
     }
 
+    /**
+     * Set configuration for metrics' disabled property, typically from a remote_settings
+     * experiment or rollout.
+     *
+     * @param json Stringified JSON map of metrics and their associated `disabled` property.
+     */
+    fun setMetricsDisabledConfig(json: String) {
+        gleanSetMetricsDisabledConfig(json)
```
badboy, line 404: @travis79 What type does the config have over in Nimbus Kotlin/Swift? Does it make sense to just take the string here or is it possible to take a dedicated object (or json object?)

badboy (commented):
this was waaaay to easy.

~~Waiting for Travis to answer.~~

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
rosahbruno: @badboy 
I chose to make it a string since we already check that the JSON is actually a `MetricsDisabledConfig` type.
https://github.com/mozilla/glean/blob/main/glean-core/src/lib.rs#L778

badboy: oh perfect. Sorry for the run-around then.

badboy (approved):

travis79 (approved):
🚀

PR #2429, opened 2023-03-23
Bug 1822482 - Switch from chrono to time 0.3

badboy:
This is a frist try at migrating from chrono to time. This compiles, but with warnings.
Tests don't compile, as they still largely make use of chrono.

Things that would need to be tested:

* Do we handle timezones correctly?
* Are we handling time measurement correctly?
  * This switched from `time::precise_time_ns()` to libstd's `Instant`. We need to verify the guarantees.
* Are we correctly (de)serializing data?
  * Datetimes are stored in the database. We need to handle the old data transparently. `DatetimeWrapper` should do that, but will require more testing.
* Truncation/Formatting
  * We use `TimeUnit` for the date/time related metrics to truncate. As this switches formatting we need to verify it's correct (especially at the nano-/micro-/millisecond level)

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -170,7 +170,7 @@ dependencies = [
  "num-integer",
  "num-traits",
  "serde",
- "time",
+ "time 0.1.45",
  "winapi",
 ]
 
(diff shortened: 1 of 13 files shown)


glandium:
> This switched from time::precise_time_ns() to libstd's Instant. We need to verify the guarantees.

FWIW, I looked into this last week or the week before, and libstd has the same, or better, precision.

PR #2433, opened 2023-03-27
Bug 1815233 - allow custom data path on mobile

rosahbruno:
For both iOS and Android we add a new parameter to `initialize` to let the user specify the specific data path they want to write to. When the user registers a custom data path, the app lifecycle observers are not setup.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,7 +8,7 @@
 * Kotlin
   * Adds the ability to record metrics on a non-main process. This is enabled by setting a `dataPath` in the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
 * iOS
-  * Adds the ability to record metrics on a non-main process. This is enabled by passing a `dataPath` to the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
+  * Adds the ability to record metrics on a non-main process. This is enabled by setting a `dataPath` in the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
 
 # v52.4.3 (2023-03-24)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,10 @@
 * General
   * On Rkv detecting a corrupted database delete the old RKV, create a new one and log the error ([#2425](https://github.com/mozilla/glean/pull/2425))
   * Add the Date header as late as possible before the uploader acts ([#2436](https://github.com/mozilla/glean/pull/2436))
+* Kotlin
+  * Adds the ability to record metrics on a non-main process. This is enabled by passing a `dataPath` to the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
+* iOS
+  * Adds the ability to record metrics on a non-main process. This is enabled by passing a `dataPath` to the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
 
 # v52.4.3 (2023-03-24)
 
@@ -48,11 +52,6 @@
 
 * General
   * No functional change from v52.3.0, just CI updates.
-
-* Kotlin
-  * Allow the user to specify a custom data path when calling `initialize`.
-* iOS
-  * Allow the user to specify a custom data path when calling `initialize`.
 
 # v52.3.0 (2023-02-23)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -49,6 +49,11 @@
 * General
   * No functional change from v52.3.0, just CI updates.
 
+* Kotlin
+  * Allow the user to specify a custom data path when calling `initialize`.
+* iOS
+  * Allow the user to specify a custom data path when calling `initialize`.
+
 # v52.3.0 (2023-02-23)
 
 [Full changelog](https://github.com/mozilla/glean/compare/v52.2.0...v52.3.0)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@
   * On Rkv detecting a corrupted database delete the old RKV, create a new one and log the error ([#2425](https://github.com/mozilla/glean/pull/2425))
   * Add the Date header as late as possible before the uploader acts ([#2436](https://github.com/mozilla/glean/pull/2436))
 * Kotlin
-  * Adds the ability to record metrics on a non-main process. This is enabled by passing a `dataPath` to the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
+  * Adds the ability to record metrics on a non-main process. This is enabled by setting a `dataPath` in the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
 * iOS
   * Adds the ability to record metrics on a non-main process. This is enabled by passing a `dataPath` to the Glean configuration ([bug 1815233](https://bugzilla.mozilla.org/show_bug.cgi?id=1815233))
 
--- a/docs/user/reference/general/initializing.md
+++ b/docs/user/reference/general/initializing.md
@@ -117,7 +117,7 @@ The Glean Kotlin SDK supports use across multiple processes. This is enabled by 
 
 Requirements for a non-main process
 - `Glean.initialize` must be called with the `dataPath` value set in the `Glean.Configuration`.
-- The default `dataPath` for Glean is `glean_data`. If you try and use `glean_data`, `Glean.initialize` will fail and throw an error.
+- The default `dataPath` for Glean is `{context.applicationInfo.dataDir}/glean_data`. If you try to use this path, `Glean.initialize` will fail and throw an error.
 
 **Note**: When initializing from a non-main process with a specified `dataPath`, the lifecycle observers will not be set up. This means you will not receive otherwise scheduled [baseline](../../user/pings/baseline.md) or [metrics](../../user/pings/metrics.md) pings.
 
@@ -182,7 +182,7 @@ The Glean Swift SDK supports use across multiple processes. This is enabled by s
 
 Requirements for a non-main process 
 - `Glean.initialize` must be called with the `dataPath` value set in the `Glean.Configuration`.
-- The default `dataPath` for Glean is `glean_data`. If you try and use `glean_data`, `Glean.initialize` will fail and throw an error.
+- On iOS devices, Glean stores data in the Application Support directory. The default `dataPath` Glean uses is `{FileManager.default.urls(for: .applicationSupportDirectory, in: .userDomainMask)[0]}/glean_data`. If you try to use this path, `Glean.initialize` will fail and throw an error.
 
 **Note**: When initializing from a non-main process with a specified `dataPath`, the lifecycle observers will not be set up. This means you will not receive otherwise scheduled [baseline](../../user/pings/baseline.md) or [metrics](../../user/pings/metrics.md) pings.
 
--- a/docs/user/reference/general/initializing.md
+++ b/docs/user/reference/general/initializing.md
@@ -113,13 +113,13 @@ class SampleApplication : Application() {
 }
 ```
 
-The Glean Kotlin SDK supports use across multiple processes. To configure this you call `initialize` as usual from your main process. From the secondary process you can call `initialize` again with the `dataPath` parameter. This allows you to store data locally in a separate location for the secondary process.
+The Glean Kotlin SDK supports use across multiple processes. This is enabled by adding a `dataPath` value to the `Glean.Configuration` object passed to `Glean.initialize`. You **do not** need to pass a `dataPath` for your main process. This configuration should only be used by a non-main process.
 
-Requirements for a secondary process
-- `initialize` must be called from the main thread
-- The reserved data path `glean_data` cannot be used
+Requirements for a non-main process
+- `Glean.initialize` must be called from the main thread.
+- The default `dataPath` for Glean is `glean_data`. If you try and use `glean_data`, `Glean.initialize` will fail and throw an error.
 
-**Note**: When initializing from a secondary process with a specified data path, the lifecycle observers will not be set up. This means you will not receive otherwise scheduled baseline or metrics pings.
+**Note**: When initializing from a non-main process with a specified `dataPath`, the lifecycle observers will not be set up. This means you will not receive otherwise scheduled baseline or metrics pings.
 
 ### Consuming Glean through Android Components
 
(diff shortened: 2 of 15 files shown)


CHANGELOG.md
```diff
@@ -2,6 +2,11 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v52.3.0...main)
 
+* Kotlin
+  * Allow the user to specify a custom data path when calling `initialize`.
```
travis79, line 6: Let's clarify this a bit and instead of "when calling `initialize`", let's say we are adding the ability to record metrics from off the main process and link to the bug explaining the what and the why

docs/user/reference/general/initializing.md
```diff
@@ -113,9 +113,13 @@ class SampleApplication : Application() {
 }
 ```
 
-The Glean Kotlin SDK does not support use across multiple processes, and must only be initialized on the application's main process. Initializing in other processes is a no-op.
+The Glean Kotlin SDK supports use across multiple processes. To configure this you call `initialize` as usual from your main process. From the secondary process you can call `initialize` again with the `dataPath` parameter. This allows you to store data locally in a separate location for the secondary process.
```
travis79, line 116: I think I'd rather the data-path be in the Configuration object with a default to the current `glean_data`. My fear is that users will see this parameter and think that they need to supply the data path for Glean when they shouldn't.

docs/user/reference/general/initializing.md
```diff
@@ -174,8 +178,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
 }
 ```
 
-The Glean Swift SDK does not support use across multiple processes,
-and must only be initialized on the application's main process.
+The Glean Swift SDK supports use across multiple processes. To configure this you call `initialize` as usual from your main process. From the secondary process you can call `initialize` again with the `dataPath` parameter. This allows you to store data locally in a separate location for the secondary process.
+
+Requirements for a secondary process
+- `initialize` must be called from the main thread
```
travis79, line 184: Does it _really_ need to be called on the main thread on a background process?

docs/user/reference/general/initializing.md
```diff
@@ -174,8 +178,13 @@ class AppDelegate: UIResponder, UIApplicationDelegate {
...
+- The reserved data path `glean_data` cannot be used
```
travis79, line 185: What happens if they try?

docs/user/user/pings/ping-schedules-and-timings.md
```diff
@@ -19,3 +19,5 @@ The `submission_timestamp` is the time the ping was received at the telemetry en
 The "Baseline 4" ping illustrates an important corner case. When "Session 2" ended, the OS also shut down the entire process, and the Glean SDK did not have an opportunity to send a `baseline` ping immediately.  In this case, it is sent at the next available opportunity when the application starts up again in "Run 2".  This `baseline` ping is annotated with the reason code `dirty_startup`.
 
 The "Metrics 2" ping likewise illustrates another important corner case. "Metrics 1" was able to be sent at the target time of 04:00 (local device time) because the application was currently running.  However, the next time 04:00 came around, the application was not active, so the Glean SDK was unable to send a `metrics` ping.  It is sent at the next available opportunity, when the application starts up again in "Run 2".  This `metrics` ping is annotated with the reason code `overdue`.
+
+**NOTE**: If you are running Glean on a secondary process where to pass a custom `dataPath` to the `initialize` call for either the Kotlin or Swift SDKs, **these life-cycles will not be set up**.
```
travis79, line 23:
```suggestion
**NOTE**: ping scheduling and other application lifecycle dependent activities are not set up when Glean is used in a secondary process. See <insert link to secondary process docs here>
```

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -38,12 +40,14 @@ data class BuildInfo(val versionCode: String, val versionName: String, val build
...
-        // the @MainThread decorator and the `assertOnUiThread` call.
-        MainScope().launch {
-            ProcessLifecycleOwner.get().lifecycle.addObserver(glean.gleanLifecycleObserver)
+        // Only set up the lifecycle observers if we don't provide a custom
+        // data path.
+        if (glean.getIsCustomDataPath() != null && glean.getIsCustomDataPath() == false) {
+            MainScope().launch {
+                ProcessLifecycleOwner.get().lifecycle.addObserver(glean.gleanLifecycleObserver)
+            }
         }
```
travis79, line 49: Maybe a less java-like and more kotlin-like way to write this would be something like:
```
glean.getIsCustomDataPath()?.takeIf { !it }?.let {
	MainScope().launch {
		ProcessLifecycleOwner.get().lifecycle.addObserver(glean.gleanLifecycleObserver)
	}
}
```

the `?` operator will handle the null for you and the `takeIf { !it }` will process on the returned value and return `null` if `getIsCustomDataPath` returns true, so the second `?` would escape. Otherwise, if it's not `null` and it returns `false` (`!it`), then it falls into the `let { ... }` block.

Checking for `null` in Kotlin is just sorta weird, but feel free to find something that is more readable if my suggestion is too arcane or something :)

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -126,6 +137,8 @@ open class GleanInternalAPI internal constructor() {
     // Store the build information provided by the application.
     internal lateinit var buildInfo: BuildInfo
 
+    private var isCustomDataPath: Boolean? = null
```
travis79, line 140: Any reason not to simplify things and just set this to `false` by default rather than making it an optional Boolean?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -141,8 +154,6 @@ open class GleanInternalAPI internal constructor() {
      * A LifecycleObserver will be added to send pings when the application goes
      * into foreground and background.
      *
-     * This method must be called from the main thread.
-     *
```
travis79, line 145: We still mention this in the documentation, why remove this comment here?

travis79 (changes requested):
I made a first pass on this, but didn't get through everything.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
badboy: `getIsCustomDataPath` should just be `isCustomDataPath`.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
rosahbruno: I did this (at first) to guard any scenario where `isCustomDataPath` might be checked before it has a value set. I didn't want it to default to false in a scenario where it was still `null` and hadn't been checked yet.

But for the two places where I check it, it should be properly initialized by then, so I will make it non-optional.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
rosahbruno: I changed this to just use `isCustomDataPath` now and it will never be null, so that should simplify the check. I appreciate the Kotlin way to do it though, I'll keep that in mind for the future!

https://github.com/mozilla/glean/pull/2433/commits/6fbe824df65632daecdba0d917d42b38d6254c76#diff-aa8825c678ddc5d873d6b93e255f5b246fe265822b922a6f50d775b0e98b1e97R45

docs/user/reference/general/initializing.md
rosahbruno: This is listed in Jan-Erik's proposal as a requirement, so I am just going based off of that.

@badboy 
Is there any specific reason it needs to be called from the main thread?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -152,28 +163,57 @@ open class GleanInternalAPI internal constructor() {
...
     @JvmOverloads
     @Synchronized
-    @MainThread
     fun initialize(
         applicationContext: Context,
         uploadEnabled: Boolean,
         configuration: Configuration = Configuration(),
-        buildInfo: BuildInfo
+        buildInfo: BuildInfo,
+        dataPath: String? = null
```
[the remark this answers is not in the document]
rosahbruno: I agree, let me give that a try. I originally had it in there and it was causing Android tests to fail for some reason. Then I looked at Python and data_path was included as an initialization parameter, so I changed these two to match.

I guess it makes more sense for Python since it is a requirement and for the mobile SDKs it is more of an option. I will update.

CHANGELOG.md
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/a0d09587cd56c122361ab8b948731b7902e96dc2

docs/user/reference/general/initializing.md
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/4ae32ae928accf9d563d33ec8d07ccdaf39e0443

docs/user/user/pings/ping-schedules-and-timings.md
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/4ae32ae928accf9d563d33ec8d07ccdaf39e0443

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -887,7 +887,7 @@ class GleanTest {
     }
 
     @Test
-    fun `Date header is set on actualy HTTP POST`() {
+    fun `Date header is set on actually HTTP POST`() {
```
badboy, line 890: I'm not good with typing, am I?

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -927,4 +927,30 @@ class GleanTest {
...
+        Glean.initialize(context, true, buildInfo = GleanBuildInfo.buildInfo)
+        assertFalse(Glean.isCustomDataPath)
+    }
+
+    @Test
+    fun `test glean does not initialize with invalid DB path`() {
+        Glean.testDestroyGleanHandle()
+
+        // The path provided here is invalid because of the leading and trailing spaces.
+        val cfg = Configuration(dataPath = " invalid db path ")
```
badboy, line 950: Hm, honestly I would expect a path with spaces to work just fine. What's the reason this doesn't work?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -163,17 +177,49 @@ open class GleanInternalAPI internal constructor() {
...
+
+            // The default database path is `glean_data`, the background process and the main
+            // process cannot write to the same file.
+            if (configuration.dataPath == "glean_data") {
+                Log.e(LOG_TAG, "Attempted to initialize Glean with an invalid database path \"glean_data\" is reserved")
+                return
+            }
+
+            // Check that the database path we are trying to write to is valid and writable.
+            if (!canWriteToDatabasePath(applicationContext.applicationInfo.dataDir, configuration.dataPath)) {
```
badboy, line 217: I just realized that in the proposal we 

> MUST pass a valid database path in the app’s data directory (context.applicationInfo.dataDir).

I understood that as "the user passes in a full path".
this code currently only essentially takes a subpath and concatenates it to the data directory.

For Python we accept a full path (obviously, the concept of a fixed "data dir" only exists for mobile application as such).
Still I still feel like having the user pass in a full path is a better choice.
Or we need to rename that config option, cause what we want is more of a directory name then (at the very least this needs to be made clear in the docs)

@travis79 what do you think?

badboy (changes requested):

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
rosahbruno: I added checks to make sure there aren't spaces at the start or end of the string. The spaces in the words are fine, its just if they are at the start or finish. It just seemed like a easy enough thing to check that would catch some weird cases.

If this doesn't seem like something we should do, then I can change that logic or have it check for other stuff.

glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -84,13 +84,49 @@ extension Date {
...
+    }
+
+    return documentsDirectory.appendingPathComponent(dataPath)
+}
+
+/// Check if the data path provided is valid and writable.
+///
+/// - parameters
+///     * customDataPath: An override for manually setting the data directory.
+func canWriteToDatabasePath(_ customDataPath: String? = "") -> Bool {
```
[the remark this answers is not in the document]
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/4961486d9b6b333743bdb5233dcddcc0e089d48f

glean-core/android/src/main/java/mozilla/telemetry/glean/utils/DataPathUtils.kt
```diff
@@ -0,0 +1,55 @@
...
+
+/**
+ * Check if the data path provided is valid and writable.
+ *
+ * @param dataDir A [String] that specifies the base directory where we will
+ *     store data.
+ * @param customDataPath An optional [String] provided by the user to specify
+ *     the specific path to store data.
+ */
+fun canWriteToDatabasePath(dataDir: String, customDataPath: String? = ""): Boolean {
```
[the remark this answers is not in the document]
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/4961486d9b6b333743bdb5233dcddcc0e089d48f

docs/user/reference/general/initializing.md
badboy: Looking at it now and I can't really say why I made it a requirement.
I think it's fine in NOT being on the main thread. We can remove the check here and adjust the proposal retroactively.

In the worst case we need to change it later, but we know the users needing this so we can help them in that case.

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/f237690dd8a2518ed2758d36a69b055bc0ff92bc

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: I agree, explicitly passing in the full path makes it even harder for someone to misuse this.

docs/user/reference/general/initializing.md
rosahbruno: Fixed in https://github.com/mozilla/glean/pull/2433/commits/90da5a9474a9fd02651560275dc902911c04e2e8

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
rosahbruno: @badboy @travis79 

What constitutes the full path here? For example, when I print the path right now after I concatenate the custom slug, I get `/data/user/0/org.mozilla.samples.gleancore/glean_sample_data`.

So when we say full path, does that mean just from the app ID onward? So they would supply `org.mozilla.samples.gleancore/glean_sample_data`?

glean-core/ios/Glean/Glean.swift
```diff
@@ -211,7 +250,7 @@ public class Glean {
         gleanSetExperimentInactive(experimentId)
     }
 
-    /// Tests wheter an experiment is active, for testing purposes only.
+    /// Tests whether an experiment is active, for testing purposes only.
```
travis79, line 253: 🐛 All these typos fixed 🎉

glean-core/ios/Glean/Utils/Utils.swift
```diff
@@ -84,13 +84,49 @@ extension Date {
...
+    return documentsDirectory.appendingPathComponent(dataPath)
+}
+
+/// Check if the data path provided is valid and writable.
+///
+/// - parameters
+///     * customDataPath: An override for manually setting the data directory.
+func canWriteToDatabasePath(_ customDataPath: String) -> Bool {
+    // Do not allow empty strings.
+    if customDataPath == "" {
```
travis79, line 114:
```suggestion
    if customDataPath.isEmpty {
```

badboy (approved):

PR #2436, opened 2023-03-30
Bug 1825298 - Add the `Date` header as late as possible before the uploader acts

badboy:


glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -883,4 +886,46 @@ class GleanTest {
...
+            )
+        )
+
+        Glean.handleBackgroundEvent()
+
+        // Before the `Date` was set on submit,
+        // so waiting here would ensure the actual sending is later.
+        // Now we set the date on actual sending,
+        // so we can observe it close to receive time further below.
+        Thread.sleep(2000)
```
badboy, line 907: Not necessarily a fan of intentionally slowing down our tests, but because this is all about time measurement we don't really have another better way

chutten (approved):
r+ with changelog.

PR #2443, opened 2023-04-06
Bug 1811253 - Invert remote metric config logic

travis79:
This is in order to simplify the understanding of how the API works to a more natural form.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,7 @@
 * General
   * On Rkv detecting a corrupted database delete the old RKV, create a new one and log the error ([#2425](https://github.com/mozilla/glean/pull/2425))
   * Add the Date header as late as possible before the uploader acts ([#2436](https://github.com/mozilla/glean/pull/2436))
+  * The logic of the Server Knobs API has been flipped. Instead of applying a list of metrics and their _disabled_ state, the API now accepts a list of metrics and their _enabled_ state ([bug 1811253](https://bugzilla.mozilla.org/show_bug.cgi?id=1811253))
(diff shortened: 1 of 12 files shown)


perrymcmanis144 (approved):
👍 

This is a nit and doc only so I won't block, but to me `// Let's convert this to the new API for backwards compatibility` could use some elaboration. It's clear from reading the code that you want to flip element wise binary representation, but I don't know if I'd agree that it is clear that that is what makes this conform to the new API. When I first read it, my expectation was actually that `inverted` probably implied `reversed`, which ok that's not the same as _inverted_, except that inverted itself is somewhat ambiguous depending on whether you are inverting the structure or the data within. My comment suggestion would maybe be:

`// Provide backwards compatibility by inverting elements, such that true represents metric enabled rather than disabled.`

travis79:
> 👍
> 
> This is a nit and doc only so I won't block, but to me `// Let's convert this to the new API for backwards compatibility` could use some elaboration. It's clear from reading the code that you want to flip element wise binary representation, but I don't know if I'd agree that it is clear that that is what makes this conform to the new API. When I first read it, my expectation was actually that `inverted` probably implied `reversed`, which ok that's not the same as _inverted_, except that inverted itself is somewhat ambiguous depending on whether you are inverting the structure or the data within. My comment suggestion would maybe be:
> 
> `// Provide backwards compatibility by inverting elements, such that true represents metric enabled rather than disabled.`

Agreed, I'll add some clarification to the comments that call out more clearly that I'm inverting the boolean values for each key contained in the map rather than adjusting the ordering.

PR #2451, opened 2023-04-19
Bug 1828528 - add TextMetric entry to private mod.rs

perrymcmanis144:
I added a test just to make sure the private version of TextMetric was being used. 

Do we have a canonical string for long entries? I'm not sure movie quotes are appropriate here

### DIFF
--- a/glean-core/rlb/src/test.rs
+++ b/glean-core/rlb/src/test.rs
@@ -10,7 +10,7 @@ use flate2::read::GzDecoder;
 use serde_json::Value as JsonValue;
 
 use crate::private::PingType;
-use crate::private::{BooleanMetric, CounterMetric, EventMetric, StringMetric};
+use crate::private::{BooleanMetric, CounterMetric, EventMetric, StringMetric, TextMetric};
 
 use super::*;
 use crate::common_test::{lock_test, new_glean, GLOBAL_APPLICATION_ID};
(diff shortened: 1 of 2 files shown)


glean-core/rlb/src/test.rs
```diff
@@ -1369,6 +1369,31 @@ fn test_boolean_get_num_errors() {
...
+        name: "text_metric".into(),
+        category: "test".into(),
+        send_in_pings: vec!["custom1".into()],
+        lifetime: Lifetime::Application,
+        disabled: false,
+        dynamic_label: Some(str::to_string("text")),
+    });
+
+    // 216 characters, which would overflow StringMetric
+    metric.set(str::to_string("I've seen things you people wouldn't believe. Attack ships on fire off the shoulder of Orion. I watched C-beams glitter in the dark near the Tannhäuser Gate. All those moments will be lost in time, like tears in rain"));
```
chutten, line 1388: Movie quotes are very acceptable, especially ones with non-ASCII characters in them.

Idiomatic Rust might be `"I've seen things...rain".into()` instead of `str::to_string("I've seen things...rain")`

chutten (approved):

glean-core/rlb/src/test.rs
perrymcmanis144: updated 👍

PR #2453, opened 2023-04-24
Bug 1829157 - Add accumulateSamples to docs

travis79:


docs/user/reference/metrics/timing_distribution.md
```diff
@@ -239,6 +239,101 @@ Glean.pages.pageLoad.stopAndAccumulate(timerId);
...
+```
+
+</div>
+<div data-lang="Rust" class="tab">
+
+```Rust
+use glean_metrics::pages;
+
+fn on_page_loaded() {
+    pages::page_load.accumulate_samples(samples);
```
chutten, line 312: Might not work. API isn't on [the trait](https://searchfox.org/glean/source/glean-core/src/traits/timing_distribution.rs) and I _think_ `glean_metrics::pages`' type is only advertised as adhering to the trait? (Certainly doesn't work on FOG, so it at least needs a disclaimer)

docs/user/reference/metrics/timing_distribution.md
```diff
@@ -239,6 +239,101 @@ Glean.pages.pageLoad.stopAndAccumulate(timerId);
...
+sample.
+
+Please note that this assumes that the provided samples are already in
+the "unit" declared by the instance of the metric type (e.g. if the
+instance this method was called on is using `TimeUnit::Second`, then
+`samples` are assumed to be in that unit).
+
+This discards any negative value in `samples` and report an `ErrorType::InvalidValue`
+for each of them. Reports an `ErrorType::InvalidOverflow` error for samples that
+are longer than maximum sample time for the given unit.
```
chutten, line 257: These should be in the `Limits` and `Recorded errors` section (like https://mozilla.github.io/glean/book/reference/metrics/custom_distribution.html#accumulatesamples )

chutten (changes requested):

docs/user/reference/metrics/timing_distribution.md
travis79: You are correct, I had added the RLB version without looking at the trait, before getting to the FOG versions and asking about it.  I'll remove this and add a link to the same bug as the FOG implementations

chutten (approved):

PR #2455, opened 2023-04-26
Apply clippy lint: Avoid potential infinite loop

badboy:
Detailed explanation in https://rust-lang.github.io/rust-clippy/master/index.html#lines_filter_map_ok We're unlikely to hit this, because `entry` is a file and thus continued read errors are unlikely.
But better be safe than sorry.


rosahbruno (approved):
lgtm

PR #2456, opened 2023-04-26
Bug 1827852 - Extend start/stop time of a ping to millisecond precision

badboy:
~~Requires https://github.com/mozilla/glean_parser/pull/592~~


travis79 (commented):
Don't forget to update the [docs](https://mozilla.github.io/glean/book/user/pings/index.html#start_time) to indicate the new precision.

glean-core/metrics.yaml
```diff
@@ -402,7 +404,7 @@ glean.client.annotation:
       - all-pings
     description: |
       An experimentation identifier derived and provided by the application
-      for the purpose of experimenation enrollment.
+      for the purpose of experimentation enrollment.
```
travis79, line 407: Oops, thanks! 😊

glean-core/metrics.yaml
```diff
@@ -498,8 +500,6 @@ glean.error:
       Only sent if the buffer ever overflows.
 
       In Version 0 this reported the total number of tasks enqueued.
-    unit:
```
travis79, line 501: o_O. Where did this come from?  Ahh, cool! Even though `unit` has not meaning here, it _still shows up in the Glean dictionary_. Maybe this would be okay to leave as I kinda like the explicit documentation this metadata provides (though the description should do that, too) 🤔 

No strong opinions on this, happy either way.

glean-core/rlb/src/lib.rs
```diff
@@ -173,7 +173,7 @@ pub fn set_experiment_inactive(experiment_id: String) {
 }
 
 /// TEST ONLY FUNCTION.
-/// Gets stored experimenation id.
+/// Gets stored experimentation id.
```
travis79, line 176: 😬 I should double check my 't' key functionality

travis79 (approved):
:shipit:

docs/user/user/collected-metrics/metrics.md
```diff
@@ -29,6 +29,7 @@ In addition to those built-in metrics, the following metrics are added to the pi
 
 | Name | Type | Description | Data reviews | Extras | Expiration | [Data Sensitivity](https://wiki.mozilla.org/Firefox/Data_Collection) |
 | --- | --- | --- | --- | --- | --- | --- |
+| glean.client.annotation.experimentation_id |[string](https://mozilla.github.io/glean/book/user/metrics/string.html) |An experimentation identifier derived and provided by the application for the purpose of experimentation enrollment.  |[Bug 1848201](https://bugzilla.mozilla.org/show_bug.cgi?id=1848201#c5)||never |1 |
```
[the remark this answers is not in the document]
badboy: yeah, we do. for others. We still have it and whether we should or not should go into a separate bug.

glean-core/metrics.yaml
badboy: I was equally surprised!
We just disallowed this in glean_parser. So we would need to retract that decision, which would require a bug, a revert and a new release.

glean-core/src/core/mod.rs
```diff
@@ -736,7 +736,7 @@ impl Glean {
 
     /// **Test-only API (exported for FFI purposes).**
     ///
-    /// Gets stored experimenation id annotation.
+    /// Gets stored experimentation id annotation.
```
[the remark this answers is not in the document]
badboy: to be fair: I reviewed this.

PR #2459, opened 2023-05-01
Bug 1812738 - allow user to set Glean log level

rosahbruno:


rosahbruno:
@travis79 
What are your thoughts on how I did this?

I set this up so that you can just call the function with your log preference. The other debug options all go through `debug.rs` and work a little differently since they rely on knowing if Glean is initialized or not. Since this is just a setting for the crate itself, it doesn't really care what is going on with Glean. I can update this to work similarly to the other debug options if we want this to feel the same.

travis79:
I think this should be fine. I actually envisioned this added as a parameter to the configuration passed into Glean.initialize, that way it's not adding the burden of calling an additional API to set the logging level

glean-core/src/glean.udl
```diff
@@ -35,6 +35,7 @@ namespace glean {
     boolean glean_set_debug_view_tag(string tag);
     boolean glean_set_source_tags(sequence<string> tags);
     void glean_set_log_pings(boolean value);
+    void glean_set_log_level(string log_level);
```
badboy, line 38: This should take an `enum LogLevel` instead of a stringy API.
See https://mozilla.github.io/uniffi-rs/udl/enumerations.html

badboy (changes requested):

badboy:
> I think this should be fine. I actually envisioned this added as a parameter to the configuration passed into Glean.initialize, that way it's not adding the burden of calling an additional API to set the logging level

Note that on Python we already pass in the log level, see https://mozilla.github.io/glean/book/user/debugging/python.html#logging-pings
We should stay consistent then and have it the same across languages.

travis79:
> > I think this should be fine. I actually envisioned this added as a parameter to the configuration passed into Glean.initialize, that way it's not adding the burden of calling an additional API to set the logging level
> 
> Note that on Python we already pass in the log level, see [mozilla.github.io/glean/book/user/debugging/python.html#logging-pings](https://mozilla.github.io/glean/book/user/debugging/python.html#logging-pings) We should stay consistent then and have it the same across languages (due to how things work the Python bindings configure the log level on the Python logger instance, not on the Rust side)

Ahh, I'd forgotten that we had this in Python 🤦. Yes, let's go with consistency!

docs/user/reference/general/initializing.md
```diff
@@ -67,6 +67,7 @@ Below are listed the configuration options available on most SDKs.
 - `httpUploader`: A custom HTTP uploader instance, that will overwrite Glean's provided uploader. Useful
 for users that wish to use specific uploader implementations. See [Custom Uploaders](#custom-uploaders)
 for more information on how and when the use this feature.
+- `logLevel`: The log level for the `log` rust crate. The level filter options in order from least to most verbose are: `Off`, `Error`, `Warn`, `Info`, `Debug`, `Trace`. The default option if nothing is provided is `Trace`. See the [`log` crate docs](https://docs.rs/log/latest/log/) for more information.
```
badboy, line 70: Interesting. `Trace` seems a bit ... much.

glean-core/src/glean.udl
```diff
@@ -69,6 +69,17 @@ dictionary InternalConfiguration {
     string app_build;
     boolean use_core_mps;
     boolean trim_data_to_registered_pings;
+    LevelFilter? log_level;
+};
+
+// An enum representing the different logging levels for the `log` crate.
+enum LevelFilter {
+    "Debug",
```
badboy, line 77: The ordering here seems a bit weird.
Probably best to keep it in the order of increasing log level, e.g. see how the Rust crate defines it: https://docs.rs/log/0.4.17/log/enum.Level.html

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -18,6 +19,8 @@ import mozilla.telemetry.glean.net.PingUploader
  *           sent along with all the pings, in the `client_info` section.
  * @property dataPath An optional [String] that specifies where to store data locally on the device.
  *           This should ONLY be used when setting up Glean on a non-main process.
+ * @property logLevel An optional [LevelFilter] that controls how verbose the logging is
+ *           from the rust `log` crate.
```
badboy, line 23: Mentioning "rust log crate" here feels not really informative.
Maybe:

```suggestion
 * @property logLevel An optional [LevelFilter] that controls how verbose the internal logging is.
```

glean-core/src/lib.rs
```diff
@@ -306,6 +309,9 @@ fn initialize_inner(
             let upload_enabled = cfg.upload_enabled;
             let trim_data_to_registered_pings = cfg.trim_data_to_registered_pings;
 
+            // Set the logging level for the `log` crate.
+            set_max_level(cfg.log_level.unwrap_or(LevelFilter::Trace));
```
badboy, line 313: We configure different levels for different platforms here: https://github.com/mozilla/glean/blob/306fb0b3d4783360768aaac852ba734945a57cd1/glean-core/src/lib.rs#L630
Default is `Debug`. On iOS release builds we actually limit to `Info`.
Maybe we should only re-set the max level if actually set
```suggestion
            cfg.log_level.map(|level| log::set_max_level(level));
```

(I think not importing the function name directly makes this more readable)

glean-core/src/lib.rs
```diff
@@ -25,6 +25,7 @@ use std::thread;
 use std::time::Duration;
 
 use crossbeam_channel::unbounded;
+use log::{set_max_level, LevelFilter};
```
badboy, line 28:
```suggestion
use log::{self, LevelFilter};
```

See below why.

badboy (changes requested):
Sorry for one more round of sending this back.

docs/user/reference/general/initializing.md
travis79, line 70: Trace does seem a bit _extra_ verbose. I don't actually think we log anything at that level anyway. I think `Debug` is plenty verbose for our default.

travis79 (approved):
I lean in @badboy's direction here on the default level of the logging, but no need to block on further review, I think.

rosahbruno:
@badboy 
I've addressed the comments you left. Now the logging level is only overwritten if provided by the user, otherwise it uses what we set elsewhere. I also updated the comments to call out "internal logging" rather than the rust crate.

badboy (approved):

PR #2461, opened 2023-05-03
Bug 1828066 - At shutdown block with a timeout and bail out if that fails.

badboy:


badboy:
## What happens in the bug?

Crash report: https://crash-stats.mozilla.org/report/index/47e42dbc-f99c-474f-8154-bbf470230413#allthreads

main thread hangs in `glean_core::dispatcher::global::block_on_queue()` during shutdown ([code](https://hg.mozilla.org/releases/mozilla-release/file/d2b2ff2c6e3ceb267e77b0163c860f7a1f13ff4a/third_party/rust/glean-core/src/lib.rs#l560)).

thread `glean.init` hangs in `dispatcher::flush_init()` ([code](https://hg.mozilla.org/releases/mozilla-release/file/d2b2ff2c6e3ceb267e77b0163c860f7a1f13ff4a/third_party/rust/glean-core/src/lib.rs#l452))
**Assumption:** some task scheduled a ping?

thread `glean.dispatcher` triggered `handle_client_active()`, thus submitting a ping synchronously.
This hangs in `std::sys::windows::fs::rename()`.

thread `glean.upload` is waiting for viaduct to make a request (why?)

## This fix

We only briefly wait for the dispatcher queue to empty. If it currently is running a task that takes too long, we just skip any further shutdown cleanup.

Guarantees?
We only wait up to 5s on the dispatcher.
If the dispatcher unblocks quicker we continue.
However we then shut down the dispatcher, which blocks again. In theory there's a window where new tasks get put on the dispatcher and could block for longer.
We also keep the blocking on the uploader shutdown (up to 30s), as well as persisting ping lifetime data (itself I/O, so could block).
So really we do catch this one instance where somehow a ping submission is in progress when shutting down, but taking long.

The 5s timeout is currently hardcoded. Should we make this compile-time/runtime configurable?

glean-core/src/dispatcher/mod.rs
```diff
@@ -161,6 +162,28 @@ impl DispatchGuard {
...
     }
 
+    /// Block on the task queue emptying, with a timeout.
+    fn block_on_queue_timeout(&self, timeout: Duration) -> Result<(), RecvTimeoutError> {
+        let (tx, rx) = crossbeam_channel::bounded(0);
+
+        // We explicitly don't use `self.launch` here.
+        // We always put this task on the unbounded queue.
+        // The pre-init queue might be full before its flushed, in which case this would panic.
+        // Blocking on the queue can only work if it is eventually flushed anyway.
```
chutten, line 172: musing aloud: This, I think, is the central idea. The only way to flush to this task would be to get through the entire preinit queue. (Theoretically, if the preinit wasn't full the "flush until this point" behaviour isn't quite preserved, as tasks enqueued after block_on_queue_timeout would need to be processed before it completes... but eh.)

glean-core/src/dispatcher/mod.rs
```diff
@@ -161,6 +162,28 @@ impl DispatchGuard {
...
+
+        let task = Command::Task(Box::new(move || {
+            // In case the calling thread times out waiting for this
+            // the channel will be dropped.
+            // But in case the work continues we don't want to panic.
```
chutten, line 177: Explain the panic for me?

glean-core/src/lib.rs
```diff
@@ -557,8 +557,18 @@ pub fn shutdown() {
         glean.set_dirty_flag(false);
     });
 
-    // We need to wait for above task to finish.
-    dispatcher::block_on_queue();
+    // We need to wait for above task to finish,
+    // but we also don't wait around forever.
+    //
+    // TODO: Make the timeout configurable?
```
chutten, line 563: In uploader_shutdown we have the TODO sorta-kinda satisfiable by looking at the data and determining what the "normal" length of time for uploader shutdown is. Would it be unreasonable to ask for a similar piece of instrumentation and a follow-up bug to look at the data?

badboy: The wonderful world of instrumentation when we kinda know we're running into trouble.
As long as we do _not_ timeout we should be ok recording this, but when running into a timeout due to I/O in the dispatcher doing more I/O might just as well fail. Nonetheless, yes, we should instrument.

glean-core/metrics.yaml
```diff
@@ -744,6 +744,22 @@ glean.validation:
       - jrediger@mozilla.com
     expires: never
 
+  shutdown_dispatcher_wait:
+    type: timing_distribution
+    time_unit: millisecond
+    description: |
+      Time waited for the dispatcher to unblock during shutdown.
```
chutten, line 751: Since we've recently learned the benefit of knowing timeouts within wait timing distributions, perhaps include a note that we expect most samples to be less than the default timeout of 10s

chutten (approved):

PR #2462, opened 2023-05-05
Synchronize Gradle dependencies with Android Components

rvandermeulen:


build.gradle
```diff
@@ -11,24 +11,24 @@ buildscript {
...
-        android_gradle_plugin: '7.4.1',
+        android_gradle_plugin: '7.4.2',
         coroutines: '1.6.4',
         jna: '5.13.0',
         junit: '4.12',
         mockito: '3.11.2',
         mockwebserver: '4.9.1', // This is different than a-c, but we're fine, it's only tests.
-        kotlin: '1.8.20',
-        robolectric: '4.9.2',
+        kotlin: '1.8.21',
```
Dexterp37, line 20: @rvandermeulen we recently were reminded that changing these values does have impact on consumers outside of Android Component consumers (e.g. VPN et al). Nothing to do here, but I thought sharing this concern with you as well :)

rvandermeulen: Interesting, that's definitely a new wrinkle!

rvandermeulen: To summarize the Slack discussion we had about this, I'll add a changelog entry for this (and future Kotlin bumps).

chutten (approved):
With the changelog, this LGTM. It's a patch (1.8.20 - 1.8.21) update, so I don't feel we need to clear it with our non-AC consumers before giving the okay.

PR #2464, opened 2023-05-09
Bug 1827452 - Skip running CI on the release branch

travis79:
Since we already run CI when merging this branch back to main, we can skip running it on the release branch

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -15,6 +15,9 @@ definitions:
         ignore: /.*/
       tags:
         only: /^v.*/
+  - ci_filters: &ci-filters
+      branches:
+        ignore: release
 
 ##########################################################################
 # COMMANDS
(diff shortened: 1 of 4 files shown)


badboy:
Any reason to choose an extra job like you did over an explicit filter like we already do for the release task?
See e.g. https://github.com/mozilla/glean/blob/main/.circleci/config.yml#L12-L17 and https://github.com/mozilla/glean/blob/main/.circleci/config.yml#L1163

travis79:
> Any reason to choose an extra job like you did over an explicit filter like we already do for the release task? See e.g. [`main`/.circleci/config.yml#L12-L17](https://github.com/mozilla/glean/blob/main/.circleci/config.yml?rgh-link-date=2023-05-10T08%3A06%3A35Z#L12-L17) and [`main`/.circleci/config.yml#L1163](https://github.com/mozilla/glean/blob/main/.circleci/config.yml?rgh-link-date=2023-05-10T08%3A06%3A35Z#L1163)

Not any specific reason

travis79:
Here's an attempt at adding the filter as a definition @badboy

badboy (approved):
Thanks, I think I like this more than the previous solution.

I'd still wish there was a better way of simply excluding a full workflow (or split it into files like GitHub Actions allows). Nothing you can do about that though.

PR #2469, opened 2023-05-10
Proposal: Make Glean Swift logs public by default

brizental:
Ok so, when I try to see Glean logs from an iOS device using `log` I just get `<private>` for all Glean messages. According to [this Stack Overflow](https://stackoverflow.com/questions/45908875/apple-iphone-debugging-with-console-private) that is because all dynamic strings are by default private. I made this change locally and I can now see the messages, would it be possible to get this merged? Thanks!


perrymcmanis144 (approved):
Passes tests, code matches stated desired behavior, 👍

PR #2472, opened 2023-05-16
Bug 1833381 - Add configuration merging for remote metric configs

travis79:
This adds the ability to merge remote metric configurations by "feature-ids", which are used by Nimbus to identify a feature. This will allow for the functionality to be shared across multiple features and avoid Nimbus feature collisions/exclusion until better multi-feature rollout and experiment capabilities are available.

### DIFF
--- a/glean-core/src/core/mod.rs
+++ b/glean-core/src/core/mod.rs
@@ -12,7 +12,8 @@ use crate::event_database::EventDatabase;
 use crate::internal_metrics::{AdditionalMetrics, CoreMetrics, DatabaseMetrics};
 use crate::internal_pings::InternalPings;
 use crate::metrics::{
-    self, ExperimentMetric, Metric, MetricType, MetricsEnabledConfig, PingType, RecordedExperiment,
+    self, ExperimentMetric, FeatureMetricConfiguration, Metric, MetricType, MetricsEnabledConfig,
+    PingType, RecordedExperiment,
 };
 use crate::ping::PingMaker;
 use crate::storage::{StorageManager, INTERNAL_STORAGE};
@@ -153,8 +154,8 @@ pub struct Glean {
     debug: DebugOptions,
     pub(crate) app_build: String,
     pub(crate) schedule_metrics_pings: bool,
-    pub(crate) remote_settings_epoch: AtomicU8,
-    pub(crate) remote_settings_metrics_config: Arc<Mutex<MetricsEnabledConfig>>,
+    pub(crate) remote_settings_metrics_config:
+        Arc<Mutex<HashMap<String, FeatureMetricConfiguration>>>,
 }
 
 impl Glean {
@@ -207,8 +208,7 @@ impl Glean {
             app_build: cfg.app_build.to_string(),
             // Subprocess doesn't use "metrics" pings so has no need for a scheduler.
             schedule_metrics_pings: false,
-            remote_settings_epoch: AtomicU8::new(0),
-            remote_settings_metrics_config: Arc::new(Mutex::new(MetricsEnabledConfig::new())),
+            remote_settings_metrics_config: Arc::new(Mutex::new(HashMap::new())),
         };
 
         // Ensuring these pings are registered.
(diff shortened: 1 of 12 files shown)


glean-core/src/core/mod.rs
```diff
@@ -711,14 +711,52 @@ impl Glean {
...
+    /// the associated feature.
+    ///
+    /// # Arguments
+    ///
+    /// * `metric_id` - The metric identifier in the format <category>.<name>
+    pub(crate) fn get_feature_id_for_metric(&self, metric_id: String) -> Option<String> {
+        let remote_settings_metrics_config = self.remote_settings_metrics_config.lock().unwrap();
+        if let Some((feature_id, _)) = remote_settings_metrics_config.iter().find(|&(_, config)| {
+            if let Ok(config) = config.config.lock() {
+                config.metrics_enabled.keys().any(|id| *id == metric_id)
```
badboy, line 751: Why `any`? That iterates through everything and thus is kinda expensive.
But `metrics_enabled` is already a hash map, so you can use [`contains_key`](https://doc.rust-lang.org/std/collections/struct.HashMap.html#method.contains_key))

glean-core/src/core/mod.rs
```diff
@@ -711,14 +711,52 @@ impl Glean {
...
+
+    /// Performs a lookup by metric_id (<category>.<name>) to get the feature identifier
+    /// associated with it if the metric has a remote configuration applied. Used by
+    /// created metrics to check if there is an updated remote configuration to apply for
+    /// the associated feature.
+    ///
+    /// # Arguments
+    ///
+    /// * `metric_id` - The metric identifier in the format <category>.<name>
+    pub(crate) fn get_feature_id_for_metric(&self, metric_id: String) -> Option<String> {
```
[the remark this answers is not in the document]
travis79: Yeah, nothing prevents that aside from careful experiment planning at the moment. This is a known limitation that will go away once Nimbus completes the multi-feature work that is currently in progress. The current risks are minimal because right now there is only one team (Search & Suggest) that's using this so it's been possible to coordinate the upcoming experiments.

glean-core/src/core/mod.rs
travis79: Because I guess I overlooked `contains_key`. Thanks!

travis79:
@badboy I think this is ready for another look

glean-core/src/metrics/mod.rs
```diff
@@ -187,35 +189,50 @@ pub trait MetricType {
...
+                    .lock()
+                    .unwrap()
+                    .get(&feature_id)
+                {
+                    remote_settings_epoch = config.epoch.load(Ordering::Acquire);
+
+                    if epoch == remote_settings_epoch {
+                        return disabled == 0;
+                    }
+                    // The epoch's didn't match so we need to look up the disabled flag
```
badboy, line 217:
```suggestion
                    // The epochs didn't match so we need to look up the disabled flag
```

glean-core/src/metrics/mod.rs
```diff
@@ -187,35 +189,50 @@ pub trait MetricType {
...
+                if let Some(config) = glean
+                    .remote_settings_metrics_config
+                    .lock()
+                    .unwrap()
+                    .get(&feature_id)
+                {
+                    remote_settings_epoch = config.epoch.load(Ordering::Acquire);
+
+                    if epoch == remote_settings_epoch {
+                        return disabled == 0;
```
badboy, line 215: Reading this code I'm so annoyed we called that field `disabled` and not `enabled` :D

badboy (commented):
The code looks fine, but after reviewing I can't get the one thought out of my head:

Why does Glean even need to know about the different feature IDs?
We don't really use the feature ID internally and we bascially always search through the nested hash map anyway to find if a metric is mentioned.
In the Nimbus parts it can still be split across multiple features

If Glean only gets a list of metrics we could merge that into the current config (instead of overwriting as the old code does).

I guess that leaves unenrollment?
What happens on unenrollment? Does that only happen on startup?
If so that means we would simply not get the config anyway, and thus nothing to merge

Or do we get an unerollment update during normal runtime?
Is `set_metrics_enabled_config` called with an empty config then? 

Sorry for bringing this only up now, but better to clarify that before we merge.

glean-core/src/metrics/mod.rs
travis79: +1

travis79:
@badboy I do agree with you and I don't think Glean should care about the feature ids but in this case this is a workaround in order to be able to share this functionality for a client enrolled in two experiments that are using it at the same time. This is only temporary, until Nimbus multi-feature (with features that don't "collide") is ready. I'm told this will be early H2.

After having a good think about it, and knowing that this doesn't really solve the problem of two features trying to set the same metric, maybe the extra layer of feature-ids is extra complication and isn't really doing much. You are right, most of the time the clients will enroll in an experiment and apply the configs at startup. This is especially true on mobile, and the only time it might not be true is on Desktop, where a long-running client checks for experiments every 6 minutes or so, and so this _could_ be applied sometime after startup, but with the only component using this right now being Search & Suggest, and the fact that they are more than likely to be attempting to turn the same metrics on for any given experiment, maybe a simpler approach is better here.

Rather than rework this one to be simpler, I'll probably start over and do a simple approach that just merges configs without the feature-id layer over the top.

travis79:
Abandoning this more complicated approach for something simpler in #2480

PR #2473, opened 2023-05-16
Bug 1647630 - expose throttling config

chutten:


badboy:
Code's ok but handing over to @travis79 as this changes the API and so should have SDK lead sign-off.

glean-core/rlb/src/test.rs
```diff
@@ -43,18 +43,10 @@ fn send_a_ping() {
...
-        upload_enabled: true,
-        max_events: None,
-        delay_ping_lifetime_io: false,
-        server_endpoint: Some("invalid-test-host".into()),
-        uploader: Some(Box::new(FakeUploader { sender: s })),
-        use_core_mps: false,
-        trim_data_to_registered_pings: false,
-        log_level: None,
-    };
+    let cfg = ConfigurationBuilder::new(true, tmpname, GLOBAL_APPLICATION_ID)
```
travis79, line 46: 🤩

travis79 (commented):
A couple of quick questions:
1. I don't see any tests that specifically test setting a rate limit through this, is that because it is difficult to test?
2. Can this be used to disable rate-limiting altogether (i.e. can we set a `PingRateLimit` with a `seconds_per_interval = 0`?

travis79:
Other than the questions above, I don't have any specific concerns with the implementation.

travis79 (approved):

chutten:
> A couple of quick questions:
> 
>     1. I don't see any tests that specifically test setting a rate limit through this, is that because it is difficult to test?

Fair enough. The behaviour is tested in the unit, but the new capability of setting it globally wasn't. It is now.
 
>     2. Can this be used to disable rate-limiting altogether (i.e. can we set a `PingRateLimit` with a `seconds_per_interval = 0`?

`seconds_per_interval = 0` will not disable rate-limiting if `pings_per_interval` is also `0`, and won't disable it entirely even for non-zero `pings_per_interval`. It will still proclaim a state of `Throttled` every time you hit `pings_per_interval` even if it is `Throttled(0)` (which becomes a `PingUploadTask::Wait(0)`). Depending on the uploader, this might have some minimum throttling behaviour (e.g. RLB will sleep for 0ms, which has [platform-specific behaviour](https://doc.rust-lang.org/std/thread/fn.sleep.html#platform-specific-behavior)).

If we wanted a way to configure for disabled throttling behaviour, I'd recommend a breaking API change on the `PingRateLimit` type to become an `enum PingRateLimit { Absolute { seconds_per_interval: u64, pings_per_interval: u32 }, Unlimited };` or something. Then it'd drill down to the upload manager which could avoid the `RateLimiterState::Throttled` entirely. But I'd like to see the use case first.

PR #2475, opened 2023-05-18
Bug 1826973 - Add Data Control Plane user documentation

travis79:
This adds documentation about the Glean Data Control Plane functionality (a.k.a. Server Knobs).

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 257 utf-8
+personal_ws-1.1 en 261 utf-8
 AAR
 AARs
 ABI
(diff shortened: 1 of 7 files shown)


badboy:
This wasn't added to `docs/user/SUMMARY.md` and so is not available in the book yet.

travis79:
> This wasn't added to `docs/user/SUMMARY.md` and so is not available in the book yet.

Stealth docs!  I'll correct this ASAP, been a while since I added anything to the docs so I just forgot all the steps. Maybe it's time to think about adding docs on how to add docs to the docs

badboy:
image

PR #2477, opened 2023-05-22
Bug 1832324 - Add a way to await ping submission in iOS 

travis79:
This simply awaits the internal dispatcher which is where the pings are submitted from.

### DIFF
--- a/glean-core/ios/GleanTests/GleanTests.swift
+++ b/glean-core/ios/GleanTests/GleanTests.swift
@@ -9,6 +9,7 @@ import XCTest
 
 private typealias GleanInternalMetrics = GleanMetrics.GleanInternalMetrics
 
+// swiftlint:disable type_body_length
 class GleanTests: XCTestCase {
     var expectation: XCTestExpectation?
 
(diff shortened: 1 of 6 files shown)


badboy:
While this might get the job done for the particular feature request, I wonder if this also exposes too much internals so to speak?
We already have `Glean.shutdown()` on the RLB, which does a bunch more, but IMO is the more appropriate API to expose.
Also the Swift dispatcher is only used for the uploader anyway, so this really just exposes waiting for upload anyway.
The shutdown API is more robust, including timeouts, etc.

what do you think?

travis79:
> While this might get the job done for the particular feature request, I wonder if this also exposes too much internals so to speak? We already have `Glean.shutdown()` on the RLB, which does a bunch more, but IMO is the more appropriate API to expose. Also the Swift dispatcher is only used for the uploader anyway, so this really just exposes waiting for upload anyway. The shutdown API is more robust, including timeouts, etc.
> 
> what do you think?

I considered that and have no issues with exposing it through the Swift API, and I agree, it is a more robust solution. The direction I went in this PR currently was only an attempt at the simplest possible solution. 

I'll rework this exposing `shutdown()` and see how it looks.

glean-core/ios/GleanTests/GleanTests.swift
```diff
@@ -367,4 +368,63 @@ class GleanTests: XCTestCase {
         Glean.shared.initialize(uploadEnabled: true, buildInfo: stubBuildInfo())
         XCTAssertFalse(Glean.shared.isCustomDataPath)
     }
+
+    func testShutdown() {
+        // This test relies on Glean not being initialized
+        Glean.shared.testDestroyGleanHandle()
+
+        // We expect only a single ping later
```
badboy, line 376: Weeeell, we ignore a baseline ping, so it's more like 2 pings.

badboy (changes requested):
Changelog entry please.

badboy (approved):

PR #2481, opened 2023-05-25
Bug 1833870 - Increase StringList metric type limits

travis79:
This increases the length of strings allowed in the list from 50 to 100 utf8 bytes to match the `StringMetric` type, as well as increasing the total number of strings allowed in the list from 20 to 100.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 262 utf-8
+personal_ws-1.1 en 263 utf-8
 AAR
 AARs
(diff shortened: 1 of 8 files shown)


travis79:
> 

Arrgh! I knew I missed someplace :)

badboy (approved):

perrymcmanis144 (approved):
Does what it says on the box, so LGTM 👍 

one possible addition to CHANGELOG but not necessary so approving

PR #2484, opened 2023-05-27
Update tooltool.py to the latest upstream version

rvandermeulen:
The main motivation to do this is to point to the now-canonical repo for tooltool.py. As an added bonus, it now has proper python3 support, so we can remove python2 from the Docker image we're using. That would in turn have ended up blocking our ability to update to a newer Ubuntu base as it's no longer distributed with 22.04.


Dexterp37 (approved):
Stealing the review :-) Thanks for this one Ryan, looks like a reasonable change and CI is green. Merging.

PR #2487, opened 2023-06-02
Replace once_cell usage with libstd's OnceLock

badboy:
Rust 1.70.0 stabilized OnceCell and OnceLock[1].
No more need for `once_cell`!

Note that:

* Only since 1.70, which is far ahead of our current MSRV.
* `LazyCell` and `LazyLock` are not yet stabilized, so we can't abandon the `once_cell` crate
* `glean_parser` generates references to `once_cell`'s `OnceCell`, so we would also need to update our generators when we land this.

[1]: https://blog.rust-lang.org/2023/06/01/Rust-1.70.0.html

---

Opening this mostly for @travis79 to see when he's back. Once seen we can close this, it's not actionable yet until we raise our MSRV anyway.

### DIFF
--- a/glean-core/src/core/mod.rs
+++ b/glean-core/src/core/mod.rs
@@ -1,10 +1,10 @@
 use std::collections::HashMap;
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU8, Ordering};
+use std::sync::OnceLock;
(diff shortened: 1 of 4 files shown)


travis79:
Yes! Nice!  I think I saw this news in a recent "This Week in Rust".  Reducing our dependency count is always good in my book.

PR #2493, opened 2023-06-07
Dependency updates & vetting

badboy:


perrymcmanis144 (approved):
See no reason to object, 👍

PR #2497, opened 2023-06-12
bug 1837996 - Document new rate_limit configuration property [doc only]

chutten:


badboy:
> Weirdly, though I titled `[doc only]` it appears to be running tests on CI

That would have to be in the last commit message, we don't use the PR title at the moment.

badboy (approved):

PR #2500, opened 2023-06-13
Bug 1838174 - Glean plugin: Try newer accessor name for exec result

badboy:
This makes it possible to run with both Gradle <8 & 8+

---

I haven't verified that this actually works with Gradle 8.


rosahbruno (approved):
lgtm

badboy:
This doesn't actually work, but until we convert to Gradle 8 I don't have any way to test it.

PR #2502, opened 2023-06-13
Synchronize Gradle dependencies with AC

rvandermeulen:


badboy:
rust beta failure unrelated.

badboy (approved):

PR #2503, opened 2023-06-15
Bug 1838453 - Shutdown timeout: Gracefully handle the waiting thread going away

badboy:
Bringing this back. Not sure how to best test this.

There's in theory now an edge case where other things could crash because the dispatcher could continue to run, but Glean already being gone.


chutten (approved):
r+wc (the c stands for Changelog)

I'm guessing the manual test doesn't translate to an automated one?

badboy:
> r+wc (the c stands for Changelog)
> 
> I'm guessing the manual test doesn't translate to an automated one?

Unfortuantely not. We hard-code the timeout to 10s and the only way to actually observe the behavior right now is through the logs.

PR #2505, opened 2023-06-16
Python: Avoid variable shadowing to avoid type confusion

badboy:
mypy found this for me.

```
glean-core/python/glean/metrics/datetime.py:107: error: Incompatible types in assignment (expression has type "datetime", variable has type "Optional[Datetime]")  [assignment]
glean-core/python/glean/metrics/datetime.py:117: error: Incompatible return value type (got "Datetime", expected "Optional[datetime]")  [return-value]
```


chutten (approved):
`value` is a bit bland, but its use is entirely within a single terminal's height, so it's probably fine.

PR #2507, opened 2023-06-16
Update ktlint to version 0.49.1

rvandermeulen:
The formatting patch was completely auto-generated by ktlintFormat. I did a sanity check and didn't see anything obviously wonky with it, though.

It's.....a lot.....however. If we end up taking this as-is, maybe we should consider adding the commit to `.git-blame-ignore-revs` to reduce the blame pollution?


badboy (approved):
Haven't checked all changes, but given that they were done automatically that's fine..

> It's.....a lot.....however. If we end up taking this as-is, maybe we should consider adding the commit to `.git-blame-ignore-revs` to reduce the blame pollution?

Sounds good to me, can you add that? (I can merge with a merge-commit, so the commit IDs are stable). This now needs a rebase anyway, after I merged the detekt update.

PR #2513, opened 2023-06-26
Bug 1840355 - Add configuration to add a `glean_timestamp` extra to all events

badboy:


CHANGELOG.md
```diff
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v53.2.0...main)
 
+* General
+  * Experimental: Add configuration to add a `glean_timestamp` extra to all events ([#2513](https://github.com/mozilla/glean/issues/2513))
```
travis79, line 6: Experimental, as in you are planning to test this with Nimbus? If so, that's nifty!

glean-core/android/src/main/java/mozilla/telemetry/glean/config/Configuration.kt
```diff
@@ -31,6 +31,7 @@ data class Configuration @JvmOverloads constructor(
     val httpClient: PingUploader = HttpURLConnectionUploader(),
     val dataPath: String? = null,
     val logLevel: LevelFilter? = null,
+    val enableEventTimestamps: Boolean = false,
```
travis79, line 34: Please add the property to the comment above, for completeness

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -30,12 +31,14 @@ public struct Configuration {
         channel: String? = nil,
         serverEndpoint: String? = nil,
         dataPath: String? = nil,
-        logLevel: LevelFilter? = nil
+        logLevel: LevelFilter? = nil,
+        enableEventTimestamps: Bool = false
```
travis79, line 35: Add this to the comment above please

travis79 (approved):
r+wc

CHANGELOG.md
badboy: Yup, that's the plan! (but also experimental as in: don't use it, we give no guarantees)

PR #2527, opened 2023-07-11
Bug1626969: Expose `locale` through the RLB so that it can be set by consumers

travis79:
This exposes the `locale` from the internal `ClientInfoMetrics` in order to allow it to be set by consumers of the RLB.


docs/user/reference/metrics/test.c
```diff
@@ -0,0 +1,194 @@
+
```
badboy, line 1: ❓

glean-core/rlb/src/core_metrics.rs
```diff
@@ -13,6 +13,9 @@ pub struct ClientInfoMetrics {
     pub app_display_version: String,
     /// The product-provided release channel (e.g. "beta").
     pub channel: Option<String>,
+    /// The locale of the application during initialization (e.g. "es-ES").
+    /// If the locale can't be determined on the system, the value is "und", to indicate "undetermined".
```
badboy, line 17: That doesn't apply here though, does it? If this field is unset we don't set it in the client info.
The foreign language bindings set it to `und`.

badboy (changes requested):

docs/user/reference/metrics/test.c
travis79: Must have had some random new file I opened in VSCode and accidentally committed it.

travis79: Oh, haha I know what it is...  it's a bitmap of a bird logo

travis79:
Ugh, starting from scratch

PR #2528, opened 2023-07-11
Bug 1841075 - re-export PingRateLimit for RLB's

perrymcmanis144:


CHANGELOG.md
```diff
@@ -2,12 +2,16 @@
...
+* Rust
+  * The Ping Rate Limit type is now accessible in the Rust Language Binding ([#2528](https://github.com/mozilla/glean/pull/2528))
+  
 # v53.1.0 (2023-06-28)
 
 [Full changelog](https://github.com/mozilla/glean/compare/v53.0.0...v53.1.0)
 
 * General
-  * Gracefully handle the waiting thread going away during shutdown ([#2503](https://github.com/mozilla/glean/pull/2503))
+  * Gracefully handle the waiting thread going away during shutdown ([#2503](/2503))
```
badboy, line 14: Mistakenly modified this link?

perrymcmanis144: Hmm appears so good catch. Not sure why that happened =/

chutten (approved):
Could land as-is, but could you also jump onto `src/configuration.rs` and `test.rs` replace their `glean_core::PingRateLimit` with `crate::PingRateLimit`?

The test one is to make sure we're covered. The `configuration.rs` one ensures if the RLB's `PingRateLimit` and glean-core's `PingRateLimit` ever diverge, we're guaranteeing we're exporting the one we need.

chutten (commented):
(See also `pub struct Builder` about forty lines below in configuration.rs)

PR #2530, opened 2023-07-12
Address beta warnings

badboy:


travis79 (approved):
No more `vec!` macros, huh?  Neat!

PR #2531, opened 2023-07-12
Bug1626969: Expose `locale` through the RLB so that it can be set by consumers

travis79:


chutten (approved):
Looks good. We might eventually want a Gleannotation on https://dictionary.telemetry.mozilla.org/apps/firefox_desktop/metrics/locale to point at the discussion on the bug about what's going on (and to let folks know when it became available for Firefox Desktop).

travis79:
That's a great idea, I'll update the Dictionary with a Gleannotation once this is available in Desktop

PR #2533, opened 2023-07-14
Suppress Swift warnings

mcleinman:
Frequently, we don't want to use the return of a UUID metric's `generateAndSet` function, or the result from `setDebugViewTag`.

However, this leaves warnings in Xcode and at build time:
Screenshot 41

[Swift has](https://docs.swift.org/swift-book/documentation/the-swift-programming-language/attributes/#discardableResult) the `discardableResult` attribute to suppress these warnings, commonly used for functions whose return will often be unused - like these two.

### DIFF
--- a/glean-core/ios/Glean/Glean.swift
+++ b/glean-core/ios/Glean/Glean.swift
@@ -329,7 +329,7 @@ public class Glean {
     ///
     /// - parameters:
     ///     * value: The value of the tag, which must be a valid HTTP header value.
-    public func setDebugViewTag(_ tag: String) -> Bool {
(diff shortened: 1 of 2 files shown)


glean-core/ios/Glean/Glean.swift
```diff
@@ -329,7 +329,7 @@ public class Glean {
     ///
     /// - parameters:
     ///     * value: The value of the tag, which must be a valid HTTP header value.
-    public func setDebugViewTag(_ tag: String) -> Bool {
+    @discardableResult public func setDebugViewTag(_ tag: String) -> Bool {
```
badboy, line 332: heh, we don't even document what the return value means.

travis79 (approved):

PR #2539, opened 2023-07-19
Use Python's modern `build`

badboy:
I seem to continuously run into trouble with the old way ([and here's why](https://blog.ganssle.io/articles/2021/10/setup-py-deprecated.html)), though it does not seem to be an issue on CI yet? Anyway, this then usually fixes it, though it's slightly more heavy-handed because it rebuilds the Rust code base over and over on `make build-python`.

But maybe a stop-gap until we land #2345


badboy:
well, apparently CI dislikes it.

travis79:
I'll hold off on reviewing then, until CI is happier

badboy:
Ok, I give up. Landing maturin is the better way forward and locally I rely on this fix to run it.

PR #2540, opened 2023-07-20
Python: Add a minimal Glean sample app to quickly run glean-python

badboy:
No CI tasks yet, but I used that locally to run the Glean.shutdown tests before.

---

`GLEAN_DEBUG_VIEW_TAG=jer-py make run` -> <https://debug-ping-preview.firebaseapp.com/pings/jer-py>


samples/python/glean-sample/__main__.py
```diff
@@ -0,0 +1,27 @@
...
+    upload_enabled=True,
+    data_dir="./data",
+    log_level=logging.DEBUG,
+    configuration=config,
+)
+
+metrics = load_metrics("metrics.yaml")
+pings = load_pings("pings.yaml")
+
+metrics.test.metrics.sample_boolean.set(True)
```
chutten, line 24: Huh. You don't have to conjugate the category to snake_case? That seems like magic.

chutten (approved):

samples/python/glean-sample/__main__.py
badboy: It is! I was equally surprised and added 5 debug statements to glean-py before I realized that it was `test.metrics` and not `test_metrics`

PR #2545, opened 2023-07-24
RLB: Gracefully handle uploader thread start failure

badboy:
There might be cases where we can't properly start the thread, e.g. when out of memory (as it seems to happen on low-end 32bit versions of Firefox). We now avoid the crash.
The data should be safely stored on disk and on the next upload it will try again (and hopefully succeed).

---

Still needs some testing. Not sure yet how to inject a thread start failure.


badboy:
I'm still undecided whether I want to land this example test, but at least the RLB code itself is working.

travis79 (approved):
"THIS IS HIGHLY UNSAFE", always a good comment 💯. Glad it's just an example for testing purposes. This looks good to me.

PR #2547, opened 2023-07-25
Bug 1838002 - Document testGetValue takes a (sometimes-optional) parameter

perrymcmanis144:


travis79 (approved):

badboy:
I took the liberty to rebase & force-push this to squash that into a single commit.

---

Content-wise this reads a bit more complicated than I'd like and is also _after_ the return values. Usually I would go for the order `function description, input arguments, return value` as we do it in code docs as well.

If we'd really want to overhaul this we should change the headers to include the arguments,
so 

```
### `testGetValue`
```

becomes

```
### `testGetValue(pingName)`
```

but then we should change that for all functions and that's a bigger change.

I'm ok merging as is, as my proposals are a bigger change; let's file a followup to clean this up though.

travis79:
I took the liberty of filing the follow-up bug for the enhancements suggested by @badboy. I'm going to go ahead and merge this so that it doesn't get stale.

PR #2551, opened 2023-07-31
Bug 1844533 - Fixes Glean ping uploading from background services in Kotlin

travis79:
When Glean attempted to schedule an upload task on the WorkManager from a background process
it would get the main-process Glean when the task was executed. Since the main-process Glean
didn't have the data directory from the background process, it would not find the ping to be
able to upload it. This changes the uploader to run on the internal Glean Dispatcher when it
is being run from a background service or process and continues to rely on WorkManager for
the main process execution.

### DIFF
--- a/samples/android/app/build.gradle
+++ b/samples/android/app/build.gradle
@@ -41,6 +41,8 @@ dependencies {
 
     implementation "androidx.appcompat:appcompat:$rootProject.versions.androidx_appcompat"
     implementation "androidx.browser:browser:$rootProject.versions.androidx_browser"
+    implementation "androidx.work:work-runtime-ktx:$rootProject.versions.androidx_browser"
+
 
     androidTestImplementation "androidx.test:core-ktx:$rootProject.versions.androidx_test"
     androidTestImplementation "androidx.test:runner:$rootProject.versions.androidx_test"
--- a/samples/android/app/build.gradle
+++ b/samples/android/app/build.gradle
@@ -41,7 +41,6 @@ dependencies {
 
     implementation "androidx.appcompat:appcompat:$rootProject.versions.androidx_appcompat"
     implementation "androidx.browser:browser:$rootProject.versions.androidx_browser"
-    implementation "androidx.work:work-runtime-ktx:$rootProject.versions.androidx_browser"
 
 
     androidTestImplementation "androidx.test:core-ktx:$rootProject.versions.androidx_test"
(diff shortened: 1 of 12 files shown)


samples/android/app/build.gradle
```diff
@@ -42,6 +42,7 @@ dependencies {
     implementation "androidx.appcompat:appcompat:$rootProject.versions.androidx_appcompat"
     implementation "androidx.browser:browser:$rootProject.versions.androidx_browser"
 
+
```
badboy, line 45: huh? mistaken newline added?

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/BackgroundPingTest.kt
```diff
@@ -0,0 +1,63 @@
...
+import org.junit.Rule
+import org.junit.Test
+import org.junit.runner.RunWith
+import org.mozilla.samples.gleancore.SampleBackgroundProcess
+
+// Note that this is ignored because for some reason the mock server
+// doesn't seem to be able to catch the background ping when it is
+// sent, despite the logs clearly stating that the ping is being sent
+// and from testing with the Glean Debug View also clearly showing
+// that the ping is being sent.
```
badboy, line 26: `GleanTestLocalServer` calls `Glean.testSetLocalEndpoint(localPort)`, which of course only happens in the main process then.
You probably would need to change the config manually for the subprocess: https://github.com/mozilla/glean/blob/main/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt#L588-L591

badboy (approved):
Code looks good. With respect to the test see the inline comment. Maybe we can pass around the info so the test app sets the right things?

samples/android/app/build.gradle
travis79: Probably

samples/android/app/src/androidTest/java/org/mozilla/samples/gleancore/pings/BackgroundPingTest.kt
travis79: Ha, worth a shot

travis79: Bug 1846518 filed to follow up on this, as it turned out to be non-trivial and we are due a Glean release for many things

PR #2586, opened 2023-08-04
Dependabot: Group all Python package updates

badboy:
This will reduce individual PRs in favor of a single PR updating multiple dependencies.

See https://docs.github.com/en/code-security/dependabot/dependabot-version-updates/configuration-options-for-the-dependabot.yml-file#groups

[doc only]


perrymcmanis144 (approved):
passes tests, desired behavior matches what the change does 👍

PR #2587, opened 2023-08-04
Possible fix for start up ping inconsistency

OrlaM:
This is a potential fix for the start up ping issue on iOS.

The current theory is that the app is launching while in the background and then getting shut down without the user ever bringing the app into the foreground (and therefore never triggering the background event). There are two places this could be happening that were added in v113, one is a background task is scheduled to run after 4 hours and the other is a local notification that fires after 2 days. Both use cases would relaunch the app if it was terminated while keeping it in the background and if the user doesn't use the app soon after then the app would be terminated in the background resulting in Glean firing an active ping without a corresponding inactive ping.

We are working to gather more evidence to confirm this theory but in the mean time I wanted to get feedback if a solution like this would be acceptable or to start the conversation about how this might be solved otherwise as the use cases on the iOS side that are causing this are valid ones that we can't remove.

I'm also unable to run locally so I have no idea if this even runs or breaks tests 🙈

### DIFF
--- a/glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
+++ b/glean-core/ios/Glean/Scheduler/GleanLifecycleObserver.swift
@@ -23,9 +23,11 @@ class GleanLifecycleObserver {
             object: nil
         )
 
-        // We handle init the same as an foreground event,
+        // We handle init the same as an foreground event if the app is in the foreground,
         // as we won't get the enter-foreground notification.
-        Glean.shared.handleForegroundEvent()
+        if UIApplication.shared.applicationState == .active {
+            Glean.shared.handleForegroundEvent()
+        }
     }
 
     @objc func appWillEnterForeground(notification _: NSNotification) {


OrlaM:
cc @travis79 @badboy

badboy:
If I understand right then `UIApplication.shared.applicationState == .active` is very unlikely to ever be true in a handler for `willEnterForegroundNotification`.
So this is not really a solution.

However there's also [`didBecomeActiveNotification`](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification) which maybe is the notification we should be handling instead?

travis79:
> If I understand right then `UIApplication.shared.applicationState == .active` is very unlikely to ever be true in a handler for `willEnterForegroundNotification`. So this is not really a solution.
> 
> However there's also [`didBecomeActiveNotification`](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification) which maybe is the notification we should be handling instead?

I don't think using `didBecomeActive` is a good idea because it would trigger any time the app gains focus, so any overlay window that pops up and steals focus would trigger this when it was dismissed and the focus returned to the app. I investigated this some time back but I can't seem to find the bug now. This would also imply that we would need to pair it with the appropriate matching lifecycle event `willResignActive` which means:
> The app moves to the inactive state because of temporary interruptions like an incoming phone call or SMS message

I think this would change the meaning of `duration` for iOS to be different from Android significantly.

badboy:
Ah we did try it in the past: https://github.com/mozilla/glean/pull/1477

OrlaM:
> If I understand right then `UIApplication.shared.applicationState == .active` is very unlikely to ever be true in a handler for `willEnterForegroundNotification`. So this is not really a solution.
> 
> However there's also [`didBecomeActiveNotification`](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification) which maybe is the notification we should be handling instead?

It's not checking the state in the event handler, it's checking the state in the init, which is fired on app launch regardless of if the app is in the foreground or background.

badboy:
> > If I understand right then `UIApplication.shared.applicationState == .active` is very unlikely to ever be true in a handler for `willEnterForegroundNotification`. So this is not really a solution.
> > However there's also [`didBecomeActiveNotification`](https://developer.apple.com/documentation/uikit/uiapplication/1622953-didbecomeactivenotification) which maybe is the notification we should be handling instead?
> 
> It's not checking the state in the event handler, it's checking the state in the init, which is fired on app launch regardless of if the app is in the foreground or background.

oh d'oh! that's on me for reading this first thing in the morning.

badboy:
Got an alternative here: https://github.com/mozilla/glean/pull/2589

PR #2589, opened 2023-08-08
iOS: Handle `didBecomeActiveNotification` to track first active state of an app

badboy:
We used to always call `handleForegroundEvent` after initialize, when setting up the lifecycle observer.
Howerver there are cases where the app starts up but is NOT made active.

The only way to really capture that is to listen to `didBecomeActiveNotification`. Now `didBecomeActiveNotification` is also triggered after the user leaves an overlay (popup, notification), but that's okay in our case: `handleForegroundEvent` keeps track of the `active` status, but the above cases don't trigger `didEnterBackgroundNotification`!

---

Alternative to #2587 probably

### DIFF
--- a/glean-core/ios/GleanTests/TestUtils.swift
+++ b/glean-core/ios/GleanTests/TestUtils.swift
@@ -50,46 +50,11 @@ func stubServerReceive(callback: @escaping (String, [String: Any]?) -> Void) {
 /// This also prevents outgoing network requests during unit tests while
 /// still allowing us to use the default telemetry endpoint.
 func resetGleanDiscardingInitialPings(testCase: XCTestCase, tag: String, clearStores: Bool = true) {
-    let expectation = testCase.expectation(description: "\(tag): Ping Received")
-
-    // We are using OHHTTPStubs combined with an XCTestExpectation in order to capture
-    // outgoing network requests and prevent actual requests being made from tests.
-    // We wait for 2 pings, in this order:
-    //
-    // 1. baseline ping with reason "dirty-startup"
-    // 2. baseline ping with reason "active"
-    //
-    // Once we received the "active" baseline ping we're good to go.
-    // All subsequent pings should be from the test itself.
-    stubServerReceive { pingType, json in
-        if pingType != "baseline" {
-            return
-        }
-        let pingInfo = json?["ping_info"] as? [String: Any]
-        let reason = pingInfo?["reason"] as? String
-        if reason != "active" {
-            return
-        }
-
-        // Fulfill test's expectation once we parsed the incoming data.
-        DispatchQueue.main.async {
-            // Let the response get processed before we mark the expectation fulfilled
-            expectation.fulfill()
-        }
-    }
-
-    // We may recieve more than one ping, using this function means we don't care about any of them
-    expectation.assertForOverFulfill = false
-
     // Force NO overdue metrics ping. We only expect baseline pings.
     let mps = MetricsPingScheduler(true)
     mps.updateSentDate(Date())
 
     Glean.shared.resetGlean(clearStores: clearStores)
-
-    testCase.waitForExpectations(timeout: 5.0) { error in
-        XCTAssertNil(error, "Test timed out waiting for upload: \(error!)")
-    }
 
     // In case this isn't the first test, there might be pending operations.
     // We wait for them to finish, so they don't clutter the next tests.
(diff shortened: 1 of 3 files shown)


OrlaM:
This won't capture the first app foreground event when the app is launched normally.
In a normal lifecycle where the app starts from a terminated state the user opens the app, the app will fire off all of it's lifecycle methods, then we initalise glean, glean listens for the didBecomeActive event, but since it has already fired at this point it won't be triggered for the current launch so we won't be able to track that foreground event.

badboy:
> This won't capture the first app foreground event when the app is launched normally. In a normal lifecycle where the app starts from a terminated state the user opens the app, the app will fire off all of it's lifecycle methods, then we initalise glean, glean listens for the didBecomeActive event, but since it has already fired at this point it won't be triggered for the current launch so we won't be able to track that foreground event.

It did in my manual tests in the iOS simulator. Of course I can imagine it's still not guaranteed.

But then what is the correct way to _know_ when an app is actually visible to the user after startup?

OrlaM:
You are right, this will fire on the first launch. I was thinking with init glean in the same lifecycle method but we don't we trigger it in `willFinishLaunchingWithOptions`. 🙈 

Are you sure this won't cause extra noise in the data and make the problem worse? I don't see any change in how `handleForegroundEvent` is implemented and the issue with the DAU tracking for the incident is that we have extra start up pings due to the app triggering `handleForegroundEvent` on init while the app is in the background and then again when it's launched due to listing to the `willEnterForeground` event. I could be missing something but it looks like `handleForegroundEvent` will be called a lot more often with this change.

badboy:
> Are you sure this won't cause extra noise in the data and make the problem worse? I don't see any change in how `handleForegroundEvent` is implemented and the issue with the DAU tracking for the incident is that we have extra start up pings due to the app triggering `handleForegroundEvent` on init while the app is in the background and then again when it's launched due to listing to the `willEnterForeground` event. I could be missing something but it looks like `handleForegroundEvent` will be called a lot more often with this change.

Mostly (I won't deploy this before being 100% sure). We internally track `active` status already (since https://github.com/mozilla/glean/pull/2368), so calling it twice in a row from both events shouldn't be a problem. And my local tests seemed fine.
However this is a slight change in behavior, so I'm working on getting that documented & agreed upon before we land anything.

OrlaM:
Ah right, it's only an issue for the incident because the app is terminated before the next launch which results in it firing twice.
This change looks good to me. I'll close my PR.
I'm about 99% sure this will fix the issue, but I'm still trying to actually reproduce it locally to be 100% sure.

glean-core/ios/GleanTests/TestUtils.swift
```diff
@@ -50,47 +50,12 @@ func stubServerReceive(callback: @escaping (String, [String: Any]?) -> Void) {
 /// This also prevents outgoing network requests during unit tests while
 /// still allowing us to use the default telemetry endpoint.
 func resetGleanDiscardingInitialPings(testCase: XCTestCase, tag: String, clearStores: Bool = true) {
-    let expectation = testCase.expectation(description: "\(tag): Ping Received")
```
travis79, line 53: Why remove this part of the test? I mean, I know we test this in glean-core also but was the change you made interfering with this?

travis79 (approved):
This looks like what we discussed, so I'm happy with it if you are happy with it (and once the checks pass)

glean-core/ios/GleanTests/TestUtils.swift
badboy: We have no initial pings anymore!
Because now we're only sending the first baseline ping on an `active` app, which tests don't have (there's no UI -> no `didBecomeActiveNotification` -> no ping).
So keeping this in now breaks tests, because it waits for pings that never arrive.

OrlaM:
I have been unable to locally reproduce any scenario that launches the app from a terminated state while in the background. The background task scheduling and notification code added in 113 don't seem to be the cause after doing some testing. 
I think this change is still worth landing to see if it fixes things, because whatever is happening does seem to be coming from some sort of background activity because Gleans internal active check guarantees that even if the client fires the set up more than once from the same session it won't result in duplicate start up pings. So unless there is something else going wrong further down the chain after the ping has been registered I can't see anywhere else this kind of issue would be introduced even if I can't nail down the exact use case that's triggering it.

badboy:
Closing this for the time being, we first want to measure this in some other way.

PR #2596, opened 2023-08-15
Bug 1848616 - Update the Server Knobs documentation

travis79:
This updates the current Server Knobs documentation to reflect recent Nimbus updates which add support for multiple features in a single rollout or experiment.

### DIFF
--- a/docs/user/user/metrics/data-control-plane/example-scenarios.md
+++ b/docs/user/user/metrics/data-control-plane/example-scenarios.md
@@ -25,6 +25,8 @@ Through [Nimbus], we have the ability to sample the population by setting the au
 
 This can be used to slowly roll out instrumentations to the population in order to validate the data we are collecting before measuring the entire population and potentially avoiding costs and  overhead by collecting data that isn’t useful.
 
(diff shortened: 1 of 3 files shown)


docs/user/user/metrics/data-control-plane/example-scenarios.md
```diff
@@ -25,6 +25,8 @@ Through [Nimbus], we have the ability to sample the population by setting the au
 
 This can be used to slowly roll out instrumentations to the population in order to validate the data we are collecting before measuring the entire population and potentially avoiding costs and  overhead by collecting data that isn’t useful.
 
+> *Note:* this functionality is intended to be used to _sample metrics for a subset of the population_, not for turning metrics on for the entire population. To do that you can just enable the metrics by default by setting `disabled: false` in the metrics.yaml. Then you can "down-sample" if necessary (see Scenario 2 below).
```
Dexterp37, line 28:
```suggestion
> *Note:* if planning to use this feature for permanently keeping a data collection on for the whole population, please consider enabling the metrics by default by setting `disabled: false` in the metrics.yaml. Then you can "down-sample" if necessary (see Scenario 2 below).
```

What about this?

Moreover, can you use one of the "important style" boxes we have in other docs?

Dexterp37 (approved):

PR #2606, opened 2023-08-28
Bug 1848201 - Introduce API to allow setting an experimentation ID

travis79:
This adds an API to allow consumers of Glean to set an "experimentation identifier" which is then sent along with all pings. This is particularly useful for client/server applications using Cirrus (Nimbus) in a client/server architecture so the experimentation id can be used to unify experiment telemetry across both the client and the server.

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -406,7 +406,7 @@ client.annotation:
     bugs:
       - https://bugzilla.mozilla.org/show_bug.cgi?id=1848201
     data_reviews:
-      - TBD
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1848201#c5
     data_sensitivity:
       - technical
     notification_emails:
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -394,7 +394,7 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
-client.annotation:
+glean.client.annotation:
   experimentation_id:
     type: string
     lifetime: application
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -394,6 +394,25 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+client.annotation:
+  experimentation_id:
+    type: string
+    lifetime: application
+    send_in_pings:
+      - all-pings
+    description: |
+      An experimentation identifier derived and provided by the application
+      for the purpose of experimenation enrollment.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1848201
+    data_reviews:
+      - TBD
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+
 glean.error:
   invalid_value:
     type: labeled_counter
--- a/glean-core/python/glean/glean.py
+++ b/glean-core/python/glean/glean.py
@@ -407,15 +407,13 @@ class Glean:
     @classmethod
     def set_experimentation_id(cls, experimentation_id: str) -> None:
         """
-        Set an experimentation identifier derived and provided by the application
-        for the purpose of enrollment and unification of telemetry across a
-        client/server experiment.
+        Set an experimentation identifier.
 
-        Note: Glean will add this experimentation id annotation to all pings that
-        are sent, and the experimentation id is not persisted between runs.
+        Note: Glean will add this experimentation id to all pings that are sent,
+        and the experimentation id is not persisted between runs.
 
         Args:
-            experimentation_id (str): The experimentation id to annotate all pings with.
+            experimentation_id (str): The experimentation id to set.
         """
         _uniffi.glean_set_experimentation_id(experimentation_id)
 
--- a/glean-core/python/glean/glean.py
+++ b/glean-core/python/glean/glean.py
@@ -405,6 +405,35 @@ class Glean:
             raise RuntimeError("Experiment data is not set")
 
     @classmethod
+    def set_experimentation_id(cls, experimentation_id: str) -> None:
+        """
+        Set an experimentation identifier derived and provided by the application
+        for the purpose of enrollment and unification of telemetry across a
+        client/server experiment.
+
+        Note: Glean will add this experimentation id annotation to all pings that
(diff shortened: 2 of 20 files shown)


travis79:
@badboy: For when you are back, this should be ready for review. I tried to break this down into smaller commits to make it easier to review. This does include a proper CHANGELOG.md entry, along with updates to the Experiment API documentation in the Glean Book. I've flagged @chutten for data-review in the associated [Bug 1848201](https://bugzilla.mozilla.org/show_bug.cgi?id=1848201#c5).

glean-core/metrics.yaml
```diff
@@ -394,6 +394,25 @@ glean.internal.metrics:
       - glean-team@mozilla.com
     expires: never
 
+client.annotation:
```
badboy, line 397: `client.annotation` is in no way special and thus has the potential to clash with apps
Should this be in the glean namespace? `glean.client.annotation`? (`glean.*` is reserved by us)

glean-core/python/glean/glean.py
```diff
@@ -404,6 +404,35 @@ def test_get_experiment_data(cls, experiment_id: str) -> "RecordedExperiment":
         else:
             raise RuntimeError("Experiment data is not set")
 
+    @classmethod
+    def set_experimentation_id(cls, experimentation_id: str) -> None:
+        """
+        Set an experimentation identifier derived and provided by the application
+        for the purpose of enrollment and unification of telemetry across a
+        client/server experiment.
```
badboy, line 412: As above

glean-core/src/internal_metrics.rs
```diff
@@ -112,6 +116,15 @@ impl AdditionalMetrics {
                 },
                 TimeUnit::Millisecond,
             ),
+
+            experimentation_id: StringMetric::new(CommonMetricData {
+                name: "experimentation_id".into(),
+                category: "client.annotation".into(),
+                send_in_pings: vec!["all-pings".into()],
```
badboy, line 123: Oh. I guess this works because our other "all-pings" metrics are actually special and handled differently internally.

We should just keep that in mind if we ever start generating code from our own metrics.yaml (we have no plans for that).

badboy (changes requested):
Code-wise this looks good.
I have some open concerns/questions though overall:

* This is its own function call. If this is called late we will get pings without the data. Maybe we need some guidance around that?
  * In the bug Christian said maybe to set this as part of intialization. What are arguments against that?
  * Are there any edge cases where during Glean init we would send pings without that ID, even if `setExperimentationId` is called before init?
* How does this play with custom pings that have `include_client_id: false`? Suddenly these pings could send a unique identifier without the ability to avoid that. Can this lead to problems?
* Is this id ending up in the deletion-request ping? Should it?

glean-core/src/internal_metrics.rs
travis79: Yeah, this wasn't _great_ but it was an already reserved "store" that the ingestion pipeline would treat the way it needed to be treated so it was fairly obvious it could be easily leveraged for the send-in-all pings purpose. I should leave a comment or document how/why it works somewhere, though.

travis79:
> Code-wise this looks good. I have some open concerns/questions though overall:
> 
>     * This is its own function call. If this is called late we will get pings without the data. Maybe we need some guidance around that?
>       
>       * In the bug Christian said maybe to set this as part of intialization. What are arguments against that?

These are both good points, let me go ahead and modify what I have here to be a Configuration parameter that's passed into Glean.initialize().


>       * Are there any edge cases where during Glean init we would send pings without that ID, even if `setExperimentationId` is called before init?

Yes, there would be cases that this would be an issue, mostly with startup pings. I think moving this to a configuration parameter solves that issue.


>     * How does this play with custom pings that have `include_client_id: false`? Suddenly these pings could send a unique identifier without the ability to avoid that. Can this lead to problems?

It currently doesn't consider that, but you are right, we probably don't want to include this when client_id isn't being sent either. I'll update the code and documentation to reflect this.
 

>     * Is this id ending up in the deletion-request ping? Should it?

Yes, it should be there as it might be necessary to delete data from other datasets where this same id is recorded (but perhaps with a different client-id, such as the "server" portion of Cirrus' data)

badboy:
> > ```
> > * How does this play with custom pings that have `include_client_id: false`? Suddenly these pings could send a unique identifier without the ability to avoid that. Can this lead to problems?
> > ```
> 
> It currently doesn't consider that, but you are right, we probably don't want to include this when client_id isn't being sent either. I'll update the code and documentation to reflect this.

Sorry to be the showstopper here, I have spend time on these questions but not even what my own answer would be. So this might warrant a brief discussion before the technical implementation. (Though in the end it's your call and I'm fine in either direction here).
Question is if unconditionally NOT putting it in pings with `include_client_id: false` the right thing? Do we need a new setting? Do we need doc updates instead? 

> > ```
> > * Is this id ending up in the deletion-request ping? Should it?
> > ```
> 
> Yes, it should be there as it might be necessary to delete data from other datasets where this same id is recorded (but perhaps with a different client-id, such as the "server" portion of Cirrus' data)

Does the current implementation guarantee that though?

travis79:
> > > ```
> > > * How does this play with custom pings that have `include_client_id: false`? Suddenly these pings could send a unique identifier without the ability to avoid that. Can this lead to problems?
> > > ```
> > 
> > 
> > It currently doesn't consider that, but you are right, we probably don't want to include this when client_id isn't being sent either. I'll update the code and documentation to reflect this.
> 
> Sorry to be the showstopper here, I have spend time on these questions but not even what my own answer would be. So this might warrant a brief discussion before the technical implementation. (Though in the end it's your call and I'm fine in either direction here). Question is if unconditionally NOT putting it in pings with `include_client_id: false` the right thing? Do we need a new setting? Do we need doc updates instead?
> 

Fair enough, I have a meeting next week with Christian and Jared, let me see if I can get a little more context around this and I'll switch to a proposal if necessary to get the proper input. If this requires a change of implementation, then I can abandon/rework this PR.

> > > ```
> > > * Is this id ending up in the deletion-request ping? Should it?
> > > ```
> > 
> > 
> > Yes, it should be there as it might be necessary to delete data from other datasets where this same id is recorded (but perhaps with a different client-id, such as the "server" portion of Cirrus' data)
> 
> Does the current implementation guarantee that though?

I'll add some tests around this to help ensure it makes it in there.

travis79:
Closing this in favor of #2615

PR #2620, opened 2023-09-14
Bug 1852217 - Glean.js python env var docs

rosahbruno:
Docs only PR companion for https://github.com/mozilla/glean.js/pull/1768

### DIFF
--- a/docs/user/language-bindings/javascript/cli.md
+++ b/docs/user/language-bindings/javascript/cli.md
@@ -17,6 +17,22 @@ by setting the `VIRTUAL_ENV` environment variable. Example:
 VIRTUAL_ENV="my/venv/path" npx glean --help
 ```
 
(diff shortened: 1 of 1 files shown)


docs/user/language-bindings/javascript/cli.md
```diff
@@ -17,6 +17,13 @@ by setting the `VIRTUAL_ENV` environment variable. Example:
 VIRTUAL_ENV="my/venv/path" npx glean --help
 ```
 
+By default Glean will try and use specific binaries based on the platform: `python.exe` on windows and `python3` everywhere else. You can customize the Python binary used through environment variables. You are able to define both `GLEAN_PYTHON` and `GLEAN_PIP`. Example:
+
+```bash
+export GLEAN_PYTHON=<path_to_python_binary>
+export GLEAN_PIP=<path_to_pip_binary>
+```
```
travis79, line 25: Would it be good to have a windows (CMD or Powershell) example also?

rosahbruno: Updated!

travis79 (approved):

PR #2621, opened 2023-09-15
bug 1847950 - Trigger upload after scanning pending pings

chutten:


chutten:
This is the one where we need to give a think to how and when we should use the global STATE inside the SDK. Should `global_state()` be fallible? Should the `Glean` instance have references to some pieces of the state?

+Cc @badboy who was going to give this a noodle.

badboy:
> This is the one where we need to give a think to how and when we should use the global STATE inside the SDK. Should `global_state()` be fallible? Should the `Glean` instance have references to some pieces of the state?
> 
> +Cc @badboy who was going to give this a noodle.

Right right. And I kinda did a bit and nothing satisfying came out of this.
As said in channel we're essentially crossing abstraction boundaries here, but there's also no good way to handle it better without re-designing a lot of this.
We could provide a fallible version of it and use that there, fully aware that it's imperfect. If the Kotlin and Swift tests pass then I'm more confident it works. But writing an appropriate test is also hard (because ... ugh ... this is all kinda async and stuff).

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -190,17 +193,15 @@ class CustomPingTest {
...
-        // But if the custom ping is specifically submitted,
+        // If the custom ping is specifically submitted,
         // it should be received.
         customPing.submit()
         // Trigger work manager once.
         // This should launch one worker that handles all pending pings.
         triggerWorkManager(context)
         var request = server.takeRequest(2L, TimeUnit.SECONDS)!!
+        // Assert we only got the one:
+        assertNull(server.takeRequest(2L, TimeUnit.SECONDS))
```
badboy, line 204: Note that this means a 2s wait time here now (slower tests), but I guess it's our only option.

badboy (approved):
triggered iOS CI. Code looks good.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
chutten: Yup. Unless someone wants to build a way for the Kotlin LB to synchronously determine whether pings are submitted

PR #2623, opened 2023-09-18
bug 1762859 - Silently ignore add(0) on counter and labeled_counter

chutten:


chutten:
(I'm guessing one or two of the language bindings probably have their own tests that will fail that I missed by only testing Rust and Python locally, but we'll see)

travis79 (approved):
This matches what I expect from the discussion in the Bug. r+

docs/user/reference/metrics/counter.md
```diff
@@ -123,8 +124,8 @@ Glean.controls.refreshPressed.add(5);
 
 ### `testGetValue`
 
-Gets the recorded value for a given counter metric.  
-Returns the count if data is stored.  
+Gets the recorded value for a given counter metric.
+Returns the count if data is stored.
```
badboy, line 128: Note that the double-spacing at line-end here was load-bearing: It enforces line breaks in the generated HTML.

glean-core/src/metrics/counter.rs
```diff
@@ -62,16 +63,23 @@ impl CounterMetric {
...
-            );
-            return;
-        }
+        match amount.cmp(&0) {
+            Ordering::Less => {
+                record_error(
+                    glean,
+                    &self.meta,
+                    ErrorType::InvalidValue,
+                    format!("Added negative or zero value {}", amount),
```
badboy, line 72:
```suggestion
                    format!("Added negative value {}", amount),
```

badboy (changes requested):

PR #2626, opened 2023-09-22
Avoid logging an error on an empty metrics config passed in

badboy:


glean-core/src/lib.rs
```diff
@@ -860,6 +860,12 @@ pub fn glean_test_get_experiment_data(experiment_id: String) -> Option<RecordedE
 ///
 /// See [`core::Glean::set_metrics_enabled_config`].
 pub fn glean_set_metrics_enabled_config(json: String) {
+    // An empty config means it is not set,
+    // so we avoid logging an error about it.
+    if json.is_empty() {
+        return;
```
travis79, line 866: Setting an empty config used to clear the current config...  🤔 I'm not sure returning here would keep that behavior but I agree we should avoid logging an error.  Let me look into this some more and see how best to accomplish this

travis79 (changes requested):

glean-core/src/lib.rs
travis79: I guess since I added the merging of configs, an empty config wouldn't `extend()` anything or overwrite (or even know _what_ to overwrite), so this is probably the correct behavior.

travis79 (approved):

PR #2627, opened 2023-09-22
Bug 1854678 - maxEvents=1 docs clarifying comments

rosahbruno:
Docs only

This issue was reported by agibson when using the newest version (v2.0.2) of Glean.js.


perrymcmanis144 (approved):
Doc only change, adds important clarity LGTM 👍

PR #2633, opened 2023-10-02
Bug 1854207 - Update Glean docs to show how to access extras in test API

perrymcmanis144:


docs/user/reference/metrics/event.md
```diff
@@ -215,6 +223,21 @@ auto optEvents = mozilla::glean::views::login_opened.TestGetValue();
 auto events = optEvents.extract();
 ASSERT_EQ(2UL, events.Length());
 ASSERT_STREQ("login_opened", events[0].mName.get());
+
+// Note that the list of extra key/value pairs can be in any order.
```
perrymcmanis144, line 227: @chutten I'm not sure how best to show what folks need to know so I heavily borrowed from[ a FOG test](https://searchfox.org/mozilla-central/source/toolkit/components/glean/tests/gtest/TestFog.cpp#156) since it was illuminating to me. Is this too much? Not Enough?

docs/user/reference/metrics/event.md
```diff
@@ -187,6 +190,10 @@ var snapshot = views::login_opened.test_get_value(None).unwrap();
 assert_eq!(2, snapshot.len());
 let first = &snapshot[0];
 assert_eq!("login_opened", first.name);
+
+let extra = event.extra.unwrap();
+assert_eq!(1, extra.len());
```
chutten, line 195: We put a `len()` function on the event extra impl?

docs/user/reference/metrics/event.md
```diff
@@ -187,6 +190,10 @@ var snapshot = views::login_opened.test_get_value(None).unwrap();
...
+assert_eq!("toolbar", extra["source_of_login"]);
```
chutten, line 196: Does this syntax work? I don't think Rust has subscript operator overloading like this and we provide concrete structs, so this should maybe be `extra.source_of_login`?

docs/user/reference/metrics/event.md
```diff
@@ -199,7 +206,8 @@ import * as views from "./path/to/generated/files/views.js";
 const snapshot = await views.loginOpened.testGetValue();
 assert.strictEqual(2, snapshot.length);
 const first = snapshot[0]
```
chutten, line 208: semicolon here, too, please

docs/user/reference/metrics/event.md
chutten: Well, looping after asserting it's an array of length 1 is a little odd. I'd probably `auto extra = events[0].mExtra[0];` and then straight up assert the key STREQ "source_of_login" and the value STREQ "toolbar"

chutten (changes requested):

docs/user/reference/metrics/event.md
perrymcmanis144: So I cargo culted this from the Rust test for the event metric, as it happens:

https://github.com/mozilla/glean/blob/main/glean-core/tests/event.rs#L80

this both checks the length and then grabs the extra ala the syntax I used in the next line. I'm happy to change this though, if we don't like that.

docs/user/reference/metrics/event.md
perrymcmanis144: Fair enough. If you think that is adequately descriptive for someone to understand then I am fine with it personally. I was mostly motivated by the lack of a guarantee re ordering, maybe for a C++ dev this is an exercise best left to the reader? I will update this

docs/user/reference/metrics/event.md
perrymcmanis144: Updated this to reflect your suggestion. I don't think the length assertion is at all useful and if there is a direct access I think we should rec that as it will potentially work with folks' autocomplete and maybe the compiler will be more useful in that case.

chutten: Oh, huh, the `test_get_value` on the trait returns a Vec<RecordedEvent> instead of a Vec<T>. Weird.

docs/user/reference/metrics/event.md
```diff
@@ -187,6 +190,9 @@ var snapshot = views::login_opened.test_get_value(None).unwrap();
 assert_eq!(2, snapshot.len());
 let first = &snapshot[0];
 assert_eq!("login_opened", first.name);
+
+let extra = event.extra.unwrap();
+assert_eq!("toolbar", extra.source_of_login);
```
chutten, line 195: Yeah, so apparently you had this more right the first time. Sorry about that. `event.extra` is a `Option<HashMap<String, String>>`

chutten (changes requested):

chutten (approved):

PR #2637, opened 2023-10-04
Revert "Dependabot: Group all Python package updates"

badboy:
This reverts commit cbe5ef9be06359a7b218c008834c3bae145b88c9

---

The bundling didn't play out as I hoped it would.
The PR titles are now less descriptive ("Bump the python-packages group in /glean-core/python with 1 update") and still mostly contain only a single dependency update.
Feels like reverting this thus is better.

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -15,7 +15,3 @@ updates:
     ignore:
       # Updated in lockstep across all implementations
       - dependency-name: "glean_parser"
-    groups:
-      python-packages:
-        patterns:
-          - "*"


travis79 (approved):

travis79 (changes requested):
I approved this, but I don't see where it reverts the specific commit it says it was rolling back 🤔

badboy:
> I approved this, but I don't see where it reverts the specific commit it says it was rolling back 🤔

D'oh! I rebased, but opened the PR before that and so github picked the wrong commit as the PR title

travis79 (approved):

PR #2648, opened 2023-10-19
Update to glean_parser v10.0.0

badboy:
Yes, we skipped a release.


badboy:
Oh actually: wrapping this in #2456

PR #2649, opened 2023-10-19
Update dependencies to bring in line with mozilla-central

badboy:
* [x] test against m-c


badboy:
grrrrr, the CI failure will be fixed by #2456 , but I'll send a separate PR to change how this is installed.

travis79 (approved):

PR #2654, opened 2023-10-23
Experimentation ID was actually released in v55

badboy:
[doc only]


perrymcmanis144 (approved):
doc only change, lgtm 👍

PR #2655, opened 2023-10-25
Remove libc 0.2.419 audit

badboy:
Our audits might be consumed by others, so calling out an unused target might not apply to other consumers.
Instead we rely on m-c audits of this.

[doc only]


badboy:
See also https://bugzilla.mozilla.org/show_bug.cgi?id=1860966

travis79 (approved):

PR #2665, opened 2023-11-13
Bug 1862640, bug 1862641 - Use correct argument name in initialization docs

badboy:


badboy:
beta failure fixed in https://github.com/mozilla/glean/pull/2667

abhi-agg (approved):

PR #2669, opened 2023-11-15
Update to Xcode 15

badboy:
This is missing the cargo-vet, I'll add that once I know CI works.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -537,7 +537,7 @@ jobs:
             DEVICES=$(xcrun xctrace list devices 2>&1)
             echo "$DEVICES"
             # Pick a device and start it
-            UUID=$(echo "$DEVICES" | grep --max-count=1 'iPhone 15 (17' | awk -F'[()]' '{print $4}')
+            UUID=$(echo "$DEVICES" | grep --max-count=1 'iPhone 15 Simulator (17' | awk -F'[()]' '{print $4}')
             xcrun simctl boot "$UUID"
             # Store build type for use in cache key
             if [ -z "${CIRCLE_TAG}" ]; then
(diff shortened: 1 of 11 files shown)


.circleci/config.yml
```diff
@@ -528,32 +521,24 @@ jobs:
 
             # For some reason everything works fine if we use the host clang,
             # not the Xcode-bundled clang.
-            echo '[target.x86_64-apple-darwin]' >> ~/.cargo/config
-            echo 'linker = "/usr/bin/clang"' >> ~/.cargo/config
+            echo '[target.aarch64-apple-darwin]' >> ~/.cargo/config
```
travis79, line 524: Do you see any reason that we might want to continue to test for x86 Apple hardware or is updating to Apple silicon here cover us good enough?  (not blocking, just curious if you see any need to maintain coverage over old hardware, though I guess most of us that build Glean have upgraded by now)

travis79 (approved):
Please update the build setup docs in the developer documentation:

Here: https://github.com/mozilla/glean/blob/b009bfa2216e863b8968169c596f65f512c1fe99/samples/ios/app/README.md?plain=1#L12

and here: https://github.com/mozilla/glean/blob/b009bfa2216e863b8968169c596f65f512c1fe99/docs/dev/ios/setup-ios-build-environment.md?plain=1#L5

After that, I'm good with upgrading to v15

.circleci/config.yml
badboy: For Apple aarch is where the future is.
CircleCI is dropping the x64 intel machines in the summer next year. I wanted to try and see if they work for us today. They _are_ more expensive ([twice to be precise](https://circleci.com/product/features/resource-classes/#macos)), so I might drop back to x64 for the time being.

badboy: oh one more reason: we don't have an intel mac anymore for dev work! Which means it's extremely hard for us to debug when CI breaks on it as we won't be able to reproduce it locally correctly.

PR #2671, opened 2023-11-20
Install setuptools in Builder

bendk:
This is needed for Python 3.12 (https://github.com/python/cpython/issues/95299)

(I hit this after upgrading to Fedora 39 and trying to build app-services)

### DIFF
--- a/glean-core/build/src/lib.rs
+++ b/glean-core/build/src/lib.rs
@@ -110,6 +110,7 @@ impl Builder {
         let venv = VirtualEnv::new(&sh, "py3-glean_parser")?;
 
         let glean_parser = format!("glean_parser~={GLEAN_PARSER_VERSION}");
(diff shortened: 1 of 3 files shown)


glean-core/build/src/lib.rs
```diff
@@ -110,6 +110,7 @@ impl Builder {
         let venv = VirtualEnv::new(&sh, "py3-glean_parser")?;
 
         let glean_parser = format!("glean_parser~={GLEAN_PARSER_VERSION}");
+        venv.pip_install("setuptools")?;
```
badboy, line 113:
```suggestion
        // TODO: Remove after we switched glean_parser away from legacy setup.py
        venv.pip_install("setuptools")?;
```

badboy (approved):

PR #2677, opened 2023-11-28
Bug 1815253 - Record what caused an RKV db load failure

perrymcmanis144:
WIP for instrumenting which RKV failure caused a new rkv to be created

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -706,6 +706,20 @@ glean.database:
       - glean-team@mozilla.com
     expires: never
 
+  rkv_load_error:
+    type: string
+    description: |
+      If there was an error loading the RKV database, record it.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1815253
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1815253
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+
 glean.validation:
   foreground_count:
     type: counter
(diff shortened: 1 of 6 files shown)


perrymcmanis144:
@badboy alright probably ready for a look just to make sure it matches your expectations while I work through test failures.

glean-core/metrics.yaml
```diff
@@ -706,6 +706,20 @@ glean.database:
...
     expires: never
 
+  rkv_load_result:
+    type: string
+    description: |
+      The result of attempting to load the RKV database.
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1815253
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=todo
```
badboy, line 716: Needs an update here.

glean-core/metrics.yaml
```diff
@@ -706,6 +706,20 @@ glean.database:
       - glean-team@mozilla.com
     expires: never
 
+  rkv_load_result:
+    type: string
+    description: |
+      The result of attempting to load the RKV database.
```
badboy, line 712: Should we mention here more precisely that it's only the errors and that it is not reported on successful loads?

glean-core/src/internal_metrics.rs
```diff
@@ -277,6 +280,15 @@ impl DatabaseMetrics {
                 },
                 MemoryUnit::Byte,
             ),
+
+            rkv_load_result: StringMetric::new(CommonMetricData {
+                name: "rkv_load_result".into(),
+                category: "glean.error".into(),
+                send_in_pings: vec!["metrics".into()],
+                lifetime: Lifetime::Application,
```
badboy, line 288: The metrics.yaml doesn't set a lifetime, so it defaults to `ping`, which I think makes more sense? We don't want this to be sent in every subsequent metrics ping until a restart, do we?

badboy (changes requested):

badboy (approved):

PR #2678, opened 2023-11-28
Update to UniFFI 0.25.2

badboy:
Need to double-check that pulling that into m-c along with a-s works.


badboy:
Checked that in m-c. It seems to work, there's a-s changes that need to happen but bdk will take care of that.

PR #2689, opened 2023-12-13
Bug 1809317 - Rust: Remove `collect_string` method

badboy:
While technically part of the API since forever, it was never part of the supported public API and not used within the SDK either.


perrymcmanis144 (approved):
removal of unused code, passes tests so LGTM

PR #2692, opened 2023-12-15
Upgrade CI to Taskgraph 7.0.0

ahal:
Taskgraph 7 has several backwards incompatible changes, so I'm going through and updating projects to keep everyone in sync and also to do it while the changes are fresh in my mind.

While upgrading, I noticed we weren't setting a checkout cache for the Decision task, so added that as well.


badboy (approved):
CI ran as expected, thus good to go.

PR #2693, opened 2023-12-19
CI: Upgrade Taskgraph to 7.0.2

ahal:
This fixes a regression from 7.0.0 that was causing actions to fail.


ahal:
Decision failure is just pending a ci-config deploy.. will re-run shortly.

badboy (approved):

PR #2680, opened 2023-11-29
Update AGP to version 8.2.1 or newer

rvandermeulen:

PR #2697, opened 2023-12-28
Bug 1870048 - Replace whenTaskAdded with configureEach to avoid unnecesary configuration.

mcarare:
Tested locally and this seems to be enough for moving forward with https://bugzilla.mozilla.org/show_bug.cgi?id=1870050.
Also, before going even further https://github.com/mozilla/glean/pull/2680 IMO this would be a good test to see if moving to Task Configuration Avoidance would be smooth.


travis79 (approved):
Thanks!

travis79:
Okay, great. Seems like the failure was just transient. Feel free to land this whenever you like @mcarare, and please let me know when you need to see this in a Glean release and I'll see if I can schedule one.

mcarare:
I don't think I have landing rights on this repo. This would be nice to have as soon as possible because it unblocks some other pending work I have on Firefox.

PR #2702, opened 2024-01-05
Bug 1873076 - Add dynamic API for experimentation id

travis79:


glean-core/src/lib.rs
```diff
@@ -869,6 +869,18 @@ pub fn glean_test_get_experiment_data(experiment_id: String) -> Option<RecordedE
...
 
+/// Set an experimentation identifier dynamically.
+///
+/// Note: it's probably a good idea to unenroll from any experiments when identifiers change.
+pub fn glean_set_experimentation_id(experimentation_id: String) {
+    core::with_glean(|glean| {
+        glean
+            .additional_metrics
+            .experimentation_id
+            .set_sync(glean, experimentation_id.to_string());
```
badboy, line 880: This is a normal metric and thus this would cause IO on the main thread, because this is not dispatched like other metrics. Is that intended?

travis79: Nope, this doesn't need to do IO on the main thread, I'll change this to use the proper `with_glean`

badboy (approved):
code-wise ok, but should we extend the tests somewhere? I'd be fine if this is at least tested in one of the bindings.

travis79:
> code-wise ok, but should we extend the tests somewhere? I'd be fine if this is at least tested in one of the bindings.

Absolutely, let me add some test coverage for this and I'll ping you again once it's ready to review.

PR #2707, opened 2024-01-10
Bug 1873911 - Update rkv to 0.19.0

saschanaz:
I see a couple of failures locally on Windows:

* database::test::safe_mode::migration_doesnt_overwrite
* database::test::safe_mode::migration_ignores_empty_database

But these two fail even without the update, so 🤷

### DIFF
--- a/glean-core/Cargo.toml
+++ b/glean-core/Cargo.toml
@@ -30,7 +30,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.44"
-rkv = { version = "0.18.4", default-features = false, features = ["lmdb"] }
+rkv = { version = "0.19.0", default-features = false, features = ["lmdb"] }
 bincode = "1.2.1"
 log = "0.4.8"
 uuid = { version = "1.0", features = ["v4"] }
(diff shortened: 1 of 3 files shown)


badboy:
cargo-vet issue: we should mark rkv as maintained by us, so it passes.
MSRV: we can probably raise that. I can do that separately though.

supply-chain/config.toml
```diff
@@ -19,6 +19,9 @@ url = "https://raw.githubusercontent.com/divviup/libprio-rs/main/supply-chain/au
 [imports.mozilla]
 url = "https://raw.githubusercontent.com/mozilla/supply-chain/main/audits.toml"
 
+[imports.zcash]
+url = "https://raw.githubusercontent.com/zcash/rust-ecosystem/main/supply-chain/audits.toml"
```
badboy, line 23: uhm. Nope, let's please not do that. So far Mozilla doesn't import zcash's audits, so unless I hear otherwise Glean won't do it either.
I can take care of `cargo-vet` for this.

saschanaz: Ah, I was doing that and saw the suggestion. Will roll back. 0.4.19 is totally fine while I need to take a look at 0.4.20

badboy (approved):

glean-core/Cargo.toml
```diff
@@ -30,7 +30,7 @@ maintenance = { status = "actively-developed" }
 [dependencies]
 serde = { version = "1.0.104", features = ["derive"] }
 serde_json = "1.0.44"
-rkv = { version = "0.18.4", default-features = false, features = ["lmdb"] }
+rkv = { version = "0.19.0", default-features = false, features = ["lmdb"] }
```
saschanaz, line 33: Does Glean still need lmdb migration path, btw?

badboy: probably not, but I need to write the doc to inform people still. We can do that in a followup.

PR #2719, opened 2024-01-19
Prototyping a Glean Event Listener

travis79:
This adds a method for a consuming application to register an event listener to receive notifications when events are recorded.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v56.1.0...main)
 
+* General
+  * Added an experimental event listener API ([#2719](https://github.com/mozilla/glean/pull/2719)) 
 * Android
   * BREAKING CHANGE: Update JNA to version 5.14.0. Projects using older JNA releases may encounter errors until they update. ([#2727](https://github.com/mozilla/glean/pull/2727))
   * Set the target Android SDK to version 34 ([#2709](https://github.com/mozilla/glean/pull/2709))
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
@@ -378,6 +378,28 @@ open class GleanInternalAPI internal constructor() {
     }
 
     /**
+     * EXPERIMENTAL: Register a listener to receive event recording notifications
+     *
+     * NOTE: Only one listener may be registered for a given tag. Each subsequent registration with
+     * that same tag replaces the currently registered listener.
+     *
+     * @param tag a tag to use when unregistering the listener
+     * @param listener implements the `GleanEventListener` interface
+     */
+    fun registerEventListener(tag: String, listener: GleanEventListener) {
+        gleanRegisterEventListener(tag, listener)
+    }
+
+    /**
+     * Unregister an event listener
+     *
+     * @param tag the tag used when registering the listener to be unregistered
+     */
+    fun unregisterEventListener(tag: String) {
+        gleanUnregisterEventListener(tag)
+    }
+
+    /**
      * Initialize the core metrics internally managed by Glean (e.g. client id).
      */
     internal fun getClientInfo(configuration: Configuration, buildInfo: BuildInfo): ClientInfoMetrics {
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
@@ -19,6 +19,7 @@ import mozilla.telemetry.glean.delayMetricsPing
 import mozilla.telemetry.glean.getContext
 import mozilla.telemetry.glean.getMockWebServer
 import mozilla.telemetry.glean.getPlainBody
+import mozilla.telemetry.glean.internal.GleanEventListener
 import mozilla.telemetry.glean.resetGlean
 import mozilla.telemetry.glean.testing.ErrorType
 import mozilla.telemetry.glean.testing.GleanTestRule
@@ -523,5 +524,79 @@ class EventMetricTypeTest {
         // No events should be left, thus we don't receive it.
         ping.submit()
         assertNull(waitForPingContent(pingName, null, server))
+    }
+
+    @Test
+    fun `Event listener correct reports event recordings`() {
+        // Define some click events to listen for in the listener
+        val click1 = EventMetricType<ClickExtras>(
+            CommonMetricData(
+                disabled = false,
+                category = "ui",
+                lifetime = Lifetime.PING,
+                name = "click1",
+                sendInPings = listOf("store1"),
+            ),
+            allowedExtraKeys = listOf(),
+        )
+        val click2 = EventMetricType<ClickExtras>(
+            CommonMetricData(
+                disabled = false,
+                category = "ui",
+                lifetime = Lifetime.PING,
+                name = "click2",
+                sendInPings = listOf("store1"),
+            ),
+            allowedExtraKeys = listOf(),
+        )
+        val click3 = EventMetricType<ClickExtras>(
+            CommonMetricData(
+                disabled = false,
+                category = "ui",
+                lifetime = Lifetime.PING,
+                name = "click3",
+                sendInPings = listOf("store1"),
+            ),
+            allowedExtraKeys = listOf(),
+        )
+
+        class TestEventListener : GleanEventListener {
+            val listenerTag = "TestEventListener"
+
+            var lastSeenId: String = ""
+            var count: Int = 0
+
+            override fun onEventRecorded(id: String) {
+                this.lastSeenId = id
+                this.count += 1
+            }
+        }
+
+        val listener = TestEventListener()
+
+        // Register the listener
+        Glean.registerEventListener(listener.listenerTag, listener)
+
+        // Check that events are properly reported
+        click1.record()
+        assertEquals("ids must match", "ui.click1", listener.lastSeenId)
+        assertTrue("count must be correct", listener.count == 1)
+        click2.record()
+        assertEquals("ids must match", "ui.click2", listener.lastSeenId)
+        assertTrue("count must be correct", listener.count == 2)
+        click3.record()
+        assertEquals("ids must match", "ui.click3", listener.lastSeenId)
+        assertTrue("count must be correct", listener.count == 3)
+
+        // Unregister the listener
+        Glean.unregisterEventListener(listener.listenerTag)
+
+        // Check that events are no longer reported
+        click1.record()
+        assertEquals("ids must match", "ui.click3", listener.lastSeenId)
+        assertTrue("count must be correct", listener.count == 3)
+        click2.record()
+        assertEquals("ids must match", "ui.click3", listener.lastSeenId)
+        assertTrue("count must be correct", listener.count == 3)
     }
 }
--- a/glean-core/src/glean.udl
+++ b/glean-core/src/glean.udl
@@ -36,6 +36,16 @@ namespace glean {
     void glean_set_experimentation_id(string experimentation_id);
     string? glean_test_get_experimentation_id();
 
+    // EXPERIMENTAL: Register a listener to receive notification of event recordings
+    //
+    // tag: String value used later to unregister the listener
+    // listener: An object which implements the GleanEventListener interface
+    void glean_register_event_listener(string tag, GleanEventListener listener);
+    // EXPERIMENTAL: Unregister a previously registered event listener
+    //
+    // tag: The tag used when registering the listener
+    void glean_unregister_event_listener(string tag);
+
     // Server Knobs API
     void glean_set_metrics_enabled_config(string json);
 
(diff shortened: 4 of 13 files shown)


glean-core/src/lib.rs
```diff
@@ -229,6 +229,25 @@ fn setup_state(state: State) {
...
+static EVENT_LISTENERS: OnceCell<Mutex<Vec<Box<dyn GleanEventListener>>>> = OnceCell::new();
+
+fn event_listeners() -> &'static Mutex<Vec<Box<dyn GleanEventListener>>> {
+    if EVENT_LISTENERS.get().is_none() && EVENT_LISTENERS.set(Mutex::new(vec![])).is_err() {
+        log::error!(
+            "Global Glean event listeners object is initialized already. This probably happened concurrently."
+        )
+    }
+
+    return EVENT_LISTENERS.get().unwrap();
```
badboy, line 243: then this becomes just:

```suggestion
    return &EVENT_LISTENERS;
```

badboy (commented):
drive-by comment, but I noticed this

glean-core/src/lib.rs
travis79: Nice! I was imitating the other callbacks, good to know this can be simplified. :)

badboy: yeah, it's const since Rust 1.63: https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.new
We might update the others in a followup, though we should check that it's actually the same behavior.

glean-core/android/src/test/java/mozilla/telemetry/glean/private/EventMetricTypeTest.kt
```diff
@@ -524,4 +525,63 @@ class EventMetricTypeTest {
...
+                category = "ui",
+                lifetime = Lifetime.PING,
+                name = "click3",
+                sendInPings = listOf("store1"),
+            ),
+            allowedExtraKeys = listOf(),
+        )
+
+        class TestEventListener : GleanEventListener {
+            var id: String = ""
```
badboy, line 564: would probably name this `lastSeenId` or `lastId` to make it obvious that we overwrite this with every call.

travis79:
Hmm, same iOS failure I got locally. Let me look into this a little more and I'll either fix any problems caused by this PR or file a bug to fix it if it is something else that broke this test.

glean-core/src/glean.udl
```diff
@@ -164,6 +165,12 @@ callback interface OnGleanEvents {
     void shutdown();
 };
 
+// A callback handler that receives the IDs of recorded events
+callback interface GleanEventListener {
+    // Called when an event is recorded, indicating the id of the event
+    void on_event_recorded(string id);
```
chutten, line 171: I'd prefer either spelling out "identifier" or explaining that "id" means the category and name concatenated with a `.`.

We, alas, don't have specific nomenclature about how to talk about FQMNs (fully-qualified metric names) yet.

glean-core/src/lib.rs
```diff
@@ -285,6 +298,12 @@ pub trait OnGleanEvents: Send {
     }
 }
 
+/// A callback handler that receives the IDs of recorded events
+pub trait GleanEventListener: Send {
+    /// Called when an event is recorded, indicating the id of the event
+    fn on_event_recorded(&self, id: String);
```
chutten, line 304: Why `String` and not `&str`? I don't imagine listeners will want to own the FQMN if they can avoid it?

chutten (commented):
I have some questions:
1) Do we want to advertise guarantees (or lack of guarantees) about the (a-)synchronicity of the callbacks wrt the event being recorded? Or their order? Or the order of event listeners being called for a specific event?
2) How do I unregister an event listener?

chutten:
Oh, and where will this be documented?

glean-core/src/lib.rs
travis79: I don't think `&str` is possible through UniFFI...  and the metric `base_identifier()` function returns a `String` rather than `&str`.

Nimbus is planning on using this value as a key lookup to increment counters in a map structure, so their "ownership" of this value will be quickly garbage collected, anyway.

travis79:
> Oh, and where will this be documented?

I was tempted to leave this as an "undocumented" feature, but in all practicality this will be documented in the Glean Book along with everything else. I'll add a commit with documentation and changelog entries.

travis79:
Okay @chutten, I think I've addressed all of your feedback and am ready for you to take another look.

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v56.1.0...main)
 
+* General
+  * Added an experimental event listener API
```
chutten, line 6: Be sure to link the PR
```suggestion
  * Added an experimental event listener API ([#2719](https://github.com/mozilla/glean/pull/2719/files))
```

docs/user/user/metrics/glean-event-listener.md
```diff
@@ -0,0 +1,67 @@
+# Glean Event Listener
+
+> **Note:** This API is currently experimental and subject to change or elimination. Please reach out to the Glean Team if you are planning on using this API in its experimental state.
+
+## Summary
+
+Glean provides an API by which to register a callback by which a consumer can be notified of an event being recorded.
```
chutten, line 7:
```suggestion
Glean provides an API to register a callback by which a consumer can be notified of all event metrics being recorded.
```

docs/user/user/metrics/glean-event-listener.md
```diff
@@ -0,0 +1,67 @@
+# Glean Event Listener
```
chutten, line 1: not a metrics API, this is a General API, so maybe it should be in `user/reference/general/` like `experiments-api.md`

docs/user/user/metrics/glean-event-listener.md
```diff
@@ -0,0 +1,67 @@
...
+
+> **Note:** This API is currently experimental and subject to change or elimination. Please reach out to the Glean Team if you are planning on using this API in its experimental state.
+
+## Summary
+
+Glean provides an API by which to register a callback by which a consumer can be notified of an event being recorded.
+
+## Usage
+
+Consumers can register a callback through this API and then the callback will be called with the base identifier of the event when it is recorded. The base identifier of the event consists of the category name and the event name with a dot separator:
```
chutten, line 11:
```suggestion
Consumers can register a callback through this API which will be called with the base identifier of each event metric when it is recorded. The base identifier of the event consists of the category name and the event name with a dot separator:
```

docs/user/user/metrics/glean-event-listener.md
```diff
@@ -0,0 +1,67 @@
...
+
+`<event_category>.<event_name>`
+
+Glean will execute the registered callbacks on a background thread independent of the thread in which the event is being recorded in order to not interfere with the collection of the event.
+
+Glean will utilize the same dispatching mechanisms used to ensure events are recorded in the order they are received to ensure that event recordings are reported to listeners in the same order that they are recorded.
```
chutten, line 17: Lead with the impact.
```suggestion
Glean will ensure that event recordings are reported to listeners in the same order that they are recorded by using the same dispatching mechanisms used to ensure events are recorded in the order they are received.
```

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -377,6 +377,25 @@ open class GleanInternalAPI internal constructor() {
         return gleanTestGetExperimentationId() ?: throw NullPointerException("Experimentation Id is not set")
     }
 
+    /**
+     * Register a listener to receive event recording notifications
+     *
+     * @param tag a tag to use when unregistering the listener
```
chutten, line 383: What happens if a second listener is registered with a previous tag?

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -377,6 +377,25 @@ open class GleanInternalAPI internal constructor() {
...
+     * @param listener implements the `GleanEventListener` interface
+     */
+    fun registerEventListener(tag: String, listener: GleanEventListener) {
+        gleanRegisterEventListener(tag, listener)
+    }
+
+    /**
+     * Unregister an event listener
+     *
+     * @param tag a tag to use when unregistering the listener
```
chutten, line 393:
```suggestion
     * @param tag the tag used when registering the listener to be unregistered
```

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
```diff
@@ -377,6 +377,25 @@ open class GleanInternalAPI internal constructor() {
         return gleanTestGetExperimentationId() ?: throw NullPointerException("Experimentation Id is not set")
     }
 
+    /**
+     * Register a listener to receive event recording notifications
```
chutten, line 381: Lacks the `EXPERIMENTAL` tag from the Swift API

glean-core/src/lib.rs
```diff
@@ -229,6 +229,29 @@ fn setup_state(state: State) {
...
+/// to receive event recording notifications.
+static EVENT_LISTENERS: OnceCell<Mutex<HashMap<String, Box<dyn GleanEventListener>>>> =
+    OnceCell::new();
+
+fn event_listeners() -> &'static Mutex<HashMap<String, Box<dyn GleanEventListener>>> {
+    if EVENT_LISTENERS.get().is_none() && EVENT_LISTENERS.set(Mutex::new(HashMap::new())).is_err() {
+        log::error!("Event Listeners object is already set, this probably happened concurrently");
+    }
+
+    EVENT_LISTENERS.get().unwrap()
```
chutten, line 242: Is `get_or_init` not suitable here? `once_cell` has [a decent example](https://docs.rs/once_cell/latest/once_cell/#lazy-initialized-global-data).

glean-core/src/lib.rs
```diff
@@ -1066,6 +1096,23 @@ pub fn glean_submit_ping_by_name_sync(ping_name: String, reason: Option<String>)
     core::with_glean(|glean| glean.submit_ping_by_name(&ping_name, reason.as_deref()))
 }
 
+/// Register a listener object to recieve notifications of event recordings.
```
chutten, line 1099: Docstrings lacking `EXPERIMENTAL` note

glean-core/src/lib.rs
```diff
@@ -1066,6 +1096,23 @@ pub fn glean_submit_ping_by_name_sync(ping_name: String, reason: Option<String>)
...
+///
+/// * `tag` - A string identifier used to later unregister the listener
+/// * `listener` - Implements the `GleanEventListener` trait
+pub fn glean_register_event_listener(tag: String, listener: Box<dyn GleanEventListener>) {
+    register_event_listener(tag, listener);
+}
+
+/// Unregister an event listener from recieving notifications.
+///
+/// Does not panic if the listener doesn't exist.
```
chutten, line 1111: Needs an `Arguments` list I think

glean-core/src/metrics/event.rs
```diff
@@ -84,6 +84,14 @@ impl EventMetric {
                 }
             }
         });
+
+        let id = self.meta().base_identifier();
+        crate::launch_with_glean(move |_| {
+            let event_listeners = crate::event_listeners().lock().unwrap();
```
chutten, line 90: Guess this means any run with events in it will init that empty `OnceCell<Mutex<HashMap<String, Listener>>>` eventually. Probably not worth it to special case a "Don't init this if it isn't init` path, though, huh.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
travis79: It would be hijacked, I suppose. I don't expect anyone other than Nimbus to use this (famous last words...), so I really don't _think_ this will be a problem (more famous last words...).

glean-core/src/lib.rs
travis79: Ahh, nifty! You get used to the patterns you see and just keep repeating them. This seems to do the trick nicely.

glean-core/src/metrics/event.rs
travis79: We could, I don't see much overhead for this at this point but happy to keep this optimization in my back pocket, should it turn out to be useful.

glean-core/android/src/main/java/mozilla/telemetry/glean/Glean.kt
chutten: Then we should document it. "Only one listener may be registered for a given tag. Each subsequent registration with a given tag unregisters the previous listener for that tag" or whatever.

glean-core/src/lib.rs
```diff
@@ -229,6 +229,26 @@ fn setup_state(state: State) {
...
 }
 
+/// A global singleton that stores listener callbacks registered with Glean
+/// to receive event recording notifications.
+static EVENT_LISTENERS: OnceCell<Mutex<HashMap<String, Box<dyn GleanEventListener>>>> =
+    OnceCell::new();
+
+fn event_listeners() -> &'static Mutex<HashMap<String, Box<dyn GleanEventListener>>> {
+    EVENT_LISTENERS.get_or_init(|| Mutex::new(HashMap::new()))
+    // EVENT_LISTENERS.get().unwrap()
```
chutten, line 239: Presumably not intending to leave this in.

chutten (approved):
r+wc

PR #2720, opened 2024-01-19
Fix incorrect method name in Kotlin

travis79:
Fixes an incorrect method name that slipped through review.


badboy (approved):

badboy:
Woops. Changelog entry?

PR #2722, opened 2024-01-22
Rust: Remove `OnceCell`s around Mutexes

badboy:
Now that Mutex, Vec and String have const `new` methods, we don't need the OnceCell anymore and can just use the statics directly.

This should still be safe:
* Everything still behind a Mutex, and thus thread-safe
* We previously didn't reset those values, and we still don't other than setting the Mutex-locked wrapped value

---

This is the followup I promised in #2719 comments


rosahbruno (approved):
Makes sense, LGTM

PR #2737, opened 2024-02-12
Update Kotlin Compiler to version 1.9.22

rvandermeulen:


rvandermeulen:
So, this builds successfully as-is but hits NewApi linter failures which are usually tied to AGP mismatches. And sure enough, those go away if I bump AGP to 8.2.2, but then we hit the build issues already noted in the other PR 😞

badboy (approved):

PR #2744, opened 2024-02-15
test AGP update

mcarare:


mcarare:
I don't have much info about what was automatically published before the update, libraries that need to be published manually now. Closing this and it can be taken over by someone with more info on the mentioned issue.

PR #2746, opened 2024-02-18
Use a Gradle version catalog for managing dependencies

rvandermeulen:


chutten:
Gonna step out of the way of someone more familiar with Android build configuration than I am.

From what I understand this is a boon to organization supporting grouping and modularization while also splitting the version tracking from the dependency graph building? I'm broadly in favour, inasmuch as I'm not misunderstanding.

badboy (approved):

PR #2751, opened 2024-02-22
Ensure glean-build's test location is inside the workspace's target directory

badboy:


chutten (approved):
This looks sensible: setting `CARGO_TARGET_DIR` to be the package root (which is two levels up from the manifest).

And there's two minor or patch updates to related deps riding along.

...but why? What made the status quo unbearable? There's some info in the commit messages suggesting that we're interested in a venv lock and also "ignoring"... but what venv and which ignoring?

badboy:
> This looks sensible: setting `CARGO_TARGET_DIR` to be the package root (which is two levels up from the manifest).
> 
> And there's two minor or patch updates to related deps riding along.
> 
> ...but why? What made the status quo unbearable? There's some info in the commit messages suggesting that we're interested in a venv lock and also "ignoring"... but what venv and which ignoring?

fine points.
the old situation is not unbearable, it's actually all good and no bug or anything like that.
But I noticed as I was trying out [`jj`](https://martinvonz.github.io/jj/), which flagged `glean-core/build/target` as modified files. It's actually ignored by git.
But that made me realize that it puts stuff in there instead of the shared target directory as all other cargo stuff does.
Then I noticed that I added new stuff to xshell-venv a while ago but never released it.

So really this is just a tiny cleanup.
I could -- should -- put all that info properly into the commit message I guess.

PR #2756, opened 2024-02-28
Bug 1866559 - Support "no-info"-configured pings

chutten:
This is a prerequisite for OHTTP support. Brings in glean_parser v13 to support it.

Only Rust's uploader does anything with this information. I suppose we could log/throw/record data if a noinfo ping makes it into an unsupported LB's uploader, but I figured just sending it normally might be the more kind approach.

Lotta changed files in this one, I'm afraid. If you'd like to follow the code changes chronologically by how the data flows, the individual commits are arranged in that manner (though Github's review UI is shite at handling that, so my condolences).

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,7 +3,7 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v57.0.0...main)
 
 * General
-  * Update `glean_parser` to v12.0.1 ([release notes](https://github.com/mozilla/glean_parser/releases/tag/v12.0.0))
+  * Update `glean_parser` to v13.0.0 ([release notes](https://github.com/mozilla/glean_parser/releases/tag/v13.0.0))
 * Rust
   * New metric type: Object ([#2489](https://github.com/mozilla/glean/pull/2489))
 * Android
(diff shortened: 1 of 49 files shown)


CHANGELOG.md
```diff
@@ -3,9 +3,11 @@
 [Full changelog](https://github.com/mozilla/glean/compare/v57.0.0...main)
 
 * General
-  * Update `glean_parser` to v12.0.1 ([release notes](https://github.com/mozilla/glean_parser/releases/tag/v12.0.0))
+  * Update `glean_parser` to v12.0.1 ([release notes](https://github.com/mozilla/glean_parser/releases/tag/v12.0.1))
+  * Update `glean_parser` to v13.0.0 ([release notes](https://github.com/mozilla/glean_parser/releases/tag/v13.0.0))
```
travis79, line 7: I think I would prefer to update the existing line to show the version in the release rather than having two separate glean_parser versions listed here.

travis79 (approved):
r+wc and passing tests

CHANGELOG.md
chutten: okily dokily

travis79 (approved):
I must say I rather like how the refactoring of the uploader code came out.

PR #2763, opened 2024-03-05
Python: Switch to ruff for linting and formatting

badboy:


pyproject.toml
```diff
@@ -36,12 +36,8 @@ module-name = "glean._uniffi"
 bindings = "uniffi"
 manifest-path = "glean-core/bundle/Cargo.toml"
 
-[tool.flake8]
-max-line-length = 100
-exclude = ["glean-core/python/glean/_uniffi/glean.py"]
-ignore = [
-    "W503" # line break before binary operator - black formats it that way
-]
```
Dexterp37, line 44: Why aren't we migrating `max-line-length` and `ignore`?

badboy: line-length I just forgot.
ignore because now the same tool lints and formats and so there's no disagreement about which way is right.

Dexterp37 (approved):

PR #2764, opened 2024-03-06
Bug 1881297 - Single Sample accumulation for distribution methods that don't already have it.

perrymcmanis144:
This is a go at adding a method for single sample accumulation.

Does not touch Memory Dist as it already has one: https://github.com/mozilla/glean/blob/main/glean-core/src/metrics/memory_distribution.rs#L82

[X] impl
[X] test
[ ] docs


CHANGELOG.md
```diff
@@ -2,6 +2,13 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v58.0.0...main)
 
+* Rust
+  * BREAKING CHANGE: Timing distribution and Custom distribution traits now expose `accumulate_single_sample`. Consumers that make use of the trait will need to implement the new functions ([Bug 1881297](https://bugzilla.mozilla.org/show_bug.cgi?id=1881297))
```
travis79, line 6: The only consumer that should be implementing this trait is the RLB, right?

docs/user/reference/metrics/timing_distribution.md
```diff
@@ -276,17 +276,94 @@ void onPageLoaded(Event e) {
...
-    Pages.pageLoad.accumulateSamples(samples)
+class PageHandler:
+    def on_page_loaded(self, event):
+        metrics.pages.page_load.accumulate_samples(samples)
+```
+
+</div>
+<div data-lang="Rust" class="tab">
+
+This API is not currently exposed in Rust, see [Bug 1829745](https://bugzilla.mozilla.org/show_bug.cgi?id=1829745).
```
travis79, line 294: Are you still planning on updating this since we determined it is now implemented for the RLB?

travis79 (commented):
I made a complete pass on this and had just a couple more documentation suggestions/questions. Otherwise this is looking good to me on the implementation.

CHANGELOG.md
perrymcmanis144: The last time we did this, it came with this message. If it's not a concern, I'll gladly remove it :D

travis79 (approved):

glean-core/android/src/test/java/mozilla/telemetry/glean/private/TimingDistributionMetricTypeTest.kt
```diff
@@ -182,6 +182,17 @@ class TimingDistributionMetricTypeTest {
         assertEquals(1L, snapshot.values[984625593])
         assertEquals(1L, snapshot.values[1969251187])
         assertEquals(1L, snapshot.values[2784941737])
+
+        // Assure the single sample API properly records.
+        metric.accumulateSingleSample(4L)
+
+        // Check that this new data was properly recorded in the second ping.
+        val snapshotTwo = metric.testGetValue("store1")!!
```
travis79, line 190: This is just a nit, but you could have reused the `snapshot` above by changing it from a `val` to a`var`.  This is perfectly fine as is, I'm not too worried about a new instantiation of a variable in a test by any means :)

travis79 (approved):

PR #2767, opened 2024-03-07
Bug 1875814 - Default-enable wall-clock event timestamps

badboy:


travis79 (approved):
r+ with a CHANGELOG entry please :)

PR #2778, opened 2024-03-15
Bug 1877830 - Keep Glean Package.swift iOS version in sync with supported version in Glean core

travis79:


docs/dev/ios/upgrading-supported-ios-platform.md
```diff
@@ -0,0 +1,12 @@
+# Upgrade the supported Xcode Version
+
+When it is necessary to update the supported Xcode version in Glean, the bin/update-xcode-version.sh script
```
badboy, line 3:
```suggestion
When it is necessary to update the supported Xcode version in Glean, the `bin/update-xcode-version.sh` script
```

badboy (approved):

PR #2786, opened 2024-03-25
Bug 1855257 - Add a configuration option to disable internal pings

badboy:
This will mark the internal pings metrics, baseline and events as
disabled and will not send them.
Data recorded for these pings is still recorded, but never cleared out.

This setting should only be used for apps that definitely do not want
any of these pings and do not record any metrics of their own into those
pings.

---

Ok, I came up with another idea how to implement this: we finally introduce the `enabled` flag on pings (but only internally!).
We then set it to `false` when asked to do and check it along with the normal flow.
This PR is not ready for review yet (code duplication, no testing, etc), but putting it up to see if that's the direction we want it to. 
I also will take this and try it on m-c for the linked bug.

### DIFF
--- a/glean-core/src/internal_pings.rs
+++ b/glean-core/src/internal_pings.rs
@@ -19,9 +19,9 @@ pub struct InternalPings {
 }
 
 impl InternalPings {
-    pub fn new() -> InternalPings {
+    pub fn new(enabled: bool) -> InternalPings {
         InternalPings {
-            baseline: PingType::new(
+            baseline: PingType::new_internal(
                 "baseline",
                 true,
                 true,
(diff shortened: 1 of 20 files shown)


badboy:
ah right, it also doesn't work for anything but Rust

glean-core/src/internal_pings.rs
```diff
@@ -19,9 +19,9 @@ pub struct InternalPings {
 }
 
 impl InternalPings {
-    pub fn new() -> InternalPings {
+    pub fn new(enabled: bool) -> InternalPings {
         InternalPings {
-            baseline: PingType::new(
+            baseline: PingType::new_internal(
```
travis79, line 24: Add another layer to the 🧅!

travis79 (commented):
I have no opposition to this approach, provided tests pass, etc.

glean-core/src/internal_pings.rs
badboy: yeah, I don't like it really, but due to how we handle those pings with an `Arc` internally it's hard to modify their `enabled` value otherwise.

travis79 (approved):

PR #2791, opened 2024-04-04
Bug 1804711 - Add schedules_pings property to pings so that pings can be scheduled to be sent alongside other pings.

mikeconley:
This is for https://bugzilla.mozilla.org/show_bug.cgi?id=1804711.

This relies on https://github.com/mozilla/glean_parser/pull/682.

What this does it make it possible for a Glean ping to declare that it wants to be sent when another ping is sent. Supposing some product wanted to sent a ping as frequently as the baseline ping, then that ping could use the `ping_schedule` metadata property and list `baseline`.

This pull request is the part that takes the reverse-mapping that glean_parser creates, and actually does the lookup for other scheduled pings when a ping is sent.

I note this in one of the commit messages, but I think it's worth calling out here too:

While glean_parser will try to prevent the base-case of pings trying to schedule _themselves_ when being sent, nothing in this stack takes care of the problem of recursive cycles of pings.

So if we have PingA scheduled to be sent when PingB is sent, and PingB is scheduled to be sent when PingC is sent, and PingC is scheduled to be sent when PingA is sent... things get messy indeed, and I think we might get stuck in a ping-sending loop.

### DIFF
--- a/glean-core/src/metrics/ping.rs
+++ b/glean-core/src/metrics/ping.rs
@@ -138,7 +138,7 @@ impl PingType {
         self.0.include_info_sections
     }
 
-    pub(crate) fn schedules_pings(&self) -> &Vec<String> {
+    pub(crate) fn schedules_pings(&self) -> &[String] {
         &self.0.schedules_pings
     }
 
@@ -259,6 +259,17 @@ impl PingType {
                     "The ping '{}' was submitted and will be sent as soon as possible",
                     ping.name
                 );
+
+                if !ping.schedules_pings.is_empty() {
+                    log::info!(
+                        "The ping '{}' is being used to schedule other pings: {:?}",
+                        ping.name,
+                        ping.schedules_pings
+                    );
+                    for scheduled_ping_name in &ping.schedules_pings {
+                        glean.submit_ping_by_name(scheduled_ping_name, reason);
+                    }
+                }
 
                 true
             }
--- a/glean-core/src/metrics/ping.rs
+++ b/glean-core/src/metrics/ping.rs
@@ -29,6 +29,8 @@ struct InnerPing {
     pub precise_timestamps: bool,
     /// Whether to include the {client|ping}_info sections on assembly.
     pub include_info_sections: bool,
+    /// Other pings that should be scheduled when this ping is sent.
+    pub schedules_pings: Vec<String>,
     /// The "reason" codes that this ping can send
     pub reason_codes: Vec<String>,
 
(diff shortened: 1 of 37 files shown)


mikeconley:
So at least some of these failures are due to the fact that the CI isn't using the glean_parser patch in https://github.com/mozilla/glean_parser/pull/682. I'm not sure what to do about that.

glean-core/src/metrics/ping.rs
```diff
@@ -130,6 +138,10 @@ impl PingType {
         self.0.include_info_sections
     }
 
+    pub(crate) fn schedules_pings(&self) -> &Vec<String> {
```
badboy, line 141:
```suggestion
    pub(crate) fn schedules_pings(&self) -> &[String] {
```

it's unusual to return a reference to a vec.

glean-core/src/metrics/ping.rs
```diff
@@ -248,6 +260,17 @@ impl PingType {
...
                 );
 
+                if !ping.schedules_pings.is_empty() {
+                    log::info!(
+                        "The ping '{}' is being used to schedule other pings: {:?}",
+                        ping.name,
+                        ping.schedules_pings
+                    );
+                    for scheduled_ping_name in &ping.schedules_pings {
+                        glean.submit_ping_by_name(scheduled_ping_name, reason);
```
badboy, line 270: Hm, that means we also use the same reason. The reason for the actual submitted ping might not be specified for the linked ping.
We should at the very least document that related pings should use the same reasons (this might be done in a followup).

badboy (commented):
time 'til someone creates a ping cycle: .... 

kidding, but it technically is a risk.
We should definitely come up with some documentation around this. 

(oh and one note: argh, so many little changes necessary for a new parameter. not your fault though!)

badboy:
> So at least some of these failures are due to the fact that the CI isn't using the glean_parser patch in [mozilla/glean_parser#682](https://github.com/mozilla/glean_parser/pull/682). I'm not sure what to do about that.

That's fine, we can review the code and once we're happy we do the glean_parser release and pull that into this PR as well (it's an unfortunate shortcoming of the repo split)

travis79 (commented):
This all looks fine to me, I'm looking at glean_parser changes next but provided we add the documentation requested and a changelog entry, I think this should be fine. I'm headed to take a look at the glean_parser changes next.

badboy:
Merged as part of https://github.com/mozilla/glean/pull/2792

PR #2792, opened 2024-04-05
Bug 1889114 - Updating Server Knobs to work with pings

travis79:
This will require the Glean Parser patch here: https://github.com/mozilla/glean_parser/pull/681


docs/user/user/pings/data-control-plane/experimenter-configuration.md
```diff
@@ -0,0 +1,23 @@
+# Experimenter Configuration
+
+The structure of this configuration is a key-value collection with the full metric identification of the Glean metric serving as the key in the format <metric_category.metric_name>.
+
+The values of the key-value pair are booleans which represent whether the metric is enabled (`true`) or not (`false`).
+
+In the example below `gleanMetricConfiguration` is the name of the variable defined in the Nimbus feature.
+
+This configuration would be what is entered into the branch configuration setup in Experimenter when defining an experiment or rollout.
```
chutten, line 9: This appears to be describing only the `metrics_enabled` substructure of `gleanMetricConfiguration`, not the entirety (which I'm given to understand has both a `metrics_enabled` and a `ping_enabled` portion)

glean-core/android/src/test/java/mozilla/telemetry/glean/GleanTest.kt
```diff
@@ -987,29 +987,16 @@ class GleanTest {
...
-        Glean.setMetricsEnabledConfig(metricConfig)
+        Glean.setRemoteSettingsConfig(metricConfig)
 
         // This should result in the metric being set to "foo"
         stringMetric.set("foo")
         assertNotNull(stringMetric.testGetValue())
         assertEquals("foo", stringMetric.testGetValue())
-
-        // Set a metric configuration which will disable the telemetry.string_metric
-        // again, this time using the deprecated API to ensure backwards compatibility
```
chutten, line 1001: Please call out the removal of the deprecated API by name in the CHANGELOG.

glean-core/src/core/mod.rs
```diff
@@ -758,19 +758,40 @@ impl Glean {
             .get_value(self, None)
     }
 
-    /// Set configuration to override the default metric enabled/disabled state, typically from a
+    /// Set configuration to override the default state, typically initiated from a
     /// remote_settings experiment or rollout
```
chutten, line 762: "override the default state"... of what? And how?

glean-core/src/core/mod.rs
```diff
@@ -758,19 +758,40 @@ impl Glean {
...
     ///
     /// # Arguments
     ///
-    /// * `json` - The stringified JSON representation of a `MetricsEnabledConfig` object
-    pub fn set_metrics_enabled_config(&self, cfg: MetricsEnabledConfig) {
-        // Set the current MetricsEnabledConfig, keeping the lock until the epoch is
+    /// * `json` - The stringified JSON representation of a `RemoteSettingsConfig` object
+    pub fn set_remote_settings_config(&self, cfg: RemoteSettingsConfig) {
+        // Set the current RemoteSettingsConfig, keeping the lock until the epoch is
         // updated to prevent against reading a "new" config but an "old" epoch
```
chutten, line 769: _software architect hat on_ To me this suggests that the lock should cover both the epoch and the config, since we want to forbid using one without the other. Might be a follow-up.

glean-core/src/lib_unit_tests.rs
```diff
@@ -1203,3 +1213,48 @@ fn internal_pings_can_be_disabled() {
...
+
+    let dir = tempfile::tempdir().unwrap();
+    let (mut glean, _t) = new_glean(Some(dir));
+
+    let disabled_ping = PingType::new("custom-disabled", true, false, true, true, vec![], false);
+    glean.register_ping_type(&disabled_ping);
+    let enabled_ping = PingType::new("custom-enabled", true, false, true, true, vec![], true);
+    glean.register_ping_type(&enabled_ping);
+
+    // We need to store a metric as an empty ping is not stored.
```
chutten, line 1229: Even if you change the second boolean to `true` in the ctor?

glean-core/src/metrics/remote_settings_config.rs
```diff
@@ -16,25 +16,29 @@ use serde::{Deserialize, Serialize};
...
     /// the default of the metric will be overriden and set to the
     /// enabled state.
-    #[serde(flatten)]
-    pub metrics_enabled: HashMap<String, bool>,
+    pub metrics_enabled: Option<HashMap<String, bool>>,
+
+    /// This is a `HashMap` consisting of ping names as keys and
+    /// boolean values representing on override for the default
+    /// enabled state of the ping of the same name.
+    pub pings_enabled: Option<HashMap<String, bool>>,
```
chutten, line 31: If I remember my `std::collections` rules of thumb correctly, in most cases we're to not bother with `Option<ACollectionType>` and instead use emptiness (since empty collections do not allocate). As a bonus it ought to make the metrics setting code quite a lot simpler.

samples/rust/pings.yaml
```diff
@@ -20,3 +20,4 @@ prototype:
     - N/A
   notification_emails:
     - CHANGE-ME@example.com
+  enabled: true
```
chutten, line 23: If it's optional, we can omit.

travis79: Yep, this was an artifact from some of my testing that I overlooked when cleaning things up.

glean-core/src/metrics/remote_settings_config.rs
travis79: Then I would need an intermediate representation to assume the serde deserialization. Here Optional is also letting serde know that this may or may not be present in the configuration that is passed through the API.

glean-core/src/lib_unit_tests.rs
travis79: I think I was intending to extend this to ensure that the metric was cleared when `submit()` was called on the disabled ping, but I think I forgot to come back and do that. Let me do that now, lest I forget again.

travis79: Or remove it and test that in a separate test

glean-core/src/core/mod.rs
travis79: This may eventually grow to encompass other settings, I made this comment generic and non-specific to allow for that and not create another place that needed kept in sync. The RemoteSettingsConfig seemed a better place to document the what (if not the how) a little closer to the source.

docs/user/user/metrics/data-control-plane/product-integration.md
```diff
@@ -80,7 +80,7 @@ features:
 Once the Feature Variable has been defined, the final step is to fetch the configuration from Nimbus and supply it to the Glean API. This can be done during initialization and again any time afterwards, such as in response to receiving an updated configuration from Nimbus. Only the latest configuration provided will be applied and any previously configured metrics that are omitted from the new configuration will not be changed. An example call to set a configuration from the “homescreen” Nimbus Feature could look like this:
 
 ```Swift
-Glean.setMetricsEnabledConfig(FxNimbus.features.homescreen.value().metricsEnabled)
+Glean.setRemoteSettingsConfig(FxNimbus.features.homescreen.value().metricsEnabled)
```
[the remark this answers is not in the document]
travis79: I can see where there may be some confusion. I (re)named this to reflect the _source_ of the the config that was to be supplied here, but it might be misconstrued that this is somehow setting some internal remote-settings setup?

Let me ponder on what might be a more concise name for this...

glean-core/src/metrics/remote_settings_config.rs
chutten: Can you go into more detail? I assume `#[serde(default)]` solves for configurations that are missing `pings_enabled` or `metrics_enabled`?

chutten: ref. https://serde.rs/field-attrs.html#default

docs/user/user/metrics/data-control-plane/product-integration.md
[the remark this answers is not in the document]
travis79: I've settled on an entirely different verb and a internally focused noun to disambiguate this from RemoteSettings: `applyServerKnobsConfig`

docs/user/user/pings/data-control-plane/experimenter-configuration.md
```diff
@@ -0,0 +1,23 @@
...
+```json
+{
+  "gleanMetricConfiguration": {
+    "pings_enabled": {
+      "baseline": false,
+      "events": false,
+      "metrics": false
+    }
+  }
+}
```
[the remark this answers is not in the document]
travis79: The metrics are shown in this example in the Server Knobs for metrics section, I narrowed the scope here to avoid someone thinking that it might be necessary to include both sections.

docs/user/user/pings/data-control-plane/experimenter-configuration.md
travis79: Either way the configuration is entered into the branch payload in Experimenter. Is there some specific wording here that would help to make this more clear to you?

docs/user/user/pings/data-control-plane/product-integration.md
```diff
@@ -0,0 +1,8 @@
+# Product Integration
+
+Glean provides a general Nimbus feature named `glean` that can be used for configuration of pings. Simply select the `glean` feature along with your Nimbus feature from the Experimenter UI when configuring the experiment or rollout (see https://experimenter.info for more information on multi-feature experiments).
+
+The `glean` Nimbus feature requires the `gleanMetricConfiguration` variable to be used to provide the require metric configuration. The format of the configuration is defined in the [Experimenter Configuration](./experimenter-configuration.md) section. If a ping is not included, it will default to the value found in the pings.yaml.
```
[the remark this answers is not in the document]
travis79: Not directly, internal pings disabled with the builtin_pings switch would still be diabled, but you could use Server Knobs PE (Pings Edition) to override that setting

docs/user/user/pings/data-control-plane/experimenter-configuration.md
chutten: "the full metric identification of the Glean metric serving as the key...." is metrics stuff, not pings stuff, right?

docs/user/user/pings/data-control-plane/product-integration.md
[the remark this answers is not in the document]
chutten: Okay, so we might want to throw a comment to that effect here. Though if SKPE (pronounced "skippy") doesn't have a use case to actually be used against builtins... maybe we forbid that? ...I guess it's a fair thing to leave in, in retrospect, since we might want to turn of "baseline" and might as well use SKPE to do it at the speed of Nimbus while waiting for `disable_builtin_pings` to ride trains...

travis79:
Ready for you to take another look, @chutten.  Then I can see about merging the conflicting changes from #2791 into this and merging those changes along with these, on top of the glean_parser version update needed to make tests pass.

docs/user/user/pings/data-control-plane/experimenter-configuration.md
```diff
@@ -0,0 +1,21 @@
+# Experimenter Configuration
+
+The structure of this configuration is a key-value collection with the name of the Glean ping serving as the key and the values are booleans representing whether the ping is enabled (`true`) or not (`false`).
```
chutten, line 3: Grammar nit: plural disagreement "ping serving as the key" "values are booleans"

docs/user/user/pings/data-control-plane/index.md
```diff
@@ -0,0 +1,20 @@
+# Data Control Plane (a.k.a. Server Knobs)
```
chutten, line 1: Probably worth cross-linking to and from `docs/user/user/metrics/data-control-plane`, huh

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -221,4 +226,54 @@ class CustomPingTest {
...
+
+        // Define a new custom ping inline.
+        val customPing = PingType<NoReasonCodes>(
+            name = "custom-ping",
+            includeClientId = true,
+            sendIfEmpty = true,
+            preciseTimestamps = true,
+            includeInfoSections = true,
+            reasonCodes = emptyList(),
+            enabled = true,
```
chutten, line 250: I was about to be confused that the order didn't need changing here, but I forgot that Kotlin has named args.

glean-core/ios/GleanTests/Net/BaselinePingTests.swift
```diff
@@ -178,4 +178,46 @@ final class BaselinePingTests: XCTestCase {
...
+}
+"""
+        Glean.shared.applyServerKnobsConfig(metricConfigStringifiedJson)
+
+        // Resetting Glean doesn't trigger lifecycle events in tests so we must call the method
+        // invoked by the lifecycle observer directly. We must also reset the `isActive` flag in
+        // order to have the correct state to test.
+        Glean.shared.isActive = false
+        Glean.shared.handleForegroundEvent()
+        waitForExpectations(timeout: 5.0) { error in
```
chutten, line 219: Ugh, so we wait for 5 whole seconds to ensure nothing happens? Oof.

glean-core/python/glean/metrics/ping.py
```diff
@@ -17,6 +17,7 @@ def __init__(
         send_if_empty: bool,
         precise_timestamps: bool,
         include_info_sections: bool,
+        # enabled: bool,  -- Not currently enabled for python bindings
```
chutten, line 20: Should we take any steps to ensure no one's trying to disable this on Python?

glean-core/rlb/src/private/ping.rs
```diff
@@ -27,13 +27,18 @@ impl PingType {
     /// * `name` - The name of the ping.
     /// * `include_client_id` - Whether to include the client ID in the assembled ping when.
     /// * `send_if_empty` - Whether the ping should be sent empty or not.
+    /// * `precise_timestamps` - Whether the ping should use precise timestamps for the start and end time.
+    /// * `include_info_sections` - Whether the ping should include the client/ping_info sections.
     /// * `reason_codes` - The valid reason codes for this ping.
+    /// * `enabled` - Whether or not this ping is enabled. Note: Data that would be sent on a disabled
```
chutten, line 33: Should probably move this over `reason_codes`

glean-core/src/core/mod.rs
```diff
@@ -758,19 +758,26 @@ impl Glean {
...
-    /// Set configuration to override the default metric enabled/disabled state, typically from a
+    /// Set configuration to override the default state, typically initiated from a
     /// remote_settings experiment or rollout
     ///
     /// # Arguments
     ///
-    /// * `json` - The stringified JSON representation of a `MetricsEnabledConfig` object
-    pub fn set_metrics_enabled_config(&self, cfg: MetricsEnabledConfig) {
-        // Set the current MetricsEnabledConfig, keeping the lock until the epoch is
+    /// * `json` - The stringified JSON representation of a `RemoteSettingsConfig` object
```
chutten, line 766: There is no argument called `json`

glean-core/src/internal_pings.rs
```diff
@@ -21,53 +21,54 @@ pub struct InternalPings {
...
                 ],
-                enabled,
             ),
             deletion_request: PingType::new(
                 "deletion-request",
                 true,
                 true,
                 true,
                 true,
+                true, // The deletion-request should not be disabled
```
chutten, line 71: Hrm, where else can we catch this... Guess we can't put this logic in Experimenter where it'd be most helpful?

glean-core/src/lib_unit_tests.rs
```diff
@@ -1203,3 +1208,38 @@ fn internal_pings_can_be_disabled() {
...
+    )
+    .to_string();
+    glean
+        .apply_server_knobs_config(RemoteSettingsConfig::try_from(remote_settings_config).unwrap());
+
+    assert!(disabled_ping.submit_sync(&glean, None));
+    assert!(!enabled_ping.submit_sync(&glean, None));
+
+    // assert!(counter.test_get_value(Some("custom-disabled".into())).is_some());
+    // assert!(counter.test_get_value(Some("custom-enabled".into())).is_none());
```
chutten, line 1244: some leftovers?

glean-core/src/metrics/ping.rs
```diff
@@ -64,43 +62,28 @@ impl PingType {
     /// * `name` - The name of the ping.
     /// * `include_client_id` - Whether to include the client ID in the assembled ping when submitting.
     /// * `send_if_empty` - Whether the ping should be sent empty or not.
+    /// * `precise_timestamps` - Whether the ping should use precise timestamps for the start and end time.
+    /// * `include_info_sections` - Whether the ping should include the client/ping_info sections.
     /// * `reason_codes` - The valid reason codes for this ping.
+    /// * `enabled` - Whether or not this ping is enabled. Note: Data that would be sent on a disabled
```
chutten, line 68: Ditto this comment above `reason_codes`

glean-core/src/metrics/ping.rs
```diff
@@ -208,47 +198,59 @@ impl PingType {
...
-                        json_body: content,
-                        headers: Some(ping.headers),
-                        body_has_info_sections: self.0.include_info_sections,
-                        ping_name: self.0.name.to_string(),
-                    };
-
-                    glean.upload_manager.enqueue_ping(glean, ping);
-                    return true;
-                }
+                if self.enabled(glean) {
```
chutten, line 201: It might be the m-c C++ style guide in my head, but I think I'd prefer this to take the form of an early return rather than a 52-line context.

samples/rust/pings.yaml
```diff
@@ -19,4 +19,4 @@ prototype:
   data_reviews:
     - N/A
   notification_emails:
-    - CHANGE-ME@example.com
+    - CHANGE-ME@example.com
```
chutten, line 22: Looks like an unintentional change here.

chutten (changes requested):
We're nearly there!

docs/user/user/metrics/data-control-plane/experimenter-configuration.md
```diff
@@ -13,9 +13,11 @@ This configuration would be what is entered into the branch configuration setup
 ```json
 {
   "gleanMetricConfiguration": {
-    "urlbar.abandonment": true,
-    "urlbar.engagement": true,
-    "urlbar.impression": true
+    "metrics_enabled": {
```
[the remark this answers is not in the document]
travis79: The variable name `gleanMetricsConfiguration` is easily changed, but right now this matches what's defined in the apps, so I'll leave it until we update those to be more in line with a more generic use of Server Knobs. Ultimately, I'd like to rename the variable to `glean_configuration` so then the casing and concepts will better match.  I'll file a bug to remind myself and assume that satisfies your concern here.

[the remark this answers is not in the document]
travis79: Bug Filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1891567

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
travis79: Yep, this didn't break when I built so it didn't register. I'll update for consistency.

glean-core/ios/GleanTests/Net/BaselinePingTests.swift
travis79: No, it happens almost immediately since the expectation is `fullfill`ed above. If this were more async, and awaiting pings to be submitted and uploaded, then it would time out if `fullfill()` wasn't called in the timeout here.

glean-core/python/glean/metrics/ping.py
travis79: I don't think it's too particularly risky as we don't have any Python consumers using Server Knobs or planning on using Server Knobs that I know of. The first use-case will likely be Cirrus and I'll be working closely with them on that so that's probalby when this will be enabled for Python.

glean-core/src/internal_pings.rs
travis79: We can incorporate a validation for this in the Server Knobs "configurator" tooling that is coming eventually. I'll add a note to the relevant bug.

glean-core/python/glean/metrics/ping.py
chutten: But can Python consumers disable a ping through `enabled: false` in the YAML?

glean-core/src/metrics/ping.rs
chutten: I don't mean completely early, just taking this line and appending `return false; }` after it instead of indenting the following 5 handfuls of lines?

glean-core/python/glean/metrics/ping.py
travis79: I should have stripped this from the output params, so it could right now but it would be easy enough to correct. I'll file a bug and fix that before we cut a glean parser release, but again, I don't see anyone using python making use of this property anytime soon.

travis79: Or I can just enable it here, either way

travis79: So, the _loader.py doesn't pass the `enabled` property to the constructor, so even if it found one in the pings.yaml file for a python Glean app, it wouldn't do anything with it. I think it's safe to leave as is for now, until time to enable this for python.

glean-core/src/metrics/ping.rs
travis79: Okay, inverted the if to check for the early out to reduce nesting depth.

travis79:
@chutten, I think this is ready for a final look, then I'll merge mike's PR into this and update glean_parser

chutten (approved):
r+wc, it appears as though the kotlin named arg reordering and pings.rs comments reordering (in both pings.rs-es) and the default `enabled: true` are still hanging on. It can land with those there, but I thought we'd settled on resolving those.

travis79:
Okay, I've fixed up the things missing from the latest commits and this should be "good to land".

Now, I'll merge #2791 into this branch, followed by updating the glean_parser version to the one with the changes associated with this and the other PR.

travis79:
Okay, this now has the changes from Mike's PR, and an update to glean_parser v14.0.1.  The kotlin/android build is still failing, due an unresolved reference to `ObjectSerialize`. I expect that's something that will be resolved with @badboy's inbound patch for ObjectMetricType stuff in glean-core.

badboy:
rebased mine on top: https://app.circleci.com/pipelines/github/mozilla/glean/12171/workflows/f9641ae6-60fd-41e2-a6bf-e17d1807da15

Looks all good, but your PR needs one more patch:

```patch
--- a/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
+++ b/glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
@@ -249,6 +249,7 @@ class CustomPingTest {
             sendIfEmpty = true,
             preciseTimestamps = true,
             includeInfoSections = true,
+            schedulesPings = emptyList()
             reasonCodes = emptyList(),
             enabled = true,
         )
```

We should land this then and follow up with my PR. CI will be broken for a moment.

glean-core/android/src/test/java/mozilla/telemetry/glean/pings/CustomPingTest.kt
```diff
@@ -221,4 +229,54 @@ class CustomPingTest {
...
```
badboy, line 253:
```suggestion
            enabled = true,
            schedulesPings = emptyList(),
            reasonCodes = emptyList(),
```

badboy: Ah, finally loaded so I could send this as an actual suggestion

PR #2800, opened 2024-04-16
Bug 1775420 - Glean Gradle plugin: Upgrade to Miniconda 24.3.0 (and Python 3.11)

badboy:
The super-old version of Miniconda we were using included only Python 3.7 Python 3.7 is EoL and glean_parser dropped support for it.

---

Tested on my macbook and on my Linux machine.


gradle-plugin/src/main/groovy/mozilla/telemetry/glean-gradle-plugin/GleanGradlePlugin.groovy
```diff
@@ -53,7 +53,7 @@ class GleanPlugin implements Plugin<Project> {
     private String GLEAN_PARSER_VERSION = "13.0"
     // The version of Miniconda is explicitly specified.
     // Miniconda3-4.5.12 is known to not work on Windows.
-    private String MINICONDA_VERSION = "4.5.11"
+    private String MINICONDA_VERSION = "24.3.0-0"
```
travis79, line 56: Woohoo!

travis79 (approved):

PR #2817, opened 2024-04-30
Bug 1893207 - CI: Switch to M1 medium workers

badboy:
These are twice as costly per usage minute.
macos.x86.medium.gen2 will be gone in June 2024.
Maybe the raw performance makes up for the increase in costs.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -506,14 +506,14 @@ jobs:
       - setup-rust-toolchain
       - restore_cache:
           name: Restore rubygems cache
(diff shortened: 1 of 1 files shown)


badboy:
I triggered the one part that's not run by default: the macos pypi release part: https://app.circleci.com/pipelines/github/mozilla/glean/12215/workflows/adcb09ff-428a-41dd-8191-071f510a070d/jobs/236006

Looks all good!

travis79 (approved):

PR #2818, opened 2024-04-30
event database: Use a Mutex instead of a RwLock

badboy:
We use this lock to synchronize the file access outside of it.
Thus we always need an exclusive lock, which is why we always used a `write` lock.
But if that's all we need a `Mutex` is really just good enough.
And newer Clippy was complaining about that


chutten (approved):
A note: this PR also contains another lintfix for changing deprecated `max_value()` calls to `MAX` constants,

PR #2819, opened 2024-05-02
Bug 1894595 - Python: Depend on setuptools

badboy:
This fixes the issue of missing `pkg_resources` on Python 3.12


badboy:
#2820 is the better fix.

PR #2820, opened 2024-05-02
Bug 1894635 - Python: Use importlib.metadata instead of deprecated pkg_resources

badboy:
importlib.metadata is available since Python 3.8, the minimum Python we support.


badboy:
Doing the same for glean_parser: https://github.com/mozilla/glean_parser/pull/699

PR #2821, opened 2024-05-03
Revert "Python: Use importlib.metadata instead of deprecated pkg_reso…

badboy:
…urces"

This reverts commit b384d2cfdfa7117e4e5f923af1171c90e9d5b3d7.

--

let's see about those windows python failures in CI


badboy:
still broken, `./bin/build-win.sh` works on my machine.
good: it's not importlib's fault.
bad: I now need to debug that on CI.

PR #2823, opened 2024-05-03
CI: Build a bcryptprimitives shim for loading it in Python

badboy:
This brings along a minimal shim for bcryptprimitives.dll. It's really minimal and _only_ necessary for the Wine build.
How I got there? Well, I found https://github.com/rust-lang/rust/commit/07633821ed63360d4d7464998c29f4f588930a03
I'm not even sure if we need the `ProcessPrng` implementation, but I'm not here to find out.

What didn't work? Setting `WINEPATH` to the directory. It still won't find the library. Preloading it once though fixes it all.

It now seems to work in CI again.
And it does work in a Docker container as well.
The reason it worked on my machine: On Fedora wine brings along the bcryptprimitives.dll

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -749,8 +749,15 @@ jobs:
             wget https://www.python.org/ftp/python/3.8.2/python-3.8.2-embed-amd64.zip
             mkdir winpython
             unzip python-3.8.2-embed-amd64.zip -d winpython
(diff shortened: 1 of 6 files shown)


travis79 (approved):
Oh wow... this is crazy. Thanks for doing this.

PR #2836, opened 2024-05-14
Remove double-slash in doc link

badboy:
GitHub doesn't normalize a double-slash, and thus the link doesn't work correctly.

[doc only]


travis79 (approved):
whoops, thanks! 😊

PR #2845, opened 2024-05-21
Bug 1896193 - Collect samples of Rkv commit timing

badboy:
Due to this being deep within the database we need to do things slightly differently.

We will measure the time ourselves and keep the samples. When we later assemble a ping we record those samples into the databse (incidentally triggering _another_ write for now). Then it just goes the usual way.

Maybe this should be optimized:

* Only collect the samples when actually assembling the `metrics` ping
* Don't put those values into the database, just add them to the generated data right away.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,7 @@
 
 * General
   * Bump the string length limit to 255 characters ([#2857](https://github.com/mozilla/glean/pull/2857))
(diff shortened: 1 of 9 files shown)


badboy:
I updated this and added a test.
I think this can land, at least we can ship it for a while and look at the data.

travis79:
I guess we can wait and see if we can use the data before we add it to the monitoring :)

badboy:
> Is there a bug for, or does the scope of this bug include adding this to the monitoring dashboards?

I can definitely start preparing a query at the very least.

PR #2848, opened 2024-05-24
Bug 1898336 - (Memory|Timing) Distribution: snapshot only contains non-zero buckets now

badboy:
**BREAKING CHANGE**:
This changes the ping payload for memory/timing distributions.

---

I _think_ this is all it would require on the client-side to do.


travis79 (approved):
That cleans things up nicely.

PR #2854, opened 2024-05-31
Merge back release v60.3.0

badboy:
Let's get this out too, so we're in a consistent state.


badboy:
in progress: https://github.com/mozilla/glean/releases/tag/v60.3.0

PR #2857, opened 2024-06-03
Bug 1900313 - Bump the string length limit to 255 characters

badboy:

PR #2858, opened 2024-06-03
Bug 1900308 - Android: Delay log init until Glean is getting initialized

badboy:
This should avoid loading the native library early


badboy:
Woops, due to local rebasing shenanigans this was merged as part of #2857, including the docs.
Commit: 0f447e3ada1e445469577deb073d7add53b71351

PR #2871, opened 2024-06-17
Bug 1902992 - Flush database time-based when delayPingLifetimeIo=true

badboy:


badboy:
currently based on #2877, so will be rebased when that lands.

glean-core/src/database/mod.rs
```diff
@@ -173,9 +179,13 @@ use crate::Result;
 /// before data is flushed to disk.
 ///
 /// Only considered if `delay_ping_lifetime_io` is set to `true`.
-#[cfg(target_os = "android")]
+#[cfg(any(target_os = "android", feature = "auto_flush"))]
 const PING_LIFETIME_THRESHOLD: usize = 1000;
 
+// Save atleast every 2 seconds.
+#[cfg(any(target_os = "android", feature = "auto_flush"))]
+const PING_LIFETIME_MAX_TIME: Duration = Duration::from_millis(2000);
```
badboy, line 187: I chose this value arbitrarily, same as the 1000 before. Should we make this a config option by now?

glean-core/rlb/examples/ping-lifetime-flush.rs
```diff
@@ -0,0 +1,134 @@
...
+        app_build: env!("CARGO_PKG_VERSION").to_string(),
+        app_display_version: env!("CARGO_PKG_VERSION").to_string(),
+        channel: None,
+        locale: None,
+    };
+
+    glean::initialize(cfg, client_info);
+
+    // Wait for init to finish,
+    // otherwise we might be to quick with calling `shutdown`.
```
chutten, line 107:
```suggestion
    // otherwise we might be too quick with calling `shutdown`.
```

glean-core/src/database/mod.rs
```diff
@@ -173,9 +179,13 @@ use crate::Result;
 /// before data is flushed to disk.
 ///
 /// Only considered if `delay_ping_lifetime_io` is set to `true`.
-#[cfg(target_os = "android")]
+#[cfg(any(target_os = "android", feature = "auto_flush"))]
 const PING_LIFETIME_THRESHOLD: usize = 1000;
 
+// Save atleast every 2 seconds.
```
chutten, line 185:
```suggestion
// Save at least every 2 seconds.
```

glean-core/src/database/mod.rs
chutten: What kind of config? At-init runtime config? Or compile-time?

chutten (approved):

glean-core/src/database/mod.rs
badboy: at-init runtime means we could configure that from the fenix side without a glean release.
(Rust is not good at compile-time configuration options so I'd rather not come up with a method to do that right now)

badboy:
I ended up adding it as a top-level configuration in this branch: https://github.com/mozilla/glean/compare/1902992/flush-time-based-with-config?expand=1

It still has some TODOs, mostly docs, but I think that's a viable approach. But I won't get that done this week.

travis79 (approved):

PR #2877, opened 2024-06-26
CI: Run tests for all crates in the workspace

badboy:
This probably breaks right now.


glean-core/src/lib.rs
```diff
@@ -1131,10 +1131,19 @@ pub fn glean_test_destroy_glean(clear_stores: bool, data_path: Option<String>) {
...
+            {
+                storage
+                    .map(|storage| storage.persist_ping_lifetime_data())
+                    .is_some()
+            }
+
+            #[cfg(not(target_os = "android"))]
+            {
+                storage.is_some()
+            }
```
chutten, line 1146: This seems like it's liable to blow up in our faces. Why wouldn't we want to persist ping lifetime data on shutdown in tests? We [do it in real life](https://searchfox.org/glean/source/glean-core/src/lib.rs#690).

chutten (changes requested):

glean-core/src/lib.rs
badboy: Mostly to get the test fixed. But yes.
Unfortunately that means I might need to disable the test and potentially move it to one of the instrumentation tests instead.

chutten: Well, since this is test-only code how about we leave it as you wrote and we annotate it with the bug number of the follow-up for "Make this work in consistently" you're about to file?

badboy: too late

glean-core/rlb/examples/delayed-ping-data.rs
```diff
@@ -0,0 +1,126 @@
...
+pub static PrototypePing: Lazy<PingType> =
+    Lazy::new(|| PingType::new("prototype", true, true, false, true, true, vec![], vec![]));
+
+fn main() {
+    env_logger::init();
+
+    let mut args = env::args().skip(1);
+
+    let data_path = PathBuf::from(args.next().expect("need data path"));
+    let state: u32 = args.next().as_deref().unwrap_or("0").parse().unwrap();
```
chutten, line 86: I think I'd prefer it if `state` were a string: "accumulate_one_and_pretend_crash", "accumulate_ten_and_orderly_shutdown", "submit_ping"

chutten (approved):
r+wc

PR #2881, opened 2024-06-27
Correct argument ordering: first the time, then how far ahead that is

badboy:


chutten (changes requested):
This is actually an intentional ordering for logs like "Scheduling for 12h after 2024-06-28 0958, reason tomorrow"

If you say "Scheduling for <now>" then it looks like we're scheduling for, y'know, _now_

badboy:
> This is actually an intentional ordering for logs like "Scheduling for 12h after 2024-06-28 0958, reason tomorrow"
> 
> If you say "Scheduling for " then it looks like we're scheduling for, y'know, _now_

🤔 now that I re-read this you are right and I'm not sure why I considering it wrong. Somehow my mind made this wrong?

PR #2889, opened 2024-07-08
Bug 1906365 - Gradle plugin: support using an external venv.

nicolas-guichard:
mozilla-central uses a vendored copy of glean_parser except for the Android parts which currently download their own copy at build time through the Gradle plugin.

This provides a way to re-use an existing Python virtualenv, in the case of mozilla-central managed by mach, instead of setting up a bespoke one.

This has been tested locally and in [this Try run](https://treeherder.mozilla.org/jobs?repo=try&revision=fd0a81c9c14b1d9bad8e3656a4f564c3fcecdb08) where we can see that [the gradle-dependencies task](https://firefox-ci-tc.services.mozilla.com/tasks/cNe0bMCLSDi_bYCr3FNw7g) (using `gradle --online`) doesn't use Miniconda anymore and [the Android Searchfox task](https://firefox-ci-tc.services.mozilla.com/tasks/R6Rn09OBQ9-sqUblJdu_yA) (using `gradle --offline`) succeeds without supplying the glean_parser wheels.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 283 utf-8
+personal_ws-1.1 en 284 utf-8
 AAR
 AARs
 ABI
@@ -271,6 +271,7 @@ urlbar
 validator
 vendored
 vendoring
+virtualenv
 walkthrough
 walkthroughs
 webextension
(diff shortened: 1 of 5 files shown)


badboy (approved):
Thank you, even including updating the docs right away!

PR #2891, opened 2024-07-09
Bug 1887023 - Prepare to remove `service-glean`

travis79:
In order to remove `service-glean` we need to first add an alias to accommodate glean_parser. We also need to make a few functions public so that they can be accessed by consumers.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/HistogramBase.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/HistogramBase.kt
@@ -22,3 +22,9 @@ interface HistogramBase {
      */
     fun accumulateSamples(samples: List<Long>)
 }
+
+// glean_parser template currently expects `HistogramMetricBase` as the name
+// and since this alias was defined in `service-glean` in android-components,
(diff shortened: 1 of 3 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/HistogramBase.kt
```diff
@@ -22,3 +22,9 @@ interface HistogramBase {
      */
     fun accumulateSamples(samples: List<Long>)
 }
+
+// glean_parser template currently expects `HistogramMetricBase` as the name
+// and since this alias was defined in `service-glean` in android-components,
+// we need to keep the alias until the parser template is updated also.
+// TODO: See Bug ####### for more information.
```
chutten, line 29: Fill in the bug number before landing, if you would

chutten (approved):

glean-core/android/src/main/java/mozilla/telemetry/glean/private/HistogramBase.kt
travis79: Whoops, I forgot about this. Thanks for reminding me :)

PR #2895, opened 2024-07-11
New object metric in samples to ensure we generate valid code

badboy:
This was broken in an older glean_parser.

---

~~glean_parser changes pending before this can land.~~


chutten (approved):
Moar tests. Good

PR #2896, opened 2024-07-11
Bug1657947 New metric types: labeled_{custom|memory|timing}_distribution

chutten:
Will also need `glean_parser` changes and schema changes before they can be used. Probably best to land the schema changes before landing this PR.


chutten:
Whoops, this breaks Python. No need for a review yet. (dammit Github, how do I mark this as "Changes Planned"...)

badboy:
> Whoops, this breaks Python. No need for a review yet. (dammit Github, how do I mark this as "Changes Planned"...)

you can turn it into a draft:

image

chutten:
I _think_ the iOS testfail might be an interkitten (it's the sample app's `testDeletionRequestPing()`). I'm rerunning it to check.

chutten:
(Ha, I was right)

glean-core/src/metrics/labeled.rs
```diff
@@ -24,6 +28,46 @@ pub type LabeledBoolean = LabeledMetric<BooleanMetric>;
...
+/// A labeled timing_distribution.
+pub type LabeledTimingDistribution = LabeledMetric<TimingDistributionMetric>;
+
+/// The metric data needed to construct inner submetrics.
+///
+/// Different Labeled metrics require different amounts and kinds of information to
+/// be constructed.
+pub enum LabeledMetricData {
+    /// The common case: just a CMD.
+    #[allow(missing_docs)]
```
badboy, line 46: I guess not documenting the fields here is fine.
You could move this above the `enum LabeledMetricData` and save the annotations per variant.

glean-core/src/metrics/labeled.rs
```diff
@@ -24,6 +28,46 @@ pub type LabeledBoolean = LabeledMetric<BooleanMetric>;
...
+pub type LabeledMemoryDistribution = LabeledMetric<MemoryDistributionMetric>;
+
+/// A labeled timing_distribution.
+pub type LabeledTimingDistribution = LabeledMetric<TimingDistributionMetric>;
+
+/// The metric data needed to construct inner submetrics.
+///
+/// Different Labeled metrics require different amounts and kinds of information to
+/// be constructed.
+pub enum LabeledMetricData {
```
badboy, line 44: It's unfortunate that this makes the memory size larger for the common case, but ... eh. it's gonna be fine.

badboy (approved):
That was pleasant to review. Not much to say, it works!
Might benefit from a smoke-test in LabeledMetricTypeTest.kt to ensure the Kotlin API is correct (same for .swift I guess), I'd be fine seeing that in a followup though

glean-core/src/metrics/labeled.rs
chutten: Ah, because it needs to store which variant it is for runtime type inspection?

glean-core/src/metrics/labeled.rs
chutten: Eh, I like it this way. Forces the next person to decide whether or not they want to document.

chutten:
> That was pleasant to review. Not much to say, it works! Might benefit from a smoke-test in LabeledMetricTypeTest.kt to ensure the Kotlin API is correct (same for .swift I guess), I'd be fine seeing that in a followup though

Alrighty, tests for unsupported languages tracked here: https://bugzilla.mozilla.org/show_bug.cgi?id=1908760

glean-core/src/metrics/labeled.rs
badboy: yeah, enums are always as big as the biggest variant (plus a tag somewhere so it knows which variant is active).

PR #2898, opened 2024-07-16
Gradle plugin: Force offline mode if env dir is specified

badboy:


travis79 (approved):
r+ with documentation and a changelog entry, please!

badboy:
> r+ with documentation and a changelog entry, please!

no release since the feature was introduced, so it's covered by that new documentation and this is merely a bugfix that noone but me could have run into yet.

PR #2899, opened 2024-07-16
Bug 1908157 - Android sample: Extra nested array

badboy:
This tests the generation of nested arrays.
This currently lacks a test that it correctly serializes.

---

This needs https://github.com/mozilla/glean_parser/pull/729
and I would like to add a proper test of the serialization.


badboy:
I didn't find any good way to test this on Android. Our sample is compiled but never run (and right now it seems to fail anyway)

travis79 (approved):

PR #2901, opened 2024-07-18
Swift: Use String(decoding:as:) to directly decode the UTF8 bytes

badboy:
This is otherwise a `non_optional_string_data_conversion` swiftlint warning (in swiftlint 0.55.1 at least, we might be using a different swiftlint in CI right now).


chutten (approved):
Seems good, but I don't know Swift so you may want Travis to take a look as well.

travis79 (approved):

PR #2903, opened 2024-07-19
Update Android NDK & SDK to the latest releases

rvandermeulen:


rvandermeulen:
@badboy Do you have idea what the issue is with CircleCI? The Taskcluster-based Android build/tests works and the lints all pass locally for me. Looks like some kind of environment setup issue.

rvandermeulen:
Ugh, looks like it's a known issue that's fixed in newer AGPs.
https://issuetracker.google.com/issues/294137077

chutten:
According to the issue tracker, should be good now? But it'll need a rebase.

chutten:
(and I guess m-c's on 27.2.12479018 now)

rvandermeulen:
The NDK could probably be bumped any time, but the SDK bump is still blocked on updating AGP in #2680. Happy to update this PR once that's done, though!

chutten:
Yeah, okay, that looks arcane enough that I should leave it until Jan-Erik gets back. Deep magic out there.

I will note in case it helps that merely updating `ndkVersion` in `build.gradle` was enough to allow me to build the Glean SDK on the command line with m-c's bootstrapped ndk, sdk, and java. (with the appropriate envvars and `local.properties`)

PR #2914, opened 2024-08-01
Bug 1909246 - Set the default pre-init task queue size to 1 million tasks

travis79:
This also removes the associated Rust feature: `preinit_million_queue`. This feature was previously used to conditionally compile with the 1 million task queue size only when building for gecko.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,8 @@
   * Make auto-flush behavior configurable and time-based ([#2871](https://github.com/mozilla/glean/pull/2871))
 * Android
   * Update to Gradle v8.9 ([#2909](https://github.com/mozilla/glean/pull/2909))
+* Rust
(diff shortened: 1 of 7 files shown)


badboy (approved):
a wooping 2 extra minutes for android tests and +3 minutes for Rust tests 😭
but I have no good idea how to avoid that.

badboy:
Hm, what do you think about for now ignoring those tests (`#[ignore] // oops, long test`)?
Or file a bug to followup on this, see if we can get better.

travis79:
> a wooping 2 extra minutes for android tests and +3 minutes for Rust tests 😭 but I have no good idea how to avoid that.

I considered removing the Kotlin test since this is already tested in Rust. Since the queue itself is in glean_core, do we need to test it again in a binding?

travis79:
> Hm, what do you think about for now ignoring those (Rust) tests (`#[ignore] // oops, long test`)? Or file a bug to followup on this, see if we can get better.

It's a fairly low risk place in the code for a regression, I think. I'd be okay with ignoring it for now and filing a bug to try and make it better and re-enable it in the future.

travis79:
Bug filed: https://bugzilla.mozilla.org/show_bug.cgi?id=1911350

PR #2915, opened 2024-08-02
Bug 1787234 - Fix `GleanTestLocalServer` test rule leaking between tests

travis79:
This adds an override for the `TestWatcher` `finished` method that ensures that the underlying `WorkManager` executor is properly terminated, and then closes the `WorkManager` database.


chutten (changes requested):
Requires CHANGELOG entry

chutten (approved):

PR #2916, opened 2024-08-02
Update to glean_parser v14.5.0

chutten:


samples/ios/app/metrics.yaml
```diff
@@ -28,6 +28,9 @@ browser.engagement:
       key2:
         type: string
         description: "This is key two"
+      And1WithUnusualCASING:
```
travis79, line 31: Lots of repetition...  I wonder if we could make better use of anchors at some point to cut down on this 🤔  (a problem for the future, obviously)

PR #2920, opened 2024-08-05
Upgrade to UniFFI 0.28

badboy:
This updates some other dependencies.
To avoid a dependency on clap (and all of its dependencies) we bring along our own uniffi-bindgen that implements only the mininmal part. We did that before as well, but now that part is slightly bigger.

Note: Ben asked me to get that into a Glean release soon. I want another Glean release before this lands though.


travis79 (approved):

travis79:
Do we have documentation of the uniffi_bindgen stuff in our developer docs?

badboy:
> Do we have documentation of the uniffi_bindgen stuff in our developer docs?

no. what kind of docs do you expect?

PR #2939, opened 2024-08-21
CI: Use newer Rust on publish

badboy:


badboy:
I did it manually for the latest release.

PR #2942, opened 2024-08-23
Kotlin: Tasks queue to delay certain early metric recordings

badboy:
Details in commit messages.
Not sure we actually want this but it is one way to address bug 1892234


travis79 (commented):
What about the other metric types? Are these just the current types we have a problem with? I am just wondering if we need to do this for all metric types or just on a "as needed" basis.

badboy:
> What about the other metric types? Are these just the current types we have a problem with? I am just wondering if we need to do this for all metric types or just on a "as needed" basis.

I'm going with this on a "as-needed" basis because I'm not really the biggest fan.
It also reduces the work right now and we can see if it addresses things before we have to do it for more.
(and there's a whole bunch of extra edgecases for other metrics and how they are used).
I'm going to post a summary in the bug.

travis79:
> > What about the other metric types? Are these just the current types we have a problem with? I am just wondering if we need to do this for all metric types or just on a "as needed" basis.
> 
> I'm going with this on a "as-needed" basis because I'm not really the biggest fan. It also reduces the work right now and we can see if it addresses things before we have to do it for more. (and there's a whole bunch of extra edgecases for other metrics and how they are used). I'm going to post a summary in the bug.

This works for me. If eventually we find a solution we are bigger fans of, it is less to undo later :)

badboy:
Details now in https://bugzilla.mozilla.org/show_bug.cgi?id=1892234#c24

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
```diff
@@ -45,6 +47,68 @@ internal object Dispatchers {
...
+            // while a `addTaskToQueue` is being executed. For this reason, we need synchronized
+            // access to the queue. However, we can't simply wrap the task execution in a sync
+            // block: suspending functions are not allowed inside critical sections.
+            val queueCopy: ConcurrentLinkedQueue<() -> Unit> = ConcurrentLinkedQueue()
+            synchronized(dispatcherObject) {
+                queueCopy.addAll(taskQueue)
+                taskQueue.clear()
+            }
+
+            queueCopy.forEach { task ->
```
travis79, line 98: I assume this iterates through the queue in a FIFO way?

travis79 (approved):
No real blockers, just a question. So: r+w?

glean-core/android/src/main/java/mozilla/telemetry/glean/Dispatchers.kt
badboy: Yup, this is the same code as the old dispatcher used.

travis79: Sorry, I guess I must have blocked that out 😅

PR #2948, opened 2024-09-02
Bug 1916673 - Buffered API for memory/timing distributions

badboy:
This is the implementation for what's being designed in bug-1915388

- [x] final ack in the design bug
- [ ] ~~docs~~ will be done separately (bug-1917809)
- [x] experimental marker

### DIFF
--- a/glean-core/rlb/tests/custom_distribution_buffered.rs
+++ b/glean-core/rlb/tests/custom_distribution_buffered.rs
@@ -1,0 +1,82 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This integration test should model how the RLB is used when embedded in another Rust application
+//! (e.g. FOG/Firefox Desktop).
+//!
+//! We write a single test scenario per file to avoid any state keeping across runs
+//! (different files run as different processes).
+
+mod common;
+
+use glean::ConfigurationBuilder;
+
+mod metrics {
+    use glean::private::*;
+    use glean::{HistogramType, Lifetime};
+    use glean_core::CommonMetricData;
+    use once_cell::sync::Lazy;
+
+    #[allow(non_upper_case_globals)]
+    pub static measure: Lazy<CustomDistributionMetric> = Lazy::new(|| {
+        CustomDistributionMetric::new(
+            CommonMetricData {
+                name: "measure".into(),
+                category: "sample".into(),
+                send_in_pings: vec!["validation".into()],
+                lifetime: Lifetime::Ping,
+                disabled: false,
+                ..Default::default()
+            },
+            0,
+            100,
+            100,
+            HistogramType::Linear,
+        )
+    });
+}
+
+/// Test scenario: Ensure buffered accumulation works.
+#[test]
+fn buffered_memory_distribution_works() {
+    common::enable_test_logging();
+
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().to_path_buf();
+
+    let cfg = ConfigurationBuilder::new(true, tmpname, "firefox-desktop")
+        .with_server_endpoint("invalid-test-host")
+        .build();
+    common::initialize(cfg);
+
+    let mut buffer = metrics::measure.start_buffer();
+
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+
+    // Nothing accumulated while buffer uncommitted.
+    assert_eq!(None, metrics::measure.test_get_value(None));
+
+    // Commit buffer by dropping it.
+    drop(buffer);
+
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    let mut buffer = metrics::measure.start_buffer();
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+    // Don't record any of this data into the metric.
+    buffer.abandon();
+
+    // Metric is unchanged.
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    glean::shutdown(); // Cleanly shut down at the end of the test.
+}
--- a/glean-core/rlb/tests/memory_distribution_buffered.rs
+++ b/glean-core/rlb/tests/memory_distribution_buffered.rs
@@ -1,0 +1,95 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This integration test should model how the RLB is used when embedded in another Rust application
+//! (e.g. FOG/Firefox Desktop).
+//!
+//! We write a single test scenario per file to avoid any state keeping across runs
+//! (different files run as different processes).
+
+mod common;
+
+use glean::{ConfigurationBuilder, ErrorType};
+
+mod metrics {
+    use glean::private::*;
+    use glean::{Lifetime, MemoryUnit};
+    use glean_core::CommonMetricData;
+    use once_cell::sync::Lazy;
+
+    #[allow(non_upper_case_globals)]
+    pub static measure: Lazy<MemoryDistributionMetric> = Lazy::new(|| {
+        MemoryDistributionMetric::new(
+            CommonMetricData {
+                name: "measure".into(),
+                category: "sample".into(),
+                send_in_pings: vec!["validation".into()],
+                lifetime: Lifetime::Ping,
+                disabled: false,
+                ..Default::default()
+            },
+            MemoryUnit::Byte,
+        )
+    });
+}
+
+/// Test scenario: Ensure buffered accumulation works.
+#[test]
+fn buffered_memory_distribution_works() {
+    common::enable_test_logging();
+
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().to_path_buf();
+
+    let cfg = ConfigurationBuilder::new(true, tmpname, "firefox-desktop")
+        .with_server_endpoint("invalid-test-host")
+        .build();
+    common::initialize(cfg);
+
+    let mut buffer = metrics::measure.start_buffer();
+
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+
+    // Nothing accumulated while buffer uncommitted.
+    assert_eq!(None, metrics::measure.test_get_value(None));
+
+    // Commit buffer by dropping it.
+    drop(buffer);
+
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    let mut buffer = metrics::measure.start_buffer();
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+    // Don't record any of this data into the metric.
+    buffer.abandon();
+
+    // Metric is unchanged.
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    // Check error recording.
+    let mut buffer = metrics::measure.start_buffer();
+    let large_sample = (1 << 40) + 1;
+    for _ in 0..10 {
+        buffer.accumulate(large_sample);
+    }
+    drop(buffer);
+
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(13, data.count);
+
+    assert_eq!(
+        10,
+        metrics::measure.test_get_num_recorded_errors(ErrorType::InvalidValue)
+    );
+
+    glean::shutdown(); // Cleanly shut down at the end of the test.
+}
--- a/glean-core/src/histogram/mod.rs
+++ b/glean-core/src/histogram/mod.rs
@@ -135,4 +135,33 @@ impl<B: Bucketing> Histogram<B> {
         }
         res
     }
+
+    /// Clear this histogram.
+    pub fn clear(&mut self) {
+        self.sum = 0;
+        self.count = 0;
+        self.values.clear();
+    }
+}
+
+impl<B> Histogram<B>
+where
+    B: Bucketing,
+    B: std::fmt::Debug,
+    B: PartialEq,
+{
+    /// Merges data from one histogram into the other.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the two histograms don't use the same bucketing.
+    pub fn merge(&mut self, other: &Self) {
+        assert_eq!(self.bucketing, other.bucketing);
+
+        self.sum += other.sum;
+        self.count += other.count;
+        for (&bucket, &count) in &other.values {
+            *self.values.entry(bucket).or_insert(0) += count;
+        }
+    }
 }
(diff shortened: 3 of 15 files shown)


badboy:
Usage of it WIP: https://phabricator.services.mozilla.com/D221017

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
             test_get_num_recorded_errors(glean, self.meta(), error).unwrap_or(0)
         })
     }
+
+    /// Start a new histogram buffer associated with this timing distribution metric.
+    ///
+    /// A histogram buffer accumulates in-memory.
+    /// Data is recorded into the metric on drop.
```
mxinden, line 312:
```suggestion
    /// Data is recorded into the metric on [`Drop::drop`].
```

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+                    Metric::TimingDistribution(hist)
+                });
+        });
+    }
+}
+
+/// A histogram buffer associated with a specific instance of a [`MemoryDistributionMetric`].
+///
+/// Accumulation happens in-memory.
+/// Data is merged into the metric on `Drop`.
```
mxinden, line 338:
```suggestion
/// Data is merged into the metric on [`Drop::drop`].
```

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+#[derive(Debug)]
+pub struct HistogramBuffer<'a> {
+    histogram: Histogram<Functional>,
+    metric: &'a MemoryDistributionMetric,
```
mxinden, line 342: In the case of `firefox-on-glean` will `'a` be `'static`? Asked differently, how do I embed `HistogramBuffer<'a>` in a `struct` without exposing the lifetime `'a` on that struct?

mxinden (commented):
Thank you for the work here.

If I understand this pull request correctly, it only adds a buffered API to `MemoryDistribution` and `TimingDistribution`.

In https://phabricator.services.mozilla.com/D220062  I want to add 4 new metrics:


1. http3_udp_datagram_segment_size_sent
2. http3_udp_datagram_segment_size_received
3. http3_udp_datagram_size_received
4. http3_udp_datagram_num_segments_received


(1), (2) and (3) I can model with a `MemoryDistribution`.  (4) on the other hand counts the number of segments received, where values range between `1` and `128`.

Here is metrics definition draft from the above Phabricator patch.

``` yaml
  http3_udp_datagram_num_segments_received:
    type: custom_distribution
    unit: integer
    range_min: 0
    # Maximum number of UDP segments per datagram.
    #
    # See e.g. Linux https://github.com/torvalds/linux/blob/20371ba120635d9ab7fc7670497105af8f33eb08/include/linux/udp.h#L111.
    range_max: 128
    bucket_count: 100
    histogram_type: exponential
    description: >
      HTTP3 UDP number of segments per datagram received.
    bugs:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1906853
    data_reviews:
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1906853
    data_sensitivity:
      - technical
    notification_emails:
      - necko@mozilla.com
      - minden@mozilla.com
    expires: never
```

I assume this is best modeled by the `CustomDistribution`. Would a buffered API for `CustomDistribution` be included in this patch?

(Note, in case you want to ship a buffered API for one distribution only for now, I assume I could as well model (1), (2) and (3) with a `CustomDistribution` instead of a `MemoryDistribution`.)

glean-core/src/histogram/mod.rs
```diff
@@ -136,3 +136,25 @@ impl<B: Bucketing> Histogram<B> {
...
+where
+    B: Bucketing,
+    B: std::fmt::Debug,
+    B: PartialEq,
+{
+    /// Merges data from one histogram into the other.
+    ///
+    /// ## Panics
+    ///
+    /// Panics if the two histograms don't use the same bucketing.
```
chutten, line 150: Really? Panics? That seems a little harsh. Though I suppose we're the only ones who should be making these, so we can ensure everything's okay.

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+            glean
+                .storage()
+                .record_with(glean, &metric.meta, move |old_value| {
+                    let mut hist = match old_value {
+                        Some(Metric::TimingDistribution(hist)) => hist,
+                        _ => Histogram::functional(LOG_BASE, BUCKETS_PER_MAGNITUDE),
+                    };
+
+                    hist.merge(&histogram);
+                    Metric::TimingDistribution(hist)
```
chutten, line 329: Uh, `Metric::MemoryDistribution(hist)`, surely?

chutten: Is this function not covered by tests?

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+        );
+    }
+}
+
+impl Drop for HistogramBuffer<'_> {
+    fn drop(&mut self) {
+        // We want to move that value.
+        // A `0/0` histogram doesn't allocate.
+        let buffer = mem::replace(&mut self.histogram, Histogram::functional(0.0, 0.0));
+        self.metric.commit_histogram(buffer);
```
chutten, line 378: We can't `self.metric.commit_histogram(self.histogram);`? We're in `drop`, we can't use any piece of `self` after the method completes.

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+    ///
+    /// Accumulation happens in-memory only.
+    pub fn accumulate(&mut self, sample: u64) {
+        self.histogram.accumulate(sample)
+    }
+
+    /// Abandon this histogram buffer and don't commit accumulated data.
+    pub fn abandon(mut self) {
+        // Replace any recordings with an empty histogram.
+        _ = mem::replace(
```
chutten, line 366: Is this better than implementing and calling a `pub fn clear()` on `impl<B: Bucketing> Histogram<B>`?

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -576,6 +577,141 @@ impl TimingDistributionMetric {
...
+        crate::launch_with_glean(move |glean| {
+            glean
+                .storage()
+                .record_with(glean, &metric.meta, move |old_value| {
+                    let mut hist = match old_value {
+                        Some(Metric::TimingDistribution(hist)) => hist,
+                        _ => Histogram::functional(LOG_BASE, BUCKETS_PER_MAGNITUDE),
+                    };
+
+                    hist.merge(&histogram);
```
chutten, line 629: All this moving of `histogram` across closures just so we can take a reference.

glean-core/src/metrics/memory_distribution.rs
```diff
@@ -304,4 +305,76 @@ impl MemoryDistributionMetric {
...
+    /// Create a new histogram buffer referencing the timing distribution it will record into.
+    fn new(metric: &'a MemoryDistributionMetric) -> Self {
+        let histogram = Histogram::functional(LOG_BASE, BUCKETS_PER_MAGNITUDE);
+        Self { histogram, metric }
+    }
+
+    /// Accumulates one sample into the histogram.
+    ///
+    /// The provided sample must be in the "unit" declared by the instance of the metric type
+    /// (e.g. if the instance this method was called on is using [`crate::TimeUnit::Second`], then
```
chutten, line 355: Example should probably use `MemoryUnit`

samples/rust/src/main.rs
```diff
@@ -113,6 +113,16 @@ fn main() {
         glean_metrics::party::drinks.test_get_num_recorded_errors(ErrorType::InvalidValue)
     );
 
+    {
+        let mut buffer = glean_metrics::test_metrics::timings.start_buffer();
+
+        for _ in 0..100 {
+            let now = Instant::now();
+            thread::sleep(Duration::from_millis(10));
+            buffer.accumulate(now.elapsed().as_millis() as u64);
```
chutten, line 122: Any reason for using `now` and `now.elapsed` and all this `sleep` instead of having a loop-invariant `Duration::from_millis(10)` that we then `as_millis() as u64`? We don't need to involve clocks at all, I think.

chutten (changes requested):

glean-core/src/metrics/memory_distribution.rs
badboy: Oops. Should be changed.
Fun fact: doesn't matter much for the payload, at least in how the data looks.

glean-core/src/metrics/memory_distribution.rs
badboy: I think I need to do some additional type dance to do it.

glean-core/src/metrics/memory_distribution.rs
badboy: Good idea. Looks cleaner.

glean-core/src/histogram/mod.rs
badboy: Sure. We control the use here. A histogram is an internal use. The only way to get there is through the buffered API and we make sure you can't mix that up.

samples/rust/src/main.rs
badboy: I don't even know anymore why I was sleeping here. To make it "realistic"? No idea, fixed values work just as good.

glean-core/src/metrics/timing_distribution.rs
badboy: We need to hold onto the data somewhere. But it doesn't make much sense for `merge` to consume it in the end, because we cannot merge two independent allocations anyway.

glean-core/src/metrics/memory_distribution.rs
badboy: A test would have caught this. So there will be a test now.

glean-core/src/metrics/memory_distribution.rs
badboy: Yes, because the metric are generated as statics (well, `OnceCell`-statics), so this can get you a `LocalMemoryDistribution<'static>` out.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
```diff
@@ -74,3 +88,32 @@ class TimingDistributionMetricType(meta: CommonMetricData, timeUnit: TimeUnit) :
...
+     * `sample` is assumed to be in seconds).
+     *
+     * Accumulation happens in-memory only.
+     */
+    fun accumulate(sample: Long) = inner.accumulate(sample.toULong())
+
+    /**
+     * Record data from the histogram buffer into the associated metric.
+     *
+     * This instance must not be used after `commit()` is called.
```
chutten, line 110: Is there a way to prevent it from being used after `commit()`? I think the Java-like pattern is to`throw new IllegalStateException("Use after commit");`

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
```diff
@@ -74,3 +88,32 @@ class TimingDistributionMetricType(meta: CommonMetricData, timeUnit: TimeUnit) :
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
     fun testGetNumRecordedErrors(error: ErrorType) = inner.testGetNumRecordedErrors(error)
 }
+
+class LocalTimingDistribution(
+    val metric: TimingDistributionMetricType,
+    val inner: FfiLocalTimingDistribution,
+) : AutoCloseable {
```
chutten, line 95: Might want `protected fun finalize() { /* TODO: bug 1691073 */ report_some_error(); ffi_abandon_buffer(this);` or whatever to help report misuse and prevent memory leaks.

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -576,6 +577,137 @@ impl TimingDistributionMetric {
...
+    /// A histogram buffer accumulates in-memory.
+    /// Data needs to be recorded into the metric using
+    /// [`TimingDistributionMetric::ffi_commit_buffer`].
+    #[doc(hidden)]
+    pub fn ffi_start_buffer(&self) -> Arc<FfiLocalTimingDistribution> {
+        Arc::new(FfiLocalTimingDistribution::new())
+    }
+
+    /// Record data from the histogram buffer into this metric.
+    pub fn ffi_commit_buffer(&self, buffer: Arc<FfiLocalTimingDistribution>) {
```
chutten, line 600: Should we clear the buffer to stop folks accidentally committing the same buffer twice? Or is that prevented somewhere/somehow else? For non-FFI we rely on the "called at most once" behaviour of drop, but I don't see us accounting for the lack of that guarantee (or testing that case) here.

chutten (changes requested):
As much as I'd like this to be a r+with_followups, I think we need to decide about the FFI and Kotlin interfaces for certain before we ship this, as subsequent changes will be breaking.

glean-core/android/src/main/java/mozilla/telemetry/glean/private/TimingDistributionMetricType.kt
badboy: Hmmm.
Then we need to track that state.

I'm considering to remove the Kotlin part. The first user is Rust only.

badboy:
Ok, I split Kotlin out so we can tackle that separately.

glean-core/rlb/tests/custom_distribution_buffered.rs
```diff
@@ -0,0 +1,82 @@
...
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    let mut buffer = metrics::measure.start_buffer();
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+    // Don't record any of this data into the metric.
+    buffer.abandon();
```
chutten, line 74: Drop buffer after abandoning it?

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -576,6 +577,108 @@ impl TimingDistributionMetric {
...
+    /// A histogram buffer accumulates in-memory.
+    /// Data is recorded into the metric on drop.
+    pub fn start_buffer(&self) -> LocalTimingDistribution<'_> {
+        LocalTimingDistribution::new(self)
+    }
+
+    fn commit_histogram(&self, histogram: Histogram<Functional>, errors: usize) {
+        let metric = self.clone();
+        crate::launch_with_glean(move |glean| {
+            let max_sample_time = metric.time_unit.as_nanos(MAX_SAMPLE_TIME);
```
chutten, line 592: May as well move this inside `if errors > 0`, right?

chutten (approved):
r+wc

glean-core/rlb/tests/memory_distribution_buffered.rs
```diff
@@ -0,0 +1,79 @@
...
+    let data = metrics::measure.test_get_value(None).unwrap();
+    assert_eq!(3, data.count);
+    assert_eq!(30, data.sum);
+
+    let mut buffer = metrics::measure.start_buffer();
+    for _ in 0..3 {
+        buffer.accumulate(10);
+    }
+    // Don't record any of this data into the metric.
+    buffer.abandon();
```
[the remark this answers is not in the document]
badboy: `abandon` takes `self`, so it consumes the buffer already, thus dropping it.
I will add a check though so we avoid work when the histogram is empty.

PR #2965, opened 2024-10-01
Small fixes to the iOS build

badboy:
This will actually likely currently fail on CI, because the `swiftlint` we use there is older?
I have swiftlint 0.57 locally.

We might wanna update that check to Xcode 16 soon.

### DIFF
--- a/build-scripts/xc-universal-binary.sh
+++ b/build-scripts/xc-universal-binary.sh
@@ -21,7 +21,6 @@ if [[ -n "${DEVELOPER_SDK_DIR:-}" ]]; then
 #  # In this case, we need to add an extra library search path for build scripts and proc-macros,
 #  # which run on the host instead of the target.
 #  # (macOS Big Sur does not have linkable libraries in /usr/lib/.)
-  local libpath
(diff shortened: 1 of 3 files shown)


badboy:
Oh it actually doesn't fail, because CI uses swiftlint 0.53 which is so old, it doesn't have that (somewhat broken) new lint!

travis79 (changes requested):
Why are we reverting this when it was just changed to this 3 months ago? 

ref: https://github.com/mozilla/glean/commit/ad9855581e06c0190eabd1c2ebe06ca2eeefaa9a

badboy:
> Why are we reverting this when it was just changed to this 3 months ago?
> 
> ref: [ad98555](https://github.com/mozilla/glean/commit/ad9855581e06c0190eabd1c2ebe06ca2eeefaa9a)

Because I should have included the link: https://github.com/realm/SwiftLint/issues/5263#issuecomment-2115182747

tl;dr: the way we changed it to is swallowing errors, and thus swiftlint actually reversed the decision and lints AGAINST this now.
That only happens with a newer swiftlint than we use on CI, but I have that newer one locally and thus `make lint-swift` fails for me. Given that we eventually upgrade CI and it doesn't fail right now we can land this.

PR #2968, opened 2024-10-03
Improved the readability

Varma0604:


badboy:
That file is a list of words to ignore for spellchecking, not a definition list.
Notably the test spits out an error on this:

> Error: The file "/tmp/./.dictionary" is not in the proper format.

(it interestingly does _not_ fail, which .. uh .. suboptimal)

PR #2979, opened 2024-10-21
Bug 1926095 - Add how-to with tips on investigating data anomalies

travis79:


badboy (changes requested):
some nits.

Would it maybe make sense which columns to query? A lot of them are standardized, so we could list them.

travis79:
> some nits.
> 
> Would it maybe make sense which columns to query? A lot of them are standardized, so we could list them.

Well, I suppose I could add that info...  let me do that before you review again

travis79:
Okay, I added column names where appropriate and other links. Ready for another look @badboy

badboy (approved):

PR #2980, opened 2024-10-23
Add docs for new metric type 'labeled_quantity'

chutten:


docs/user/SUMMARY.md
```diff
@@ -85,6 +85,7 @@
     - [Custom Distribution](reference/metrics/custom_distribution.md)
     - [Labeled Custom Distributions](reference/metrics/labeled_custom_distributions.md)
     - [Quantity](reference/metrics/quantity.md)
+    - [Labeled Quantities](reference/metrics/labeled_quantity.md)
```
travis79, line 88: This differs slightly in name from all the other places this is referred to: Quantity vs. Quantities here. Not sure if this is intentional or even a request to change, just something I noticed as I went through it.

travis79 (approved):

PR #2982, opened 2024-10-24
New API on UUID metric: generateOnce

badboy:
TODO:

* [ ] Decision if we want/need this
* [ ] Docs
* [ ] Kotlin


badboy:
not for now.

PR #2983, opened 2024-10-24
Allow client_info section to be reduced

badboy:
First draft of how to reduce the client info fields.
Requires https://github.com/mozilla/glean_parser/pull/766


badboy:
this won't work as is anyway, because the schema has those fields as required: https://github.com/mozilla-services/mozilla-pipeline-schemas/blob/64f8ce4660d6d0a978afb9274792be7430994e32/schemas/glean/glean/glean.1.schema.json#L91-L98

PR #2986, opened 2024-10-24
Clarify OS and Device info suitability in Data Investigation Guide

chutten:
Some small nits that occurred to me while reading this excellent guide.


travis79 (approved):
Thanks!

PR #2998, opened 2024-11-12
Allow timespans to track across upload toggles

badboy:
This is a pre-requisite to allow the coming per-ping toggle:

At the `start` time we don't know _which_ pings to record for.

It's fine to do this because:
* Timespans, if at all, usually collect a span over a single action, with no chance of a upload toggle happening in between
* There's very few timespans in use right now
* It's often the wrong metric type anyway, and timing distributions should be used
* Timing Distributions already allow this


travis79 (approved):
Looks like what we discussed. r+

PR #2999, opened 2024-11-13
Bug 1918149 - Add measurements of the impact of disabing Glean event batching to the docs

travis79:


docs/user/user/howto/real-time-events/real-time-events.md
```diff
@@ -14,6 +14,21 @@ Glean ["events" ping](../../pings/events.md) submission can be configured either
 Setting the maximum event threshold to a value of `1` will configure the Glean SDK to submit an "events" ping for each and every [event](../../../reference/metrics/event.md) as they
 are recorded. By default, the Glean SDK will batch 500 events per "events" ping.
 
+As of November 2024, Desktop Release:
+Median user per day: 67 events / 3 pings
+The impact of turning on one event per ping based on the median user would result in an increase of approximately 21 times more event ping volume.
```
badboy, line 19: We should find a better formatting for this. mdbook treats this all as a single line, so at the very least it needs a `.` (dot) after `3 pings` or something (see rest in review comment)

badboy (changes requested):
I have some hesitation r+'ing this.

For one I'm not 100% sure this is the right place to put it.
This is something very specific, while the rest talks about the general approach.
At the very least this should be marked as such, maybe using an info block? That would make it clear it's looking at the current real product state.

Second, are we comfortable publishing these numbers as is? (They are now somewhat public in this PR).
How quickly will they be out of date? Is this relevant to us/others reading this page in 3 months/6 months/etc.
Maybe having that in the bug, not in the docs, is "good enough"?

docs/user/user/howto/real-time-events/real-time-events.md
travis79: I tried to clean it up a bit, see what you think now.

travis79:
In regards to where this should live: Yes, I agree it's not exactly comfortable here but it is the closest to the information that would make it relevant (turning off event batching), so this is where I saw it living for now.

In regards to the numbers? Yeah, any measurement I made would at some point be out of date but I think that the general trend of number of events to number of events pings being about 20-40 : 1 ratio is going to be fairly consistent, and anyone thinking of turning this on I would hope would take more recent measurements into consideration.

badboy (approved):

PR #3002, opened 2024-11-18
In tests register all pings by instantiating them

badboy:


chutten (approved):
Any reason you went for "build and ref the ping" as the design instead of "create a test-only method for adding ping names to the registry"?

badboy:
> Any reason you went for "build and ref the ping" as the design instead of "create a test-only method for adding ping names to the registry"?

I updated it and actually moved a bit of duplication into the setup function for the RLB tests.
I'm keeping the instantiation in the examples, because they are ... well, examples. So they are closer to real code.
I also file a bug that we should expose a "register all pings" function in the generated code, so that no one has to ever manually do the `_ = &*...` dance.

PR #3004, opened 2024-11-19
Bug 1921976 - Adds methods for accessing Glean debug settings as well as the list of currently registered pings

travis79:


glean-core/src/core/mod.rs
```diff
@@ -724,6 +724,15 @@ impl Glean {
             .insert(ping.name().to_string(), ping.clone());
     }
 
+    /// Gets a list of currently registered ping names.
+    ///
+    /// # Returns
+    ///
+    /// The list of ping names that are currently registered.
+    pub fn get_registered_ping_names(&self) -> Vec<&String> {
+        self.ping_registry.keys().collect()
```
badboy, line 733: `&String` is an unuasl type to see in Rust.
This can be just `Vec<&str>`. Might need a `.map(String::as_str)` in there below.
```suggestion
    pub fn get_registered_ping_names(&self) -> Vec<&str> {
        self.ping_registry.keys().map(String::as_str).collect()
```

glean-core/src/lib.rs
```diff
@@ -874,6 +874,22 @@ pub(crate) fn register_ping_type(ping: &PingType) {
     }
 }
 
+/// Gets a list of currently registered ping names.
+///
+/// # Returns
+///
+/// The list of ping names that are currently registered.
+pub fn glean_get_registered_ping_names() -> Vec<String> {
+    block_on_dispatcher();
```
badboy, line 883: I think this is the crux why we never implemented it.
This would block if initialize is never called. Or at least block until it is. 

Should we call this out more explicitly? At the very least this needs to be explicitly called out as a debug API?

badboy (changes requested):

glean-core/src/lib.rs
travis79: I'll document the behavior. I don't see this being used before Glean is initialized, I really only see the one use case with debugging UI's and tooling so marking this as for debug use seems logicial.

glean-core/src/lib.rs
```diff
@@ -874,6 +874,22 @@ pub(crate) fn register_ping_type(ping: &PingType) {
...
+    core::with_glean(|glean| {
+        let mut pings: Vec<String> = Vec::new();
+        glean.get_registered_ping_names().iter().for_each(|ping| {
+            pings.push(ping.to_string());
+        });
+        pings
```
badboy, line 889: fwiw this could be changed to

```suggestion
        glean.get_registered_ping_names().iter().map(|ping| {
            ping.to_string()
        }).collect()
```

(I think)
But fine either way.

PR #3005, opened 2024-11-19
Bug 1928288 Support interrupting `Wait` upload tasks during shutdown

chutten:


glean-core/rlb/src/net/mod.rs
```diff
@@ -59,6 +60,8 @@ struct Inner {
     uploader: Box<dyn PingUploader + 'static>,
     thread_running: AtomicState,
     handle: Mutex<Option<JoinHandle<()>>>,
+    rx: Receiver<()>,
```
badboy, line 63: We shouldn't need to store this in the struct, do we?
Should be good enough to be moved into the thread.

glean-core/rlb/src/net/mod.rs
```diff
@@ -192,6 +204,9 @@ impl UploadManager {
         let thread = handle.take();
 
         if let Some(thread) = thread {
+            // poke the thread in case it's in `Wait`.
+            let _ = self.inner.tx.send(());
```
badboy, line 208: fun fact, this would work just as well:

```suggestion
            _ = self.inner.tx.send(());
```

(though if clippy doesn't complain I don't complain)

glean-core/rlb/tests/interruptible_shutdown.rs
```diff
@@ -0,0 +1,146 @@
...
+        .with_uploader(ReportingUploader {
+            calls: AtomicUsize::new(0),
+            sender: tx,
+        })
+        .build();
+    cfg.rate_limit = Some(PingRateLimit {
+        seconds_per_interval: 600, // Needs only be longer than the timeout in `glean::uploader_shutdown()`.
+        pings_per_interval: 1,     // throttle thyself immediately.
+    });
+    common::initialize(cfg);
```
badboy, line 118: while you're here, can you add the `_ = &*pings::validation` dance, while I figure out how to make that less of a problem in the future?

glean-core/rlb/tests/interruptible_shutdown.rs
```diff
@@ -0,0 +1,146 @@
...
+        match calls {
+            // First goes through immediately.
+            0 => {
+                self.sender.send(decode(body)).unwrap();
+                net::UploadResult::http_status(200)
+            }
+            // Second SHOULD NEVER SEND
+            1 => panic!("We should shutdown before getting to the second ping"),
+            // Any others ought to be impossible.
+            _ => panic!("Wat."),
```
badboy, line 84: [wat.](https://www.destroyallsoftware.com/talks/wat)

badboy (approved):
test looks good! (and works!)
minimal nits. r+wc.

glean-core/rlb/src/net/mod.rs
chutten: Meaning... `tx` becomes `Option<Sender<()>>` which is `None` when there's no thread? Synchronizing modifications to this via `handle`, I guess.

Wait. While I'm looking at this data... why is `inner` not wrapped in a `Mutex`?

glean-core/rlb/tests/interruptible_shutdown.rs
chutten: Not sure it's needed? Registering the ping will happen immediately before submitting it. We don't even record anything into that ping.

badboy: Ah fine, I might have been overeager then because I've been doing that so much in my code lately :D

glean-core/rlb/src/net/mod.rs
badboy: Because we never need to change `inner`, only some parts of it (which are either atomic or a Mutex itself)

chutten: Trying to do this gets a little complex since `inner` isn't `mut` so giving it its `tx` late is a big ol' nope.

Mind if I land this as-is?

PR #3008, opened 2024-11-22
Don't run tests for uniffi-bindgen wrapper

badboy:


chutten (approved):
Seems like a very reasonable optimization to me.

PR #3013, opened 2024-11-26
Bug 1933679 - Docs for collection-enabled mode

badboy:
[doc only]


bin/build-rust-docs.sh
```diff
@@ -49,6 +49,9 @@ echo '<meta http-equiv=refresh content=0;url=./adding-glean-to-your-project/inde
 # General API
 echo '<meta http-equiv=refresh content=0;url=../reference/general/index.html>' > build/docs/book/user/general-api.html
 echo '<meta http-equiv=refresh content=0;url=../reference/general/experiments-api.html>' > build/docs/book/user/experiments-api.html
+# General reference
+mkdir -p build/docs/book/reference/general
+echo '<meta http-equiv=refresh content=0;url=toggling-collection-status.html>' > build/docs/book/reference/general/toggling-upload-status.html
```
travis79, line 54: This references `toggling-upload-status.html` but elsewhere you have changed the page title to `toggling-collection-status`

docs/user/reference/general/initializing.md
```diff
@@ -86,7 +86,7 @@ To learn about SDK specific configuration options available, refer to the [Refer
 >
 > Always pass the user preference, e.g. `Glean.initialize(uploadEnabled=userSettings.telemetryEnabled)` or the equivalent for your application.
 >
-> Calling `Glean.setUploadEnabled(false)` at a later point will trigger [`deletion-request` pings](../../user/pings/deletion-request.md) and regenerate client IDs.
+> Calling `Glean.setCollectionEnabled(false)` at a later point will trigger [`deletion-request` pings](../../user/pings/deletion-request.md) and regenerate client IDs.
```
travis79, line 89: I guess _technically_ we don't regenerate the client_id until `setCollectionEnabled(true)` is called. Might be worthwhile to discuss what we do with the data here also.

docs/user/reference/general/toggling-collection-status.md
```diff
@@ -121,14 +130,20 @@ uploadSwitch.addEventListener("change", event => {
 });
 ```
 
+{{#include ../../../shared/blockquote-info.html}}
+
+## Glean.js still uses `setUploadEnabled`
+
+> Glean.js did not yet switch to the new naming and continues to use `setUploadEnabled` unchanged.
```
travis79, line 137: Do we at least have a bug for the Glean.js work?

bin/build-rust-docs.sh
badboy: That's the redirect. The old `toggling-upload-status.html` will auto-redirect to the new `toggling-collection-status.html>`

badboy:
Let's merge this, it's sitting around for too long already!

PR #3032, opened 2025-01-02
Bug 1892234 - Kotlin: Dispatch experiment API on the task queue

badboy:
Extending 9d3a300c44 to also apply to the experiment API as that's (obviously) called by Nimbus early as well.

---

I should take this for a proper test drive again before releasing.


badboy:
I don't know how to properly test this in a debug build without A LOT of effort.
Enabling an experiment via the debug menu seems to not lead nimbus through the code path of `recordExperimentTelemetry`

travis79 (approved):
Looks good to me! Depending the testing you are doing, go ahead and proceed.

PR #3045, opened 2025-01-13
Update flate2 and dependencies

badboy:
This might unbreak builds on Rust 1.84.0

Which is a bit weird, but ... eh.


badboy:
.. aaaand ... it doesn't.

PR #3048, opened 2025-01-16
CI: Use nextest to run tests

badboy:
nextest is a [next-generation test runner for Rust projects.](https://nexte.st/).

This isn't to say we _should_ do it.
But it was worth a look.

It's marginally faster (though our rust tests run in ~6s once built).
nextest does have junit output, which CircleCI has integration for, see the Tests tab on https://app.circleci.com/pipelines/github/mozilla/glean/12980/workflows/14ac220e-7661-411e-a7bc-76f44421ffa7/jobs/251666/tests
When all passes it's not interesting. But when a test fails you get it separated out nicely.
And because every test runs in a separate process these additional tests failing because the global lock is poisoned might go away too.

The change here _does_ rely on two external binaries (nextest and the circleci-junit-fix), so that's a bit of a risk (as always) (but I do the usual: checksum validation against locally verified archives)

I figured I put this up as a PR at least, then we can decide what we do.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -69,10 +69,38 @@ commands:
             sudo apt install --yes --no-install-recommends \
               python3-pip \
               python3.10-venv
+
+      - run:
+          name: Install nextest
+          command: |
+            NEXTEST_SHA256=bde8d231e435099b068e654c591224defe686c784b4920682ee12784b6bfcf9e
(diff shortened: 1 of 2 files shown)


travis79 (approved):
🤩

badboy:
Let's merge then and keep an eye on it for a bit.

PR #3058, opened 2025-01-31
bug 1944694 - Add more target_arch based on those seen in the Legacy Telemetry Environment

chutten:


chutten:
Query: https://sql.telemetry.mozilla.org/queries/105271/source
This covers everything we've seen in the 1% sample of "main" pings in the past six months. Except `amd64` which `target_arch` rolls into `x86_64`. Oh well.

chutten:
When this is vendored to m-c, we can increment `Glean Count` in `TelemetryEnvironment.sys.mjs` by one as `client_info.architecture` will now cover `environment.build.architecture`.

PR #3059, opened 2025-02-04
Remove verbose trace log about auto-flushing

badboy:


chutten (approved):
Should we mention it in CHANGELOG?

badboy:
> Should we mention it in CHANGELOG?

its a trace log. that's just for ... debugging. I don't think we need one.

PR #3060, opened 2025-02-05
Bug 1946133 - GleanDebugActivity fixes for StrictMode violation in Fenix

travis79:
After doing a little research on this, I have a couple of things that should help

- Eliminates StrictMode violations (by ensuring immutable extras).
- Ensures safe intent launching (by verifying activity exports).


travis79:
cc @pollym if you wouldn't mind taking a look since you are familiar

pollym (approved):
seems reasonable to me!
it's not 100% clear from the docs what qualifies as a safely constructed `Intent`... but this certainly looks better than copying across the whole `component`... we can try it and see :)

badboy (approved):

PR #3065, opened 2025-02-21
bug 1920732 - Support uploader_capabilities

chutten:


chutten:
Oh, I see @badboy in #3069 fixed the beta rust lints differently. I'll readjust my stack once that lands.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -96,12 +99,19 @@ public class HttpPingUploader {
     ///     * data: The serialized text data to send
     ///     * headers: Map of headers from Glean to annotate ping with
     ///     * callback: A callback to return the success/failure of the upload
-    func upload(path: String, data: Data, headers: [String: String], callback: @escaping (UploadResult) -> Void) {
+    func upload(request: PingRequest, callback: @escaping (UploadResult) -> Void) {
+        // First check the uploader capabilities against the capabilites required by the ping
+        guard self.capabilities.allSatisfy(request.uploaderCapabilities.contains) else {
+            callback(.incapable(unused: 0))
+            return
+        }
```
badboy, line 107: For all other implementations we wrap the request as a `CapablePingUploadRequest` on which we put the `.capable` check. Here we don't.

Any specific reason? It is a bit of additional boilerplate, but it does nicely encapsulate that behavior in the type system.
Plus it would make the code consistent between languages.

Apart from that I think the `.allSatisfy( ... contains)` is really hard to read and grok.
As written it's also I think wrong? `request.uploaderCapabilities` needs to be a subset of `self.capabilities`,
that is: the uploader must _at least_ support all requested capabilities.
If I read the code right, it checks that all capabilities defined for the uploader are requested by the pingrequest, but the pingrequest could also ask for more.

Or as the other implementations currently do it, define that it's empty? Then we can defer the logic to when we actually implement capabilities.

badboy (changes requested):
Overall this looks ok.
I have questions regarding the iOS part (cc @travis79 )

And we have some docs at https://mozilla.github.io/glean/dev/core/internal/upload.html -- those should be updated.

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: In this case `allSatisfy` is checking that the `self.capabilities` satisfy all of the items in `request.uploaderCapabilities`, i.e. is `uploaderCapabilities` a subset of `self.capabilities`?

I guess I can see how that could be read inversely.

I can add the extra boilerplate, I just didn't figure it was necessary until we expose the ability to provide an application uploader implementation.

glean-core/ios/GleanTests/UploaderCapabilityTests.swift
```diff
@@ -0,0 +1,83 @@
...
+        // We are explicitly setting the test mode to true here to force the uploader to not
+        // run in the background, which can make this test take a long time.
+        var testValue: UploadResult?
+        stubServerReceive { _, json in
+            XCTAssert(json != nil)
+            XCTAssertEqual(json?["ping"] as? String, "test")
+        }
+        expectation = expectation(description: "Completed upload")
+
+        let ohttpCapableUploader = self.getUploader(["ohttp"])
```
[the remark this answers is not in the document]
travis79: These tests were already fragile in triggering intermittent interactions with other tests, but I'd be happy to provide more thorough test coverage as part of this. We really should get to the bottom of that, grumble grumble shared state grumble.

glean-core/ios/Glean/Net/HttpPingUploader.swift
badboy: I'm good with not introducing that yet, but then we should probably just leave out the generic check here and check that the ping request capabilities are empty.

badboy (approved):

PR #3066, opened 2025-02-24
Implement iOS uploader capabilities support

travis79:


glean-core/ios/GleanTests/UploaderCapabilityTests.swift
```diff
@@ -0,0 +1,89 @@
...
+                .incapable(unused: 0),
+                result,
+                "upload should return .incapable when capabilities don't match"
+            )
+        }
+    }
+
+    func testUploaderCapable() {
+        var request = testRequest
+        request.uploaderCapabilities = ["ohttp"]
```
chutten, line 60: Seems like, since everywhere you use `testRequest` you're giving it the sample `uploaderCapabilities` that we could instead set the `uploaderCapabilities: ["ohttp"]` in its decl and call it `ohttpRequest`.

chutten (approved):

glean-core/ios/GleanTests/UploaderCapabilityTests.swift
travis79: We could, these tests were originally just grouped with the other HttpUploaderTests and so I had made a way to pass capabilities as needed. Since Xcode so nicely forced me to move these to their own test class due to some conflict with unrelated tests, that flexibility is no longer needed.

travis79: Let me fix that up right quick

PR #3074, opened 2025-03-08
Update Kotlin to version 2.1.10

rvandermeulen:


rvandermeulen:
Looks mostly OK. I think that things will look better after rebasing this on top of an updated AGP release.

PR #3077, opened 2025-03-12
Create pylint.yml

RichardValerosoUniverse:


badboy:
We already have lint tasks and we don't use GhA for that.

PR #3087, opened 2025-03-19
Bug 1950869 - Document how GeckoView content processes record data in Glean on Android

travis79:
Adding some better docs and links to the Firefox source docs for FOG IPC

### DIFF
--- a/docs/user/language-bindings/android/gecko-content-processes-and-glean.md
+++ b/docs/user/language-bindings/android/gecko-content-processes-and-glean.md
@@ -1,0 +1,16 @@
+## GeckoView Content Processes and Glean
+
+This is a brief overview of how GeckoView content processes record data in Glean, despite Glean being required to be called only from the main process.
+
+### GeckoView Content Processes
+
+These are processes in which web page's content is rendered and collect data such as [use counters](https://firefox-source-docs.mozilla.org/dom/use-counters.html).
+
+Content processes are [launched as a "child" of the main GeckoView UI process](https://firefox-source-docs.mozilla.org/dom/ipc/process_model.html).
+
+### Recording Telemetry in Content Processes
+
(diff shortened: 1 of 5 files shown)


docs/user/language-bindings/android/gecko-content-processes-and-glean.md
```diff
@@ -0,0 +1,16 @@
+## GeckoView Content Processes and Glean
+
+This is a brief overview of how GeckoView content processes record data in Glean, despite Glean being required to be called only from the main process.
```
chutten, line 3: Wellllll.... It's more like "Only one Glean can access a single datareporting dir at a time, and since the datareporting dir is set for the whole app this means only one process can have an active Glean instance" aka the "One Glean to Rule Them All" convention

docs/user/language-bindings/android/gecko-content-processes-and-glean.md
```diff
@@ -0,0 +1,16 @@
...
+
+### GeckoView Content Processes
+
+These are processes in which web page's content is rendered and collect data such as [use counters](https://firefox-source-docs.mozilla.org/dom/use-counters.html).
```
chutten, line 7:
```suggestion
These are processes in which web page's content is rendered and data is collected such as [use counters](https://firefox-source-docs.mozilla.org/dom/use-counters.html).
```

chutten (approved):
Could land as-is, but I have some notes

docs/user/language-bindings/android/gecko-content-processes-and-glean.md
travis79: We don't really go into that in other places in the docs, we only _require_ that Glean be initialized in the main process, otherwise special care must be taken (such as passing an alternative dataDir, ala VPN's daemon process), or like this, using some actual IPC. I feel like that might be getting a little too into the weeds for the expected audience of this doc.

PR #3091, opened 2025-03-27
Bug 1955428 - Implement attribution and distribution

chutten:


glean-core/src/ping/mod.rs
```diff
@@ -175,10 +175,38 @@ impl PingMaker {
         if let Some(client_info) =
             StorageManager.snapshot_as_json(glean.storage(), "glean_client_info", true)
         {
+            log::error!("db glean_client_info: {:?}", client_info);
```
badboy, line 178: unconditional error logging again

glean-core/src/ping/mod.rs
```diff
@@ -175,10 +175,38 @@ impl PingMaker {
...
             let client_info_obj = client_info.as_object().unwrap(); // safe unwrap, snapshot always returns an object.
-            for (_key, value) in client_info_obj {
-                merge(&mut map, value);
+            for (_metric_type, metrics) in client_info_obj {
+                merge(&mut map, metrics);
             }
+            log::error!("flattened client_info: {:?}", map);
```
badboy, line 183: unconditional error logging again

glean-core/src/ping/mod.rs
```diff
@@ -175,10 +175,38 @@ impl PingMaker {
...
+            let map = map.as_object_mut().unwrap(); // safe unwrap, we created the object above.
+            let mut attribution = serde_json::Map::new();
+            let mut distribution = serde_json::Map::new();
+            map.retain(|name, value| {
+                let split: Vec<_> = name.split('.').collect();
```
badboy, line 188: I stumbled here at first.
Technically the category can contain dots.
I guess here it is correct because we control all metrics in client_info and we don't have dotted categories.

But do we need to collect? That definitely allocates. If we just make use of the iterator (`split.next()`) we could avoid that.

glean-core/src/ping/mod.rs
```diff
@@ -175,10 +175,38 @@ impl PingMaker {
...
+                    true
+                }
+            });
+            if !attribution.is_empty() {
+                map.insert("attribution".into(), serde_json::Value::from(attribution));
+            }
+            if !distribution.is_empty() {
+                map.insert("distribution".into(), serde_json::Value::from(distribution));
+            }
+            log::error!("reconstituted client_info: {:?}", map);
```
badboy, line 209: unconditional error logging again

badboy (changes requested):

glean-core/src/ping/mod.rs
chutten: I wanted a nice way to name the parts of the split, but you're right. I'll work around it.

chutten:
This will now no longer build until mozilla/glean_parser#793 makes it into a release and we pull that down.

badboy (approved):

PR #3097, opened 2025-04-01
Revert "Bug 1946133 - GleanDebugActivity fixes for StrictMode violation in Fenix

badboy:
This reverts commit 24165f9fe177ccd1c0b43cbb2db1434fe9da0d00.

---

It seems this is currently causing issues in the raptor/browsertime tests. Or so I think. But the only way to find out is to release and get that into m-c for a try run


travis79 (commented):
approved, of course, though late

PR #3098, opened 2025-04-01
Sync AGP and Kotlin versions with m-c

rvandermeulen:


badboy (approved):
Rebased, to make sure the changelog is in the right place.

PR #3101, opened 2025-04-02
Bug 1892234 - Kotlin: Dispatch ping instantiation and submit on the task queue

badboy:
This also ensures we run through the Kotlin task queue _before_ intializing Glean to keep the order correct.

This should mitigate yet another early libxul loading in Fenix due to Nimbus sending an early ping.

---

I tested this locally in a Fenix build, profiling startup using [simpleperf and samply](https://firefox-source-docs.mozilla.org/performance/profiling_with_simpleperf.html).
The profile doesn't show any `com.sun.jna.Native.load` or glean-core call on the `NimbusDbScope-t` stack.

### DIFF
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
@@ -5,6 +5,7 @@
 package mozilla.telemetry.glean.private
 
 import androidx.annotation.VisibleForTesting
+import mozilla.telemetry.glean.Dispatchers
 import mozilla.telemetry.glean.internal.PingType as GleanPingType
 
 /**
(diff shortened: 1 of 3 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/PingType.kt
```diff
@@ -109,8 +112,10 @@ class PingType<ReasonCodesEnum> (
         }
         this.testCallback = null
 
-        val reasonString = reason?.let { this.reasonCodes[it.code()] }
-        this.innerPing.submit(reasonString)
```
badboy, line 113: I wasn't actually sure whether to also dispatch the callback here?
I guess in tests it won't really matter much, because the time between calling submit and glean being initialized and thus running the glean-core submit should be small.

chutten: RLB doesn't (because it doesn't have access to the glean_core dispatcher, and likely also for the reasons you've identified) so I'd keep it sync here for consistency.

chutten (changes requested):
Needs test, CHANGELOG

badboy:
> Needs test, CHANGELOG

Changelog yes.
Tests: it's covered. Tests were breaking when I messed up the implementation at first. Not sure what specific test to write for this.

travis79 (approved):

chutten (approved):

PR #3105, opened 2025-04-08
Correct length limits for metric categories, names and labels everywhere

badboy:
[doc only]


glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -65,7 +65,7 @@ class LabeledMetricType<T>(
      * only the first 16 unique labels will be used. After that, any additional
      * labels will be recorded under the special `__other__` label.
      *
-     * Labels must be snake_case and less than 30 characters. If an invalid label
+     * Labels must not exceed 71 characters in length, and may comprise any printable ASCII characters.
```
travis79, line 68: "not exceed 71 characters..."   Do you mean I can _use_ 71 chars and 72 would "exceed" that?  This isn't clear to me as 71 seems like an odd number, I would have expected us to settle on 70 instead and say: "not exceed 70 chars" or "must be 70 chars or less" or something.

PR #3109, opened 2025-04-10
bug 1959771 - Ensure there's a global Glean before using it

chutten:
I _would_ prefer to merely check the global_glean, but in tests we reset INITIALIZE_CALLED but we don't reset core::GLEAN, so here we are.


travis79 (approved):
Yep, here we are.

PR #3118, opened 2025-05-02
Update NDK to r28b

rvandermeulen:
Not a big update, but makes it a bit easier to work on the Glean repo if you have a bootstrapped Firefox environment since it's on r28b now also.


badboy (approved):
CI failure looks weird, rerunning it before landing.

PR #3120, opened 2025-05-05
Bug 1960592 - Measure memory usage of the Glean global object

badboy:


badboy:
Will be rebased once #3163 lands

PR #3130, opened 2025-05-21
Python: Properly export rate/text metric

badboy:
Turns out we never actually exposed them and thus they were unusable, even though we claimed it exists!

(currently based against #3129, will merge cleanly into main after it)


badboy:
wait what. lol. Ok, auto-merge is not a good idea when it doesn't go into `main` :D

PR #3135, opened 2025-05-22
[WIP] DualLabeledCounter implementation as replacement for keyed categorical legacy type

travis79:
cc @badboy

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -2,11 +2,14 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use std::fmt::Display;
+use std::ops::Deref;
 use std::sync::atomic::{AtomicU8, Ordering};
 
 use malloc_size_of_derive::MallocSizeOf;
 
 use crate::error::{Error, ErrorKind};
+use crate::metrics::dual_labeled_counter::validate_dynamic_key_and_or_category;
 use crate::metrics::labeled::validate_dynamic_label;
 use crate::Glean;
 use serde::{Deserialize, Serialize};
@@ -72,7 +75,52 @@ pub struct CommonMetricData {
     /// metric to be recorded to, dynamic labels are stored in the specific
     /// label so that we can validate them when the Glean singleton is
     /// available.
-    pub dynamic_label: Option<String>,
+    pub dynamic_label: Option<DynamicLabelType>,
+}
+
+// TODO FIX ERRORS. Similar errors see for all variants, not sure how to specify this in UniFFI idl:
+/*
+error[E0769]: tuple variant `Self::Label` written as struct variant
+   --> <redacted>/glean/target/debug/build/glean-core-7b31c5607feb2bbe/out/glean.uniffi.rs:125:1
+    |
+125 | #[::uniffi::udl_derive(Enum)]
+    | ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
+    |
+    = note: this error originates in the attribute macro `::uniffi::udl_derive` (in Nightly builds, run with -Z macro-backtrace for more info)
+
+*/
+
+#[derive(Debug, Clone, Deserialize, Serialize, MallocSizeOf)]
+pub enum DynamicLabelType {
+    Label(String),
+    KeyOnly(String),
+    CategoryOnly(String),
+    KeyAndCategory(String),
+}
+
+impl Default for DynamicLabelType {
+    fn default() -> Self {
+        Self::Label(String::new())
+    }
+}
+
+impl Deref for DynamicLabelType {
+    type Target = str;
+
+    fn deref(&self) -> &Self::Target {
+        match self {
+            DynamicLabelType::Label(label) => label,
+            DynamicLabelType::KeyOnly(key) => key,
+            DynamicLabelType::CategoryOnly(category) => category,
+            DynamicLabelType::KeyAndCategory(key_and_category) => key_and_category,
+        }
+    }
+}
+
+impl Display for DynamicLabelType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.to_string())
+    }
 }
 
 #[derive(Default, Debug, MallocSizeOf)]
@@ -139,7 +187,17 @@ impl CommonMetricDataInternal {
         let base_identifier = self.base_identifier();
 
         if let Some(label) = &self.inner.dynamic_label {
-            validate_dynamic_label(glean, self, &base_identifier, label)
+            match label {
+                DynamicLabelType::Label(label) => {
+                    validate_dynamic_label(glean, self, &base_identifier, label)
+                }
+                _ => validate_dynamic_key_and_or_category(
+                    glean,
+                    self,
+                    &base_identifier,
+                    label.clone(),
+                ),
+            }
         } else {
             base_identifier
         }
--- a/glean-core/src/metrics/dual_labeled_counter.rs
+++ b/glean-core/src/metrics/dual_labeled_counter.rs
@@ -1,0 +1,418 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::borrow::Cow;
+use std::collections::HashMap;
+use std::mem;
+use std::sync::{Arc, Mutex};
+
+use malloc_size_of::MallocSizeOf;
+
+use crate::common_metric_data::{CommonMetricData, CommonMetricDataInternal, DynamicLabelType};
+use crate::error_recording::{record_error, test_get_num_recorded_errors, ErrorType};
+use crate::metrics::{CounterMetric, Metric, MetricType};
+use crate::Glean;
+
+const MAX_LABELS: usize = 16;
+const OTHER_LABEL: &str = "__other__";
+const MAX_LABEL_LENGTH: usize = 111;
+const RECORD_SEPARATOR: u8 = 0x1E;
+
+pub trait DualLabeledMetricType {
+    /// Create a new metric from this with a specific label.
+    fn with_dynamic_key(&self, _key: String) -> Self
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+
+    /// Create a new metric from this with a specific label.
+    fn with_dynamic_category(&self, _category: String) -> Self
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+
+    fn with_dynamic_key_and_category(&self, _key: String, _category: String) -> Self
+    where
+        Self: Sized,
+    {
+        unimplemented!()
+    }
+}
+
+/// A dual labled metric, mirrors the Legacy keyed-categorical type.
+///
+/// Dual labled metrics allow recording multiple sub-metrics of the same type, in relation
+/// to two dimensions rather than the single label provided by the standard labeled type.
+#[derive(Debug)]
+pub struct DualLabeledCounterMetric {
+    keys: Option<Vec<Cow<'static, str>>>,
+    categories: Option<Vec<Cow<'static, str>>>,
+    /// Type of the underlying metric
+    /// We hold on to an instance of it, which is cloned to create new modified instances.
+    counter: CounterMetric,
+
+    /// A map from a unique ID for the dual labeled submetric to a handle of an instantiated
+    /// metric type.
+    dual_label_map: Mutex<HashMap<(String, String), Arc<CounterMetric>>>,
+}
+
+impl ::malloc_size_of::MallocSizeOf for DualLabeledCounterMetric {
+    fn size_of(&self, ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+        // TODO figure out how to measure the sizeof these metrics
+
+        // let map = self.dual_label_map.lock().unwrap();
+        // let keys: Vec<String> = map.keys().collect();
+        // let categories = map.values().for_each(|category| -> String  { category.keys() });
+
+        // // Copy of `MallocShallowSizeOf` implementation for `HashMap<K, V>` in `wr_malloc_size_of`.
+        // // Note: An instantiated submetric is behind an `Arc`.
+        // // `size_of` should only be called from a single thread to avoid double-counting.
+        // let shallow_size = if ops.has_malloc_enclosing_size_of() {
+        //     map.values()
+        //         .next()
+        //         .map_or(0, |v| unsafe { ops.malloc_enclosing_size_of(v) })
+        // } else {
+        //     map.capacity()
+        //         * (mem::size_of::<String>() + mem::size_of::<T>() + mem::size_of::<usize>())
+        // };
+
+        // let mut map_size = shallow_size;
+        // for (k, v) in map.iter() {
+        //     map_size += k.size_of(ops);
+        //     map_size += v.size_of(ops);
+        // }
+
+        // self.labels.size_of(ops) + self.submetric.size_of(ops) + map_size
+        0
+    }
+}
+
+impl DualLabeledCounterMetric {
+    /// Creates a new dual labeled counter from the given metric instance and optional list of labels.
+    pub fn new(
+        meta: CommonMetricData,
+        keys: Option<Vec<Cow<'static, str>>>,
+        catgories: Option<Vec<Cow<'static, str>>>,
+    ) -> DualLabeledCounterMetric {
+        let submetric = CounterMetric::new(meta);
+        DualLabeledCounterMetric::new_inner(submetric, keys, catgories)
+    }
+
+    fn new_inner(
+        counter: CounterMetric,
+        keys: Option<Vec<Cow<'static, str>>>,
+        categories: Option<Vec<Cow<'static, str>>>,
+    ) -> DualLabeledCounterMetric {
+        let dual_label_map = Default::default();
+        DualLabeledCounterMetric {
+            keys,
+            categories,
+            counter,
+            dual_label_map,
+        }
+    }
+
+    /// Creates a new metric with a specific label.
+    ///
+    /// This is used for dynamic labels where we have to actually validate and correct the
+    /// label later when we have a Glean object.
+    fn new_counter_metric(&self, key: &str, category: &str) -> CounterMetric {
+        // Determine which type of dynamic label it is
+        if self.keys.is_none() && self.categories.is_none() {
+            self.counter
+                .with_dynamic_label(DynamicLabelType::KeyAndCategory(
+                    make_label_from_key_and_category(key, category),
+                ))
+        } else if self.keys.is_none() {
+            let static_category = self.static_category(category);
+            self.counter.with_dynamic_label(DynamicLabelType::KeyOnly(
+                make_label_from_key_and_category(key, static_category),
+            ))
+        } else if self.categories.is_none() {
+            let static_key = self.static_key(key);
+            self.counter
+                .with_dynamic_label(DynamicLabelType::CategoryOnly(
+                    make_label_from_key_and_category(static_key, category),
+                ))
+        } else {
+            // Both labels are static and can be validated now
+            let static_key = self.static_key(key);
+            let static_category = self.static_category(category);
+            let name = combine_base_identifier_and_labels(
+                self.counter.meta().base_identifier().as_str(),
+                static_key,
+                static_category,
+            );
+            self.counter.with_name(name)
+        }
+    }
+
+    /// Creates a static label for the key dimension.
+    ///
+    /// # Safety
+    ///
+    /// Should only be called when static labels are available on this metric.
+    ///
+    /// # Arguments
+    ///
+    /// * `key` - The requested key
+    ///
+    /// # Returns
+    ///
+    /// The requested key if it is in the list of allowed labels.
+    /// Otherwise `OTHER_LABEL` is returned.
+    fn static_key<'a>(&self, key: &'a str) -> &'a str {
+        debug_assert!(self.keys.is_some());
+        let keys = self.keys.as_ref().unwrap();
+        if keys.iter().any(|l| l == key) {
+            key
+        } else {
+            OTHER_LABEL
+        }
+    }
+
+    /// Creates a static label for the category dimension.
+    ///
+    /// # Safety
+    ///
+    /// Should only be called when static labels are available on this metric.
+    ///
+    /// # Arguments
+    ///
+    /// * `category` - The requested category
+    ///
+    /// # Returns
+    ///
+    /// The requested category if it is in the list of allowed labels.
+    /// Otherwise `OTHER_LABEL` is returned.
+    fn static_category<'a>(&self, category: &'a str) -> &'a str {
+        debug_assert!(self.categories.is_some());
+        let categories = self.categories.as_ref().unwrap();
+        if categories.iter().any(|l| l == category) {
+            category
+        } else {
+            OTHER_LABEL
+        }
+    }
+
+    /// Gets a specific metric for a given key/category combination.
+    ///
+    /// If a set of acceptable labels were specified in the `metrics.yaml` file,
+    /// and the given label is not in the set, it will be recorded under the special `OTHER_LABEL` label.
+    ///
+    /// If a set of acceptable labels was not specified in the `metrics.yaml` file,
+    /// only the first 16 unique labels will be used.
+    /// After that, any additional labels will be recorded under the special `OTHER_LABEL` label.
+    ///
+    /// Labels must have a maximum of 111 characters, and may comprise any printable ASCII characters.
+    /// If an invalid label is used, the metric will be recorded in the special `OTHER_LABEL` label.
+    pub fn get<S: AsRef<str>>(&self, key: S, category: S) -> Arc<CounterMetric> {
+        let key = key.as_ref();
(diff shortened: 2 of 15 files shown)


glean-core/src/common_metric_data.rs
```diff
@@ -72,7 +75,52 @@ pub struct CommonMetricData {
...
+
+*/
+
+#[derive(Debug, Clone, Deserialize, Serialize, MallocSizeOf)]
+pub enum DynamicLabelType {
+    Label(String),
+    KeyOnly(String),
+    CategoryOnly(String),
+    KeyAndCategory(String),
+}
```
badboy, line 99: That's because in the UDL you have it defined as

    enum DynamicLabelType {
        "Label",
        "KeyAndCategory",
    };

which is not the same.
What we _really_ need is not expressible in UDL though, so we can do this in UDL:

```
typedef enum DynamicLabelType;
```

Then the following should tell UniFFI about it:

```rust
#[derive(Debug, Clone, Deserialize, Serialize, MallocSizeOf, uniffi::Enum)]
pub enum DynamicLabelType {
    Label(String),
    KeyOnly(String),
    CategoryOnly(String),
    KeyAndCategory(String),
}
```

glean-core/src/common_metric_data.rs
```diff
@@ -72,7 +75,52 @@ pub struct CommonMetricData {
...
+            DynamicLabelType::KeyAndCategory(key_and_category) => key_and_category,
+        }
+    }
+}
+
+impl Display for DynamicLabelType {
+    fn fmt(&self, f: &mut std::fmt::Formatter<'_>) -> std::fmt::Result {
+        write!(f, "{}", self.to_string())
+    }
 }
```
badboy, line 124: I don't think we need display (plus this potentially double allocates due to the `to_string()`). Leave it out.

```suggestion
```

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,418 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
```
badboy, line 3: Picking an arbitrary place to comment here.

As written storing data might work, but currently the storage manager has no knowledge about those dual labeled metrics.
`DualLabeledCounter` in `metrics/mod.rs` is also _the first_ labeled metric that gets its own entry as a metric.

Notably we don't treat "labeled" metrics as its own data type in the database.
We encode the label into the _key_, so the value is a "pure" metric (counter, bool, etc).

This should work similar for dual-labeled metrics.

The key would be something like:

```
ping#category.name\x{1E}key\x{1E}category
```

and the value just be a number (for counter), just as before.
See also https://mozilla.github.io/glean/dev/core/internal/database.html

This will require changes to handle it properly in `src/storage/mod.rs`.
See `snapshot_labeled_metrics` and its use.

glean-core/src/metrics/mod.rs
```diff
@@ -336,6 +343,9 @@ impl Metric {
             Metric::Object(s) => {
                 serde_json::from_str(s).expect("object storage should have been json")
             }
+            Metric::DualLabeledCounter(s) => {
+                json!(s)
```
badboy, line 347: See the other comment, we probably don't need this.
But apart from that, this wouldn't even work. We cannot serialize a `HashMap<(String, String), ...>` to JSON, as all JSON object keys have to be strings.

Again, we need to handle serialization in some other place anyway.

badboy (commented):
I gave this a first look. Definitely in a rough shape and far from being ready.
First comments inline.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,418 @@
...
+    /// If a set of acceptable labels were specified in the `metrics.yaml` file,
+    /// and the given label is not in the set, it will be recorded under the special `OTHER_LABEL` label.
+    ///
+    /// If a set of acceptable labels was not specified in the `metrics.yaml` file,
+    /// only the first 16 unique labels will be used.
+    /// After that, any additional labels will be recorded under the special `OTHER_LABEL` label.
+    ///
+    /// Labels must have a maximum of 111 characters, and may comprise any printable ASCII characters.
+    /// If an invalid label is used, the metric will be recorded in the special `OTHER_LABEL` label.
+    pub fn get<S: AsRef<str>>(&self, key: S, category: S) -> Arc<CounterMetric> {
```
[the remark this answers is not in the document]
travis79: That sounds reasonable, it just seemed like `get(key, category)` felt more "proper" than subscript access when I started. I have no strong preference of subscript vs `get`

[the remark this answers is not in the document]
badboy: the double-subscript will happen in the foreign languages. In Rust we might end up with `metric.get(key).get(category)` I guess in the RLB?
Either way, `get(key, category)` in `glean-core` should allow us to implement this regardless.

docs/user/reference/metrics/dual_labeled_counters.md
```diff
@@ -0,0 +1,213 @@
+# Dual Labeled Counters
+
+Dual labeled counters are used to record different related counts that should sum up to a total.
```
badboy:
```suggestion
Dual labeled counters allow two levels of labels for a counter. This metric type allows recording counts along two dimensions.
```

The "sum up to a total" sounds weird. I took the above phrasing from the proposal. I'm open for other phrasings though.

docs/user/reference/metrics/dual_labeled_counters.md
```diff
@@ -0,0 +1,213 @@
...
+</div>
+
+<div data-lang="Firefox Desktop" class="tab">
+
+**C++**
+```cpp
+#include "mozilla/glean/GleanUpload.h"
+
+mozilla::glean::glean_upload::failures.Get("metrics"_ns, "recoverable network error"_ns).Add(1);
+mozilla::glean::glean_upload::failures.Get("baseline"_ns, "4xx").Add(3);
```
badboy: Note that this is different from what the proposal specified.
It had this:

```c++
send_failure_type_per_ping.Get("baseline"_ns).EnumGet(mozilla::glean::uploader::SendFailureTypePerPingCategory::eOk).Add(1);
```

docs/user/reference/metrics/dual_labeled_counters.md
```diff
@@ -0,0 +1,213 @@
...
+```
+
+**JavaScript**
+```js
+Glean.gleanUpload.failures["baseline"]["4xx"].add(3);
```
badboy: Somehow we get the double-index version here?
Maybe it makes sense to leave out the Desktop API until we have it implemented.

glean-core/rlb/src/test.rs
```diff
@@ -1282,6 +1284,38 @@ fn test_boolean_get_num_errors() {
...
+            lifetime: Lifetime::Ping,
+            ..Default::default()
+        },
+        Some(vec!["key1".into()]),
+        Some(vec!["category1".into()]),
+    );
+
+    metric.get("key1", "category1").add(1);
+
+    // Check for both the valid and invalid "__other__" key values
```
badboy: The comment doesn't match the code.

glean-core/src/storage/mod.rs
```diff
@@ -42,6 +42,33 @@ fn snapshot_labeled_metrics(
     obj.insert(label.into(), metric.as_json());
 }
 
+/// Labeled metrics are stored as `<metric id>/<label>`.
```
badboy: Copy-pasta. Should be adjusted for dual labeled.

glean-core/src/storage/mod.rs
```diff
@@ -42,6 +42,33 @@ fn snapshot_labeled_metrics(
...
+/// add to it.
+fn snapshot_dual_labeled_metrics(
+    snapshot: &mut HashMap<String, HashMap<String, JsonValue>>,
+    metric_id: &str,
+    metric: &Metric,
+) {
+    let ping_section = format!("dual_labeled_{}", metric.ping_section());
+    let map = snapshot.entry(ping_section).or_default();
+
+    // Safe unwrap, the function is only called when the id does contain a '/'
```
badboy: Wrong comment. There is no unwrap here.

glean-core/src/storage/mod.rs
```diff
@@ -42,6 +42,33 @@ fn snapshot_labeled_metrics(
...
+    let parts = metric_id.split('\x1e').collect::<Vec<&str>>();
```
badboy: Might be good to reuse the defined constant. But that requires to make it `pub(crate)` over in `glean-core/src/metrics/dual_labeled_counter.rs` and then also a `use` statement.

glean-core/src/storage/mod.rs
```diff
@@ -42,6 +42,33 @@ fn snapshot_labeled_metrics(
...
+
+    let obj = map
+        .entry(parts[0].into())
+        .or_insert_with(|| json!({}))
+        .as_object_mut()
+        .unwrap();
+    // safe unwrap, we constructed the object above
```
badboy: That comment should probably be inline with the `unwrap`

glean-core/src/storage/mod.rs
```diff
@@ -42,6 +42,33 @@ fn snapshot_labeled_metrics(
...
+    let key_obj = obj.entry(parts[1].to_string()).or_insert_with(|| json!({}));
+    let key_obj = key_obj.as_object_mut().unwrap();
+    key_obj.insert(parts[2].into(), metric.as_json());
```
badboy, line 67: We should have _some_ check to ensure we actually have 2 separators and thus 2 parts in the id.
The function below only checks that it contains at least one.

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    let metric = dual_labeled_counter.get("non-ASCII-key�", "non-ASCII-cat�");
+    metric.add_sync(&glean, 1);
+
+    let snapshot = StorageManager
+        .snapshot_as_json(glean.storage(), "store1", true)
+        .unwrap();
+
+    // We end up with 2x InvalidLabel errors, one each for key and category which both contain
+    // invalid characters.
```
badboy: Why? I thought we allow any UTF-8 characters for dynamic labels?

badboy: Ah, the `validate_dynamic_key_and_or_category` -> `label_is_valid` call chain has the `is_ascii` check. That should probably go away.

badboy: Once that is gone we should also have an explict test that uses a label that has `\x1e` in there. We don't want _those_ labels (and I guess we should call that out in the docs?)

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+    counter: CounterMetric,
+
+    /// A map from a unique ID for the dual labeled submetric to a handle of an instantiated
+    /// metric type.
+    dual_label_map: Mutex<HashMap<(String, String), Arc<CounterMetric>>>,
+}
+
+// TODO figure out how to measure the sizeof these metrics
+// impl ::malloc_size_of::MallocSizeOf for DualLabeledCounterMetric {
+//     fn size_of(&self, ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
```
badboy: Can you implement this as just returning `0`? Then at least for now it should work in m-c.
I can fix that up in a followup.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+use crate::error_recording::{record_error, test_get_num_recorded_errors, ErrorType};
+use crate::metrics::{CounterMetric, Metric, MetricType};
+use crate::Glean;
+
+const MAX_LABELS: usize = 16;
+const OTHER_LABEL: &str = "__other__";
+const MAX_LABEL_LENGTH: usize = 111;
+const RECORD_SEPARATOR: char = '\x1E';
+
+/// A dual labled metric, mirrors the Legacy keyed-categorical type.
```
badboy: We shouldn't refer to the legacy type here. Glean doesn't know anything about Legacy.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+///
+/// Dual labled metrics allow recording multiple sub-metrics of the same type, in relation
+/// to two dimensions rather than the single label provided by the standard labeled type.
+#[derive(Debug)]
+pub struct DualLabeledCounterMetric {
+    keys: Option<Vec<Cow<'static, str>>>,
+    categories: Option<Vec<Cow<'static, str>>>,
```
badboy, line 27: Do we need the `Option` here? Can we treat an empty Vec as no static keys/categories?
(Size-wise it doesn't matter, thanks to niche optimizations both types are 24 bytes)

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+            categories,
+            counter,
+            dual_label_map,
+        }
+    }
+
+    /// Creates a new metric with a specific label.
+    ///
+    /// This is used for dynamic labels where we have to actually validate and correct the
+    /// label later when we have a Glean object.
```
badboy: This comment is slightly incorrect.
For labeled metrics we split things into two functions, but here both static and dynamic labels are handled in this function.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+            // Both labels are static and can be validated now
+            let static_key = self.static_key(key);
+            let static_category = self.static_category(category);
+            let name = combine_base_identifier_and_labels(
+                self.counter.meta().inner.name.as_str(),
+                static_key,
+                static_category,
+            );
+            self.counter.with_name(name)
+        }
```
badboy: I find if/elseif/else chains a bit hard to read.
What do you think about using a match here?

```rust
        match (&self.keys, &self.categories) {
            (None, None) => /* ... */
            (None, _) => /* ... */
            (_, None) => /* ... */
            _  => /* ... */
        }
```

glean-core/src/metrics/dual_labeled_counter.rs
badboy: Ah I guess we do the same for single-labeled metrics. Fine keeping it consistent then.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+pub fn combine_base_identifier_and_labels(
+    base_identifer: &str,
+    key: &str,
+    category: &str,
+) -> String {
+    format!(
+        "{}{}",
+        base_identifer,
+        make_label_from_key_and_category(key, category)
+    )
```
badboy, line 228: I kinda hate the double allocation here, but fine for now.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+/// Separate label into key and category components
+pub fn separate_label_into_key_and_category(label: &str) -> Option<(&str, &str)> {
+    label
+        .strip_prefix(RECORD_SEPARATOR)
+        .unwrap_or(label)
+        .split_once(RECORD_SEPARATOR)
+        .map(|s| (s.0, s.1))
```
badboy: This can go, you already get a `Option<(&str, &str)>` from `split_once`

```suggestion
```

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+fn label_is_valid(label: &str, glean: &Glean, meta: &CommonMetricDataInternal) -> bool {
+    if label.len() > MAX_LABEL_LENGTH {
+        let msg = format!(
+            "label length {} exceeds maximum of {}",
+            label.len(),
+            MAX_LABEL_LENGTH
+        );
+        record_error(glean, meta, ErrorType::InvalidLabel, msg, None);
+        false
+    } else if label.chars().any(|c| !c.is_ascii() || c.is_ascii_control()) {
```
badboy: As mentioned before we should not need the ascii check here.
Though we should check for `\x1e` and reject it.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+}
+
+fn count_keys_and_categories(meta: &CommonMetricDataInternal, glean: &Glean) -> (usize, usize) {
+    let base_identifier = &meta.base_identifier();
+    let prefix = format!("{base_identifier}{RECORD_SEPARATOR}");
+    let mut seen_keys: Vec<String> = Vec::new();
+    let mut seen_categorires: Vec<String> = Vec::new();
+    let mut snapshotter = |mut metric_id: &[u8], _: &Metric| {
+        let mut metric_id_str = String::new();
+        _ = metric_id.read_to_string(&mut metric_id_str);
```
badboy: You should be able to use this:

```suggestion
        let mut metric_id_str = str::from_utf8(metric_id);
```

or potentially even better:

```suggestion
	let metric_id_str = String::from_utf8_lossy(metric_id);
```

No allocation needed if this is utf-8 (which it should be, because that's what we put in).

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+
+        if parts.len() == 2 {
+            if !seen_keys.contains(&parts[0].into()) {
+                seen_keys.push(parts[0].into());
+            }
+            if !seen_categorires.contains(&parts[1].into()) {
+                seen_categorires.push(parts[1].into());
+            }
+        } else {
+            println!("DualLabeledCounter label does not contain exactly 2 parts");
```
badboy: Stray `println`. Should be an error or something.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+        let mut metric_id_str = String::new();
+        _ = metric_id.read_to_string(&mut metric_id_str);
+
+        // Split full identifier on the ASCII Record Separator (\x1e)
+        let parts: Vec<&str> = metric_id_str.split(RECORD_SEPARATOR).collect();
+
+        if parts.len() == 2 {
+            if !seen_keys.contains(&parts[0].into()) {
+                seen_keys.push(parts[0].into());
+            }
```
badboy: Use a `HashSet` for `seen_keys` and `seen_categories` instead. Then all you need is `insert`.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+    } else {
+        true
+    }
+}
+
+fn count_keys_and_categories(meta: &CommonMetricDataInternal, glean: &Glean) -> (usize, usize) {
+    let base_identifier = &meta.base_identifier();
+    let prefix = format!("{base_identifier}{RECORD_SEPARATOR}");
+    let mut seen_keys: Vec<String> = Vec::new();
+    let mut seen_categorires: Vec<String> = Vec::new();
```
badboy:
```suggestion
    let mut seen_categories: Vec<String> = Vec::new();
```

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -0,0 +1,375 @@
...
+    };
+
+    let lifetime = meta.inner.lifetime;
+    for store in &meta.inner.send_in_pings {
+        glean
+            .storage()
+            .iter_store_from(lifetime, store, Some(&prefix), &mut snapshotter);
+    }
+
+    (seen_keys.len(), seen_categorires.len())
```
badboy: A thing I realized here:
This means we count categories, regardless of under which `key` they are.
So if we have

`key: a, categories: [0, 1, ... 14]`
`key: b, categories: [15]`

then `key: c` will _always_ end up recording to `__other__`.
that seems a bit unintuitive.
I assumed it's 16 categories per unique key.
Of course that's harder to implement.

The proposal is vague about this, so I could be convinced the implemented behavior is correct.

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -0,0 +1,527 @@
...
+    let metric = dual_labeled_counter.get("key1", "category1");
+    metric.add_sync(&glean, 1);
+
+    let metric = dual_labeled_counter.get("key1", "category2");
+    metric.add_sync(&glean, 2);
+
+    let metric = dual_labeled_counter.get("key2", "category1");
+    metric.add_sync(&glean, 3);
+
+    let metric = dual_labeled_counter.get("key2", "category2");
```
badboy: To cover one more case you could make this `category3` to show that the categories are independent per key.

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -0,0 +1,527 @@
...
+
+    // Make sure that the errors have been recorded
+    assert_eq!(
+        Ok(1),
+        test_get_num_recorded_errors(&glean, metric.meta(), ErrorType::InvalidValue)
+    );
+}
+
+#[test]
+fn labels_are_checked_against_static_list() {
```
badboy, line 130: Two test that are missing:

* Static keys, dynamic categories
* Dynamic keys, static categories

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -0,0 +1,527 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
```
badboy, line 4: Regardless of which behavior we decide with the 16 categories/16 keys issue I mentioned above, we should have a test for that behavior.

badboy (changes requested):

glean-core/src/metrics/dual_labeled_counter.rs
travis79: I discussed this with @chutten and this was the behavior we arrived at. I thought this too but you can have 16 distinct keys and 16 distinct categories, not 16 distinct categories _per_ key.

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -525,3 +483,46 @@ fn caching_metrics_with_dynamic_labels_across_pings() {
...
+    let metric = dual_labeled_counter.get("record\x1eseparator", "category1");
+    metric.add_sync(&glean, 1);
+
+    let snapshot = StorageManager
+        .snapshot_as_json(glean.storage(), "store1", true)
+        .unwrap();
+
+    // We end up with 1x InvalidLabel error because the key contains an ASCII record separator, but both
+    // labels end up as other because the use of the record separator prevents us from knowing which one
+    // contained it (or both) at this point in exection.
```
badboy, line 660: damn. Guess that's truly an edge case, so not worth "fixing". In the future we might properly store labels separately and then we don't need to care about separators anymore.

docs/user/reference/metrics/dual_labeled_counters.md
```diff
@@ -63,7 +49,7 @@ Glean.gleanUpload.failures["baseline"]["4xx"].add(3);
     These labels may contain any UTF-8 characters.
   * 4096 keys and 4096 categories if specified as static labels in `metrics.yaml`, see [keys](#keys) and [categories](#categories).
     Unknown keys and categories will be recorded under the special label `__other__`.
-    These labels may only be made of printable ASCII characters.
+    Each label may contain any utf-8 character _except_ for the ASCII record separator (0x1E).
```
badboy: I think I was wrong here.
The static labels _are_ limited to ASCII characters.

The record separator part must go into the above dynamic labels point.

glean-core/tests/dual_labeled_counter.rs
travis79: That required putting the storage on every metric because of where it was located (`CommonMetricData`), so I figured that would have negatively impacted memory size for Glean.

badboy: Yup. It's fine for now.

glean-core/tests/dual_labeled_counter.rs
```diff
@@ -411,6 +411,154 @@ fn caching_metrics_with_dynamic_labels() {
...
+            Some(4),
+            recorded_value,
+            "Other value should be 4, found {recorded_value:?} for key{i}"
+        );
+    }
+
+    // The above would have asked for "key5", but since that doesn't exist as a static key it should have fallen into
+    // `__other__` as a key, the way `get()` works the above assert wouldn't fail but the actual key would have been
+    // renamed. Let's check for that explicitly, just to verify.
+    let other = dual_labeled_counter.get(format!("__other__"), "__other__".into());
```
badboy:
```suggestion
    let other = dual_labeled_counter.get("__other__".into(), "__other__".into());
```

badboy: ah hey, that's what clippy is complaining about

travis79: whoops, forgot to swap this one out

docs/user/reference/metrics/dual_labeled_counters.md
travis79: Leaving this out until the Desktop C++/JS API is ironed out

badboy (approved):
All my concerns have been addressed. Needs a rebase against `main` and fix up the changelog entry then.

PR #3157, opened 2025-06-06
Bug 1970856 - Count distinct labels before spilling into __other__ bucket

badboy:
If a labeled metrics records in MULTIPLE pings every ping counts as a use of a label.

That means a single labeled metric recorded into 16+ pings can only ever use a single label, before new labels being changed to `__other__`.

We don't have any metric that goes into 16 pings.
But we probably have some that go into 2 pings.
That means this metric can only use 8 labels maximum.

That's bad.


badboy:
I added a commit that would fix the test. I still need to rewrite the test to not expose internals (through `pub`).

PR #3162, opened 2025-06-10
Run `cargo nextest` when available

badboy:
This falls back to regular ol' `cargo test` if that command is not available. Should make local testing more convenient.
On CI we still invoke `cargo nextest` explicitly.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -72,7 +72,11 @@ glean-core/python/glean/_uniffi/__init__.py:
 test: test-rust
 
 test-rust: ## Run Rust tests for glean-core and glean-ffi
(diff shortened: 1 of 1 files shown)


badboy:
> TIL `cargo nextest`
> 
> Do we want to update the ["Making Code Changes" section](https://mozilla.github.io/glean/dev/contributing.html) to suggest the installation of `nextest`?

Woops, this was set to auto-merge-on-r+.
We do mention it on https://mozilla.github.io/glean/dev/testing.html which even comes _before_ contributong (guess that whole thing should be really re-ordered)

PR #3165, opened 2025-06-13
chore: Update `time` to 0.3.41

larseggert:
And deal with the fallout. The code now calls some deprecated-but-still-supported functions; I made no effort to switch to the recommended replacements.

With this, we can get Gecko off `time@0.1`.

### DIFF
--- a/glean-core/Cargo.toml
+++ b/glean-core/Cargo.toml
@@ -34,14 +34,14 @@ rkv = { version = "0.19.0", default-features = false }
 bincode = "1.2.1"
 log = "0.4.8"
 uuid = { version = "1.0", features = ["v4"] }
-chrono = { version = "0.4.10", features = ["serde"] }
+chrono = { version = "0.4.41", features = ["serde"] }
 once_cell = "1.18.0"
 flate2 = "1.0.19"
 zeitstempel = "0.1.0"
 crossbeam-channel = "0.5"
 thiserror = "1.0.4"
 uniffi = { version = "0.29.0", default-features = false }
-time = "0.1.40"
+time = "0.3"
 env_logger = { version = "0.10.0", default-features = false, optional = true }
 malloc_size_of_derive = "0.1.3"
 malloc_size_of = { version = "0.2.1", package = "wr_malloc_size_of", default-features = false, features = ["once_cell"] }
--- a/glean-core/src/metrics/datetime.rs
+++ b/glean-core/src/metrics/datetime.rs
@@ -15,7 +15,7 @@ use crate::util::{get_iso_time_string, local_now_with_offset};
 use crate::CommonMetricData;
 use crate::Glean;
 
-use chrono::{DateTime, Datelike, FixedOffset, TimeZone, Timelike};
+use chrono::{DateTime, Datelike, FixedOffset, NaiveTime, TimeZone, Timelike};
 use malloc_size_of_derive::MallocSizeOf;
 
 /// A datetime type.
@@ -97,7 +97,7 @@ impl MetricType for DatetimeMetric {
 
 impl From<ChronoDatetime> for Datetime {
     fn from(dt: ChronoDatetime) -> Self {
-        let date = dt.date();
+        let date = dt.date_naive();
         let time = dt.time();
         let tz = dt.timezone();
         Self {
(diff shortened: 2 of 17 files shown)


glean-core/Cargo.toml
```diff
@@ -1,6 +1,6 @@
 [package]
 name = "glean-core"
-version = "64.4.0"
+version = "64.4.1"
```
badboy: Please don't change the Glean version number here.

glean-core/Cargo.toml
```diff
@@ -34,14 +34,14 @@ rkv = { version = "0.19.0", default-features = false }
...
 log = "0.4.8"
 uuid = { version = "1.0", features = ["v4"] }
-chrono = { version = "0.4.10", features = ["serde"] }
+chrono = { version = "0.4.41", features = ["serde"] }
 once_cell = "1.18.0"
 flate2 = "1.0.19"
 zeitstempel = "0.1.0"
 crossbeam-channel = "0.5"
-thiserror = "1.0.4"
+thiserror = "2"
```
badboy: Handled in https://github.com/mozilla/glean/pull/3163
We need other updates first.

glean-core/src/util.rs
```diff
@@ -49,44 +49,10 @@ pub fn get_iso_time_string(datetime: DateTime<FixedOffset>, truncate_to: TimeUni
...
-            let now: DateTime<Utc> = Utc::now();
-            let utc_offset = FixedOffset::east(0);
-            return now.with_timezone(&utc_offset);
-        }
-    }
-
+    // See https://bugzilla.mozilla.org/show_bug.cgi?id=1611770.
+    //
+    // It's not clear if this bug on Windows still exist with the latest versions of
+    // the `time` crate. Removed the workaround.
```
chutten, line 55: It is true that `tm_to_datetime` is no longer even in the crate source, but this may prove to be a bold way to test that it has improved matters.

We should release and vendor this change early in a nightly cycle so we have lots of bake time to catch crashes like 1611770.

glean-core/tests/datetime.rs
```diff
@@ -39,9 +39,12 @@ fn datetime_serializer_should_correctly_serialize_datetime() {
...
 
         // `1983-04-13T12:09:14.274+00:00` will be truncated to Minute resolution.
-        let dt = FixedOffset::east(0)
-            .ymd(1983, 4, 13)
-            .and_hms_milli(12, 9, 14, 274);
+        let dt = FixedOffset::east_opt(0)
+            .unwrap()
+            .with_ymd_and_hms(1983, 4, 13, 12, 9, 14)
+            .unwrap()
+            .with_nanosecond(274)
```
chutten: note: Technically this is 274e6 ns in the original, but the test is still as useful with ns instead of ms.

glean-core/src/metrics/datetime.rs
```diff
@@ -201,33 +203,36 @@ impl DatetimeMetric {
...
-            TimeUnit::Hour => d.date().and_hms_nano_opt(time.hour(), 0, 0, 0),
-            TimeUnit::Day => d.date().and_hms_nano_opt(0, 0, 0, 0),
-        }
+            TimeUnit::Minute => d
+                .date_naive()
+                .and_hms_nano_opt(time.hour(), time.minute(), 0, 0),
+            TimeUnit::Hour => d.date_naive().and_hms_nano_opt(time.hour(), 0, 0, 0),
+            TimeUnit::Day => d.date_naive().and_hms_nano_opt(0, 0, 0, 0),
+        };
+        dt.map(|dt| Utc.from_utc_datetime(&dt).into())
```
chutten: This appears to be incorrect (Python test failures reinforce this, I think). `d` will have the timezone offset in it.

chutten (changes requested):

glean-core/tests/datetime.rs
larseggert: Good catch; I'll fix it.

glean-core/src/metrics/datetime.rs
larseggert: I tried a few things, but I think I'm not understanding the logic here. Would you have a hint for a fix?

chutten:
( + @badboy for an extra pair of eyes, who is now back from PTO )

supply-chain/imports.lock
```diff
@@ -90,6 +244,13 @@ user-id = 48
 user-login = "badboy"
 user-name = "Jan-Erik Rediger"
 
+[[audits.bytecode-alliance.wildcard-audits.bumpalo]]
```
martinthomson, line 247: This seems strictly weaker than https://searchfox.org/mozilla-central/rev/ec8a326713f60dec138a3e3383b03ac739870fc7/supply-chain/audits.toml#1118

Nick is the author of the crate, so maybe it's OK.

larseggert:
I messed up the squash. Let me fix that.

PR #3166, opened 2025-06-13
Bug 1957085 - Add dlc to rust sample

travis79:


travis79:
This won't work without https://github.com/mozilla/glean_parser/pull/797

samples/rust/metrics.yaml
```diff
@@ -233,3 +233,93 @@ glean.distribution:
...
+      key:
+        description: >
+          The key for the dual labeled counter
+        labels:
+          - key1
+          - key2
+      category:
+        description: >
+          The category for the dual labeled counter
+        labels: []
```
chutten: The current preferred form for dynamic labels is to have an absent list not an empty one.

samples/rust/metrics.yaml
```diff
@@ -233,3 +233,93 @@ glean.distribution:
...
+          - category1
+          - category2
+    bugs:
+      - https://bugzilla.mozilla.org/11137353
+    data_reviews:
+      - http://example.com/reviews
+    notification_emails:
+      - CHANGE-ME@example.com
+    expires: 2100-01-01
+  dynamic_dynamic:
```
chutten: Please add whitespace between metric definitions

samples/rust/src/main.rs
```diff
@@ -132,6 +132,74 @@ fn main() {
...
+    glean_metrics::test_dual_labeled::static_static
+        .get("party", "balloons")
+        .add(1);
+    assert_eq!(
+        Some(1),
+        glean_metrics::test_dual_labeled::static_static
+            .get("__other__", "__other__")
+            .test_get_value(None)
+    );
+
```
chutten, line 202: Gonna test "invalid" (not present in static list) labels as well, or are those covered by the more unit-ish tests in the impl commit?

chutten (approved):
r+wc

samples/rust/src/main.rs
travis79: InvalidLabel is tested elsewhere, and we don't really test negative cases for other metric types in the sample.

PR #3169, opened 2025-06-20
Fix dual_labeled_counter yaml example

chutten:
[doc only]


chutten:
Actually, I think it's worse than this. It should be `key` and `category` singular, no?

PR #3173, opened 2025-06-24
Bug 1924793 - CI: Build a Python wheel for aarch64-linux

badboy:
This uses `zig` under the hood for easy cross-compiling. maturin ensures
to pick the right minimal glibc to support.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -854,7 +854,44 @@ jobs:
             .venv3.8/bin/python3 -m twine upload target/wheels/*
       - install-ghr-linux
       - run:
-          name: Publish to Github
+          name: Publish to GitHub
+          command: |
+            # Upload to GitHub
+            ./ghr -replace ${CIRCLE_TAG} target/wheels
+
+  pypi-linux-aarch64-release:
+    docker:
+      - image: cimg/python:3.13
+    steps:
+      - install-rustup
+      - setup-rust-toolchain
+      - checkout
+      - run:
+          name: Setup Python env
+          command: |
+            make setup-python
+            .venv3.13/bin/pip install ziglang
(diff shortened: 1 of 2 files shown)


.circleci/config.yml
```diff
@@ -859,6 +859,43 @@ jobs:
...
+            cargo build -p glean-bundle
+
+            . .venv3.13/bin/activate
+            make build-python-wheel GLEAN_BUILD_TARGET=aarch64-unknown-linux-gnu GLEAN_BUILD_EXTRA="--zig"
+      - run:
+          name: Upload Linux wheel
+          command: |
+            # Requires that the TWINE_USERNAME and TWINE_PASSWORD environment
+            # variables are configured in CircleCI's environment variables.
+            .venv3.13/bin/python3 -m twine upload target/wheels/*
```
chutten, line 891: Aren't we using `uv` now?

.circleci/config.yml
```diff
@@ -859,6 +859,43 @@ jobs:
...
+      - install-ghr-linux
+      - run:
+          name: Publish to Github
```
chutten:
```suggestion
          name: Publish to GitHub
```

chutten (approved):
I'm no CI or Python whiz, but this seems sensible

.circleci/config.yml
badboy: [not for Glean (yet)](https://github.com/mozilla/glean/blob/01c0f74fac8c253e7b2212325acdf1245ae67004/.circleci/config.yml#L823).

.circleci/config.yml
badboy: copy-pasta! I adjust it everywhere.

PR #3175, opened 2025-06-26
Bug 1950143 - Expose an interface to supply an external ping uploader on iOS

travis79:


CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 * Rust
   * Added `MetricType` implementation for `DualLabeledCounter` ([Bug 1973287](https://bugzilla.mozilla.org/show_bug.cgi?id=1973287))
+* Swift
```
badboy, line 16: Note: will need a rebase as we're gonna do a release before this merges.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -35,78 +17,53 @@ public class HttpPingUploader {
...
-    func upload(request: PingRequest, callback: @escaping (UploadResult) -> Void) {
+    ///     * request: A `CapablePingUploadRequest` containing the information needed to perform the upload
+    ///     * callback: A callback to return the success/failure of the upload to the scheduler
+    public func upload(
+        request: CapablePingUploadRequest,
+        callback: @escaping (UploadResult) -> Void
+    ) {
+        // Get the internal `PingUploadRequest` to gather all the things we need from it, shadowing
+        // `request` to make it less confusing
+        let request = request.request
```
badboy: Shouldn't we avoid this and make proper use of the `.capable(f)` API?
Something like this (untested):

```swift
guard let request = request.capable({ capabilities in return capabilities.isEmpty}) else {
  logger.info(...)
  return
}
```

travis79: Yes, at the very least it would serve as a good example of the API in use

CHANGELOG.md
travis79: I always expect to rebase, so no worries :)

glean-core/ios/Glean/Glean.swift
```diff
@@ -36,7 +36,7 @@ class OnGleanEventsImpl: OnGleanEvents {
 
     func triggerUpload() {
         // If uploading is disabled, we need to send the deletion-request ping
-        HttpPingUploader.launch(configuration: self.glean.configuration!)
+        glean.pingUploadScheduler?.process()
```
lmarceau: I am not sure I understand this comment with the line below. Is the comment still up-to-date?

glean-core/ios/Glean/Net/PingUploader.swift
```diff
@@ -0,0 +1,52 @@
...
+public struct CapablePingUploadRequest {
+    let request: PingUploadRequest
+
+    init(_ request: PingUploadRequest) {
+        self.request = request
+    }
+
+    func capable(f: (_ uploaderCapabilities: [String]) -> Bool)
+        -> PingUploadRequest? {
+        if f(request.uploaderCapabilities) {
```
lmarceau: Could we find another name for that variable? `f` is a bit short and I'm not sure what it means here

glean-core/ios/Glean/Net/PingUploadScheduler.swift
```diff
@@ -0,0 +1,117 @@
...
+
+import Foundation
+
+/// `true` to allow the uploader to process pings.
+/// Note that this does not mean that an uploader is actually running.
+/// It will be invoked when a ping is submitted.
+///
+/// `false` to stop the uploader from starting new uploads.
+/// An already running uploader will finish work and then stop.
+var stateRunAllowed: AtomicBoolean = AtomicBoolean(false)
```
lmarceau, line 13: Can we avoid a global variable here? I am unsure where `shutdownUploader` and `startUploader` are used, but we should avoid global functions and variables.

glean-core/ios/Glean/Glean.swift
travis79: Yes, if upload is disabled, we need to process the pending pings as there should be a pending "deletion-request" ping to send. This merely triggers the scheduler in order to trigger uploading of that one ping.

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -135,13 +92,14 @@ public class HttpPingUploader {
     /// Internal function that builds the request used for uploading the pings.
     ///
     /// - parameters:
-    ///     * path: The URL path to append to the server address
+    ///     * url: The URL, including the path, to use for the destination of the ping
```
ih-codes, line 88: unrelated nit: Might be useful to clarify "the path" is the server endpoint (if I'm understanding the change correctly) 👀

glean-core/ios/Glean/Net/PingUploadScheduler.swift
ih-codes: Nice, looks like maybe they already have a ticket to move this to a singleton based on the comment below: `TODO(bug 1816403): Move this and the associated global state`

glean-core/ios/Glean/Net/PingUploader.swift
```diff
@@ -0,0 +1,52 @@
...
+ * file, You can obtain one at http://mozilla.org/MPL/2.0/. */
+
+typealias HeadersList = [String: String]
+
+/// The interface defining how to send pings.
+public protocol PingUploader {
+    /**
+     * Synchronously upload a ping to a server.
+     *
+     * @param request the ping upload request, locked within a uploader capability check
```
ih-codes, line 12:
```suggestion
     * @param request the ping upload request, locked within an uploader capability check
```

glean-core/ios/Glean/Net/PingUploadScheduler.swift
```diff
@@ -0,0 +1,117 @@
...
+                }
+            }
+
+            while true {
+                // Limits are enforced by glean-core to avoid an infinite loop here.
+                // Whenever a limit is reached, this binding will receive `.done` and step out.
+                switch gleanGetUploadTask() {
+                case let .upload(request):
+                    var body = Data(capacity: request.body.count)
+                    body.append(contentsOf: request.body)
```
ih-codes, line 84: Do these two lines actually perform some work? `body` is not used below in this scope. 🤔

glean-core/ios/Glean/Config/Configuration.swift
```diff
@@ -51,9 +54,11 @@ public struct Configuration {
         enableInternalPings: Bool = true,
         pingLifetimeThreshold: Int = 0,
         pingLifetimeMaxTime: Int = 0,
-        pingSchedule: [String: [String]] = [:]
+        pingSchedule: [String: [String]] = [:],
+        httpClient: PingUploader? = nil
```
ih-codes: One thing I'd personally like to see is for a default `PingUploader` concrete type to be provided here in the `Configuration`'s init vs. hidden in the implementation of Glean's `initialize` method > `PingUploadScheduler` init.

From a code perspective, I was briefly concerned that since the configuration had a default `nil` for the `httpClient`, there was a chance of Glean's `pingUploadScheduler` being `nil` and therefore `triggerUpload()` silently failing to actually execute any work. 🤔  (After some digging I saw this wasn't actually the case)

That said, I understand why it's built this way... the change I propose would require that the `HttpPingUploader` not take a `Configuration` as well. But, as far as I can tell, the `config` property is unused in `HttpPingUploader.swift` and could possibly be removed... 👀 

Anyway, that's a lot of text but it's honestly just a nitpick, so don't feel obligated to restructure anything. It's just nice to see the default values from the client's side (as with `enableInternalPings = true` for example).

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -35,78 +17,53 @@ public class HttpPingUploader {
...
 
     /// Initialize the HTTP Ping uploader from a Glean configuration object
-    /// and a URLSession
     ///
     /// - parameters:
-    ///     * configuration: The Glean `Configuration` to use.
-    ///     * session: A `URLSession` that will be reused to upload pings
-    public init(configuration: Configuration, session: URLSession) {
+    ///     * configuration: The Glean `Configuration` to use
+    public init(configuration: Configuration) {
```
ih-codes: With respect to my other comment, is the `configuration` actually a necessary input into this class? As far as I could tell from a quick look, it is not used (yet?) for anything.

glean-core/ios/Glean/Net/PingUploader.swift
```diff
@@ -0,0 +1,52 @@
...
+        callback: @escaping (UploadResult) -> Void
+    )
+}
+
+struct PingUploadRequest {
+    let documentId: String
+    let url: String
+    let data: [UInt8]
+    let headers: HeadersList
+    let uploaderCapabilities: [String]
```
ih-codes, line 27: I assume the range of valid `uploadCapabilities` keys will come later? 👀

ih-codes (commented):
Hi @travis79! I took a quick look. I'm totally outside the context of this work, so take my thoughts with a grain of salt. But looks good to me, and I appreciate all the documentation and the Glean architecture! 💪

IMO this seems like an acceptable external uploader API, especially since we can reference the `HttpPingUploader` implementation as a concrete example.

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: Always happy to add clarifications, comments and documentation! And you are correct in your understanding that this is the ingestion endpoint url including the API "path". Previously we referred to the URL and the path separately and just concatenated them before uploading to create the full ingestion endpoint URL for the connection. This change moves that concatenation to happen earlier and makes this function no longer need to know about the Glean Configuration object and was mostly to simplify splitting the scheduling logic from the uploader logic.

glean-core/ios/Glean/Net/PingUploadScheduler.swift
travis79: heh, trading one evil (global vars) for another (singletons) 😅

travis79: The bug mentioned does cover our intent to clean this up, we never were happy with needing to use a global to control the state here.

glean-core/ios/Glean/Net/PingUploadScheduler.swift
travis79: Good catch, the logic behind this mostly stayed in the http uploader and I missed removing it here (and didn't catch that it was unused because of the call to `append`

glean-core/ios/Glean/Config/Configuration.swift
travis79: That makes perfect sense, and if I can avoid passing the `Configuration` around that wouldn't be a bad thing. Let me see if I can rework this to be more in line with your suggestion.

glean-core/ios/Glean/Glean.swift
travis79: I can clarify what's going on here in the comment

glean-core/ios/Glean/Net/HttpPingUploader.swift
travis79: Another good catch! Originally the configuration object contained information about endpoint url and that's now handled by the scheduler so this shouldn't be needed (and can make it possible to provide the concrete type and init in the `Configuration` object as you suggested.

glean-core/ios/Glean/Net/PingUploader.swift
travis79: Glean doesn't really know about these, it merely acts as a conduit from the ping definition to the uploader implementation. These capabilities are application/ping specific and right now have a range of 1 ( `- ohttp`).

The purpose of a `uploaderCapabilites` is because we've let pings with data that should be only sent over OHTTP to be sent over HTTP in the past, forcing us to delete some data due to the incident. This was our response to that: Requiring pings that need a particular uploader capability (such as ohttp upload) to define these in the pings.yaml definition, and to require uploader implementations to advertise their capabilities to Glean so we can avoid giving an uploader implemenation a ping it cannot correctly process/upload. It's not a very good "contract" because the uploader implementation could just lie and say it has a capability but rather than continuing with no safeguards at all, we opted to put up the safeguards we could and rely on the good intentions of the consuming applications with regards to uploader implementations.

tl;dr: Rather than making this OHTTP specific, we opted to make it more flexible to handle future capabilities which Glean is mostly or totally agnostic of, so the range is defined by the application and not limited by Glean.

glean-core/ios/Glean/Net/PingUploader.swift
travis79: Yes, I fear I may have just cargo-culted the `f` parameter naming and it can definitely be more descriptive.

glean-core/ios/Glean/Net/PingUploader.swift
ih-codes: Ahh thanks for the history behind this work, @travis79! 🙏  That totally makes sense to me, and I think it was a good approach to keep things flexible for the future.

glean-core/ios/GleanTests/Net/HttpPingUploaderTests.swift
```diff
@@ -128,13 +128,79 @@ class HttpPingUploaderTests: XCTestCase {
...
+        }
+
+        XCTAssertEqual(
+            .incapable(unused: 0),
+            testValue,
+            "upload should return .incapable when capabilities don't match"
+        )
+    }
+
+    func testUploaderReturnsRequestWhenCapbilitiesSupported() {
```
lmarceau: nit, typo
```suggestion
    func testUploaderReturnsRequestWhenCapabilitiesSupported() {
```

travis79: 🙈 thank you!

glean-core/ios/Glean/Net/HttpPingUploader.swift
```diff
@@ -35,78 +16,50 @@ public class HttpPingUploader {
...
-            let session = URLSession(configuration: config)
-
-            HttpPingUploader(configuration: configuration, session: session).process()
-
-            if backgroundTaskId != .invalid {
-                UIApplication.shared.endBackgroundTask(backgroundTaskId)
-                backgroundTaskId = .invalid
-            }
-        }
+    ///     * configuration: The Glean `Configuration` to use
```
badboy: This `init` doesn't have any parameters anymore.

badboy (approved):
one last nit, but then good to go!

PR #3178, opened 2025-07-04
chore: Update deps to remove dependency on `paste` crate

larseggert:
Which is unmaintained, see https://rustsec.org/advisories/RUSTSEC-2024-0436.


larseggert:
Looks like it! Should have checked first. Ping me please when you vendor a glean version with this PR in it into Firefox?

PR #3179, opened 2025-07-07
Bug 1975812 - Cache opened event databases

badboy:
We kept re-opening the event database file on every event recording. That's costly.

This PR includes an example that records 500k events (configurable).

Before the change that takes 2.8s on my Linux machine.
The `open` call alone takes 41% of the time. Closing the file (`drop`) another 11%.
Profile: https://share.firefox.dev/4lnTYXG

After it takes 1.7s
Profile: https://share.firefox.dev/4lMhI7D
1.1% of time is spent finding the (already opened) file descriptor.
At the cost of one more thing to track (and a non-contended lock around it).

---

This is not yet ready to land.
We do have _another_ lock around all operations and I want to make sure we're not running into problems there.

### DIFF
--- a/glean-core/src/event_database/mod.rs
+++ b/glean-core/src/event_database/mod.rs
@@ -351,7 +351,6 @@ impl EventDatabase {
                 .unwrap()
                 .entry(store_name.to_string())
                 .or_insert_with(|| {
-                    log::info!("opening event store for {store_name}");
                     Arc::new(
                         OpenOptions::new()
                             .create(true)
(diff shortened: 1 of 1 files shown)


chutten (approved):
So we're trading the runtime costs of opening and closing the `File` on each write for the memory and fd cost of keeping a `File` for each store that has events written to it.

Given volumes of pings per app (dozens, with only handfuls having events) I think this ought to be fine, but it wouldn't be unreasonable to close the `File`s (wipe the cache) on ping lifetime io flush. And, heck, `event` metrics are all ping lifetime anyway... could skip `flush` calls...

But that's the very definition of premature optimization. We can wait until we have a need to speed things up further.

badboy:
> And, heck, `event` metrics are all ping lifetime anyway... could skip `flush` calls...

True. I kept the `flush` to at least hopefully give us some guarantees. (Though it is more strict than before, because `close` doesn't imply `flush`)

PR #3186, opened 2025-07-14
Kotlin: 2 more dispatch optimizations

badboy:
1. Dispatch applying the server knobs config
2. Move all calls into glean-core into the dispatcher


badboy:
> On desktop we ran into a case with Nimbus recording an event before the Server Knobs config was applied, I wonder if this will increase the window where we do not have that operational safety valve available to us to be able to turn off noisy events because of delaying Glean init and API calls.
> 
> Nothing to block this PR on, just thinking about how this might affect things like that.

the `applyServerKnobsConfig` already has no effect until Glean is initialized. This only avoids the call into glean-core up until we're somewhat sure it's ok to load libxul (or even better libxul is already loaded). The order of operations should still be the same.

PR #3188, opened 2025-07-16
Bug 1976479 - Report a glean.ping.uploader_capabilities list in pings

badboy:
Because I already prototyped that here's it as a PR.
Would benefit from more extensive tests.
And also checking that my logic is right to attach it only when there's already other data. 
And making sure it doesn't overwrite existing `string_list` metrics.
And also it hard-codes the `"string_list"` string in the code, which I don't like.

All in all I'd appreciate if someone takes this over and finishes it.


travis79 (approved):
Ping me for a data-review, if you like, before landing this.

PR #3189, opened 2025-07-21
Swift: Build on Xcode 16.2

badboy:


badboy:
Guess ffx-ios is on 26 already: https://github.com/mozilla-mobile/firefox-ios/commit/0120fd7728577d748c541ff79cbc115a6ada03e3

PR #3190, opened 2025-07-21
Bug 1930572: add testGetValue for LabeledMetrics

jeddai:
This PR updates the `testGetValue` function to be a trait that can be exported across the FFI boundary. 

It also defines a new `testGetLabeledValues` function on `LabeledMetrics` that is exported across the FFI boundary for metrics on which it is implemented.

### DIFF
--- a/docs/user/reference/metrics/labeled_booleans.md
+++ b/docs/user/reference/metrics/labeled_booleans.md
@@ -118,8 +118,8 @@ import org.mozilla.yourApplication.GleanMetrics.Accessibility
 
 val values = Accessibility.features.testGetValue()
 // Do the booleans have the expected values?
-assertEquals(True, values["category.accessibility/screen_reader"])
-assertEquals(False, values["category.accessibility/high_contrast"])
+assertEquals(True, values["screen_reader"])
+assertEquals(False, values["high_contrast"])
 ```
 </div>
 
@@ -130,8 +130,8 @@ import org.mozilla.yourApplication.GleanMetrics.Accessibility;
 
 Map<String, ?> values = Accessibility.INSTANCE.features().testGetValue();
 // Do the booleans have the expected values?
-assertEquals(True, values["category.accessibility/screen_reader"]);
-assertEquals(False, values["category.accessibility/high_contrast"]);
+assertEquals(True, values["screen_reader"]);
+assertEquals(False, values["high_contrast"]);
 ```
 </div>
 
@@ -140,8 +140,8 @@ assertEquals(False, values["category.accessibility/high_contrast"]);
 ```Swift
 let values = Accessibility.features.testGetValue()
 // Do the booleans have the expected values?
-XCTAssertEqual(true, values["category.accessibility/screen_reader"])
-XCTAssertEqual(false, values["category.accessibility/high_contrast"])
+XCTAssertEqual(true, values["screen_reader"])
+XCTAssertEqual(false, values["high_contrast"])
 ```
 </div>
 
@@ -153,8 +153,8 @@ metrics = load_metrics("metrics.yaml")
 
 values = metrics.accessibility.features.testGetValue()
 # Do the booleans have the expected values?
-assert values["category.accessibility/screen_reader"]
-assert not values["category.accessibility/high_contrast"]
+assert values["screen_reader"]
+assert not values["high_contrast"]
 ```
 </div>
 
@@ -165,8 +165,8 @@ use glean_metrics::accessibility;
 
 let values = accessibility::features.test_get_value(None).unwrap();
 // Do the booleans have the expected values?
-assert!(values["category.accessibility/screen_reader"]);
-assert!(!values["category.accessibility/high_contrast"]);
+assert!(values["screen_reader"]);
+assert!(!values["high_contrast"]);
 ```
 </div>
 
--- a/docs/user/reference/metrics/labeled_booleans.md
+++ b/docs/user/reference/metrics/labeled_booleans.md
@@ -102,8 +102,10 @@ Glean.accessibility.features["high_contrast"].set(false);
 ### `testGetValue`
 
 Gets the recorded value for a given label in a labeled boolean metric.  
-Returns the count if data is stored.  
-Returns a language-specific empty/null value if no data is stored.
+Returns the boolean value if data is stored. The Glean SDK will return a map of each label with a
+stored value to its value.  
+Returns a language-specific empty/null value if no data is stored. The Glean SDK will always 
+return a map, but it will be empty if no data is stored.
 Has an optional argument to specify the name of the ping you wish to retrieve data from, except
 in Rust where it's required. `None` or no argument will default to the first value found for `send_in_pings`.
 
@@ -114,9 +116,10 @@ in Rust where it's required. `None` or no argument will default to the first val
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Accessibility
 
+val values = Accessibility.features.testGetValue()
 // Do the booleans have the expected values?
-assertEquals(True, Accessibility.features["screen_reader"].testGetValue())
-assertEquals(False, Accessibility.features["high_contrast"].testGetValue())
+assertEquals(True, values["category.accessibility/screen_reader"])
+assertEquals(False, values["category.accessibility/high_contrast"])
 ```
 </div>
 
@@ -125,18 +128,20 @@ assertEquals(False, Accessibility.features["high_contrast"].testGetValue())
 ```Java
 import org.mozilla.yourApplication.GleanMetrics.Accessibility;
 
+Map<String, ?> values = Accessibility.INSTANCE.features().testGetValue();
 // Do the booleans have the expected values?
-assertEquals(True, Acessibility.INSTANCE.features()["screen_reader"].testGetValue());
-assertEquals(False, Acessibility.INSTANCE.features()["high_contrast"].testGetValue());
+assertEquals(True, values["category.accessibility/screen_reader"]);
+assertEquals(False, values["category.accessibility/high_contrast"]);
 ```
 </div>
 
 <div data-lang="Swift" class="tab">
 
 ```Swift
+let values = Accessibility.features.testGetValue()
 // Do the booleans have the expected values?
-XCTAssertEqual(true, Accessibility.features["screen_reader"].testGetValue())
-XCTAssertEqual(false, Accessibility.features["high_contrast"].testGetValue())
+XCTAssertEqual(true, values["category.accessibility/screen_reader"])
+XCTAssertEqual(false, values["category.accessibility/high_contrast"])
 ```
 </div>
 
@@ -146,9 +151,10 @@ XCTAssertEqual(false, Accessibility.features["high_contrast"].testGetValue())
 from glean import load_metrics
 metrics = load_metrics("metrics.yaml")
 
+values = metrics.accessibility.features.testGetValue()
 # Do the booleans have the expected values?
-assert metrics.accessibility.features["screen_reader"].test_get_value()
-assert not metrics.accessibility.features["high_contrast"].test_get_value()
+assert values["category.accessibility/screen_reader"]
+assert not values["category.accessibility/high_contrast"]
 ```
 </div>
 
@@ -157,9 +163,10 @@ assert not metrics.accessibility.features["high_contrast"].test_get_value()
 ```Rust
 use glean_metrics::accessibility;
 
+let values = accessibility::features.test_get_value(None).unwrap();
 // Do the booleans have the expected values?
-assert!(accessibility::features.get("screen_reader").test_get_value(None).unwrap());
-assert!(!accessibility::features.get("high_contrast").test_get_value(None).unwrap());
+assert!(values["category.accessibility/screen_reader"]);
+assert!(!values["category.accessibility/high_contrast"]);
 ```
 </div>
 
--- a/glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
+++ b/glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
@@ -162,13 +162,13 @@ class LabeledMetricType<T>(
      * @return The values for each label in the labeled metric.
      */
     @VisibleForTesting(otherwise = VisibleForTesting.NONE)
-    fun testGetLabeledValues(pingName: String? = null): Map<String, Any> {
+    fun testGetValue(pingName: String? = null): Map<String, Any> {
         return when (this.inner) {
-            is LabeledBoolean -> this.inner.metric.testGetLabeledValues(pingName)
-            is LabeledCounter -> this.inner.metric.testGetLabeledValues(pingName)
-            is LabeledString -> this.inner.metric.testGetLabeledValues(pingName)
-            is LabeledQuantity -> this.inner.metric.testGetLabeledValues(pingName)
+            is LabeledBoolean -> this.inner.metric.testGetValue(pingName)
+            is LabeledCounter -> this.inner.metric.testGetValue(pingName)
+            is LabeledString -> this.inner.metric.testGetValue(pingName)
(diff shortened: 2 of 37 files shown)


glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -150,4 +150,25 @@ class LabeledMetricType<T>(
...
+     * metric.
+     *
+     * This doesn't clear the stored value.
+     *
+     * @param pingName The optional name of the ping to retrieve the metrics for. Defaults to the
+     *   first value in `send_in_pings`.
+     * @return The values for each label in the labeled metric.
+     */
+    @VisibleForTesting(otherwise = VisibleForTesting.NONE)
+    fun testGetLabeledValues(pingName: String? = null): Map<String, Any> {
```
chutten: bikeshed: Why `testGetLabeledValues` instead of `testGetValue`?

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
```diff
@@ -150,4 +150,25 @@ class LabeledMetricType<T>(
...
+        return when (this.inner) {
+            is LabeledBoolean -> this.inner.metric.testGetLabeledValues(pingName)
+            is LabeledCounter -> this.inner.metric.testGetLabeledValues(pingName)
+            is LabeledString -> this.inner.metric.testGetLabeledValues(pingName)
+            is LabeledQuantity -> this.inner.metric.testGetLabeledValues(pingName)
+            else -> error("Can not create a labeled version of this metric type")
+        }
```
chutten: Discussion: I was going to leave a comment here that this ought to be an inheritance game, using the language's dynamic dispatch to sort out. However, it seems as though this is in keeping with the rest of the class' impl so I'll merely reference @travis79 here to weigh in on why we wrote it this way and any future plans. (See also Swift)

glean-core/src/metrics/boolean.rs
```diff
@@ -100,41 +100,43 @@ impl BooleanMetric {
...
-        crate::core::with_glean(|glean| self.get_value(glean, ping_name.as_deref()))
+
+        crate::core::with_glean(|glean| {
+            test_get_num_recorded_errors(glean, self.meta(), error).unwrap_or(0)
+        })
     }
+}
 
-    /// **Exported for test purposes.**
+impl TestGetValue<bool> for BooleanMetric {
```
chutten, line 123: Discussion: What are your (and I guess @travis79's, since I've already @-ed him once) opinions on making this part of `trait MetricType` which "describes common behavior across all metrics."? And then we could include `test_get_num_recorded_errors`.

Could two-birds-one-scone this with "we want to make it harder to implement new metrics that forget to `impl MetricType`".

glean-core/src/metrics/counter.rs
```diff
@@ -137,41 +137,43 @@ impl CounterMetric {
...
-        crate::core::with_glean(|glean| self.get_value(glean, ping_name.as_deref()))
+
+        crate::core::with_glean(|glean| {
+            test_get_num_recorded_errors(glean, self.meta(), error).unwrap_or(0)
+        })
     }
+}
 
-    /// **Exported for test purposes.**
+impl TestGetValue<i32> for CounterMetric {
```
chutten, line 160: musing: One of these years we really need to switch to unsigned. The unsigned types have been stable since Kotlin 1.5 which was released in 2021.

glean-core/src/metrics/labeled.rs
```diff
@@ -351,6 +353,23 @@ where
...
+{
+    fn test_get_labeled_values(&self, ping_name: Option<String>) -> HashMap<String, S> {
+        let mut out = HashMap::new();
+        let map = self.label_map.lock().unwrap();
+        map.iter().for_each(|(label, submetric)| {
+            if let Some(v) = submetric.test_get_value(ping_name.clone()) {
+                out.insert(label.clone(), v);
+            }
+        });
+        out
```
chutten: For fun I tried to see if I could write this functionally. Not sure whether it's any more or less readable, though I do like having space to mention why we think the `unwrap()` is okay to call.

```rust
self.label_map
    .lock()
    .unwrap() // Safe unless lock's poisoned. Also: this is a test-only method: we're allowed to panic.
    .iter()
    .filter_map(|(label, submetric)| {
        submetric.test_get_value(ping_name.clone()).map(|v| (label.clone(), v))
    })
    .collect()
```

I haven't even checked to see if it would compile, but you're welcome to adapt it or drop it: your choice.

glean-core/src/metrics/mod.rs
```diff
@@ -269,6 +269,47 @@ pub trait MetricIdentifier<'a> {
     fn get_identifiers(&'a self) -> (&'a str, &'a str, Option<&'a str>);
 }
 
+/// [`TestGetValue`] describes an interface for retrieving the value for a given metric
+pub trait TestGetValue<T> {
```
chutten, line 273: I feel like the RLB should re-export this trait and use it as a supertrait for its metric traits, e.g. `pub trait Counter : TestGetValue<i32>`. That way we make it clear we never expect there to be a `Counter` that doesn't have `test_get_value(...) -> Option<i32>` on it.

Thoughts?

glean-core/src/metrics/labeled.rs
```diff
@@ -351,6 +353,23 @@ where
...
+    }
+}
+
```
chutten, line 375: I feel like there ought to be a unit test in `glean-core` somewhere. Maybe here like the metric-specific unit tests at the bottom of e.g. `counter.rs`?

chutten (changes requested):
Gosh, only seven things to say this time. I must've missed something.

Looks like it would work as designed, all my commentary is about tightening it or aligning it with potential future considerations. All those could be taken as follow-ups. 

Still marking as Request Changes for at least the unit test bit, though. Oh, and it needs a CHANGELOG entry -- biggest question I have for that is whether this'll be a BREAKING change or not (changing `glean.udl` suggests Yes, but it's strictly additive...)

chutten:
Oh, and it needs docs and probably a Python impl, too.

glean-core/src/metrics/labeled.rs
jeddai: I'm also now torn between keeping up with the piping like what you suggested or going even more imperative with a `for` on the iter. Not really sure which is more readable

jeddai: For things like this, where it's a test-only fn, I usually aim to stick with readability over performance or method chaining

jeddai:
Oh mega whoops, I completely missed Python 🤦‍♀️

glean-core/src/metrics/labeled.rs
chutten: Then let's go with as-written since there's nothing unreadable about it. Just drop in a crash-safety comment on the `unwrap()`

jeddai: I'm actually going to end up committing my change from what's here to the more imperative `for`

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
jeddai: Main reason is that there's two underlying traits in Rust, `TestGetValue<T>` and `TestGetLabeledValues<T>`.

`TestGetLabeledValues<T>` gets auto-impl'd for every metric type that impls `TestGetValue<T>`, minus the UDL piece of it. 

And the main reason I chose `TestGetLabeledValues` instead of `TestGetValues` for the name was so the names would be more distinct from one-another

jeddai: I suppose the method on `TestGetLabeledValues` could still just be `testGetValues`, but it felt weird to me for some reason

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
travis79: Yeah, this seems like it could be something like:

```
return when (val labeled = inner) {
    is LabeledBoolean,
    is LabeledCounter,
    is LabeledString,
    is LabeledQuantity -> labeled.metric.testGetLabeledValues(pingName)
    else -> error("Cannot create a labeled version of this metric type")
}
```

That would reduce the repetition, remove the unnecessary `this`, and should only bind `inner` once.

glean-core/ios/Glean/Metrics/LabeledMetric.swift
```diff
@@ -111,4 +112,28 @@ public class LabeledMetricType<T> {
...
+            return (self.inner as! LabeledBoolean).testGetLabeledValues(pingName)
+        case is LabeledString:
+            return (self.inner as! LabeledString).testGetLabeledValues(pingName)
+        case is LabeledQuantity:
+            return (self.inner as! LabeledQuantity).testGetLabeledValues(pingName)
+        default:
+            // The constructor will already throw an exception on an unhandled sub-metric type
+            assertUnreachable()
+        }
+    }
```
travis79: This could be make more idiomatic of Swift with something like:

```
public func testGetLabeledValues(_ pingName: String? = nil) -> [String: Any] {
    switch inner {
    case let labeled as LabeledCounter,
         let labeled as LabeledBoolean,
         let labeled as LabeledString,
         let labeled as LabeledQuantity:
        return labeled.testGetLabeledValues(pingName)
        
    default:
        // This shouldn't happen — assert at runtime
        assertionFailure("Unhandled labeled metric type")
        return [:]
    }
}
```

travis79: `case let x as T` both type-checks and binds the casted value, and combining cases reduces duplication

glean-core/src/metrics/boolean.rs
travis79: I'm all for adding this to the trait, if possible, or potentially a `LabeledMetric` trait?

glean-core/src/metrics/labeled.rs
travis79: Ditto

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
jeddai: `inner` is typed as `Any`, so unless I change that type the type narrowing doesn't work there

travis79: As to why it was written this way...   I'm not recalling the answer to that at the moment 🤔

glean-core/android/src/main/java/mozilla/telemetry/glean/private/LabeledMetricType.kt
chutten: Why does there have to be two underlying traits? Can't we have 'em all be `TestGetValue`? I don't think we want instrumentation devs to have to type a slightly different name when testing this one case.

jeddai: Well the return type is going to be different for the labeled metrics (`HashMap<String, T>` vs `T`), and best I recall method overloading isn't a thing in Rust, right? Or has that changed?

jeddai: Hmmm `TestGetValue<HashMap<String, T>>` might work...

chutten: I was about to say, in what way is `HashMap<String, int>` not a `T`?

glean-core/src/metrics/labeled.rs
jeddai: After chatting with @badboy, I won't be able to add one to `glean-core` because the `testGetValue` methods all use the dispatcher. But I will add one to the RLB crate

glean-core/ios/Glean/Metrics/LabeledMetric.swift
jeddai: The `Labeled<Metric>` classes need to share a parent type in order to do this, and regrettably they don't. Having serious difficulties making that happen across the FFI

travis79: Ugh, please forget my suggestions then. Next time I'll try them out before I suggest such things 😅

glean-core/ios/Glean/Metrics/LabeledMetric.swift
```diff
@@ -22,6 +22,7 @@ public class LabeledMetricType<T> {
     /// * `BooleanMetricType`
     /// * `CounterMetricType`
     /// * `StringMetricType`
+    /// * `QuantityMetric`
```
travis79, line 25: 💯

glean-core/python/glean/metrics/labeled.py
```diff
@@ -3,7 +3,7 @@
 # file, You can obtain one at http://mozilla.org/MPL/2.0/.
 
 
-from typing import Any, Optional, Set, Type
+from typing import Any, Optional, Set, Type, Dict
```
travis79:
```suggestion
from typing import Any, Dict, Optional, Set, Type
```

nit, keeping things alphabetized

glean-core/rlb/src/test.rs
```diff
@@ -1284,6 +1284,36 @@ fn test_boolean_get_num_errors() {
...
+            category: "telemetry".into(),
+            send_in_pings: vec!["store1".into()],
+            disabled: false,
+            lifetime: Lifetime::Ping,
+            ..Default::default()
+        } },
+        Some(vec!["key1".into()]),
+    );
+
+    metric.get("key1").add(1);
```
travis79, line 1307: For the other binding tests, you had two labels. Any reason you only have one here?

glean-core/src/lib.rs
```diff
@@ -76,6 +76,7 @@ pub use crate::metrics::{
     MemoryDistributionMetric, MemoryUnit, NumeratorMetric, ObjectMetric, PingType, QuantityMetric,
     Rate, RateMetric, RecordedEvent, RecordedExperiment, StringListMetric, StringMetric,
     TextMetric, TimeUnit, TimerId, TimespanMetric, TimingDistributionMetric, UrlMetric, UuidMetric,
+    TestGetValue,
 };
```
travis79:
```suggestion
    MemoryDistributionMetric, MemoryUnit, NumeratorMetric, ObjectMetric, PingType, QuantityMetric,
    Rate, RateMetric, RecordedEvent, RecordedExperiment, StringListMetric, StringMetric,
    TestGetValue, TextMetric, TimeUnit, TimerId, TimespanMetric, TimingDistributionMetric, UrlMetric, 
    UuidMetric,
};
```

nit alphabetize `use`

travis79 (approved):
Looking great! Just a handful of nits nothing to block landing this on re-review.

glean-core/rlb/src/test.rs
jeddai: Nope lol that was an accident

docs/user/reference/metrics/labeled_booleans.md
```diff
@@ -114,9 +116,10 @@ in Rust where it's required. `None` or no argument will default to the first val
 ```Kotlin
 import org.mozilla.yourApplication.GleanMetrics.Accessibility
 
+val values = Accessibility.features.testGetValue()
 // Do the booleans have the expected values?
-assertEquals(True, Accessibility.features["screen_reader"].testGetValue())
-assertEquals(False, Accessibility.features["high_contrast"].testGetValue())
+assertEquals(True, values["category.accessibility/screen_reader"])
+assertEquals(False, values["category.accessibility/high_contrast"])
```
travis79: Should this be:

```suggestion
assertEquals(True, values["Accessibility.features/screen_reader"])
assertEquals(False, values["Accessibility.features/high_contrast"])
```

PR #3191, opened 2025-07-21
Optimizing DualLabledCounter metric and adding a benchmark for it

travis79:


glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -18,59 +19,60 @@ const OTHER_LABEL: &str = "__other__";
...
+
+        // let mut map_size = shallow_size;
+        // for (k, v) in map.iter() {
+        //     map_size += k.size_of(ops);
+        //     map_size += v.size_of(ops);
+        // }
+        // n += map_size;
+
+        // n
+        0
```
travis79: It doesn't like measuring this with the hashmap key being `(Spur, Spur)`

glean-core/Cargo.toml
```diff
@@ -57,10 +58,15 @@ env_logger = { version = "0.10.0", default-features = false, features = ["humant
...
+criterion = "0.5"
 
 [build-dependencies]
 uniffi = { version = "0.29.3", default-features = false, features = ["build"] }
 
 [features]
 # Enable `env_logger`. Only works on non-Android non-iOS targets.
 enable_env_logger = ["env_logger"]
+
+[[bench]]
```
travis79, line 70: I couldn't get it to run the benchmark without adding this

PR #3194, opened 2025-07-24
Add criterion crate and benchmarks for some metric types

travis79:
This adds some very basic benchmarks to some of the labeled/dual-labeled metrics.

You can run these locally using `cargo bench`. To run a specific set of benches: `cargo bench --bench dual_labeled_counter`

### DIFF
--- a/glean-core/benches/dual_labeled_counter.rs
+++ b/glean-core/benches/dual_labeled_counter.rs
@@ -1,0 +1,41 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use criterion::{criterion_group, criterion_main, Criterion};
+use glean_core::metrics::DualLabeledCounterMetric;
+use glean_core::CommonMetricData;
+use glean_core::Lifetime;
+use std::hint::black_box;
+
+fn bench_dual_labeled_counter_get_and_add(c: &mut Criterion) {
+    let metric = DualLabeledCounterMetric::new(
+        CommonMetricData {
+            name: "test_metric".into(),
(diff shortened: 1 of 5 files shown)


glean-core/benches/dual_labeled_counter.rs
```diff
@@ -0,0 +1,41 @@
...
+
+    const N: usize = 5000;
+    let keys: Vec<(String, String)> = (0..N)
+        .map(|i| (format!("key{}", i), format!("cat{}", i)))
+        .collect();
+
+    c.bench_function("dual_labeled_counter_get_and_add", |b| {
+        b.iter(|| {
+            for (k, c) in &keys {
+                let m = black_box(metric.get(k, c));
```
badboy, line 33: This `black_box` doesn't make sense. You're using the return value, so it won't be eliminated anyway.

badboy (changes requested):
These benchmarks don't actually test what you intended.

You're using the `glean-core` types, but calling the dispatched methods.
Because Glean is never initialized, the dispatcher will never actually run and thus the majority of the work won't be executed.

I think it would make more sense to move those benchmarks into the RLB and use the `glean::private::*` types after initializing Glean.
And then we probably also need at least one call to `test_get_value` at the end of each benchmark to ensure the dispatcher is actually going through all tasks, otherwise all we measure is how fast we can put things on to the dispatcher queue.

glean-core/benches/dual_labeled_counter.rs
travis79: Ahh, I was using it under the assumption that it would encourage the compiler to not optimize this. I'll rework this and move things to the RLB as you suggest. Thanks for the pointers here, benchmarking things like this is new territory for me :)

badboy:
Closing this. We have another benchmark now. We might bring back one for labeled counters soon.

PR #3200, opened 2025-07-29
Update ktlint to version 1.7.1

rvandermeulen:
Lots of changes since the last time this was updated. I broke out the various formatting changes by rule so it's easier to see the effect each one has. If you think any of these rules are more hassle than they're worth, it should be pretty straightforward to disable them and drop the relevant patch from the series.


travis79 (approved):
I don't have any strong opinions about any of these other than the one Jan-Erik mentioned, the line wrappings change isn't my favorite either.

rvandermeulen:
Disabled standard:multiline-expression-wrapping and dropped that change. Thanks for reviewing!

PR #3206, opened 2025-07-30
Apply a high watermark for pending pings to process from disk

badboy:
If we detect more files in the pending pings directory they are removed without processing. Usually we read all files into memory to sort them and then discard the oldest ones if necessary. However if a user has too many files that's very costly (in both processing time and memory usage), so we add a high watermark. This means there's now a case where we can drop arbitrary data. However if a user has _that many pending pings_ some things might have gone awry and the data is less important than keeping the user's application running smoothly.

### DIFF
--- a/glean-core/src/upload/directory.rs
+++ b/glean-core/src/upload/directory.rs
@@ -50,6 +50,7 @@ pub struct PingPayload {
 pub struct PingPayloadsByDirectory {
     pub pending_pings: Vec<(u64, PingPayload)>,
     pub deletion_request_pings: Vec<(u64, PingPayload)>,
+    pub discarded_on_scan: u32,
 }
 
 impl MallocSizeOf for PingPayloadsByDirectory {
@@ -260,9 +261,17 @@ impl PingDirectoryManager {
 
     /// Processes both ping directories.
     pub fn process_dirs(&self) -> PingPayloadsByDirectory {
+        let mut discarded_on_scan = 0;
+        let (discarded, pending_pings) = self.process_dir(&self.pending_pings_dir);
+        discarded_on_scan += discarded;
+        let (discarded, deletion_request_pings) =
+            self.process_dir(&self.deletion_request_pings_dir);
+        discarded_on_scan += discarded;
+
         PingPayloadsByDirectory {
-            pending_pings: self.process_dir(&self.pending_pings_dir),
-            deletion_request_pings: self.process_dir(&self.deletion_request_pings_dir),
+            pending_pings,
+            deletion_request_pings,
+            discarded_on_scan,
         }
     }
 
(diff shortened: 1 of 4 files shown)


glean-core/src/upload/directory.rs
```diff
@@ -267,18 +288,20 @@ impl PingDirectoryManager {
     /// # Returns
     ///
     /// A vector of tuples with the file size and payload of each ping file in the directory.
```
travis79: Mind updating the doc-comment for the return type to show it now also returns the `discarded_files`?

glean-core/src/upload/directory.rs
```diff
@@ -17,6 +17,17 @@ use uuid::Uuid;
...
+// The maximum number of files we process.
+// If we detect more files in the pending pings directory they are removed without processing.
+// Usually we read all files into memory to sort them and then discard the oldest ones if
+// necessary.
+// However if a user has too many files that's very costly (in both processing time and
+// memory usage), so we add a high watermark.
+// This means there's now a case where we can drop arbitrary data.
+// However if a user has _that many pending pings_ some things might have gone awry and the
+// data is less important than keeping the user's application running smoothly.
+const HIGH_WATERMARK_PENDING_PINGS: u64 = super::policy::MAX_PENDING_PINGS_COUNT * 4;
```
travis79, line 27: Any reasoning behind the `* 4` we can document along with this? In case we want/need/have to change this in the future, it would be nice to understand the current value.

glean-core/src/upload/directory.rs
```diff
@@ -357,7 +392,7 @@ mod test {
         let directory_manager = PingDirectoryManager::new(dir.path());
 
         // Verify that processing the directory didn't panic
-        let data = directory_manager.process_dirs();
+        let data = directory_manager.process_dirs().1;
```
travis79: nit: I hate how un-descriptive index based access is (`.1`). Not saying we need to change it but it requires me to go and look up the process_dirs() to understand what `.1` vs. `.0` means here

glean-core/src/upload/directory.rs
badboy: I picked a number out of the hat. Open to discussion (`MAX_PEDING_PINGS_COUNT` is 250, so this would be 1000, that sounded like a nice round number, but that's all the thinking that went into it)

glean-core/src/upload/directory.rs
badboy: me too. I don't like how we need to thread through the value here now. If we do want this then I might re-do that part.

travis79: I'm all for adding this as a guardrail to keep /pending-pings in check, for what it's worth.

glean-core/src/upload/directory.rs
chutten: Saying that it was picked out of a hat is a good comment, too, so future-us realizes it wasn't an important number, just a convenient one.

Also, this isn't the maximum number of files, this is the max _per directory_.

badboy:
Not landing this now. We covered the initial issue, this change requires more changes than I want it to.

PR #3209, opened 2025-08-04
Bug 1980150: add TestGetValue trait usage to all non-labeled metrics

jeddai:
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1980150

### DIFF
--- a/glean-core/rlb/src/private/event.rs
+++ b/glean-core/rlb/src/private/event.rs
@@ -72,7 +72,7 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
 }
 
 #[inherent]
-impl<K: traits::ExtraKeys> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
+impl<K> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
     pub fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
         self.inner.test_get_value(ping_name)
     }
--- a/glean-core/rlb/src/private/event.rs
+++ b/glean-core/rlb/src/private/event.rs
@@ -73,8 +73,7 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
 
 #[inherent]
 impl<K: traits::ExtraKeys> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
-    #[allow(unused)]
-    fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
+    pub fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
         self.inner.test_get_value(ping_name)
     }
 }
--- a/glean-core/rlb/src/private/event.rs
+++ b/glean-core/rlb/src/private/event.rs
@@ -7,7 +7,7 @@ use std::{collections::HashMap, marker::PhantomData};
 
 use malloc_size_of::MallocSizeOf;
 
-use glean_core::{metrics::MetricIdentifier, traits};
+use glean_core::{metrics::MetricIdentifier, traits, TestGetValue};
 
 use crate::{ErrorType, RecordedEvent};
 
@@ -72,6 +72,14 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
 }
 
 #[inherent]
+impl<K: traits::ExtraKeys> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
+    #[allow(unused)]
+    fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
+        self.inner.test_get_value(ping_name)
+    }
+}
+
+#[inherent]
 impl<K: traits::ExtraKeys> traits::Event for EventMetric<K> {
     type Extra = K;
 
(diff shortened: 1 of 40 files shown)


glean-core/rlb/src/private/event.rs
```diff
@@ -71,6 +71,14 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
     }
 }
 
+#[inherent]
+impl<K: traits::ExtraKeys> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
+    #[allow(unused)]
```
badboy: This shouldn't be needed.

glean-core/rlb/src/private/event.rs
```diff
@@ -71,6 +71,14 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
...
+    fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
```
badboy: This should be pub, right? In which case the above `allow(unused)` is indeed unnecessary.
```suggestion
    pub fn test_get_value(&self, ping_name: Option<String>) -> Option<Vec<RecordedEvent>> {
```

glean-core/rlb/tests/custom_distribution_buffered.rs
```diff
@@ -11,6 +11,7 @@
 mod common;
 
 use glean::ConfigurationBuilder;
+use glean_core::TestGetValue;
```
badboy: Let's import it from the `glean` namespace in tests, so it looks more like outside code would also use it.
```suggestion
use glean::TestGetValue;
```

glean-core/rlb/tests/memory_distribution_buffered.rs
```diff
@@ -11,6 +11,7 @@
 mod common;
 
 use glean::{ConfigurationBuilder, ErrorType};
+use glean_core::TestGetValue;
```
badboy:
```suggestion
use glean::TestGetValue;
```

glean-core/rlb/tests/simple.rs
```diff
@@ -11,6 +11,7 @@
 mod common;
 
 use glean::ConfigurationBuilder;
+use glean_core::TestGetValue;
```
badboy:
```suggestion
use glean::TestGetValue;
```

glean-core/rlb/tests/timing_distribution_buffered.rs
```diff
@@ -11,6 +11,7 @@
 mod common;
 
 use glean::{ConfigurationBuilder, ErrorType};
+use glean_core::TestGetValue;
```
badboy:
```suggestion
use glean::TestGetValue;
```

glean-core/rlb/tests/timing_distribution_single_sample.rs
```diff
@@ -11,6 +11,7 @@
 mod common;
 
 use glean::{ConfigurationBuilder, ErrorType};
+use glean_core::TestGetValue;
```
badboy:
```suggestion
use glean::TestGetValue;
```

glean-core/rlb/src/private/event.rs
```diff
@@ -71,6 +71,13 @@ impl<K: traits::ExtraKeys> EventMetric<K> {
     }
 }
 
+#[inherent]
+impl<K: traits::ExtraKeys> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
```
badboy: We don't need the trait bound here.
```suggestion
impl<K> TestGetValue<Vec<RecordedEvent>> for EventMetric<K> {
```

glean-core/rlb/src/private/object.rs
```diff
@@ -39,6 +39,17 @@ impl<'a, K> MetricIdentifier<'a> for ObjectMetric<K> {
     }
 }
 
+impl<K: traits::ObjectSerialize> TestGetValue<JsonValue> for ObjectMetric<K> {
```
badboy: We don't need the trait bound here.
```suggestion
impl<K> TestGetValue<JsonValue> for ObjectMetric<K> {
```

badboy (changes requested):
Sorry for another round; we can leave out the trait bounds on two impls and make our life easier that way.
Once that's done we can get it in.

badboy:
fwiw, with the above changes and #3214 I have a patch for m-c that works: https://github.com/mozilla-firefox/firefox/commit/b5ef3adf76d5de67f42340dbabc9290eef10fcee

glean-core/src/metrics/timespan.rs
```diff
@@ -308,3 +284,29 @@ impl TimespanMetric {
         })
     }
 }
+
+impl TestGetValue<i64> for TimespanMetric {
+    /// **Test-only API (exported for FFI purposes).**
+    ///
+    /// Gets the currently stored value as an integer.
```
chutten, line 291: musing: Wow, this is an old comment. These days I hope we'd be specific about what the integer represents (a span of time in `time_unit` units. Definitely never negative. Only signed because of old Kotlin reasons)

glean-core/src/metrics/timing_distribution.rs
```diff
@@ -632,6 +613,27 @@ impl TimingDistributionMetric {
     }
 }
 
+impl TestGetValue<DistributionData> for TimingDistributionMetric {
+    /// **Test-only API (exported for FFI purposes).**
+    ///
+    /// Gets the currently stored value as an integer.
```
chutten, line 619: bug: Not an integer. (Not your bug, we never should've let that slip in originally).

badboy (approved):

glean-core/src/metrics/timespan.rs
badboy: we should clean up docs, but can be done afterwards.

glean-core/src/metrics/timing_distribution.rs
badboy: I'm ok if that happens in a cleanup afterwards, given it was there before

PR #3212, opened 2025-08-06
Bug 1899618 - Register Glean threads with the Gecko profiler (feature-gated)

badboy:
As I was working on some profiling today I was getting annoyed by not seeing those threads.
So I'm bringing this back now.

As noted last time in the bug doing that on Android is not possible because of ... things (which we hopefully gonna fix at some point), but for Desktop we can register just fine.

I split it up into 2 commits:
1. Refactoring things so we spawn using a single function
2. Adding the Gecko thread registration in that one unique place

The diff looks bigger than it is due to the change in indentation, but `?w=1` in the URL should make that less noisy.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,4 +1,4 @@
-personal_ws-1.1 en 301 utf-8
+personal_ws-1.1 en 302 utf-8
 AAR
 AARs
 ABI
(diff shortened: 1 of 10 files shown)


chutten (approved):
Is there any way to test that we don't accidentally add in new, unregistered threads? If it's not easy, then no worries: we don't tend to add threads all that often.

PR #3221, opened 2025-08-15
Bug 1982711 - Glean Health Ping with directory info before and after init

travis79:


glean-core/src/internal_metrics.rs
```diff
@@ -374,3 +375,60 @@ impl DatabaseMetrics {
...
+                dynamic_label: None,
+            }),
+        }
+    }
+}
+
+pub type DataDirectoryInfoObject = Vec<DataDirectoryInfoObjectItem>;
+
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Serialize, Deserialize)]
+#[allow(non_snake_case)]
```
badboy: This should be unnecessary.
```suggestion
```

glean-core/src/internal_metrics.rs
```diff
@@ -374,3 +375,60 @@ impl DatabaseMetrics {
...
+    #[serde(skip_serializing_if = "Option::is_none")]
+    pub file_count: Option<i64>,
+    #[serde(skip_serializing_if = "Vec::is_empty", default = "Vec::new")]
+    pub files: DataDirectoryInfoObjectItemItemFiles,
+}
+
+pub type DataDirectoryInfoObjectItemItemFiles = Vec<DataDirectoryInfoObjectItemItemFilesItem>;
+
+#[derive(Debug, Hash, Eq, PartialEq, Clone, Serialize, Deserialize)]
+#[allow(non_snake_case)]
```
badboy:
```suggestion
```

glean-core/src/lib.rs
```diff
@@ -1360,6 +1375,118 @@ pub fn glean_enable_logging_to_fd(fd: u64) {
...
+}
+
+fn record_dir_info_and_submit_health_ping(dir_info: Option<serde_json::Value>, reason: &str) {
+    core::with_glean(|glean| {
+        glean
+            .health_metrics
+            .data_directory_info
+            .set_sync(glean, dir_info.unwrap_or(serde_json::json!({})));
+    });
+    core::with_glean(|glean| {
```
badboy: Don't split this across two closures.
```suggestion
```

glean-core/src/lib.rs
```diff
@@ -557,6 +563,15 @@ fn initialize_inner(
...
         }
 
+        if !dispatcher::global::is_test_mode() {
+            // Now that Glean is initialized, we can capture the directory info from the pre_init phase and send it in
+            // a health ping with reason "pre_init".
+            record_dir_info_and_submit_health_ping(dir_info, "pre_init");
+
+            // Now capture a post_init snapshot of the state of Glean's data directories after initialization to send
+            // in a health ping with reason "post_init".
+            record_dir_info_and_submit_health_ping(collect_directory_info(data_path), "post_init");
```
badboy, line 579: We should probably NOT send this ping when `cfg.enable_internal_pings` is `false`.

glean-core/src/lib.rs
```diff
@@ -557,6 +563,15 @@ fn initialize_inner(
             Err(err) => log::error!("Unable to flush the preinit queue: {}", err),
         }
 
+        if !dispatcher::global::is_test_mode() {
```
badboy: Why disabling it in `test_mode`?
(I'm not even sure we reliably set that in e.g. m-c)

travis79: There were a lot of RLB tests and a few glean-core tests that failed because they didn't account for a new internal ping showing up at startup/init of Glean, so it was easier to avoid sending the ping when in test mode, and ensuring test mode was set than it was to adjust each individual test to  skip over the health ping where necessary.

travis79: So, now we _more reliably_ set test_mode, at least in glean-core and the rlb :)

Glean.xcframework.zip
badboy, line 1: shouldn't be committed

glean-core/src/lib.rs
badboy: Note that this is still a concern. If `cfg.enable_internal_pings` is `false` we should not even try to collect that info.
`enable_internal_pings` is `false` for some of the background tasks in m-c.

glean-core/metrics.yaml
```diff
@@ -970,3 +1019,136 @@ glean.ping:
...
+              properties:
+                file_name:
+                  type: string
+                file_created:
+                  type: number
+                file_modified:
+                  type: number
+                file_size:
+                  type: number
+  subdir_err:
```
chutten:
```suggestion
                  type: number

  subdir_err:
```

glean-core/rlb/examples/delayed-ping-data.rs
```diff
@@ -127,12 +127,13 @@ fn main() {
...
             log::debug!("incrementing by 10, waiting, shutdown. should trigger a flush.");
             glean_metrics::sample_counter.add(10);
+            std::thread::sleep(std::time::Duration::from_millis(100));
             glean::shutdown();
         }
         "submit_ping" => {
             log::info!("submitting PrototypePing");
             PrototypePing.submit(None);
-
+            std::thread::sleep(std::time::Duration::from_millis(100));
```
chutten: Gonna need some comments about what these sleeps are doing (ie, why they're needed here in the test, but not out in the wider world where people use the SDK)

glean-core/rlb/examples/enabled-pings.rs
```diff
@@ -219,5 +223,6 @@ fn main() {
             .is_none());
     }
 
+    std::thread::sleep(std::time::Duration::from_millis(100));
```
chutten, line 228: ditto sleep comment

glean-core/src/lib.rs
```diff
@@ -394,6 +397,15 @@ fn initialize_inner(
             log::set_max_level(level)
         }
 
+        let data_path_str = cfg.data_path.clone();
+        let data_path = Path::new(&data_path_str);
+        let internal_pings_enabled = cfg.enable_internal_pings;
+        let dir_info = if !is_test_mode() && internal_pings_enabled {
+            collect_directory_info(Path::new(&data_path))
```
chutten, line 404: Bug: This uses `core::with_glean` for the error metrics, but `core::with_glean` will not work (will panic `Global Glean object not initialized`) until `core::setup_glean`, below.

chutten (changes requested):

glean-core/src/lib.rs
travis79: 🤔   Well, I guess I'll have to cache them until later when they can be recorded

travis79: Oddly, I never saw this panic with any of the tests or examples

travis79: Perhaps I never triggered an error to be recorded in which it would panic

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
                 }
-                if let Ok(modified) = metadata.modified() {
-                    directory_info.dir_modified = Some(
-                        modified
-                            .duration_since(UNIX_EPOCH)
-                            .unwrap_or(Duration::ZERO)
-                            .as_secs() as i64,
-                    );
+                Err(error) => {
+                    let msg = format!("Unable to get metadata for subdir {}: {}", subdir, error);
```
chutten: We know which subdir it is because of `directory_info.dir_name`, so we can/should omit it.

We don't control `error`'s impl of `Display` and don't have limits or other protections on string fields of object metrics. I'm wary of including them as they might contain paths (if only we could just grab the name of the `ErrorKind` variant). I'd prefer that we either switch this to a bool (meaning "Some error happened, don't care which") like before or write a quick'n'dirty `fn as_safe_str(err: io::Error) -> &'static str { match err.kind() { NotFound => "not found", ...` to grab the ones we care about.

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                    directory_info.error_message = Some(msg.clone());
+                    log::error!("{}", msg);
```
chutten: I'm not sure we want this as an `error`. To me `error` would be fatal, whereas the code treats these errors as recoverable.

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
-                        .subdir_err
-                        .get(dir_path.to_string_lossy())
-                        .set(true);
-                });
-                log::error!("Unable to read directory {:?}", dir_path);
-                continue;
+            let entries = match fs::read_dir(&dir_path) {
+                Ok(entries) => entries,
+                Err(error) => {
+                    let msg = format!("Unable to read subdir {}: {}", subdir, error);
```
chutten: Ditto `subdir`, `directory_info.dir_name`, not trusting `impl Display for io::Error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                    directory_info.error_message = Some(msg.clone());
+                    log::error!("{}", msg);
```
chutten: ditto `error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                        file_size: None,
+                        error_message: None,
+                    };
+                let entry = match entry {
+                    Ok(entry) => entry,
+                    Err(error) => {
+                        let msg = format!("Unable to read file: {}", error);
+                        file_info.error_message = Some(msg.clone());
+                        log::error!("{}", msg);
+                        directory_info.files.push(file_info);
```
chutten: Defensively I wonder if we should push the `file_info` into `directory_info.files` on creation (in Firefox Desktop C++ land we'd use `nsTArray`'s `EmplaceBack`) so we can't forget to push it in the various loop increments.

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                        file_name: None,
+                        file_created: None,
+                        file_modified: None,
+                        file_size: None,
+                        error_message: None,
+                    };
+                let entry = match entry {
+                    Ok(entry) => entry,
+                    Err(error) => {
+                        let msg = format!("Unable to read file: {}", error);
```
chutten: Ditto not trusting `impl Display for io::Error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
-                            .subdir_entry_metadata_err
-                            .get(dir_path.to_string_lossy())
-                            .add(1);
-                    });
-                    log::error!("Unable to read entry metadata for entry {:?}", entry.path());
-                    continue;
+                let metadata = match entry.metadata() {
+                    Ok(metadata) => metadata,
+                    Err(error) => {
+                        let msg = format!("Unable to read file metadata: {}", error);
```
chutten: Ditto not trusting `impl Display for io::Error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                        file_info.file_name = Some(entry.file_name().to_string_lossy().to_string());
+                        file_info.error_message = Some(msg.clone());
+                        log::error!("Unable to read entry metadata: {}", error);
```
chutten: Ditto `error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                        file_modified: None,
+                        file_size: None,
+                        error_message: None,
+                    };
+                let entry = match entry {
+                    Ok(entry) => entry,
+                    Err(error) => {
+                        let msg = format!("Unable to read file: {}", error);
+                        file_info.error_message = Some(msg.clone());
+                        log::error!("{}", msg);
```
chutten: Ditto `error`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                            .unwrap_or(UNIX_EPOCH)
+                            .duration_since(UNIX_EPOCH)
+                            .unwrap_or(Duration::ZERO)
+                            .as_secs() as i64,
+                    );
+                    file_info.file_size = Some(metadata.len() as i64);
+                } else {
+                    let msg = format!(
+                        "Skipping non-file entry in {}: {:?}",
+                        subdir,
```
chutten: Ditto not needing `subdir`

glean-core/src/lib.rs
```diff
@@ -1399,100 +1399,119 @@ fn collect_directory_info(path: &Path) -> Option<serde_json::Value> {
...
+                        entry.file_name()
```
chutten: `file_name` doesn't `impl Display` and only `impl Debug`?

chutten: Oh, I see, it needs conversion later into `file_info.file_name`. Maybe pull `file_info.file_name` up and then you can use it in the msg?

glean-core/src/lib.rs
travis79: I tried this initially but then when I added the error message I moved the creation of the file info object up in order to just fill in the fields as I needed them and still push an object onto the array (before it just skipped the object so we wouldn't know anything about it). If I tried to push this into the files object the way I'm creating it up-front, then the borrow checker got me on trying to update the object after a partial move (when it was pushed into the files at creation).

travis79: Refactored this back to closer to what you suggest and what I had initially.

glean-core/rlb/tests/health_ping.rs
```diff
@@ -0,0 +1,115 @@
...
+                    Some((url, parsed_body, parsed_metadata))
+                } else {
+                    None
+                }
+            } else {
+                None
+            }
+        })
+        .collect()
+}
```
chutten, line 63: Seems like a twisting way to inspect ping contents. Why did you choose to roundtrip through disk instead of something like `interruptible_shutdown`'s `ReportingUploader`?

glean-core/rlb/tests/health_ping.rs
```diff
@@ -0,0 +1,115 @@
...
+/// And later the whole process is shutdown.
+#[test]
+fn test_pre_post_init_health_pings_exist() {
+    common::enable_test_logging();
+
+    // Create a custom configuration to use a validating uploader.
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().to_path_buf();
+
+    let cfg = ConfigurationBuilder::new(true, tmpname.clone(), "firefox-desktop")
```
chutten: If we're using "firefox-desktop" it seems like there's some specific meaning to it. I think this behaviour is app-id-agnostic so we should instead go with something clever ("app-ly-directly-to-the-forehead") or generic ("some-app-id").

chutten (approved):
r+wc

glean-core/rlb/tests/health_ping.rs
travis79: Copy pasta from the other test, let me change this.

glean-core/rlb/tests/health_ping.rs
travis79: Yeah, let me rework it to use a less convoluted approach.

travis79: Hmm I find the use of the crossbeam channels to be less reliable and it hangs on `recv`. I'm not sure why but the directory approach works every time so I'm sticking with it for now. Happy to rework this as a follow-up if I can figure out why the fake uploader trick wasn't working out.

PR #3223, opened 2025-08-18
Sync event databases on persist_ping_lifetime_data

badboy:
Instead of #3220 

~~I'd like to defer this until after the release.~~

### DIFF
--- a/glean-core/src/core/mod.rs
+++ b/glean-core/src/core/mod.rs
@@ -881,6 +881,8 @@ impl Glean {
             return data.persist_ping_lifetime_data();
         }
 
+        self.event_storage().sync_event_data();
+
         Ok(())
     }
 
--- a/glean-core/src/event_database/mod.rs
+++ b/glean-core/src/event_database/mod.rs
@@ -641,6 +641,15 @@ impl EventDatabase {
         result
     }
 
+    pub fn sync_event_data(&self) {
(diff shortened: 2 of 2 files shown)


chutten (changes requested):
Needs `CHANGELOG.md`

API design: Yeah, okay, it kinda tracks. Events _are_ ping-lifetime and this does persist them.

Tests: I guess we need to record events and then forceably crash the app to prove that `persist_ping_lifetime_data` is necessary to persist events to disk now?

Instrumentation: Do we want to monitor how often `sync_all` errors out? Seems likely. Though if it's recorded in a ping-lifetime metric we might not ever receive it.

Logging: Is it an `error`? There's nothing anyone can do about it. Seems like at most a `warn`. Do we want to spend some words on why this might matter (e.g. "Recently-recorded events for ping {file} have not been persisted to disk and may be lost")?

badboy:
> Logging: Is it an `error`? There's nothing anyone can do about it. Seems like at most a `warn`. Do we want to spend some words on why this might matter (e.g. "Recently-recorded events for ping {file} have not been persisted to disk and may be lost")?

The problem with that wording is: it's not accurate. if fsync fails we don't really know whether data is safely persisted or not. Or which data. Anyway, fun!

chutten (changes requested):
(about testing) Fair 'nuff.

(Still needs CHANGELOG)

PR #3232, opened 2025-08-25
Bug 1985037 - Gracefully handle inability to open the event storage file

badboy:


glean-core/src/event_database/mod.rs
```diff
@@ -344,22 +345,22 @@ impl EventDatabase {
...
-                        OpenOptions::new()
-                            .create(true)
-                            .append(true)
-                            .open(self.path.join(store_name))
-                            .unwrap(),
-                    )
-                }),
-        )
+    fn get_event_store(&self, store_name: &str) -> Result<Arc<File>, io::Error> {
+        let mut map = self.event_store_files.write().unwrap();
```
chutten, line 350: If this is a safe unwrap (I believe it is because poisoning's the only worry), then it needs a comment.

chutten (changes requested):
Needs CHANGELOG and a safety comment.

PR #3235, opened 2025-08-25
Refactor FXIOS-13276 - Adapt PingUploadRequest and CapablePingUploadRequest so they can be used to create custom uploaders

lmarceau:
[Task](https://mozilla-hub.atlassian.net/browse/FXIOS-13276)

### Context💡   
TLDR; I am looking to start using the interface added in https://github.com/mozilla/glean/pull/3175 so we can provide a custom uploader so Glean can send [OHTTP pings](https://github.com/mozilla-mobile/firefox-ios/pull/26949).

The new uploader we'll be providing to `Glean` is the [OHttpManager](https://github.com/mozilla/application-services/blob/release-v125/components/as-ohttp-client/ios/ASOhttpClient/OhttpManager.swift). I am implementing the integration between the `Glean` uploader interface and this `OHttpManager` to enable them to interoperate, as they expose different interfaces.

### Description ✏️
`Glean` creates the `PingUploadRequest` and we need the `OHttpManager` to access this `URLRequest` information so it can properly send it through the `OhttpSession`. But the request is currently private, so I am proposing here to make it and its related information public.

I also noted the `HttpPingUploader` is the one [making the check](https://github.com/mozilla/glean/blob/72a785653c6935d398796005342c0be177c3c62e/glean-core/ios/Glean/Net/HttpPingUploader.swift#L45) to ensure the ping is actually following the required capability. This is why I am proposing to make the `func capable(:)` function public. 

Let me know if I am missing something obvious 🙏

### DIFF
--- a/glean-core/ios/Glean/Net/PingUploader.swift
+++ b/glean-core/ios/Glean/Net/PingUploader.swift
@@ -36,7 +36,7 @@ public struct PingUploadRequest {
 }
 
 public struct CapablePingUploadRequest {
-    public let request: PingUploadRequest
+    private let request: PingUploadRequest
 
     init(_ request: PingUploadRequest) {
         self.request = request


glean-core/ios/Glean/Net/PingUploader.swift
```diff
@@ -36,7 +36,7 @@ struct PingUploadRequest {
 }
 
 public struct CapablePingUploadRequest {
-    private let request: PingUploadRequest
+    public let request: PingUploadRequest
```
badboy: Why does this need to be public? Isn't the whole point that `CapablePingUploadRequest` encapsulates the request and you can only get it by calling `capable`?

lmarceau: Because we need to create our own custom ping uploader, so an equivalent of `HttpPingUploader` but from the Firefox iOS codebase. We only get access to the `PingUploader.CapablePingUploadRequest` to build the `URLRequest` in Firefox iOS, and we can't build a `URLRequest` from that current `struct` since it gives no access to any of the necessary information.

There's a Slack channel discussing this, I'll add you to it 🙌

moztcampbell: Is your objection just to `request`? I think I agree with @badboy that `request` should still be hidden to ensure we always use `.capable`. 
If I understand correctly, the changes to `PingUploadRequest` are still needed so we extract raw request data _after_ we pass capability check. For desktop, we see to have all fields of PingUploadRequest public: https://searchfox.org/firefox-main/rev/856a307913c2b73765b4e88d32cf15ed05549cae/third_party/rust/glean/src/net/mod.rs#24-34

badboy: Yup, just on the wrapped `request` here.

lmarceau: If the `request` is not public, we cannot access the `url`, `data` and `headers` to build the `URLRequest` inside the [`OHttpManager`](https://github.com/mozilla/application-services/blob/release-v125/components/as-ohttp-client/ios/ASOhttpClient/OhttpManager.swift). I am unsure how this would work if we don't have this access? 

To ensure the `PingUploader` always use the `.capable` functionality, we should constrain in a slightly different manner IMO so that responsibility doesn't fall on the Client (and hence can be forgotten to be checked by any custom uploader, since it could be optional). I can propose some changes to do that if you want?

badboy: I'll take a look at this tomorrow again. I want to make this as easy as possible for the implementor, I just want to make sure that Glean exposes this in the right way then.

moztcampbell: The `capable` method should be returning the underlying `PingUploadRequest` for us, right? so we don't need to pull it out of `CapableUploadRequest` directly.

lmarceau: Yeah that would work! Just depends on the type of interface we want

badboy: So yeah, your uploader would implement `PingUploader`, just as [HttpPingUploader](https://github.com/mozilla/glean/blob/f694c339eabb0339b0e0c88cba88dfcc10f7e871/glean-core/ios/Glean/Net/HttpPingUploader.swift#L10) does.
For that you need to implement `func upload(request: CapablePingUploadRequest, callback: @escaping (UploadResult) -> Void`.

So your code gets a `CapablePingUploadRequest` and is supposed to call `capable` to get the underlying `PingUploadRequest` only after it verified it can satisfy the capabilities.

So all the other `public`s in this patch are required indeed, but the `private let request: PingUploadRequest` should stay, because an implementor is not supposed to access the `PingUploadRequest` directly.

lmarceau: I made the change! Let me know if anything else is needed 🙏

PR #3239, opened 2025-08-28
Refactor FXIOS-13276 - Missing public typealias

lmarceau:
Follow up on https://github.com/mozilla/glean/pull/3235

The headers are now public under the `PingUploadRequest`, but the typealias should be public too
```
public let headers: HeadersList
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v65.0.2...main)
 
+* Swift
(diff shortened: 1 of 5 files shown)


chutten (dismissed):
Since #3235 didn't include one, this PR should include a CHANGELOG entry for the changes involved here. I think this all is semantically `"Swift: BUGFIX: Refactor uploader interfaces to permit customization (#3235, #3239)"`, but feel free to suggest something better.

Is there a way to test this? We don't want to accidentally make future changes that'll make this interface un-extensible.

PR #3246, opened 2025-09-08
Bug 1967515 - Remove infill 0 buckets from custom distributions

badboy:
It's as easy as that.
Just some small code changes to satisfy existing tests.

not sure if we actually test for the buckets in the payload anywhere, CI will tell me.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v65.2.1...main)
 
+* General
+  * BREAKING: Remove infill 0 buckets from custom distributions ([#3246](https://github.com/mozilla/glean/pull/3246))
 * Swift
   * Make `EventMetricType`, `ObjectMetricType`, `URLMetricType` and `Ping` `Sendable`  ([#3255](https://github.com/mozilla/glean/pull/3255))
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -4,6 +4,7 @@
 
 * General
   * Added a Glean Health ping which collects telemetry health data into a single ping sent before and after initialization in order to track issues with Glean storage files and other telemetry health characteristics. ([#3221](https://github.com/mozilla/glean/pull/3221))
+  * Remove infill 0 buckets from custom distributions ([#3246](https://github.com/mozilla/glean/pull/3246))
```
badboy: Teeeechnically this is breaking I guess, as [it was when we did it before](https://github.com/mozilla/glean/blob/11d6ad6930c2fcd797585a7e01514d0b20768b47/CHANGELOG.md?plain=1#L229).

But it's not breaking in the code, "just" the pipeline.

badboy:
@chutten (for when you return) 
1. Breaking in code yes/no?
2. Does that need other work in the pipeline/GLAM and do we thus need to coordinate when releasing?

chutten:
> @chutten (for when you return)
> 
>     1. Breaking in code yes/no?

It _might_ be breaking because I believe we infill zeroes for test functions?

>     2. Does that need other work in the pipeline/GLAM and do we thus need to coordinate when releasing?

Nope. We're clear to land when we want to.

PR #3255, opened 2025-09-18
Make event metric type, object metric type, and ping Sendable

Cramsden:
Add Sendable conformance to:
- EventMetricType
- ObjectMetricType
- Ping
Some of these conformances are marked as unchecked sendable. TODO comments added to make these unchecked sendable in this bugzilla ticket https://bugzilla.mozilla.org/show_bug.cgi?id=1987503


glean-core/ios/Glean/Metrics/LabeledMetric.swift
```diff
@@ -10,7 +10,8 @@
 /// Unlike most metric types, LabeledMetricType does not have its own corresponding storage engine,
 /// but records metrics for the underlying metric type `T` in the storage engine for that type.
 /// The only difference is that labeled metrics are stored with the special key `$category.$name/$label`.
-public class LabeledMetricType<T> {
+// TODO: Make LabeledMetricType actaully sendable.
```
badboy:
```suggestion
// TODO: Make LabeledMetricType actually sendable.
```

Can you expand on why that's currently not possible?

badboy (approved):

PR #3259, opened 2025-09-22
Bug 1989902 - Rust: Use an associated type for TestGetValue to make it clear it's an output parameter

badboy:
Generic parameters are an "input".
A trait `TestGetValue<T>` can be implemented for the same type multiple times with different concrete types `T`.
That's never needed for `TestGetValue`, we only have a single type that should be returned from `test_get_value`.
An associated parameter is the way to do that.


jeddai (approved):
Ah this is definitely a good change, great call

PR #3262, opened 2025-09-24
bug 1990624 - Add running count of inits to 'health' ping

chutten:


glean-core/metrics.yaml
```diff
@@ -1100,3 +1100,21 @@ glean.health:
...
+  init_count:
+    type: counter
+    lifetime: user
+    description: |
+      A running count of how many times the Glean SDK has been initialized.
+    notification_emails:
+      - glean-team@mozilla.org
+      - chutten@mozilla.com
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?1990624
```
travis79: This and the data_review urls are invalid/don't land on the bug, you dropped the `id=` and it broke things :) 
```suggestion
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1990624
```

glean-core/metrics.yaml
```diff
@@ -1100,3 +1100,21 @@ glean.health:
...
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?1990624
```
travis79:
```suggestion
      - https://bugzilla.mozilla.org/show_bug.cgi?id=1990624
```

glean-core/metrics.yaml
```diff
@@ -1100,3 +1100,21 @@ glean.health:
                   type: number
                 error_message:
                   type: string
+
+  init_count:
+    type: counter
```
Dexterp37, line 1105: @chutten @travis79 The current approach relies on the storage (since we need to store the "init counter"). What if instead we use the "legacy telemetry" session id approach?

When Glean initializes, we generate a uuid and call that "session id". It's not persisted, but we'll slap it on the pings that need it when we assemble the ping, like we do for other glean metadata, with the exception that this doesn't have to be stored . And we can scope it to the health ping only for now, as you did.

This way we don't rely on previous runs for the proper count and have distinct ids after Glean starts, regardless of the potential bugs.

chutten: I'm not sure what benefit your approach has over combining `client_id` and `glean.health.init_count` together. On the contrary, it prevents us from counting and ordering Glean sessions (since there's no relationship between your proposed pseudo-application-lifetime session ids). 

I suppose it would help if the db were found to be volatile _within_ a session, but so far we haven't seen evidence of that in the wild (unless data platform infra isn't telling us about pings that no longer satisfy the schema because all the db-resident client_info and ping_info couldn't be found during ping assembly).

Does that make sense, @Dexterp37 ?

Dexterp37: > I'm not sure what benefit your approach has over combining `client_id` and `glean.health.init_count` together. On the contrary, it prevents us from counting and ordering Glean sessions (since there's no relationship between your proposed pseudo-application-lifetime session ids).

The suggestion doesn't prevent counting sessions (distinct uuid count per client id gets us there, unless I'm missing something). It doesn't support ordering, that is true, but that is addressed by the creation time (well, assuming time works!).

If, in the context of the incident that required this, we're good with potentially losing the count across inits, I'm fine with this!

chutten: Ah-ha! You've fallen into my trap! Counting distinct session ids per client id requires a stable client_id, which presupposes stable storage, which would mean we'd not risk losing the count across inits : D

Dexterp37: Counting sessions with a counter _also_ implies stable storage, with the additional implications of using a user-lifetime metric.

With that said, if you and Travis agree this is the path, that's ok!

PR #3264, opened 2025-09-26
Bug 1932617 - Split out test_get_num_recorded_errors for events

badboy:


badboy:
> So, now there is no `test_get_num_recorded_errors` on a public trait for Event?

Correct. For the sole reason that we need to split out the `impl` so we can later take a `dyn Trait` of it.
But let me file a bug, we could consider bringing it back as a separate trait and do that for all metrics.

PR #3272, opened 2025-09-30
Bug 1987918 - Log failing to persist data as an info note

badboy:
Logging them as errors suggest there's a way to fix those, when often they are due to error conditions we don't control but gracefully handle.


badboy:
> Should we consider instrumenting these failures?

Woops, merged without answering the question.
Weeeeeell. This is us failing to write to the database. Instrumenting them would mean we need to write to the database. Which we just encountered we can't do. I don't see a reliable way to instrument this and have it reported.

PR #3231, opened 2025-08-25
Bug 1983858 - Implement ObjectSerialize recursively

badboy:
Needs https://github.com/mozilla/glean_parser/pull/808 to go along with it.

PR #3273, opened 2025-10-01
Bug 1982687 - Implement object `oneOf` support

badboy:
The majority of the work happens in glean_parser: https://github.com/mozilla/glean_parser/pull/803
This also requires: https://github.com/mozilla/glean/pull/3231


samples/python/README.md
```diff
@@ -3,8 +3,7 @@
 ## Development
 
 ```
-make dev
-. .venv/bin/activate
+uv sync
```
jeddai, line 6: What's this do?

jeddai (approved):

samples/python/README.md
badboy: We have switched the sample project to use [`uv`](https://docs.astral.sh/uv/) some time ago. `uv sync` ensures the virtual env exists and installs all required dependencies.

PR #3274, opened 2025-10-01
CI: Terminate slow Rust tests after 2 minutes

badboy:


badboy:
Some error here kept the test running indefinitely for hours: https://app.circleci.com/pipelines/github/mozilla/glean/13828/workflows/d8ef45d1-7415-40c0-8c4c-92115a8f2d36/jobs/270813
There's no point. The Rust tests should all run in at most 10s, thus 30s is a decent buffer to consider them "slow" and a 2 minute hard-limit is good to keep our CI runtime in check.

PR #3275, opened 2025-10-01
Bug 1991201 - CI: Use a make target to unify wheel upload

badboy:


Makefile
```diff
@@ -207,3 +207,8 @@ coverage-python: build-python ## Generate a code coverage report for Python
 	$(GLEAN_PYENV)/bin/python3 -m coverage combine
 	$(GLEAN_PYENV)/bin/python3 -m coverage html
 .PHONY: coverage-python
+
+upload-wheels: setup-python
+	VIRTUAL_ENV=$(GLEAN_PYENV) \
+		$(GLEAN_PYENV)/bin/python3 -m twine --version
```
jeddai: Is there a reason we use --version here instead of calling upload?

badboy: Yes, testing! These are otherwise release tasks, so the second commit here is just to run them once and double-check that it all works as planned. It will then be removed before we land this.

badboy:
Test show that it works:
* https://app.circleci.com/pipelines/github/mozilla/glean/13832/workflows/4bafb4a8-ba48-4da7-9277-56c331530025/jobs/270922
* https://app.circleci.com/pipelines/github/mozilla/glean/13832/workflows/4bafb4a8-ba48-4da7-9277-56c331530025/jobs/270921

jeddai (approved):

PR #3279, opened 2025-10-02
Bug 1948866 - Use awake timestamps for timespan/timing distribution

badboy:
I need to check the other uses for correctness.
Right now this uses https://github.com/badboy/zeitstempel/pull/5. I will need to cut a release of that first.


glean-core/src/event_database/mod.rs
```diff
@@ -219,7 +219,7 @@ impl EventDatabase {
                     self.record(
                         glean,
                         &glean_restarted.into(),
-                        crate::get_timestamp_ms(),
+                        crate::get_awake_timestamp_ms(),
```
chutten: Isn't it odd to have glean.restarted use awake timestamp but all other events use including-sleep timestamp?

badboy: woops, that's a mistake indeed.

PR #3292, opened 2025-10-21
Bug 1994757 - Store the client ID in an additional plaintext file and report on regeneration

badboy:
This is a first approach at putting [this state diagram](https://tmp.fnordig.de/mozilla/client_id_txt.html) into code. 

This is not yet pretty. or organized. or ready to be looked at.
But it's not not wrong and I will expand upon that.

### DIFF
--- a/docs/dev/core/internal/client_id_recovery.md
+++ b/docs/dev/core/internal/client_id_recovery.md
@@ -1,0 +1,58 @@
+# Client ID recovery
+
+Currently (2025-10-31, Glean v66) we see some unexplained Glean SDK database resets.
+These are noticeable in data as client ID regenerations:
+A client application with telemetry enabled, which previously already sent data,
+regenerates its client ID on initialize and thus looks like a new client.
+
+That's undesirable and a bug.
+However we have yet to track down the actual faulty code path.
+Until that bug is found and fixed, the Glean SDK provides an extra mitigation.
+
+From Glean v66.1.0 on the SDK will store the client ID in a `client_id.txt` in the provided data path.
+Any inconsistencies in that data compared to the database will be reported
+and, if applicable, the client ID restored.
+
+**Note:** Glean v66.1.0 will only report the inconsistency, but will not restore a recovered client ID.
+This allows us to measure the impact.
+The mitigation will be enabled in a later release ([bug 1996862](https://bugzilla.mozilla.org/show_bug.cgi?id=1996862)).
+
+The exact flow of decisions is depicted in the chart below.
+The implementation is in [`glean-core/src/core/mod.rs`](https://github.com/mozilla/glean/blob/HEAD/glean-core/src/core/mod.rs#L264)
+
+```mermaid
+flowchart TD
+    A["Glean.init"] -->B
+    B{client_id.txt exists?} -->|yes| C
+    B -->|no| D
+    C["(a) load file ID"] --> E
+    D["load DB ID"] --> D3
+    D3{DB ID empty} -->|yes| D4
+    D3 -->|no| S
+    D4["generate DB ID"] --> S
+    E{valid file and ID?} -->|yes| H
+    E -->|no| G
+    G["(b) record file read error"] --> H
+    H{"(c) DB size <= 0"} -->|yes| J
+    H -->|no| F
+    J["(d) record empty DB error
+    report recovered ID: file ID"] --> Q
+    F["load DB ID"] --> N
+    L{file ID == DB ID} --> |yes| Z
+    L -->|no| T
+    N{DB ID empty?} -->|yes| O
+    N -->|no| L
+    O["(f) record regen error"] --> Q
+    P["(g) record mismatch error
+    report recovered ID: file ID"] --> S
+    Q["(e) mitigation:
+    set DB ID = file ID"] --> Z
+    S["(h) write DB ID to file"] --> Z
+    T{"DB ID == 'c0ffee'"}
+    T -->|yes| U
+    T -->|no| P
+    U["(i) record c0ffee error
+    report recovered ID: file ID"] --> Q
+    
+    Z(normal operation)
+```
--- a/docs/dev/core/internal/index.md
+++ b/docs/dev/core/internal/index.md
@@ -21,3 +21,4 @@ This includes:
 * [Debug Pings](debug-pings.md)
 * [Upload mechanism](upload.md)
 * [Implementations](implementations.md)
+* [Client ID recovery](client_id_recovery.md)
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -1158,3 +1158,92 @@ glean.health:
     expires: never
     send_in_pings:
       - health
+
+  exception_state:
+    type: string
+    lifetime: ping
+    description: |
+      An exceptional state was detected upon trying to laod the database.
+
+      Valid options are:
+        - empty-db
+        - regen-db
+        - c0ffee-in-db
+        - client-id-mismatch
+    notification_emails:
+      - glean-team@mozilla.org
+      - jrediger@mozilla.com
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757#c2
+    data_sensitivity:
+      - technical
+    expires: never
+    send_in_pings:
+      - health
+
+  recovered_client_id:
+    type: uuid
+    lifetime: ping
+    description: |
+      A client_id recovered from a `client_id.txt` file on disk.
+      Only expected to have a value for the exception states `empty-db`, `c0ffee-in-db` and `client-id-mismatch`.
+      See `exception_state` for different exception states when this can happen.
+    notification_emails:
+      - glean-team@mozilla.org
+      - jrediger@mozilla.com
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757#c2
+    data_sensitivity:
+      - technical
+    expires: never
+    send_in_pings:
+      - health
+
+  file_read_error:
+    type: labeled_counter
+    lifetime: ping
+    description: |
+      Count of different errors that happened when trying to read the `client_id.txt` file from disk.
+    notification_emails:
+      - glean-team@mozilla.org
+      - jrediger@mozilla.com
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757#c2
+    data_sensitivity:
+      - technical
+    expires: never
+    send_in_pings:
+      - health
+    labels:
+      - parse
+      - permission-denied
+      - io
+      - c0ffee-in-file
+
+  file_write_error:
+    type: labeled_counter
+    lifetime: ping
+    description: |
+      Count of different errors that happened when trying to write the `client_id.txt` file to disk.
+    notification_emails:
+      - glean-team@mozilla.org
+      - jrediger@mozilla.com
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=1994757#c2
+    data_sensitivity:
+      - technical
+    expires: never
+    send_in_pings:
+      - health
+    labels:
+      - permission-denied
+      - io
+      - not-found
--- a/glean-core/rlb/tests/health_ping_file_overwrite.rs
+++ b/glean-core/rlb/tests/health_ping_file_overwrite.rs
@@ -1,0 +1,85 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+//! This integration test should model how the RLB is used when embedded in another Rust application
+//! (e.g. FOG/Firefox Desktop).
+//!
+//! We write a single test scenario per file to avoid any state keeping across runs
+//! (different files run as different processes).
+
+mod common;
+
+use std::{fs, io::Read};
+
+use crossbeam_channel::{bounded, Sender};
+use flate2::read::GzDecoder;
+use glean::{net, ConfigurationBuilder};
+use serde_json::Value as JsonValue;
+
+// Define a fake uploader that reports when and what it uploads.
+#[derive(Debug)]
+struct ReportingUploader {
+    sender: Sender<JsonValue>,
+}
+
+impl net::PingUploader for ReportingUploader {
+    fn upload(&self, upload_request: net::CapablePingUploadRequest) -> net::UploadResult {
+        let upload_request = upload_request.capable(|_| true).unwrap();
+        let body = upload_request.body;
+        let decode = |body: Vec<u8>| {
+            let mut gzip_decoder = GzDecoder::new(&body[..]);
+            let mut s = String::with_capacity(body.len());
+
+            gzip_decoder
+                .read_to_string(&mut s)
+                .ok()
+                .map(|_| &s[..])
+                .or_else(|| std::str::from_utf8(&body).ok())
+                .and_then(|payload| serde_json::from_str(payload).ok())
+                .unwrap()
+        };
+
+        self.sender.send(decode(body)).unwrap();
+        net::UploadResult::http_status(200)
+    }
+}
+
+/// Test scenario: Write a client ID to the backup file and check that it's used after initialization.
+#[test]
+fn test_pre_post_init_health_pings_exist() {
+    common::enable_test_logging();
+
+    // Create a custom configuration to use a validating uploader.
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().to_path_buf();
+
+    let client_id = "e03cc2de-bc8b-4f9c-862f-b474d910899e";
+
+    // We write a random but fixed client ID, without there being a Glean database.
+    let clientid_txt = tmpname.join("client_id.txt");
+    fs::write(&clientid_txt, client_id.as_bytes()).unwrap();
+
+    let (tx, rx) = bounded(1);
+    let cfg = ConfigurationBuilder::new(true, tmpname.clone(), "health-ping-test")
+        .with_server_endpoint("invalid-test-host")
+        .with_use_core_mps(false)
+        .with_uploader(ReportingUploader { sender: tx })
+        .build();
+    common::initialize(cfg);
+
+    glean_core::glean_test_destroy_glean(false, Some(tmpname.display().to_string()));
+
+    // Check for the initialization pings.
+    // Wait for the ping to arrive.
+    let payload = rx.recv().unwrap();
+
+    let exception_state = &payload["metrics"]["string"]["glean.health.exception_state"];
+    assert_eq!("empty-db", exception_state);
+    let exception_uuid = &payload["metrics"]["uuid"]["glean.health_recovered_client_id"];
+    assert_eq!(&JsonValue::Null, exception_uuid);
+
+    // TODO(bug 1996862): We don't run the mitigation yet.
+    //let ping_client_id = &payload["client_info"]["client_id"];
+    //assert_eq!(client_id, ping_client_id);
+}
--- a/glean-core/src/core/mod.rs
+++ b/glean-core/src/core/mod.rs
@@ -3,6 +3,8 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 use std::collections::HashMap;
+use std::fs::{self, File};
+use std::io::{self, Write};
 use std::path::{Path, PathBuf};
 use std::sync::atomic::{AtomicU8, Ordering};
 use std::sync::{Arc, Mutex};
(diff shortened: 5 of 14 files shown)


glean-core/rlb/tests/health_ping_file_overwrite.rs
```diff
@@ -0,0 +1,112 @@
...
+};
+
+use glean::{net, ConfigurationBuilder};
+use serde_json::Value as JsonValue;
+
+// Define a fake uploader that sleeps.
+#[derive(Debug)]
+struct FakeUploader;
+
+impl net::PingUploader for FakeUploader {
```
chutten: Why'd you choose the pending ping dir style of fake uploader instead of `ReportingUploader` style? Or, heck, one that just unzips the payload and puts it in a Vec?

(( Come to think on it, why do basically all of the fakeuploader-style tests use the filesystem anyway? Why not just put stuff in memory like xpcshell's pingserver? ))

glean-core/rlb/tests/health_ping_file_overwrite.rs
```diff
@@ -0,0 +1,112 @@
...
+        .into_iter()
+        .find(|(url, body, _)| url.contains("health") && body["ping_info"]["reason"] == "pre_init")
+        .unwrap();
+
+    let exception_state = &payload["metrics"]["string"]["glean.health.exception_state"];
+    assert_eq!("empty-db", exception_state);
+    let exception_uuid = &payload["metrics"]["uuid"]["glean.health_recovered_client_id"];
+    assert_eq!(&JsonValue::Null, exception_uuid);
+
+    // We don't run the mitigation yet.
```
chutten: Please file and reference a bug number here

glean-core/src/core/mod.rs
```diff
@@ -257,6 +264,27 @@ impl Glean {
...
+            }
+            Err(ClientIdFileError::ParseError(e)) => {
+                log::trace!("reading cliend_id.txt. Could not parse into UUID: {e}");
+                None
+            }
+            Err(ClientIdFileError::IoError(e)) => {
+                // We can't handle other IO errors (most couldn't occur on this operation anyway)
+                log::trace!("reading client_id.txt. Unexpected io error: {e}");
+                None
+            }
```
chutten, line 329: What are your opinions on reporting these failures?

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
             ping_lifetime_max_time,
         )?);
 
+        {
+            // We just set it.
```
chutten: Comment is lacking context and maybe would be better removed.

glean-core/src/database/mod.rs
```diff
@@ -147,7 +147,7 @@ pub struct Database {
     pub(crate) write_timings: RefCell<Vec<i64>>,
 
     /// The database size at various phases of loading.
-    load_sizes: Option<LoadSizesObject>,
+    pub(crate) load_sizes: Option<LoadSizesObject>,
```
chutten, line 150: Guess we no longer need the accessor function then, huh.

glean-core/src/core/mod.rs
```diff
@@ -373,6 +486,40 @@ impl Glean {
         self.data_store = None;
     }
 
+    fn client_id_file_path(&self) -> PathBuf {
+        let mut path = self.data_path.clone();
+        path.push(CLIENT_ID_PLAIN_FILENAME);
+        path
```
chutten: [Docs say](https://doc.rust-lang.org/std/path/struct.PathBuf.html#method.push)
> Consider using Path::join if you need a new PathBuf instead of using this function on a cloned PathBuf.
```suggestion
        self.data_path.join(CLIENT_ID_PLAIN_FILENAME)
```

glean-core/src/core/mod.rs
```diff
@@ -373,6 +486,40 @@ impl Glean {
...
+
+        Ok(())
+    }
+
+    /// Try to load a client ID from the plain file on disk.
+    fn client_id_from_file(&self) -> Result<Uuid, ClientIdFileError> {
+        let uuid_str = fs::read_to_string(self.client_id_file_path())?;
+        // We don't write a newline, but we still trim it. Who knows who else touches that file by accident.
+        // We're also a bit more lenient in what we accept here: uppercase, lowercase, with or without dashes.
+        let uuid = Uuid::parse_str(uuid_str.trim_end())?;
```
chutten: Docs say:
> Prefer try_parse unless you need detailed user-facing diagnostics. This method will be eventually deprecated in favor[sic] of try_parse.

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
...
 
+        {
+            // We just set it.
+            let data_store = glean.data_store.as_ref().unwrap();
+            let db_load_sizes = data_store.load_sizes.as_ref().unwrap();
+            let new_size = db_load_sizes.new.unwrap_or(0);
+
+            // If we have a client ID on disk, we check the database
+            if let Some(stored_client_id) = stored_client_id {
+                if new_size <= 0 {
```
travis79, line 340: Can you map these to the states in your diagram with comments, just to make it easier to reference later?

glean-core/src/core/mod.rs
```diff
@@ -373,6 +486,40 @@ impl Glean {
...
+        fp.sync_all()?;
+
+        Ok(())
+    }
+
+    /// Try to load a client ID from the plain file on disk.
+    fn client_id_from_file(&self) -> Result<Uuid, ClientIdFileError> {
+        let uuid_str = fs::read_to_string(self.client_id_file_path())?;
+        // We don't write a newline, but we still trim it. Who knows who else touches that file by accident.
+        // We're also a bit more lenient in what we accept here: uppercase, lowercase, with or without dashes.
```
chutten: Heck, we also support `urn` and `braced`

glean-core/src/core/mod.rs
```diff
@@ -384,7 +531,8 @@ impl Glean {
             Some(uuid) => uuid == *KNOWN_CLIENT_ID,
         };
         if need_new_client_id {
-            self.core_metrics.client_id.generate_and_set_sync(self);
+            let new_clientid = self.core_metrics.client_id.generate_and_set_sync(self);
+            _ = self.store_client_id(new_clientid);
```
travis79: Not even logging here if there's an error writing the file or something?

chutten: We ought to report resulting errors, don't you think?

glean-core/src/core/mod.rs
```diff
@@ -619,6 +767,10 @@ impl Glean {
             log::warn!("Error clearing pending pings: {}", err);
         }
 
+        if let Err(e) = self.remove_stored_client_id() {
+            log::error!("Couldn't remove client ID on disk. This might lead to a resurrection of this client ID later. Error: {e}");
```
chutten, line 864: Report?

travis79 (approved):
Nothing concerning here, r+ from me pending data-review

glean-core/src/core/mod.rs
chutten: Also: what's the behaviour if we're init-ed with upload disabled? Will we log this error on every init?

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+
+#[test]
+fn reused_clientid_from_file() {
+    let temp = tempfile::tempdir().unwrap();
+    let new_uuid = Uuid::new_v4();
+
+    write_clientid_to_file(temp.path(), &new_uuid).unwrap();
+
+    let (glean, temp) = new_glean(Some(temp));
+    // We don't run the mitigation yet
```
chutten: Ref to-be-filed bug number

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+
+    let db_client_id2 = clientid_metric().get_value(&glean, None).unwrap();
+    assert_eq!(db_client_id, db_client_id2);
+
+    glean.submit_ping_by_name("health", Some("pre_init"));
+    let mut pending = get_queued_pings(temp.path()).unwrap();
+    assert_eq!(1, pending.len());
+    let payload = pending.pop().unwrap().1;
+
+    let state = payload["metrics"]["string"]["glean.health.exception_state"].as_str();
```
chutten, line 113: musing: Should we consider an absent file to be a state worth reporting... I mean, obviously, the first init after updating will always fall into this case. But if it happens later, would we want to know? Would we want to monitor this over time?

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+        let client_id_key = "glean_client_info#client_id";
+
+        let mut writer = rkv.write().unwrap();
+        user_store.delete(&mut writer, client_id_key).unwrap();
+        writer.commit().unwrap();
+    }
+
+    let (glean, temp) = new_glean(Some(temp));
+
+    // We don't run the mitigation yet
```
chutten: Ref to-be-filed bug number

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+    let (file_client_id, temp) = {
+        // Ensure we initialize once to get a client_id
+        let (glean, temp) = new_glean(None);
+        let file_client_id = clientid_from_file(temp.path()).unwrap();
+        drop(glean);
+
+        (file_client_id, temp)
+    };
+
+    // We modify the database and ONLY clear out the client id.
```
chutten: We're not clearing it out. Copy-pasta, I think

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+        let known_client_id = rkv::Value::Blob(&encoded);
+        user_store
+            .put(&mut writer, client_id_key, &known_client_id)
+            .unwrap();
+        writer.commit().unwrap();
+    }
+
+    let (glean, temp) = new_glean(Some(temp));
+
+    // We don't run the mitigation yet
```
chutten: Ref to-be-filed bug

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
...
+                                "c0ffee client_id in DB, stored_client_id={stored_client_id}"
+                            );
+                            glean
+                                .health_metrics
+                                .recovered_client_id
+                                .set_from_uuid_sync(&glean, stored_client_id);
+                            glean
+                                .health_metrics
+                                .exception_state
+                                .set_sync(&glean, ExceptionState::C0ffeeInDb);
```
chutten, line 382: We don't take any resolution actions here?

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
...
+
+                    match db_client_id {
+                        None => {
+                            log::trace!("no client_id in DB. stored_client_id={stored_client_id}");
+                            glean
+                                .health_metrics
+                                .exception_state
+                                .set_sync(&glean, ExceptionState::RegenDb);
+                        }
+                        Some(db_client_id) if db_client_id == *KNOWN_CLIENT_ID => {
```
chutten, line 370: We've stopped storing `c0ffee` in db since, what, v64? When should we stop accounting for it?

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
...
+                                .exception_state
+                                .set_sync(&glean, ExceptionState::C0ffeeInDb);
+                        }
+                        Some(db_client_id) if db_client_id == stored_client_id => {
+                            // all valid. nothing to do
+                            log::trace!("database consistent. db_client_id == stored_client_id: {db_client_id}");
+                        }
+                        Some(db_client_id) => {
+                            log::trace!(
+                                "client_id mismatch. db_client_id{db_client_id}, stored_client_id={stored_client_id}"
```
chutten: Might be good to include in the log the action we're taking as a result: "Overwriting file with db's client_id"

glean-core/src/core/mod.rs
```diff
@@ -269,6 +297,88 @@ impl Glean {
...
+            } else {
+                log::trace!("No stored client ID. Database might have it.");
+
+                let db_client_id = glean
+                    .core_metrics
+                    .client_id
+                    .get_value(&glean, Some("glean_client_info"));
+                if let Some(db_client_id) = db_client_id {
+                    if let Err(e) = glean.store_client_id(db_client_id) {
+                        log::error!("Could not write {db_client_id} to state file. Might happen on next init then. Error: {e}");
```
chutten: Report?

glean-core/src/core/mod.rs
chutten: Oh, I guess the db will have the c0ffee id scrubbed from it later... maybe?

glean-core/metrics.yaml
```diff
@@ -1158,3 +1158,46 @@ glean.health:
...
+    expires: never
+    send_in_pings:
+      - health
+
+  recovered_client_id:
+    type: uuid
+    lifetime: ping
+    description: |
+      A client_id recovered from a `client_id.txt` file on disk.
+      See `exception_state` for different exception states when this can happen.
```
chutten, line 1192: The description should probably mention that this is only expected to have a value if `exception_state` is one of "empty-db", "c0ffee-in-db", or "client-id-mismatch".

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+    //let db_client_id = clientid_metric().get_value(&glean, None).unwrap();
+    //assert_eq!(file_client_id, db_client_id);
+
+    glean.submit_ping_by_name("health", Some("pre_init"));
+    let mut pending = get_queued_pings(temp.path()).unwrap();
+    assert_eq!(1, pending.len());
+    let payload = pending.pop().unwrap().1;
+
+    let state = payload["metrics"]["string"]["glean.health.exception_state"].as_str();
+    assert_eq!(Some("c0ffee-in-db"), state);
```
chutten, line 239: You should test the restored client_id metric in this case.

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,314 @@
...
+    let file_client_id = clientid_from_file(temp.path()).unwrap();
+    assert_eq!(file_client_id, db_client_id);
+
+    glean.submit_ping_by_name("health", Some("pre_init"));
+    let mut pending = get_queued_pings(temp.path()).unwrap();
+    assert_eq!(1, pending.len());
+    let payload = pending.pop().unwrap().1;
+
+    let state = payload["metrics"]["string"]["glean.health.exception_state"].as_str();
+    assert_eq!(Some("client-id-mismatch"), state);
```
chutten, line 186: You should test the restored client_id metric in this case.

glean-core/rlb/tests/health_ping_file_overwrite.rs
badboy: copy-pasta. I just copied one of the existing test cases. You are right that the other implementations make more sense.

glean-core/src/database/mod.rs
badboy: the accessor takes the value: https://github.com/mozilla/glean/blob/da549018d1aa61fb4fddf447307a8c7704c1bb70/glean-core/src/database/mod.rs#L325-L326

we only need to access it in other places.

glean-core/src/core/mod.rs
badboy: tackled the init-with-upload-disabled. we don't need to log a not-found error.

glean-core/src/core/mod.rs
badboy: good idea to test.

glean-core/src/core/mod.rs
badboy: probably when we don't see this error reported from recent versions?

docs/dev/core/internal/client_id_recovery.md
```diff
@@ -0,0 +1,57 @@
...
+However we have yet to track down the actual faulty code path.
+Until that bug is found and fixed, the Glean SDK provides an extra mitigation.
+
+From Glean v66.1.0 on the SDK will store the client ID in a `client_id.txt` in the provided data path.
+Any inconsistencies in that data compared to the database will be reported
+and, if applicable, the client ID restored.
+
+**Note:** Glean v66.1.0 will only report the inconsistency, but will not restore a recovered client ID.
+This allows us to measure the impact.
+The mitigation will be enabled in a later release.
```
chutten: ref the bug

docs/dev/core/internal/index.md
```diff
@@ -13,10 +13,12 @@ This chapter describes aspects of the Glean SDKs that are internal implementatio
...
+* [Clearing metrics when disabling/enabling Glean](clearing.md).
 * [Database format](database.md)
-* [Payload format](payload.md)
-* [Directory structure](directory-structure.md)
+* [Payload format](payload.md).
+* [Directory & file structure](directory-structure.md)
 * [Debug Pings](debug-pings.md)
+* [Upload mechanism](upload.md)
 * [Implementations](implementations.md)
+* [Client ID recovery](client_id_recovery.md)
```
chutten, line 24: Uh, what's with the periods for some thing going on here?

glean-core/rlb/tests/health_ping_file_overwrite.rs
```diff
@@ -0,0 +1,85 @@
...
+fn test_pre_post_init_health_pings_exist() {
+    common::enable_test_logging();
+
+    // Create a custom configuration to use a validating uploader.
+    let dir = tempfile::tempdir().unwrap();
+    let tmpname = dir.path().to_path_buf();
+
+    let client_id = "e03cc2de-bc8b-4f9c-862f-b474d910899e";
+
+    // We write a random but fixed client ID, without their being a Glean database.
```
chutten:
```suggestion
    // We write a random but fixed client ID, without there being a Glean database.
```

docs/dev/core/internal/client_id_recovery.md
```diff
@@ -0,0 +1,57 @@
...
+flowchart TD
+    A["Glean.init"] -->B
+    B{client_id.txt exists?} -->|yes| C
+    B -->|no| D
+    C["(a) load file ID"] --> E
+    D["load DB ID"] --> D3
+    D3{DB ID empty} -->|yes| D4
+    D3 -->|no| S
+    D4["generate DB ID"] --> S
+    E{valid ID?} -->|yes| H
```
chutten:
```suggestion
    E{valid file and ID?} -->|yes| H
```

glean-core/src/core/mod.rs
```diff
@@ -269,6 +276,155 @@ impl Glean {
...
+
+        {
+            let data_store = glean.data_store.as_ref().unwrap();
+            let db_load_sizes = data_store.load_sizes.as_ref().unwrap();
+            let new_size = db_load_sizes.new.unwrap_or(0);
+
+            // If we have a client ID on disk, we check the database
+            if let Some(stored_client_id) = stored_client_id {
+                // state (c)
+                if new_size <= 0 {
```
chutten, line 340: Optional: This condition and the one in the state diagram are inverted.

glean-core/src/core/mod.rs
```diff
@@ -269,6 +276,155 @@ impl Glean {
...
+                    log::trace!("no database. database size={new_size}. stored_client_id={stored_client_id}");
+                    // state (d)
+                    glean
+                        .health_metrics
+                        .recovered_client_id
+                        .set_from_uuid_sync(&glean, stored_client_id);
```
chutten, line 346: The state diagram doesn't say that we report the file's client_id in state (d)

glean-core/src/core/mod.rs
chutten: Fair 'nuff.

glean-core/src/core/mod.rs
```diff
@@ -269,6 +276,155 @@ impl Glean {
...
+                                .health_metrics
+                                .exception_state
+                                .set_sync(&glean, ExceptionState::C0ffeeInDb);
+
+                            // If we have a recovered client ID we also overwrite the database.
+                            // state (e)
+                            glean
+                                .core_metrics
+                                .client_id
+                                .set_from_uuid_sync(&glean, stored_client_id);
```
chutten: Would this be considered a "mitigation" that we should avoid until later?

glean-core/tests/clientid_textfile.rs
```diff
@@ -0,0 +1,506 @@
...
+            .unwrap();
+        writer.commit().unwrap();
+    }
+
+    let (glean, temp) = new_glean(Some(temp));
+
+    // TODO(bug 1996862): We don't run the mitigation yet
+    _ = file_client_id;
+    //let db_client_id = clientid_metric().get_value(&glean, None).unwrap();
+    //assert_eq!(file_client_id, db_client_id);
```
chutten, line 231: Interestingly, I think the code _does_ run the mitigation.

glean-core/metrics.yaml
```diff
@@ -1158,3 +1158,91 @@ glean.health:
...
+    labels:
+      - parse
+      - permission-denied
+      - io
+
+  file_write_error:
+    type: labeled_counter
+    lifetime: ping
+    description: |
+      Count of different errors that happened when trying to write the `client_id.txt` file from disk.
```
chutten:
```suggestion
      Count of different errors that happened when trying to write the `client_id.txt` file to disk.
```

docs/dev/core/internal/index.md
badboy: woops. unintentional

glean-core/tests/clientid_textfile.rs
badboy: 🙈 -- thanks for noticing. I should make sure code and test agree, yes.

glean-core/src/core/mod.rs
badboy: So I guess _technically_?
But let's think through it:
1. We assume a client right now can contain a c0ffee ID
2. We also assume no `client_id.txt` exists.
3. Glean inits, writes the client ID (`c0ffee`) from the DB into the file.
4. Glean initializes core metrics. Detects `c0ffee` in the database. Generates a new UUID. Stores that into the database and the file.

If for whatever reason (4) doesn't happen, the file will also include the `c0ffee` ID.
So "restoring" means we don't change anything.
If (4) happens, both the database and the file should have the same, valid ID (except if it happens to crash right in between these two operations, which I guess ... at our scale could happen?). So maybe we should check that `stored_client_id` is not the c0ffee ID either? It's a tiny edge case, but I had enough with our edge cases 🙈

PR #3320, opened 2025-11-11
A first benchmark: Putting things on the dispatcher

badboy:
This sets up criterion, a statistics-driven micro-benchmarking tool. It also adds a first simple benchmark: Launching tasks on the global dispatcher.

---

Needs some docs I guess.
The way I've been running this:

On this commit:

```
cargo bench -p glean-core --bench dispatcher -- --save-baseline bn
```

Then on later commits (e.g. #3318)

```
cargo bench -p glean-core --bench dispatcher -- --baseline bn
```

results in output like this:

```
   Compiling glean-core v66.1.1 (/home/jer/src/mozilla/glean/glean-core)
    Finished `bench` profile [optimized] target(s) in 18.62s
     Running benches/dispatcher.rs (target/release/deps/dispatcher-b3ccfcc39ee552d0)
empty fn                time:   [61.181 ns 61.231 ns 61.285 ns]
                        change: [−23.764% −23.575% −23.390%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 13 outliers among 100 measurements (13.00%)
  5 (5.00%) low severe
  3 (3.00%) low mild
  2 (2.00%) high mild
  3 (3.00%) high severe

fn 1024                 time:   [431.57 ns 432.26 ns 432.99 ns]
                        change: [−44.413% −34.742% −23.066%] (p = 0.00 < 0.05)
                        Performance has improved.
Found 1 outliers among 100 measurements (1.00%)
  1 (1.00%) high mild
```

### DIFF
--- a/bin/benchmark-compare.sh
+++ b/bin/benchmark-compare.sh
@@ -1,0 +1,46 @@
+#!/usr/bin/env bash
+#
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# Run a benchmark comparison between a previous commit and the current one.
+
+set -eo pipefail
+
+run() {
+    [ "${VERB:-0}" != 0 ] && echo "+ $*"
+    "$@"
+}
+
+WORKSPACE_ROOT="$( cd "$(dirname "$0")/.." ; pwd -P )"
+
+if [ -z "$1" ]; then
+    echo "Usage: $(basename "$0") <start commit> [end commit]"
+    echo
+    echo "Run a benchmark on both the <start commit> and [end commit] for comparison."
+    echo "[end commit] defaults to HEAD"
+    exit 1
+fi
+
+START_COMMIT="$1"
+START_COMMIT=$(git rev-parse "$START_COMMIT")
+END_COMMIT="${2:-HEAD}"
+END_COMMIT=$(git rev-parse "$END_COMMIT")
+
+BASELINE_NAME="bn-${END_COMMIT}"
+
+stash_commit=$(git stash create)
+
+run git checkout "$START_COMMIT"
+run cargo bench -p glean-core --features benchmark -- --save-baseline "$BASELINE_NAME"
+
+run git checkout "$END_COMMIT"
+run cargo bench -p glean-core --features benchmark -- --baseline "$BASELINE_NAME"
+
+if [[ -n "$stash_commit" ]]; then
+  git stash apply "$stash_commit"
+fi
+
+run echo "You can rerun the benchmark against the baseline using:"
+run echo "cargo bench -p glean-core -- --baseline $BASELINE_NAME"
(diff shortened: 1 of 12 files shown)


glean-core/src/lib.rs
```diff
@@ -38,7 +38,8 @@ mod core_metrics;
 mod coverage;
 mod database;
 mod debug;
-mod dispatcher;
+#[doc(hidden)]
+pub mod dispatcher;
```
badboy, line 43: Benchmarks have the "downside" that they use the crate from extern, which means everything we want to test needs to be reachable. That requires me to expose the dispatcher (we don't want to make it public API, so we hide it from the docs, which is the "common" signal that, while public, is not considered public API)

supply-chain/config.toml
```diff
@@ -48,6 +48,14 @@ criteria = "safe-to-deploy"
 version = "1.3.2"
 criteria = "safe-to-deploy"
 
+[[exemptions.criterion]]
+version = "0.7.0"
```
badboy, line 52: We don't ship these benchmarks, so I decided to _not_ review criterion and its dependencies in detail right now.

bin/benchmark-compare.sh
```diff
@@ -0,0 +1,42 @@
...
+fi
+
+START_COMMIT="$1"
+START_COMMIT=$(git rev-parse "$START_COMMIT")
+END_COMMIT="${2:-HEAD}"
+END_COMMIT=$(git rev-parse "$END_COMMIT")
+
+BASELINE_NAME="bn-${END_COMMIT}"
+
+run git stash save "benchmark between ${START_COMMIT} and ${END_COMMIT}"
```
chutten: `git stash save` is deprecated in favour of `git stash push`

Also, you stash changes but you never unstash them.

glean-core/benches/dispatcher.rs
```diff
@@ -0,0 +1,111 @@
...
+
+    // Drain the dispatcher.
+    dispatcher::block_on_queue();
+
+    // Moving data into the function means we need to copy that data on to the queue too.
+    c.bench_function("fn 1024", |b| {
+        b.iter(|| {
+            let c = [0; 1024];
+            dispatcher::launch(move || {
+                black_box(c);
```
chutten, line 37: TIL

glean-core/benches/dispatcher.rs
```diff
@@ -0,0 +1,111 @@
...
+    fn cancel_uploads(&self) -> glean_core::Result<(), glean_core::CallbackError> {
+        Ok(())
+    }
+}
+
+pub fn metric_dispatcher_benchmark(c: &mut Criterion) {
+    let dir = tempfile::tempdir().unwrap();
+    let data_path = dir.path().display().to_string();
+
+    let cfg = glean_core::InternalConfiguration {
```
chutten, line 68: Oh no, do we now need an _internal_ configurationbuilder?

glean-core/src/lib.rs
chutten: Is there a pattern to set a, I dunno, "benching" feature and making certain internals external only for that build configuration?

chutten (approved):
What do the perf folks think?

glean-core/benches/dispatcher.rs
badboy: Well, technically this is also not public API, so other than our wrappers no one should use this. Just for the benchmarking purposes we reach to the lowest level.

Other benchmarks might as well then run on the RLB and use the public APIs.

PR #3326, opened 2025-11-15
Bug 1997963 - Config performance optimizations for glean-gradle-plugin

ahochheiden:
- Lazy task registration: Changed project.task() to project.tasks.register() to defer task configuration until needed
- Deferred disk I/O: Replaced configuration-time file existence checks with execution-time checks using onlyIf and .filter { it.exists() }
- Optional inputs: Used inputs.files().optional(true) instead of conditionally adding inputs based on file existence
- Configuration avoidance: Replaced .all() with .configureEach() for variant iteration
- Lazy task dependencies: Used taskProvider.configure { dependsOn() } to defer dependency setup


badboy:
I re-kicked the CI. Seems that was an intermittent. Not sure why it happened on both Android runs on different platforms

PR #3343, opened 2025-12-04
Bug 2004001 - Disable `post_init` health ping

badboy:


travis79:
oof, lots of tests will need "fixed" to rely on the new number of "startup" pings. (Sorry about that...)

glean-core/rlb/tests/health_ping_file_overwrite.rs
```diff
@@ -47,7 +47,7 @@ impl net::PingUploader for ReportingUploader {
 
 /// Test scenario: Write a client ID to the backup file and check that it's used after initialization.
 #[test]
-fn test_pre_post_init_health_pings_exist() {
+fn test_pre_init_health_pings_exist() {
```
travis79: just a nit, but maybe the name of this function should be something like the following, since there's only one ping to test now:

```suggestion
fn test_pre_init_health_ping_exists() {
```

glean-core/pings.yaml
```diff
@@ -108,8 +108,7 @@ health:
     The purpose of the health ping is to transport all of the health
     metric information. The `health` ping is automatically sent when the
     application calls Glean initialize before any operations are done on
-    the data path with a reason of `pre_init`, and once after Glean
-    initialization with a reason of `post_init`.
+    the data path with a reason of `pre_init`.
```
travis79, line 111: Let's leave a breadcrumb here in the description stating that `post_init` used to exist but has been removed, or something similar. That way it shows up in the Glean Dictionary and maybe someone looking at the data would find it and not be surprised if they encounter the old reason in the data.

PR #3347, opened 2025-12-04
Bug 2004094 - `PingUploadScheduler` does not correctly end background tasks

ih-codes:
[Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=2004094)

- Fix strict concurrency warnings in `Dispatchers` by making the type `Sendable`.
   - You only see these warnings on Swift 6 or on Swift 5 with the `SWIFT_STRICT_CONCURRENCY = complete`.
   - There are a number of other such warnings in the project but I didn't want to start changing a bunch of other files on this PR
- Fix bug in `PingUploadScheduler` for breaking out of while-switch and continuing execution to the background task handling.
- Make `PingUploadScheduler` more testable with dependency injection via protocols.
- Add unit tests for `PingUploadScheduler`. 
- Add mock types for testing.

I wanted to be a good citizen and test this thoroughly since I'm not sure how to check this against my local iOS build, but if you are not happy with all the dependency injection changes that let me use mocks, I can back out those changes.


glean-core/ios/Glean/Dispatchers.swift
```diff
@@ -5,9 +5,9 @@
 import Foundation
 
 /// This class manages a single background operation queue.
-class Dispatchers {
+public final class Dispatchers: Sendable {
     /// This is the shared singleton access to the Glean Dispatchers
-    static let shared = Dispatchers()
+    public static  let shared = Dispatchers()
```
lmarceau: nit; extra space
```suggestion
    public static let shared = Dispatchers()
```

PR #3357, opened 2025-12-15
Bug 2005557 - CI: Deploy pages using GhA

badboy:
It only runs on pushes to the `main` branch.
With this we can drop the `gh-pages` branch which bloats the entire repository.

This does the minimal thing:

It builds Rust docs, Python docs & the books.
It then deploys them.

On GitHub Actions this does NOT do:

1. Spellcheck
2. Linkcheck
3. Swift docs

(1) and (2) are still done on CI as part of the usual CI from branches/PR runs. We do not do them on merges to `main` though.

Swift docs (3) are a bit harder, as it requires Xcode and thus requires macOS. That's a bit more expensive in terms of resources. We could bring it back on GitHub Actions too if needed. But it's not fully clear how much value those rendered docs bring.

---

I have tested this over on my own fork: https://github.com/badboy/glean/actions/runs/20164355091

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -18,6 +18,11 @@ definitions:
   - ci_filters: &ci-filters
       branches:
         ignore: release
+  - ci_only: &ci-only
+      branches:
+        ignore:
+          - main
+          - release
 
 ##########################################################################
 # COMMANDS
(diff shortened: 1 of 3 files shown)


badboy:
In order to land this we will need to change to "Deploy with GitHub Actions" in https://github.com/mozilla/glean/settings/pages.
The environment already exists and isn't restricted. We can restrict it to the `main` branch later.
When that's all done we can delete the `gh-pages` branch. Over time that should get rid of any dangling commits and the GC should sweep up all that unneeded data, making the overall repo smaller.

travis79 (approved):

PR #3419, opened 2026-03-12
Bug 2017277 - [RLB] Expose client_info.os_version

badboy:
Note this by itself doesn't even change the value! We still fall back to what `whatsys` gives us.


chutten (approved):
Does it require a doc update to https://mozilla.github.io/glean/book/reference/general/initializing.html#gleaninitializeconfiguration ?

chutten (approved):
Oh, I guess it doesn't _yet_. This is still behind RLB's closed doors.

travis79 (approved):

PR #3426, opened 2026-03-30
Bug 2018944 - glean-sym - a Rust API built on top of the Glean UniFFI C FFI

badboy:
This small patch on a-s makes use of the new glean-sys crate: https://github.com/mozilla/application-services/compare/main...badboy:application-services:glean-dylib

Todo:

* [x] ~~Naming! Not sure I like `glean-sys`.~~ it was decided: glean-sym (as in glean-symbols)
* [x] Testing. Not clear how we will do that. Probably some sample crate that uses `glean-sys` and we inject a `libxul.so` into it at test time
* [x] `glean-sys/src/metrics.rs` is generated. Currently by a test. That has a bit of a problem that you need to compile the code to generate the code, which breaks if the code doesn't compile. Might want to have a non-test way to generate that code.
* [ ] `libxul.so` is hard-coded. Ok for manual testing in Fenix, not sure we want to ship that as-is
* [ ] No way to ensure we're not the first ones loading `libxul`. 
* [ ] Should we have the UniFFI-alike runtime checks? (like the version check)
* [x] Extend glean_parser (and `glean-build`) to generate the correct code usable with `glean-sys`
	* https://github.com/mozilla/glean_parser/pull/841
* [x] Dependency vetting
* [x] Rust minimum version bump probably necessary (or we need to exclude glean-sys)


glean-core/glean-sym/src/lib.rs
```diff
@@ -0,0 +1,95 @@
...
+            #[cfg(unix)]
+            pub fn load() -> std::io::Result<Self> {
+                #[cfg(any(target_os = "ios", target_os = "macos"))]
+                let name = "libxul.dylib";
+                #[cfg(target_os = "windows")]
+                let name = "libxul.dll";
+                #[cfg(not(any(target_os = "ios", target_os = "macos", target_os = "windows")))]
+                let name = "libxul.so";
+
+                let library = match unsafe { libloading::Library::new(name) } {
```
badboy, line 61: fwiw, I tried it with `libloading::os::unix::Library::this()`, but then it couldn't find the symbols, even though I'm certain `libxul` was already loaded. This might need some testing to ensure we're not the first ones loading libxul and that loading it again doesn't have any other negative impact.

badboy:
I would like to land this (experimentally) sometime soon. The plan is to test it out with a metric in appservices, see https://bugzilla.mozilla.org/show_bug.cgi?id=2034534
PoC: https://github.com/badboy/application-services/commit/2c0f79ac4e64f0685ac6986944814fb00b90719a

Given this is an experiment and not a released crate we can tackle the open questions as we go along, instead of landing the perfectly finished PR.

glean-core/glean-sym/src/metrics.rs
```diff
@@ -0,0 +1,2007 @@
...
+impl CounterMetric {
+    unsafe fn clone_handle(&self) -> u64 {
+        unsafe {
+            let mut call_status = uniffi::RustCallStatus::default();
+            std::mem::transmute(
+                (crate::GLEAN
+                    .uniffi_glean_core_fn_clone_countermetric)(
+                    std::mem::transmute(self.handle),
+                    &mut call_status,
+                ),
```
travis79: We call transmute twice in each of the `clone_handle` methods in this file, is that needed? Won't the underlying type be `u64` on both sides?

tools/glean-sym-parser/src/type_resolver.rs
```diff
@@ -0,0 +1,282 @@
...
+impl TypeResolver for weedle::types::DoubleType {
+    fn resolve(&self) -> TokenStream {
+        if self.unrestricted.is_some() {
+            panic!("we don't support `unrestricted double`");
+        }
+        quote! { f64 }
+    }
+
+    fn resolve_ffi(&self) -> TokenStream {
+        quote! { bool }
```
travis79: Should this be `bool` here? That doesn't seem correct

tools/glean-sym-parser/src/type_resolver.rs
```diff
@@ -0,0 +1,282 @@
...
+            None => {
+                quote! { uniffi::RustBuffer }
+            }
+        }
+    }
+}
+
+impl TypeResolver for weedle::term::Boolean {
+    fn resolve(&self) -> TokenStream {
+        quote! { i8 }
```
travis79: Does this mean consumers will need to cast this to a Rust `bool`?

glean-core/glean-sym/src/types.rs
```diff
@@ -0,0 +1,119 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
```
travis79, line 1: I take it this is pretty much duplication of types from glean-core? And, if we make changes there we need to make changes here too?

Be sure and document anything like that, preferably in a comment close to where these live so we don't forget to update things in both places.

glean-core/glean-sym/src/util.rs
```diff
@@ -0,0 +1,141 @@
...
+        }
+        )*
+    }
+}
+
+impl_clone_ffi_arg_primitive!(i8, i32, i64);
+
+impl CloneFfiArg<RustBuffer> for RustBuffer {
+    fn clone_for_ffi(&self) -> RustBuffer {
+        // SAFETY: trust me, bro.
```
travis79: 🙈

glean-core/glean-sym/tests/parser.rs
```diff
@@ -0,0 +1,25 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fs;
+
+use xshell::{Shell, cmd};
+
+#[test]
+fn generated_metrics_code_up_to_date() {
```
travis79, line 18: Is this test what keeps things in sync? I see you writing to a committed file and then checking for modifications, and that's what I assume. Could you add some documentation comment(s) on what this does?

supply-chain/config.toml
```diff
@@ -152,6 +148,10 @@ criteria = "safe-to-run"
 version = "0.3.7"
 criteria = "safe-to-run"
 
+[[exemptions.prettyplease]]
```
travis79, line 151: Just out of curiosity, should we document these exemptions anywhere?

.circleci/config.yml
```diff
@@ -122,6 +122,9 @@ commands:
           name: Run glean-sym sample
           command: |
             cd samples/glean-sym-test
+            python3 -m venv ${PWD}/venv
+            venv/bin/pip install "git+ssh://git@github.com/mozilla/glean_parser@push-lskqopuplvts#egg=glean-parser"
```
travis79: Might be worth filing a bug to update this once it's no longer experimental? I expect this will eventually be pinned to a proper version

glean-core/glean-sym/Cargo.toml
```diff
@@ -0,0 +1,19 @@
...
+name = "glean-sym"
+version = "0.1.0"
+authors = ["The Glean Team <glean-team@mozilla.com>"]
+description = "A Rust API on top of the Glean UniFFI C FFI"
+repository = "https://github.com/mozilla/glean"
+license = "MPL-2.0"
+edition = "2024"
+
+[dependencies]
+bytes = "1.11.1"
```
travis79: I don't see this being `use`ed anywhere

travis79 (commented):
Okay, a decent first pass with several questions. Nothing really big aside from the comments in the metrics.rs about the mismatched Double type being `bool`

glean-core/glean-sym/src/metrics.rs
badboy: you are right. this is a left-over from a previous iteration where they were (transparent) wrapper types and needed the explicit transmute.

tools/glean-sym-parser/src/type_resolver.rs
badboy: mistake. should be bool.

glean-core/glean-sym/src/types.rs
badboy: it is at the moment, will add notes.

glean-core/glean-sym/src/util.rs
badboy: I swear I mean well.

glean-core/glean-sym/tests/parser.rs
badboy: That's correct! It automatically! The generated metrics file contains info about that, but I'll copy it to the test.

.circleci/config.yml
badboy: oh yeah, that will be dropped before we merge this.

travis79 (approved):
Nothing stands out on my second pass through this, let's do it!

PR #3434, opened 2026-04-14
Bug 2031446 - Retrieve the OS version on Android.

afranchuk:
It is surprising difficult to find the API you're supposed to use for this nowadays (aside from invoking the `getprop` program). I tested that this compiles, but I did not have an environment set up to run it to see if it works as expected.

I will do some shenanigans to check whether this works.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v67.2.0...main)
 
+* Rust
+  * Add support for determining `client_info.os_version` for Android binaries ([#3434](https://github.com/mozilla/glean/pull/3434))
+
 # v67.2.0 (2026-04-06)
 
 [Full changelog](https://github.com/mozilla/glean/compare/v67.1.0...v67.2.0)
--- a/glean-core/rlb/src/system.rs
+++ b/glean-core/rlb/src/system.rs
@@ -86,7 +86,44 @@ pub fn get_os_version() -> String {
     whatsys::kernel_version().unwrap_or_else(|| "Unknown".to_owned())
 }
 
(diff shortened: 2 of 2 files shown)


badboy:
I put that code into a binary, then build it for Android and ran it:

```
[~/projects/rust/android-os-version]  (git:main#%)
❯ cargo ndk build --release --target arm64-v8a
    Building arm64-v8a (aarch64-linux-android)
    Finished `release` profile [optimized] target(s) in 0.00s

[~/projects/rust/android-os-version]  (git:main#%)
❯ adb push target/aarch64-linux-android/release/android-os-version /data/local/tmp
target/aarch64-linux-android/release/android-os-vers...shed, 0 skipped. 447.8 MB/s (463728 bytes in 0.001s)

[~/projects/rust/android-os-version]  (git:main#%)
❯ adb shell /data/local/tmp/android-os-version
14
```

looks like it does what it should.

badboy:
guess that's also the method mozglue uses: https://searchfox.org/firefox-main/rev/8332a06d47ce7d66623d807068b3410061cd29d3/mozglue/linker/ElfLoader.cpp#33-48

badboy (approved):
Not sure why the Android tests initially failed. Re-run the one on CircleCI and it passed now (I can't re-run it on TC). I assume some intermittent issue unrelated to the change (because that would be weird anyway). we probably should call this out in the changelog though.

afranchuk:
Thanks for verifying it. I tried to do so, but I didn't want to install cargo ndk, and tried to build it on my own against the mozbuild sysroot. Annoyances ensued, lesson learned. I've added a changelog entry.

badboy:
Rebased and adjusted the changelog; ready to land.

PR #3437, opened 2026-04-16
glean-core: fix 'occured' -> 'occurred' in error display strings

SAY-5:
Two `fmt::Display` implementations in `glean-core/src/error.rs` (lines 204-205) read `IO error occured` and `Parse error occured`. Fixed to `occurred`. Error-display-only change; no API or behavior change.


badboy (approved):
Thank you!

PR #3438, opened 2026-04-16
Bug2020962 adding session metadata and sampling functionality

travis79:
I attempted to structure the commits here to make this easier to review, so I recommend reviewing it that way rather than as a whole.


glean-core/benchmark/benches/dispatcher.rs
```diff
@@ -83,6 +83,9 @@ pub fn metric_dispatcher_benchmark(c: &mut Criterion) {
         ping_schedule: Default::default(),
```
badboy, line 83: Adding a note to the very first line of the diff for the first commit, because one still can't leave a comment for the whole commit (or the message).

Thanks for splitting it up into individual commits. That's certainly helpful, though unfortunately this commit is not working on its own, introducing use of the `SessionMode` before it's defined.
I think I can cope with it in review though, just comments might be a bit out of order then.

glean-core/benchmark/benches/dispatcher.rs
```diff
@@ -83,6 +83,9 @@ pub fn metric_dispatcher_benchmark(c: &mut Criterion) {
...
         ping_lifetime_threshold: 0,
         ping_lifetime_max_time: 0,
+        session_mode: glean_core::SessionMode::Auto,
+        session_sample_rate: 1.0,
+        session_inactivity_timeout_ms: 1_800_000,
```
badboy, line 90: gnah, I really wish we could make this a `Duration`. But eh, it's the UDL-exposed type, we stick with integers for now.

glean-core/src/common_metric_data.rs
```diff
@@ -75,6 +88,34 @@ pub struct CommonMetricData {
...
+    ///
+    /// `true` (the default) means this metric bypasses session sampling and
+    /// does not carry session metadata. Non-event metrics should always use
+    /// the default. Event metrics that participate in session tracking must
+    /// explicitly set this to `false`.
+    ///
+    /// Note: the naming is an inversion of the spec's `in_session` field.
+    /// `out_of_session: false` ↔ `in_session: true`.
+    #[serde(default = "default_out_of_session")]
+    pub out_of_session: bool,
```
badboy: Why the inversion?
The spec says `in_session` and that's also what ends up in the metrics.yaml in the end.
It seems confusing to me to use the inverted one here then.

> Event metrics that participate in session tracking must explicitly set this to `false`

Setting something to `false` to _participate_/opt-in seems also the wrong way around in my mind.

glean-core/src/common_metric_data.rs
```diff
@@ -75,6 +88,34 @@ pub struct CommonMetricData {
...
+            name: Default::default(),
+            category: Default::default(),
+            send_in_pings: Default::default(),
+            lifetime: Default::default(),
+            disabled: Default::default(),
+            dynamic_label: Default::default(),
+            // Non-event metrics are out-of-session by default per the spec:
+            // "By default, all non-event metrics will be in_session: false."
+            // Event metrics must explicitly set `out_of_session: false`.
+            out_of_session: true,
```
badboy: Using `in_session` instead would also make Default derivable again because the default for `bool` is `false`

glean-core/src/internal_metrics.rs
```diff
@@ -47,6 +47,9 @@ pub struct AdditionalMetrics {
 
     /// Server knobs configuration received from remote settings.
     pub server_knobs_config: ObjectMetric,
+
+    /// The total number of sessions started, regardless of sampling outcome.
```
badboy: started per what? (looks like this is ping lifetime on the health ping, so ... sessions started between two full starts of the application [I think that's the only time we sent the health ping? though the health ping's lifecycle is not as strict, we could at any point change it, making this counter even more ambiguous]

glean-core/src/internal_metrics.rs
```diff
@@ -202,17 +192,24 @@ impl AdditionalMetrics {
...
-                dynamic_label: None,
+                ..Default::default()
+            }),
+
+            sessions_seen: CounterMetric::new(CommonMetricData {
+                name: "sessions_seen".into(),
+                category: "glean".into(),
+                send_in_pings: vec!["health".into()],
+                lifetime: Lifetime::Ping,
+                out_of_session: true,
```
badboy: Why defined on a counter metric? It doesn't have an effect on non-event metrics.
And also isn't that the default?

glean-core/tests/ping.rs
```diff
@@ -137,6 +137,7 @@ fn test_pings_submitted_metric() {
                 lifetime: Lifetime::Ping,
                 disabled: false,
                 dynamic_label: None,
+                out_of_session: true,
```
badboy: Should this be rewritten to use `..Default`?

glean-core/metrics.yaml
```diff
@@ -1068,6 +1068,100 @@ glean:
...
+          - "replaced": a new session was started while one was already active
+          - "timeout": AUTO mode inactivity timeout expired
+          - "abnormal": session ended due to app crash/force-close
+          - "abnormal_inactive": session was inactive at time of crash
+        type: string
+
+  sessions_seen:
+    type: counter
+    description: |
+      The total number of sessions started during this period.
```
badboy: "this period" is a bit unclear here.

glean-core/metrics.yaml
```diff
@@ -1068,6 +1068,100 @@ glean:
...
+      Incremented for every session start, regardless of sampling.
+      This metric is marked out-of-session so it is never suppressed by
```
badboy: 🤔 once again I thought non-event metrics don't adhere to any out-of-session/in_session field

(and regardless, if we use some specific value here that should be in the definition)

glean-core/src/lib.rs
```diff
@@ -1196,6 +1212,33 @@ pub fn glean_handle_client_inactive() {
...
+}
+
+/// Ends a session manually.
+///
+/// Only has effect in `SessionMode::Manual`. Calling this in `Auto` or
+/// `Lifecycle` mode is a no-op to prevent corrupting automatic session state.
+///
+/// `reason` is an optional application-provided string attached to the
+/// `glean.session_end` boundary event for downstream analysis.
+pub fn glean_session_end(reason: Option<String>) {
```
badboy, line 1238: I assume these are the ones listed in the definition? Or can they be whatever the application dictates?
Do we validate them?

glean-core/src/ping/mod.rs
```diff
@@ -176,6 +176,12 @@ impl PingMaker {
                 .insert("server_knobs_config".to_string(), server_knobs_config);
         }
 
+        // Session metadata lives on individual event records (RecordedEvent.session),
+        // not in ping_info. The ping_info schema has additionalProperties: false, so
+        // adding a session field here would break schema validation. Per-event session
+        // metadata is the authoritative source per the spec; ping-level session fields
+        // would require a schema change coordinated across the Glean ecosystem.
+
```
badboy: Adding this comment here seems a bit out of place.

glean-core/src/metrics/mod.rs
```diff
@@ -215,6 +215,21 @@ pub trait MetricType {
...
+        // This check applies to ALL metric types, not just events.  Out-of-session
+        // metrics (`out_of_session = true`) bypass this gate entirely and always record.
+        // In-session metrics (`out_of_session = false`) are suppressed here when the
+        // active session is sampled out.
+        //
+        // EventMetric additionally uses `compute_event_context()` after this check to
+        // determine which session metadata to attach — that function is purely about
+        // metadata, not suppression, and it may be called from other metric types in
+        // future if they also need per-event session context.
+        if !self.meta().inner.out_of_session && !glean.session_manager().is_sampled_in() {
```
badboy: We should expose a getter on `CommonMetricDataInternal` instead of reaching to `inner`.
And if we flip that, then it becomes `if self.meta().in_session()`, which IMO even reads better

glean-core/src/metrics/mod.rs
```diff
@@ -257,8 +272,11 @@ pub trait MetricType {
         let new_disabled = (remote_settings_epoch << 4) | (current_disabled & 0xF);
         self.meta().disabled.store(new_disabled, Ordering::Relaxed);
 
-        // Return a boolean indicating whether or not the metric should be recorded
-        current_disabled == 0
+        if current_disabled != 0 {
+            return false;
+        }
+
+        true
```
badboy: Why was this changed?

glean-core/src/metrics/event.rs
```diff
@@ -157,9 +158,21 @@ impl EventMetric {
             map.insert("glean_timestamp".to_string(), precise_timestamp.to_string());
         }
 
+        // Determine the session context for this event.
+        //
+        // Sampling suppression has already been handled by `should_record()` above,
+        // which gates all metric types uniformly.  This call is concerned only with
+        // what metadata to attach: `OutOfSession` for out-of-session metrics and
+        // between-session events, `InSession(meta)` for events in a sampled-in session.
+        let ctx = if self.meta().inner.out_of_session {
```
badboy: With `in_session` this `if` flips around

glean-core/src/core/mod.rs
```diff
@@ -1300,11 +1346,406 @@ impl Glean {
...
+    // -----------------------------------------------------------------------
+    // Client lifecycle methods
+    // -----------------------------------------------------------------------
+
     /// Performs the collection/cleanup operations required by becoming active.
     ///
     /// This functions generates a baseline ping with reason `active`
     /// and then sets the dirty bit.
     pub fn handle_client_active(&mut self) {
+        // Session lifecycle (mode-dependent).
```
badboy: Not sure this comment is needed.

glean-core/src/core/mod.rs
```diff
@@ -1300,11 +1346,406 @@ impl Glean {
...
+        );
+
+        // Emit synthetic session_end.
+        self.record_session_end_event(&persisted_id, persisted_seq, Some(reason));
+
+        // Clear persisted session state so the recovered session won't be replayed.
+        session::clear_session_id(self);
+        session::clear_inactive_since(self);
+        session::clear_session_start_time(self);
+        session::clear_session_event_seq(self);
```
badboy: Looking at the full change, it seems these 4 functions are always called together.
Would it make sense to bundle them into `session::clear(self)` instead?

glean-core/src/core/mod.rs
```diff
@@ -1300,11 +1346,406 @@ impl Glean {
...
+            return None;
+        }
+
+        let session_id = self.session_manager.session_id?;
+        let seq = self.session_manager.session_seq;
+        let event_seq = self.session_manager.event_seq.load(Ordering::Relaxed);
+        let sample_rate = self.session_manager.sample_rate;
+        let start_time = self.session_manager.session_start_time;
+
+        // Update in-memory state.
```
badboy: Below we change persistence first, then in-memory state.
Here we do in-memory first, then persistence.
I feel like keeping the same order between these would be good. It shouldn't make a difference for the logic, but reading this it's nice to see the same pattern.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// A session is currently active.
+    Active,
+}
+
+/// Session metadata attached to each in-session event.
+///
+/// Serialized into the event payload for downstream session-level analysis.
+///
+/// `PartialEq` is derived (using `f64::eq` for `session_sample_rate`).
+/// `Eq` is implemented manually — it is sound because `session_sample_rate`
```
badboy, line 65: Teeeeechnically `session_sample_rate` is `pub`, and thouse could be set to an arbitrary value by any one that holds a `SessionMetadata`, thus making the `Eq` wrong.
However `Eq` is not an unsafe trait, a _wrong_ implementation of `PartialEq` / `Eq` will not make code unsafe, just buggy.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    /// Returns `OutOfSession` when no session is active (between sessions), or
+    /// `InSession(meta)` when an active session is present.
+    ///
+    /// `event_seq` is incremented only for `InSession` results so that
+    /// between-session events do not consume sequence numbers.
+    pub fn compute_event_context(&self) -> EventSessionContext {
+        match self.state {
+            SessionState::Inactive => EventSessionContext::OutOfSession,
+            SessionState::Active => {
+                // should_record() has already ensured sampled_in is true.
```
badboy: We could `debug_assert!` this here for a real check.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+// ---------------------------------------------------------------------------
+
+/// Converts a UUID to a deterministic sample value in [0, 1).
+///
+/// Interprets the first 8 bytes as a big-endian u64 and divides by 2^64.
+/// A session is sampled in when `uuid_to_sample_value(uuid) < sample_rate`.
+///
+/// Dividing by 2^64 (not u64::MAX) guarantees the result is strictly less than
+/// 1.0 for all inputs, so `sample_rate = 1.0` always samples every session.
+pub(crate) fn uuid_to_sample_value(uuid: &Uuid) -> f64 {
```
badboy, line 298: I wondered if this is biased. Some bits in a v4 UUID are fixed and we're only looking at half of them.
The variant seems to be in the 9th byte, the version are 4 bits of the 7th byte. The rest is random. Guess it's fine though? If not we would probably need to switch to a proper hash of it before dividing.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    arr.copy_from_slice(&bytes[..8]);
+    let n = u64::from_be_bytes(arr);
+    (n as f64) / 2.0f64.powi(64)
+}
+
+// ---------------------------------------------------------------------------
+// Persistence helpers
+// ---------------------------------------------------------------------------
+
+fn make_session_seq_metric() -> QuantityMetric {
```
badboy, line 313: For AFTER the initial PR: We should consider refactoring this to avoid the constant re-creation of these metrics. It's a lot of allocation if they get called more often.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+        .filter(|&v| v >= 0)
+        .map(|v| v as u64)
+        .unwrap_or(0)
+}
+
+/// Persists the given session sequence number.
+///
+/// `QuantityMetric` stores `i64`; the cast from `u64` is lossless for any
+/// value below `i64::MAX` (~9.2 × 10^18).  Values at or above that threshold
+/// (unreachable in practice) would silently truncate, which is preferable to
```
badboy: Actually it would wrap, not truncate.

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+/// Clears the persisted session ID.
+pub(crate) fn clear_session_id(glean: &Glean) {
+    make_session_id_metric().set_sync(glean, "");
+}
+
+/// Reads the persisted session ID, if any.
+/// Returns `None` if no session ID is stored or if it was cleared.
+pub(crate) fn read_session_id(glean: &Glean) -> Option<String> {
+    let id = make_session_id_metric().get_value(glean, INTERNAL_STORAGE)?;
+    if id.is_empty() {
```
badboy, line 388: Under which circumstances would it be the empty string?

glean-core/src/session/mod.rs
```diff
@@ -0,0 +1,464 @@
...
+    let s = make_session_start_time_metric().get_value(glean, INTERNAL_STORAGE)?;
+    if s.is_empty() {
+        return None;
+    }
+    DateTime::parse_from_rfc3339(&s).ok()
+}
+
+/// Clears the persisted session start timestamp.
+pub(crate) fn clear_session_start_time(glean: &Glean) {
+    make_session_start_time_metric().set_sync(glean, "");
```
badboy, line 452: oh. got it.

glean-core/python/tests/test_glean.py
```diff
@@ -844,8 +847,8 @@ def test_client_activity_api(tmpdir, monkeypatch, helpers):
     url_path, payload = helpers.wait_for_ping(info_path)
     assert "baseline" == url_path.split("/")[3]
     assert payload["ping_info"]["reason"] == "active"
-    # It's an empty ping.
-    assert "metrics" not in payload
+    # No duration is recorded on an active ping.
+    assert "timespan" not in payload.get("metrics", {})
```
badboy: Why these changes?

glean-core/src/lib_unit_tests.rs
```diff
@@ -1360,3 +1363,161 @@ fn pings_are_controllable_from_remote_settings_config() {
...
+    glean2.recover_session_on_dirty_flag();
+
+    let events = session_end_metric_internal()
+        .get_value(&glean2, "events")
+        .expect("expected synthetic session_end after crash recovery");
+    assert_eq!(1, events.len());
+    let extra = events[0].extra.as_ref().expect("expected extras on session_end");
+    assert_eq!(
+        "abnormal",
+        extra.get("reason").expect("reason missing from synthetic session_end"),
```
badboy: You should be able to simplify this into a single assertion like:

```suggestion
        Some("abnormal"),
        extra.get("reason"),
```

Saves us one long line and expect, gives the same good error message if it's not right.

badboy: Ok, I tried and that won't work as-is because `` expected `Option<&str>`, found `Option<&String>` ``
Maybe we keep that expect

glean-core/src/lib_unit_tests.rs
```diff
@@ -1360,3 +1363,161 @@ fn pings_are_controllable_from_remote_settings_config() {
...
+        "crashed active session must produce reason='abnormal'"
+    );
+    assert!(
+        extra.contains_key("session_id"),
+        "synthetic session_end must carry the crashed session's session_id"
+    );
+
+    // After recovery, session state must be fully cleared.
```
badboy, line 1430: Heh, I tend to also add these comments, but to be honest, the assert message has the same info. We could remove the comment (up to you though)

badboy (commented):
I gave this a first somewhat-thorough review and left lots of comments.

High-level I think this is a good approach, some clarifications in the details needed maybe.
I have not yet gone through all the new test; I'd like to do that before I'm out tomorrow.

What's currently missing though is docs, both dev docs about all the new state we're storing (and the state changes! oh is it time for a new diagram?), and user-facing docs.
Fine if they land in followups, but should be there before we get this out to applications.

glean-core/src/common_metric_data.rs
travis79: The inversion has been un-inverted. Those responsible for the inversion have been sacked (monty python reference...). 

This change request touches quite a few things, so I'm going to push it as a single commit before addressing the other comments.

glean-core/metrics.yaml
travis79: I cleaned up the wording of this comment to make it more clear, that yes this counter is out-of-session (as counters would be, by default) and why that is important as it relates to session analysis.

glean-core/src/internal_metrics.rs
travis79: Updated the description to better clarify the use of this and why it explicitly uses the default here.

travis79:
Okay, Part 1 commit of addressing change requests should cover just the refactoring to remove the inversion by changing `out_of_session` to `in_session`. Moving on to the rest of the comments now.

badboy:
Oh can you update the PR title to be a bit more descriptive?

glean-core/src/metrics/event.rs
```diff
@@ -164,7 +164,7 @@ impl EventMetric {
         // which gates all metric types uniformly.  This call is concerned only with
         // what metadata to attach: `OutOfSession` for out-of-session metrics and
         // between-session events, `InSession(meta)` for events in a sampled-in session.
-        let ctx = if self.meta().inner.out_of_session {
+        let ctx = if !self.meta().inner.in_session {
```
badboy: If you swap around the branches here you don't need the negation (IMO that often gets lost skimming the code, so doing the true case first is easier)

glean-core/src/internal_metrics.rs
travis79: Yeah, I see what you are saying. I reworded this and moved this to the metrics ping, since it is unlikely to change scheduling and is probably a better dataset for this anyway.

glean-core/metrics.yaml
travis79: Reworded this to be more clear, along with moving it to the metrics ping

glean-core/src/lib.rs
travis79: Hmm not user-provided, Glean sets this depending on the particular session state transition. I'll clarify this

travis79: Well, they are user provided in total `MANUAL` mode. Otherwise if in AUTO or LIFECYCLE modes, Glean would set one of the ones from the definition

glean-core/src/metrics/mod.rs
travis79: Should not have been changed, reverting this.

glean-core/src/session/mod.rs
travis79: I think it should be fine, and yes there are a handful of bytes that would reduce the randomness of this some. I'll verify the randomness during validation of sampling, and make changes to it once we can measure it, if needed.

glean-core/src/session/mod.rs
travis79: It gets cleared to an empty string in the persistence helper `clear_id`.

glean-core/python/tests/test_glean.py
travis79: Another mistake, reverting this

glean-core/src/lib.rs
travis79: Currently no validation on user supplied strings but I don't think that we will see any `MANUAL` mode users, especially not to begin with. I'd be happy to file a follow up bug to explore what that validation should be.

glean-core/src/session/mod.rs
travis79: Let's leave this comment unresolved as a reminder for me to file a bug for this.

glean-core/src/core/mod.rs
```diff
@@ -1137,6 +1165,25 @@ impl Glean {
...
+            //     `session_sample_rate` to `null` in the RS payload.  The next
+            //     session will use `configured_sample_rate` as the fallback.
+            //
+            // This override is intentionally NOT persisted to storage.  Remote
+            // Settings configuration is refreshed on every app startup, so the
+            // override will be re-applied before the next session begins.
+            // Persisting it would risk making a stale value sticky if the RS
+            // payload changes or is removed between restarts.
+            remote_settings_config.session_sample_rate =
+                cfg.session_sample_rate.map(|r| r.clamp(0.0, 1.0));
```
jeddai: Would it be at all beneficial to report an error on the clamp needing to be applied? Like 
```rust
if r < 0.0 || r > 1.0 { 
  // report error
  r.clamp(0.0, 1.0)
} else { r }
```

jeddai (commented):
I still have some of the larger files to review but I got through most of them

glean-core/src/core/mod.rs
travis79: We definitely can! Although there's probably not much we can do about it as I would expect validation of the sampling rate to be done where ever the sampling rate is being set (Experimenter or wherever that may be). To me that seems to be the best place to inform someone of the mistake/error.

I'm happy to at least log it for now, and perhaps add a counter to the health ping or something later, if that works for you?

CHANGELOG.md
```diff
@@ -2,6 +2,8 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v67.2.0...main)
 
+* General
```
badboy, line 5: this will need a proper rebase before landing.

glean.1.schema.json
```diff
@@ -111,6 +111,43 @@
           "timestamp": {
             "minimum": 0,
             "type": "integer"
+          },
```
badboy: Should rerun `bin/update-schema.sh HEAD` after https://github.com/mozilla-services/mozilla-pipeline-schemas/pull/869 is merged

glean-core/python/glean/net/ping_upload_worker.py
```diff
@@ -122,6 +128,9 @@ def _process(data_dir: Path, application_id: str, configuration) -> bool:
             ping_schedule={},
             ping_lifetime_threshold=0,
             ping_lifetime_max_time=0,
+            session_mode=SessionMode.AUTO,
+            session_sample_rate=1.0,
+            session_inactivity_timeout_ms=1800000,
```
badboy: `1_800_000` works in python too.

glean-core/rlb/src/configuration.rs
```diff
@@ -141,6 +154,9 @@ impl Builder {
             ping_schedule: HashMap::new(),
             ping_lifetime_threshold: 0,
             ping_lifetime_max_time: Duration::ZERO,
+            session_mode: SessionMode::Auto,
+            session_sample_rate: 1.0,
+            session_inactivity_timeout: Duration::from_secs(30 * 60),
```
badboy, line 159: Duration has `from_mins` which ... [DAMN, stable since 1.91.0](https://doc.rust-lang.org/std/time/struct.Duration.html#method.from_mins) and we use 1.90.0 as MSRV

glean-core/src/core/mod.rs
```diff
@@ -186,6 +190,8 @@ pub struct Glean {
     pub(crate) remote_settings_config: Arc<Mutex<RemoteSettingsConfig>>,
     pub(crate) with_timestamps: bool,
     pub(crate) ping_schedule: HashMap<String, Vec<String>>,
+    #[ignore_malloc_size_of = "TODO: Expose session memory allocations (bug 1960592)"]
```
badboy: Can you file a new bug for this specifically and change the todo here?

glean-core/src/core/mod.rs
```diff
@@ -248,6 +254,18 @@ impl Glean {
             remote_settings_config: Arc::new(Mutex::new(RemoteSettingsConfig::new())),
             with_timestamps: cfg.enable_event_timestamps,
             ping_schedule: cfg.ping_schedule.clone(),
+            // The SessionManager is deliberately left in its default (hollow)
+            // state for subprocesses.  `restore_session_state_from_storage()`
```
badboy: 😱 double-spaces after a dot.

glean-core/src/session/mod.rs
badboy: reminder: file a bug about this (and attach a TODO in the code)

glean-core/src/common_metric_data.rs
```diff
@@ -54,7 +54,7 @@ impl TryFrom<i32> for Lifetime {
 }
 
 /// The common set of data shared across all different metric types.
-#[derive(Default, Debug, Clone, Deserialize, Serialize, MallocSizeOf)]
+#[derive(Debug, Clone, Deserialize, Serialize, MallocSizeOf, Default)]
```
badboy: 🤔 Why the reordering?

glean-core/src/lib_unit_tests.rs
```diff
@@ -1360,3 +1363,166 @@ fn pings_are_controllable_from_remote_settings_config() {
     assert!(disabled_ping.submit_sync(&glean, None));
     assert!(!enabled_ping.submit_sync(&glean, None));
 }
+
+// ---------------------------------------------------------------------------
+// Session crash recovery (requires pub(crate) access)
+// ---------------------------------------------------------------------------
+
+/// Helper: create an EventMetric that matches glean.session_end boundary events.
```
badboy, line 1369: Maybe good to call out where this is defined internally?

glean-core/src/lib_unit_tests.rs
```diff
@@ -1360,3 +1363,166 @@ fn pings_are_controllable_from_remote_settings_config() {
     assert!(disabled_ping.submit_sync(&glean, None));
     assert!(!enabled_ping.submit_sync(&glean, None));
 }
+
+// ---------------------------------------------------------------------------
+// Session crash recovery (requires pub(crate) access)
+// ---------------------------------------------------------------------------
```
badboy: access to what?

Also this extensive commenting style isn't used before by us.
Also also the file's getting really long, we might even consider splitting up the tests here (can be done in a followup)

badboy (approved):
Re-read through this today, including looking at the tests. 
That looks very thorough in `lib_unit_tests.rs`. I think we would benefit from some tests in at least one of the bindings though to ensure e.g. the auto-mode actually works nicely with the automatic lifecycle there (guess the easiest then would be to do it for Swift, where we actually do run the full integration tests?) -- should be done in a followup (file a bug for it?)

Couple inline comments left.

Also this needs an update to the docs at https://mozilla.github.io/glean/dev/core/internal/payload.html#event
What's the plan for docs in general? I would want some landed before this goes into a release.

CHANGELOG.md
travis79: Yeah, I messed up the changelog good today 😅. I've addressed that in the latest commit.

glean.1.schema.json
travis79: Yep, this was all just smoke and mirrors to make things work. I'll land the schema bits first, then update this branch with the update-schemas script output.

glean-core/src/core/mod.rs
travis79: 😱

glean-core/src/core/mod.rs
travis79: Bug 2043355 filed, TODO updated.

glean-core/src/common_metric_data.rs
travis79: I think I'd removed it to add a specific implementation for  `out_of_session`, and then added it back in the wrong order...   will fix it back to the original

glean-core/src/session/mod.rs
travis79: bug 2043357 filed, todo added

PR #3442, opened 2026-04-22
glean-build: 2 improvements: Allow to overwrite venv dir in code & allow to change format

badboy:
Pre-work for #3426

### DIFF
--- a/glean-core/build/src/lib.rs
+++ b/glean-core/build/src/lib.rs
@@ -46,6 +46,7 @@ pub struct Builder {
     files: Vec<String>,
     out_dir: String,
     env_dir: Option<PathBuf>,
+    format: String,
 }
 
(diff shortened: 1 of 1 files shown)


jeddai (commented):
I see the `env_dir` member on `Builder` but I don't see where that can be set in the args, is there something I'm missing there? Or is it meant to only be used in code?

badboy:
> I see the `env_dir` member on `Builder` but I don't see where that can be set in the args, is there something I'm missing there? Or is it meant to only be used in code?

I see the `env_dir` member on `Builder` but I don't see where that can be set in the args, is there something I'm missing there? Or is it meant to only be used in code?

`fn env_dir` on the builder.

jeddai (approved):

PR #3449, opened 2026-05-06
Import sqlite storage as separate module

badboy:
My plan over the next few weeks is to split up #3405 into reviewable pieces and merge them piece by piece.
For that I created the [`main-sqlite`](https://github.com/mozilla/glean/tree/main-sqlite) branch, which currently is the same as `main`. This way any feature work/bug fixes against the current Glean can continue to go to `main` without issues.
Over time I can rebase/merge that branch against `main` to keep-up-to-date, but keep further sqlite work reviewable.

---

this very first part is merely adding the first pieces of code, but none of that is actually part of the compilation yet (wouldn't work because it's not fully usable yet).
The details about the sqlite setup will change in later commits. Re-arranging that into a single commit that has the perfect setup would be a lot of work of splitting and rebasing commits (and squashing and merge conflicts), so doing it this way seemed easier.

### DIFF
--- a/glean-core/src/database/sqlite.rs
+++ b/glean-core/src/database/sqlite.rs
@@ -1,0 +1,443 @@
+// This Source Code Form is subject to the terms of the Mozilla Public
+// License, v. 2.0. If a copy of the MPL was not distributed with this
+// file, You can obtain one at https://mozilla.org/MPL/2.0/.
+
+use std::fs;
+use std::num::NonZeroU64;
+use std::path::Path;
+use std::str;
+
+use rusqlite::params;
+use rusqlite::types::FromSqlError;
+use rusqlite::Transaction;
+
+use connection::Connection;
+use schema::Schema;
+
+use crate::common_metric_data::CommonMetricDataInternal;
+use crate::metrics::Metric;
+use crate::Glean;
+use crate::Lifetime;
+use crate::Result;
+
+mod connection;
+mod schema;
+
+pub struct Database {
+    /// The database connection.
+    conn: connection::Connection,
+}
+
+impl std::fmt::Debug for Database {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        fmt.debug_struct("Database")
+            .field("conn", &self.conn)
+            .finish()
+    }
+}
+
+const DEFAULT_DATABASE_FILE_NAME: &str = "glean.sqlite";
+
+impl Database {
+    /// Initializes the data store.
+    ///
+    /// This opens the underlying SQLite store and creates
+    /// the underlying directory structure.
+    pub fn new(data_path: &Path, _delay_ping_lifetime_io: bool) -> Result<Self> {
+        let path = data_path.join("db");
+        log::debug!("Database path: {:?}", path.display());
+
+        fs::create_dir_all(&path)?;
+        let store_path = path.join(DEFAULT_DATABASE_FILE_NAME);
+        let conn = Connection::new::<Schema>(&store_path).unwrap();
+
+        let db = Self { conn };
+
+        Ok(db)
+    }
+
+    /// Get the initial database file size.
+    pub fn file_size(&self) -> Option<NonZeroU64> {
+        None
+    }
+
+    /// Get the rkv load state.
+    pub fn rkv_load_state(&self) -> Option<String> {
+        None
+    }
+
+    /// Iterates with the provided transaction function
+    /// over the requested data from the given storage.
+    ///
+    /// * If the storage is unavailable, the transaction function is never invoked.
+    /// * If the read data cannot be deserialized it will be silently skipped.
+    ///
+    /// # Arguments
+    ///
+    /// * `lifetime` - The metric lifetime to iterate over.
+    /// * `storage_name` - The storage name to iterate over.
+    /// * `transaction_fn` - Called for each entry being iterated over. It is
+    ///   passed two arguments: `(metric_id: &[u8], metric: &Metric)`.
+    ///
+    /// # Panics
+    ///
+    /// This function will **not** panic on database errors.
+    pub fn iter_store<F>(&self, lifetime: Lifetime, storage_name: &str, mut transaction_fn: F)
+    where
+        F: FnMut(&[u8], &Metric),
+    {
+        let iter_sql = r#"
+        SELECT id, value
+        FROM telemetry
+        WHERE
+            lifetime = ?1
+            AND ping = ?2
+        "#;
+
+        self.conn
+            .read(|conn| {
+                let mut stmt = conn.prepare_cached(iter_sql).unwrap();
+                let rows = stmt
+                    .query_map(
+                        params![lifetime.as_str().to_string(), storage_name],
+                        |row| {
+                            let id: String = row.get(0)?;
+                            let blob: Vec<u8> = row.get(1)?;
+                            let blob: Metric = bincode::deserialize(&blob)
+                                .map_err(|_| FromSqlError::InvalidType)?;
+                            Ok((id, blob))
+                        },
+                    )
+                    .unwrap();
+
+                for row in rows {
+                    let Ok((metric_id, metric)) = row else {
+                        continue;
+                    };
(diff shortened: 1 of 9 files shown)


glean-core/src/database/sqlite/schema.rs
```diff
@@ -0,0 +1,68 @@
...
+    const MAX_SCHEMA_VERSION: u32 = 1;
+
+    type Error = SchemaError;
+
+    fn setup(conn: &mut rusqlite::Connection) -> Result<(), Self::Error> {
+        conn.execute_batch(
+            "PRAGMA journal_mode = WAL;
+             PRAGMA journal_size_limit = 512000; -- 512 KB.
+             PRAGMA temp_store = MEMORY;
+             PRAGMA auto_vacuum = INCREMENTAL;
```
chutten: Gonna want comments about why these values.

glean-core/src/database/sqlite/schema.rs
```diff
@@ -0,0 +1,68 @@
...
+            ",
+        )?;
+
+        // Set hardening flags.
```
chutten, line 37: Having now read the doc, I'd group this ("ordinary SQL able to deliberately corrupt the db") with misfeatures

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,442 @@
...
+        SELECT id, value
+        FROM telemetry
+        WHERE
+            lifetime = ?1
+            AND ping = ?2
+        "#;
+
+        self.conn
+            .read(|conn| {
+                let mut stmt = conn.prepare_cached(iter_sql).unwrap();
```
chutten, line 99: unguarded and uncommented unwrap

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,442 @@
...
+                for row in rows {
+                    let Ok((metric_id, metric)) = row else {
+                        continue;
+                    };
+                    transaction_fn(metric_id.as_bytes(), &metric);
+                }
+
+                Result::<(), ()>::Ok(())
+            })
+            .unwrap()
```
chutten, line 122: unguarded and uncommented unwrap

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,442 @@
...
+        }
+
+        let name = data.identifier(glean);
+
+        _ = self.conn.write(|tx| {
+            for ping_name in data.storage_names() {
+                if let Err(e) =
+                    self.record_per_lifetime(tx, data.inner.lifetime, ping_name, &name, value)
+                {
+                    log::error!(
```
chutten, line 183: (musing) I was gonna say "TODO: instrument this error" but I guess we have limited capacity for instrumenting the db gone awry. ...I suppose an in-memory error structure could be included in a "health" ping or something... I dunno.

glean-core/src/database/sqlite/schema.rs
```diff
@@ -0,0 +1,68 @@
...
+
+    fn create(tx: &mut Transaction<'_>) -> Result<(), Self::Error> {
+        tx.execute_batch(
+            "CREATE TABLE telemetry(
+               id TEXT NOT NULL,
+               ping TEXT NOT NULL,
+               lifetime TEXT NOT NULL,
+               labels TEXT NOT NULL, -- can't be null or ON CONFLICT won't work
+               value BLOB,
+               UNIQUE(id, ping, labels)
```
chutten, line 56: This reflects a change in semantics, doesn't it? Previously, if a metric changed lifetime we'd store both.

Also: why are `labels` part of the unique constraint?

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,442 @@
...
+        F: FnMut(Option<Metric>) -> Metric,
+    {
+        let find_sql = r#"
+        SELECT value
+        FROM telemetry
+        WHERE
+            lifetime = ?1
+            AND ping = ?2
+            AND id = ?3
+        LIMIT 1
```
chutten, line 297: I'd be more tempted to allow all rows in the result and instrument an error if we got more than 1

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,442 @@
...
+    /// * If any individual delete fails, an error is returned, but other deletions might have
+    ///   happened.
+    ///
+    /// Otherwise `Ok(())` is returned.
+    ///
+    /// # Panics
+    ///
+    /// This function will **not** panic on database errors.
+    pub fn clear_ping_lifetime_storage(&self, storage_name: &str) -> Result<()> {
+        let clear_sql = "DELETE FROM telemetry WHERE lifetime = ?1 AND ping = ?2";
```
chutten: Why not include `Lifetime::Ping.as_str()` in the statement directly?

glean-core/src/database/sqlite.rs
badboy: we're cleaning those up in later code.

glean-core/src/database/sqlite.rs
badboy: we're cleaning those up in later code.

glean-core/src/database/sqlite.rs
badboy: Hm, yeah, I can file a bug to do that later.

glean-core/src/database/sqlite/schema.rs
badboy: a) That's true, but it's kind of ... is it a bug? Unintended side-effect? We can never sent out _two_ values. And based on the order of lifetimes by which we fetch values (ping, app, user) we end up with the "longest" one.

b) Because that's what makes it unique?
A labeled metric `cat.name` will add these rows:
| id | ping | labels | value |
| -- | ---- | ------ | ----- |
| cat.name | metrics | label1 | 1 |
| cat.name | custom | label1 | 1 | 
| cat.name | metrics | label2 | 31 |
| cat.name | custom | label2| 31 |

We use `ON CONFLICT` to do an upsert and that acts on the uniqueness constraint.
If we would not put `labels` in the unique constraint, we would overwrite previous rows.

Is that more clear?

glean-core/src/database/sqlite.rs
badboy: I'll add a note and we can clean that up when the complete logic is landed.

glean-core/src/database/sqlite.rs
badboy: Guess mostly to avoid copy-n-paste errors should we ever consider renaming `Ping` :P (I'll inline it)

glean-core/src/database/sqlite.rs
badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=2042828

badboy:
Woops, I initially pushed 3 changes as 3 separate commits, but then locally already squashed them and repushed accidentally. Only the 3 changes can be seen here: https://github.com/mozilla/glean/compare/d3d4e1bbab652bf31e3c8b0c2bbb5558ccebfb49...09f39e8bd96807c9f97026221eb2a8c503030897

glean-core/src/database/sqlite/schema.rs
chutten: a) I guess the only changes in behaviour would be in the weird corners. e.g. Like if someone changed more than just the `lifetime` at the same time so that different values were acceptable. Or if someone downgraded a `user`-lifetime metric to `ping`, then with this change that wouldn't leave behind a `user`-lifetime copy.

Okay, fair enough.

b) Ohhhhhhhhh, okay, so my new problem is that the column should be called `"label"` not `"labels"`

badboy: (b) It _can_ hold multiple labels (for dual-labeled metrics)

(Sorry for merging ahead of that discussion, we can of course still tackle that in a followup; I'll pull out this discussion into a bug so it's not lost in PR comments)

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=2044932

PR #3453, opened 2026-05-08
Downgrade maturin to ensure builds work for now (and run builds in CI to ensure they work)

badboy:
Newer maturin changed how the Windows cross-compile builds work and that currently breaks for us.
The easiest is to stay with what we used before until we have time to handle the upgrade.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -861,7 +861,7 @@ jobs:
           name: Setup Python env
           command: |
             make setup-python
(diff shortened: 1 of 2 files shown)


badboy:
for the last release I ran these manually and uploaded the wheels:

```
maturin build --release --target aarch64-pc-windows-gnullvm --zig  --verbose
maturin build --release --target aarch64-unknown-linux-gnu --zig
```

travis79 (approved):

PR #3462, opened 2026-05-20
Bug 2040030 - Increase pending pings limitations to better support long network outages

travis79:
This increases the pending pings limit from 250 to 500, and also increases the pending pings directory size limit from 10MB to 50MB. In order to make this more configurable to consuming apps, I've also exposed these as properties in the Configuration.

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -816,8 +816,10 @@ glean.upload:
     description: |
       The number of pings deleted because a pending pings quota was exceeded,
       labeled by which quota triggered the deletion.
-      `count_quota`: the maximum number of pending pings was reached.
-      `size_quota`: the maximum size of the pending pings directory was reached.
+
+      * `count_quota`: the maximum number of pending pings was reached.
+      * `size_quota`: the maximum size of the pending pings directory was reached.
+
       Sent on the health ping only; use this metric to verify whether
       the configured limits are sufficient to survive connectivity outages.
     send_in_pings:
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -808,6 +808,32 @@ glean.upload:
     no_lint:
       - COMMON_PREFIX
 
+  pending_pings_deleted:
+    type: labeled_counter
+    labels:
+      - count_quota
+      - size_quota
+    description: |
+      The number of pings deleted because a pending pings quota was exceeded,
+      labeled by which quota triggered the deletion.
+      `count_quota`: the maximum number of pending pings was reached.
+      `size_quota`: the maximum size of the pending pings directory was reached.
+      Sent on the health ping only; use this metric to verify whether
+      the configured limits are sufficient to survive connectivity outages.
+    send_in_pings:
+      - health
+    bugs:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=2040030
+    data_reviews:
+      - https://bugzilla.mozilla.org/show_bug.cgi?id=2040030
+    data_sensitivity:
+      - technical
+    notification_emails:
+      - glean-team@mozilla.com
+    expires: never
+    no_lint:
+      - COMMON_PREFIX
+
   pending_pings:
     type: counter
     description: |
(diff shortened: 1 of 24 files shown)


uv.lock
```diff
@@ -0,0 +1,695 @@
+version = 1
```
badboy, line 1: uh oh, we didn't have this committed before, so I guess some change in usage?

glean-core/src/upload/policy.rs
```diff
@@ -9,15 +9,13 @@ use malloc_size_of_derive::MallocSizeOf;
...
-// With the default rate limit of 15 pings per 60s it would take roughly 16 minutes to send out all pending
-// pings.
-const MAX_PENDING_PINGS_COUNT: u64 = 250;
+                                               // Directory safety limit for managing pending pings, along with `MAX_PENDING_PINGS_COUNT`.
+                                               // The default of 50MB is sized more for outage survival than normal day-to-day ping volumes.
+const MAX_PENDING_PINGS_DIRECTORY_SIZE: u64 = 50 * 1024 * 1024;
+// Current Android P80 baseline ping volume for release is around 28 pings/client/day. (May 2026 data)
+// A limit of 500 should allow a P99 client to buffer roughly 5 days of mixed-type pings before pings
+// are deleted. Integrating apps may override this via Configuration if needed.
+const MAX_PENDING_PINGS_COUNT: u64 = 500;
```
badboy, line 19: Would have been nice to have this as a separate commit for just this increase.

glean-core/src/upload/policy.rs
```diff
@@ -9,15 +9,13 @@ use malloc_size_of_derive::MallocSizeOf;
...
-const MAX_PENDING_PINGS_DIRECTORY_SIZE: u64 = 10 * 1024 * 1024; // 10MB
-
-// The average number of baseline pings per client (on Fenix) is at 15 pings a day.
-// The P99 value is ~110.
-// With a maximum of (a nice round) 250 we can store about 2 days worth of pings.
-// A baseline ping file averages about 600 bytes, so that's a total of just 144 kB we store.
-// With the default rate limit of 15 pings per 60s it would take roughly 16 minutes to send out all pending
-// pings.
-const MAX_PENDING_PINGS_COUNT: u64 = 250;
+                                               // Directory safety limit for managing pending pings, along with `MAX_PENDING_PINGS_COUNT`.
```
badboy: This probably shouldn't be indented as is, as it applies to the const below.

glean-core/metrics.yaml
```diff
@@ -808,6 +808,32 @@ glean.upload:
...
+    type: labeled_counter
+    labels:
+      - count_quota
+      - size_quota
+    description: |
+      The number of pings deleted because a pending pings quota was exceeded,
+      labeled by which quota triggered the deletion.
+      `count_quota`: the maximum number of pending pings was reached.
+      `size_quota`: the maximum size of the pending pings directory was reached.
+      Sent on the health ping only; use this metric to verify whether
```
badboy, line 823:
```suggestion
      labeled by which quota triggered the deletion.
      
      * `count_quota`: the maximum number of pending pings was reached.
      * `size_quota`: the maximum size of the pending pings directory was reached.
      
      Sent on the health ping only; use this metric to verify whether
```

Some formatting, otherwise this will be smushed in the Dictionary display.

badboy (approved):
r+wc. Comments inline.

If you had asked me like a week ago I could have sworn we had already exposed those limits as configuration options. But I guess we didn't.

uv.lock
travis79: Ooops, my bad. I seem to miss the uv.lock file when staging changes frequently. Is this something we can put in the gitignore? we don't need to commit this anywhere do we?

badboy: Let me look into this. Maybe we _should_ commit it?

glean-core/src/upload/policy.rs
travis79: Doh, I think this was VSCode trying to be smart and failing.

glean-core/src/upload/policy.rs
travis79: sorry, increasing one without the other seemed pointless. Regardless, you are right. I could have organized this as separate commits.

uv.lock
travis79: Leaving uv.lock for now, I guess. We can always remove it later if it isn't helpful

glean-core/src/upload/policy.rs
badboy: I meant having the increases as one commit and then a second commit that adds all the machinery to make it configurable.

PR #3465, opened 2026-05-27
Use the sqlite module as the database layer

badboy:
This only integrates it into the module tree.
It compiles, but not warning-free. ~~It does not compile yet. Later changes will clean that up.~~
It fully replaces the Rkv storage. No migration implemented.

---

Done against sqlite/initial-import right now, will change to `main-sqlite` when #3449 is merged.

### DIFF
--- a/glean-core/src/database/sqlite.rs
+++ b/glean-core/src/database/sqlite.rs
@@ -6,7 +6,9 @@ use std::fs;
 use std::num::NonZeroU64;
 use std::path::Path;
 use std::str;
+use std::time::Duration;
 
+use malloc_size_of::MallocSizeOf;
 use rusqlite::params;
 use rusqlite::types::FromSqlError;
 use rusqlite::Transaction;
@@ -28,6 +30,13 @@ pub struct Database {
     conn: connection::Connection,
 }
 
+impl MallocSizeOf for Database {
+    fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+        // FIXME: Can we get the allocated size of the connection?
+        0
+    }
+}
+
 impl std::fmt::Debug for Database {
     fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
         fmt.debug_struct("Database")
@@ -43,7 +52,12 @@ impl Database {
     ///
     /// This opens the underlying SQLite store and creates
     /// the underlying directory structure.
-    pub fn new(data_path: &Path, _delay_ping_lifetime_io: bool) -> Result<Self> {
+    pub fn new(
+        data_path: &Path,
+        _delay_ping_lifetime_io: bool,
+        _ping_lifetime_threshold: usize,
+        _ping_lifetime_max_time: Duration,
+    ) -> Result<Self> {
         let path = data_path.join("db");
         log::debug!("Database path: {:?}", path.display());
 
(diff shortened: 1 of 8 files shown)


badboy:
Oh and of course tests won't pass (for a while)

glean-core/src/database/sqlite.rs
```diff
@@ -28,6 +30,13 @@ pub struct Database {
     conn: connection::Connection,
 }
 
+impl MallocSizeOf for Database {
+    fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+        // FIXME: Can we get the allocated size of the connection?
```
chutten, line 35: A quick survey of rust users of sqlite in m-c shows no one trying this yet.

glean-core/src/database/sqlite.rs
```diff
@@ -352,6 +366,10 @@ impl Database {
         })
     }
 
+    pub fn clear_lifetime_storage(&self, lifetime: Lifetime, storage_name: &str) -> Result<()> {
+        Ok(())
```
chutten, line 371: No FIXME, no `todo!` ?

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -424,7 +424,7 @@ fn get_seen_keys_and_categories(
     for store in &meta.inner.send_in_pings {
         glean
             .storage()
-            .iter_store_from(lifetime, store, Some(&prefix), &mut snapshotter);
+            .iter_store(lifetime, store, &mut snapshotter);
```
chutten, line 427: This makes `prefix` unused, doesn't it?

glean-core/src/metrics/labeled.rs
```diff
@@ -431,7 +431,7 @@ pub fn validate_dynamic_label(
     for store in &meta.inner.send_in_pings {
         glean
             .storage()
-            .iter_store_from(lifetime, store, Some(prefix), &mut snapshotter);
+            .iter_store(lifetime, store, &mut snapshotter);
```
chutten, line 434: ditto, unused `prefix`

glean-core/src/ping/mod.rs
```diff
@@ -284,16 +284,6 @@ impl PingMaker {
         info!("Collecting {}", ping.name());
         let database = glean.storage();
 
-        // HACK: Only for metrics pings we add the ping timings.
-        // But we want that to persist until the next metrics ping is actually sent.
-        let write_samples = database.write_timings.replace(Vec::with_capacity(64));
```
chutten, line 289: So `write_timings` will never be drained in rkv mode?

chutten: I mean, I know there is no "rkv mode" in this patchset, but unbounded memory writes seems worth bringing up.

chutten (approved):
r+wc Seems legit, but absent a control mechanism to switch between rkv and sqlite and migration this isn't complete. Exactly as you say in the PR summary, of course, but I guess I'm restating for the gallery that approving this is not the usual "I'm happy with this code shipping (after the small fixes)".

glean-core/src/metrics/dual_labeled_counter.rs
badboy: Correct, this is to make it compile. The functionality is restored in later commits then.

glean-core/src/metrics/labeled.rs
badboy: as above.

glean-core/src/ping/mod.rs
badboy: This metric will go away in later commits. It's already disabled anyway.

PR #3471, opened 2026-05-28
Bug 2042327 - glean-sym: Enable it on iOS by using `this` library

badboy:
This is the equivalent of using `dlsym(RTLD_DEFAULT, name)`, that is:

> looking for the desired symbols using the default shared object search order.
> The search will include global symbols in the executable and its

(from the dlsym(3) man page)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,6 +6,7 @@
   * Disable the `glean.error.event_timestamp_clamped` metric ([#3500](https://github.com/mozilla/glean/pull/3500)).
 * Rust
   * Make `rate_limit` configurable in the configuration builder ([#3495](https://github.com/mozilla/glean/pull/3495))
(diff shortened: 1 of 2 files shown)


badboy:
I tested this out with a local build of appservices and this Glean change in Firefox iOS. It's able to load the function pointers.
E.g. when adding some panic to check what it does:

> Loaded Glean symbols. ffi_glean_core_rustbuffer_from_bytes: Symbol "ffi_glean_core_rustbuffer_from_bytes"@0x108aa91e8 from "/Users/jer/Library/Developer/Xcode/DerivedData/Client-ggbdmvoodhczijcievulgqgnjkrv/Build/Products/Fennec-iphonesimulator/Glean.framework/Glean"

travis79 (approved):
Let's be sure and eventually document this difference wherever the user facing docs for glean-sym will be

PR #3472, opened 2026-05-30
Migrate the Glean Gradle plugin off the legacy AGP variant API

rvandermeulen:
The legacy AGP variant API (`applicationVariants`/`libraryVariants`) and `registerJavaGeneratingTask` were removed in AGP 9. This migrates `GleanGradlePlugin.groovy` to the stable replacements (`androidComponents.onVariants` and `sources.java.addGeneratedSourceDirectory`) and reads the variant namespace via `variant.namespace` instead of the BuildConfig provider.

These APIs exist in both AGP 8 and 9, so the plugin remains compatible with all current consumers while adding AGP 9 support. The plugin has no compile-time AGP dependency, so the Glean project itself is unaffected and stays on AGP 8.

**Testing:** Built the `:glean` library and `:glean-sample-app` application modules on AGP 8.13.2 (debug and release), and verified against mozilla-central by locally publishing the plugin and building a Glean-consuming component on Gradle 9.5.1 / AGP 8.13.2; metrics generate and compile into the expected package in all cases.

### DIFF
--- a/.dictionary
+++ b/.dictionary
@@ -1,7 +1,8 @@
-personal_ws-1.1 en 318 utf-8
+personal_ws-1.1 en 319 utf-8
 AAR
 AARs
 ABI
+AGP
 API's
 APIs
 APK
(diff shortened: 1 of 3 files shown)


badboy (approved):

badboy:
fixing spellcheck in a second

PR #3476, opened 2026-06-03
Bug 2044592 - glean-sym: Automatically check and update the embedded contract version using a test

badboy:


glean-core/glean-sym/src/lib.rs
```diff
@@ -26,6 +26,8 @@ pub mod __export {
     pub use once_cell::sync::Lazy;
 }
 
+const FFI_CONTRACT_VERSION: u32 = include!("contract_version.txt");
```
travis79, line 29: Well, that's the first time I've seen `include!` used in this way...  neat!

glean-core/glean-sym/src/lib.rs
```diff
@@ -66,13 +68,25 @@ macro_rules! library_binding {
                 let $localname = library.ok_or_else(|| {
                     std::io::Error::new(std::io::ErrorKind::NotFound, "failed to find glean library")
                 })?;
-                Ok(GleanSym { $($load)* _library: $localname })
+
+                let handle = GleanSym { $($load)* _library: $localname };
+                handle.check()?;
```
travis79, line 73: If this fails, should we log or record it somewhere?

glean-core/glean-sym/src/lib.rs
```diff
@@ -66,13 +68,25 @@ macro_rules! library_binding {
...
             pub fn load() -> std::io::Result<Self> {
                 compile_error!("This crate is not implemented for Windows");
             }
+
+            fn check(&self) -> std::io::Result<()> {
+                let ffi_contract_version = unsafe { (self.ffi_glean_core_uniffi_contract_version)() };
+                if crate::FFI_CONTRACT_VERSION == ffi_contract_version {
+                    Ok(())
+                } else {
+                    Err(std::io::Error::other("contract version mismatch"))
```
travis79, line 87: Why `stderr` instead of a log line?

glean-core/glean-sym/tests/parser.rs
```diff
@@ -31,3 +31,28 @@ fn generated_metrics_code_up_to_date() {
...
+        .expect("can't parse content of contract_version.txt");
+
+    if glean_core_version != version {
+        let version = format!("{glean_core_version}\n");
+        fs::write(embedded_version, version).unwrap();
+    }
+
+    assert_eq!(
+        glean_core_version, version,
+        "Wrong contract version in contract_version.txt. File has been automatically updated. Please commit it."
```
travis79, line 56: Ahh, disregard my comments about logs/stderr, I understand now! We aren't using this anywhere aside from tests so the logging is produced through test output.

glean-core/glean-sym/src/contract_version.txt
```diff
@@ -0,0 +1 @@
+30
```
travis79, line 1: What updates this? The version-update script?

travis79 (approved):

glean-core/glean-sym/src/lib.rs
badboy: It will actually crash on first access here: https://github.com/mozilla/glean/blob/56c7d8b3ade8727e4e783170e32aff1d878d07bc/glean-core/glean-sym/src/lib.rs#L89 
-- this would be caught in testing! It's merely a check that we compile everything correctly together.

glean-core/glean-sym/src/lib.rs
badboy: it's not stderr, it's a [`std::io::Error`](https://doc.rust-lang.org/std/io/struct.Error.html). That's because we already return that from `load`, so we don't need to invent a new error type.

glean-core/glean-sym/tests/parser.rs
badboy: well, we do call it as part of `load` and then in the `GLEAN` global static, where it will unrap and thus panic.

glean-core/glean-sym/src/contract_version.txt
badboy: the test! -> https://github.com/mozilla/glean/blob/push-orspuoqpmxwu/glean-core/glean-sym/tests/parser.rs#L49-L51

the contract version is determined by UniFFI, so it's automatically embedded in the compilation. our test merely checks that the file is matching that (so that `glean-sym` doesn't actually depend on `glean-core` directly)

badboy:
Not sure why the `Vet Dependencies` task didn't run. I'll merge manually.

PR #3478, opened 2026-06-04
sqlite: Set synchronous=NORMAL to sync less often, but at the critical moments

badboy:
See all details:
https://sqlite.org/pragma.html#pragma_synchronous

The default (FULL) syncs on every write.
That's slightly higher guarantees, but also costly. We're already using WAL (write-ahead log). It's safe from corruption in NORMAL mode and consistent.
It does lose durability, that means data might roll back following a power loss or system crash.

Note: `rkv` does NOT sync at all. It only writes to disk (and moves files around). That's strictly worse than WAL in `NORMAL` mode.


chutten (approved):
Double-r+ for the informative PR summary

PR #3480, opened 2026-06-04
sqlite: Measure database on-disk size by scanning the directory

badboy:
Same way this was done on Rkv: we just some up the size of all files in the database directory.

### DIFF
--- a/Cargo.lock
+++ b/Cargo.lock
@@ -618,6 +618,7 @@ dependencies = [
  "pulldown-cmark",
  "pulldown-cmark-to-cmark",
  "rkv",
+ "rmp-serde",
  "rusqlite",
  "serde",
  "serde_json",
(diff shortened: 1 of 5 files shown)


travis79:
Would it be good to have a per-table metric of this too?

If SQLite is compiled with `SQLITE_ENABLE_DBSTAT_VTAB`, you can query the `dbstat` table. This should return the table size (in bytes) of the table by name:

`SELECT SUM("pgsize") FROM "dbstat" WHERE name='TABLENAME';`

There's also some `PRAGMA` statement queries that you can use with the formula: page_count * page_size = database_size:

```
PRAGMA page_count;
PRAGMA page_size;
```

badboy:
> Would it be good to have a per-table metric of this too?
> 
> If SQLite is compiled with `SQLITE_ENABLE_DBSTAT_VTAB`, you can query the `dbstat` table. This should return the table size (in bytes) of the table by name:
> 
> `SELECT SUM("pgsize") FROM "dbstat" WHERE name='TABLENAME';`
> 
> There's also some `PRAGMA` statement queries that you can use with the formula: page_count * page_size = database_size:
> 
> ```
> PRAGMA page_count;
> PRAGMA page_size;
> ```

Oh not a bad idea. I would probably punt that to a followup though (I'll file a bug).

badboy:
https://bugzilla.mozilla.org/show_bug.cgi?id=2044991

travis79 (approved):

PR #3481, opened 2026-06-04
Implement the label verifier against sqlite storage

badboy:
It's now easier to do: query the column and count.
There's some complications when we get to dual-labeled metrics, but that
comes later.

### DIFF
--- a/glean-core/src/metrics/dual_labeled_counter.rs
+++ b/glean-core/src/metrics/dual_labeled_counter.rs
@@ -108,23 +108,20 @@ impl DualLabeledCounterMetric {
     /// the static or dynamic labels where needed.
     fn new_counter_metric(&self, key: &str, category: &str) -> CounterMetric {
         match (&self.keys, &self.categories) {
-            (None, None) => self
-                .counter
-                .with_dynamic_label(DynamicLabelType::KeyAndCategory(
-                    make_label_from_key_and_category(key, category),
-                )),
+            (None, None) => self.counter.with_label(DynamicLabelType::KeyAndCategory(
+                make_label_from_key_and_category(key, category),
+            )),
             (None, _) => {
                 let static_category = self.static_category(category);
-                self.counter.with_dynamic_label(DynamicLabelType::KeyOnly(
+                self.counter.with_label(DynamicLabelType::KeyOnly(
                     make_label_from_key_and_category(key, static_category),
                 ))
             }
             (_, None) => {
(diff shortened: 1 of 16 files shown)


badboy:
I'm not perfectly happy where I had to do the split in commits here. I need to give this a proper look in this shape myself.
It's somewhat incomplete in that some other logic is coming in later commits, but due to how I structured this re-organizing the commits now would require a lot of rebasing and solving more merge conflicts.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -302,79 +302,111 @@ pub fn validate_dynamic_key_and_or_category(
     base_identifier: &str,
     label: DynamicLabelType,
 ) -> String {
-    // We should have exactly 3 elements when splitting by `RECORD_SEPARATOR`, since the label should begin with one and
-    // then the key and category are separated by one. Split should contain an empty string, the key, and the category.
-    // If we have more than 3 elements, then the consuming app must have used this character as part of a label and we
-    // cannot determine whether it was the key or the category at this point, so we record an `InvalidLabel` error and
-    // return `OTHER_LABEL` for both key and category.
-    if label.split(RECORD_SEPARATOR).count() != 3 {
+    panic!("not validating dual labeled like this anymore");
```
travis79, line 305: 🫶

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -302,79 +302,111 @@ pub fn validate_dynamic_key_and_or_category(
...
-        for store in &meta.inner.send_in_pings {
-            if glean.storage().has_metric(meta.inner.lifetime, store, key) {
-                return combine_base_identifier_and_labels(base_identifier, key, category);
-            }
-        }
+    let mut existing_keys = HashSet::new();
+    let mut existing_categories = HashSet::new();
+    'checkdb: {
+        let Ok(mut stmt) = tx.prepare(existing_labels_sql) else {
+            // If we can't fetch from the database, assume the label is ok to use
```
travis79, line 337: Does this assumption change the previous behavior in any way?

travis79 (approved):
A couple of comments, but otherwise very happy with how this is looking.

glean-core/src/metrics/dual_labeled_counter.rs
badboy: Nope, we always did it that way.

PR #3482, opened 2026-06-04
bug 2043535 - Add clear_{att|dist}ribution methods to clear core attribution and distribution data.

chutten:


glean-core/src/lib.rs
```diff
@@ -1316,6 +1316,16 @@ pub fn glean_set_dirty_flag(new_value: bool) {
     core::with_glean(|glean| glean.set_dirty_flag(new_value))
 }
 
+/// Clears the core attribution data.
+/// Does not clear glean.attribution.ext (if present).
+pub fn glean_clear_attribution() {
+    if was_initialize_called() && core::global_glean().is_some() {
+        core::with_glean(|glean| glean.clear_attribution());
+    } else {
+        _ = PRE_INIT_ATTRIBUTION.lock().unwrap().take()
```
badboy, line 1379: This is not enough to actually clear values! `attribution` is user-lifetime so we need to _explicitly_ delete the value from the database.
(which also means this is missing a test for that case)

chutten: Good point. Lemme rework the test, and get clearing working.

glean-core/python/glean/glean.py
```diff
@@ -495,6 +495,14 @@ def shutdown(cls):
         # we also wait for the process dispatcher to finish.
         ProcessDispatcher._wait_for_last_process()
 
+    @classmethod
+    def clear_attribution(cls) -> None:
```
jeddai, line 526: Do we need to add these to Kotlin/Swift files at all or is it just Python where we have this sort of layer?

jeddai (approved):
r+wc

glean-core/python/glean/glean.py
chutten: Whoops! Yes we do. Good catch!

PR #3483, opened 2026-06-04
CI: Update actions used

badboy:


chutten (approved):
I trust you and everything, but it'd be nice to have a little blurb in the commit msg (or at least the PR) about what motivates this. Are there nice bonuses in the new versions? The stability of moving off `@master` is probably part of the reason. etc.

badboy:
good callout! It's deprecation warnings, e.g. like here: https://github.com/mozilla/glean/actions/runs/26963443711

> Node.js 20 actions are deprecated. The following actions are running on Node.js 20 and may not work as expected: actions/cache@v4. Actions will be forced to run with Node.js 24 by default starting June 16th, 2026. Node.js 20 will be removed from the runner on September 16th, 2026

so while at it it I made sure we use the same (latest) versions of all actions.

PR #3489, opened 2026-06-08
sqlite: tests! Reenabling and adding a whole lot of them.

badboy:


glean-core/src/database/sqlite.rs
```diff
@@ -218,7 +218,7 @@ impl Database {
                 })
                 .optional()
             })
-            .unwrap_or(None)
+            .unwrap_or(None) // TODO: Should we handle the error here properly?
```
travis79: //TODO file a bug to capture this consideration

travis79 (approved):

PR #3487, opened 2026-06-08
sqlite: Further cleanup how labels are handled

badboy:
This is the rest of #3481 basically, bringing it all into good shape.

PR #3503, opened 2026-06-17
Address outstanding review comments

badboy:
Everything chutten had on #3487


glean-core/src/error_recording.rs
```diff
@@ -152,7 +152,11 @@ pub fn record_error_sqlite(
     error: ErrorType,
     num_errors: i32,
 ) {
-    assert!(num_errors > 0);
+    debug_assert!(num_errors > 0);
+    if num_errors == 0 {
```
chutten: It's signed, might want a <= 0 here.

chutten (approved):

PR #3504, opened 2026-06-17
Bug 2034605 - Switch to an SQLite storage backend

badboy:
This is essentially #3405 but from the branch we've been slowly merging into:

main <- main-sqlite.

This is what will finally get merged.
~~It won't need a full review again, all indidivual pieces have been reviewed previously.~~
This branch is rebased against `main` to ensure we do not lose any commits from `main`.


glean-core/rlb-tests/metrics.yaml
```diff
@@ -0,0 +1,257 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the metrics that are recorded by the Glean SDK. They are
```
chutten: copy-pasta comment?

glean-core/rlb-tests/pings.yaml
```diff
@@ -0,0 +1,22 @@
+# This Source Code Form is subject to the terms of the Mozilla Public
+# License, v. 2.0. If a copy of the MPL was not distributed with this
+# file, You can obtain one at http://mozilla.org/MPL/2.0/.
+
+# This file defines the built-in pings that are recorded by the Glean SDK. They
```
chutten: ditto?

glean-core/rlb/src/private/object.rs
```diff
@@ -34,7 +34,7 @@ impl<K> MallocSizeOf for ObjectMetric<K> {
 }
 
 impl<'a, K> MetricIdentifier<'a> for ObjectMetric<K> {
-    fn get_identifiers(&'a self) -> (&'a str, &'a str, Option<&'a str>) {
+    fn get_identifiers(&'a self) -> (&'a str, &'a str, Option<String>) {
```
chutten, line 37: musing: I do wonder how much more expensive this will prove to be in practice

glean-core/src/core/mod.rs
```diff
@@ -307,6 +307,29 @@ impl Glean {
...
+                .database_metrics
+                .failed_metrics
+                .add_sync(&glean, state.failed_metrics);
+            glean
+                .database_metrics
+                .migration_duration
+                .set_raw_sync(&glean, state.duration);
+        }
+
+        if let Some(()) = glean.data_store.as_mut().unwrap().migration_error.take() {
```
chutten: This looks goofy. "If the error is something, but a zero-sized something, report a single error." Not sure there's any improvement to be made, I guess, but it looked odd so I thought I'd err on the side of saying something.

glean-core/src/core/mod.rs
```diff
@@ -911,13 +932,34 @@ impl Glean {
...
-            _ = data.clear_lifetime_storage(Lifetime::User, "glean_internal_info");
-            _ = data.remove_single_metric(Lifetime::User, "glean_client_info", "client_id");
+            let warn_on_error = |result, msg| {
+                if let Err(e) = result {
+                    log::warn!("{msg}: {e}");
+                }
+            };
+
+            warn_on_error(
+                data.clear_lifetime_storage(Lifetime::User, "glean_internal_info"),
```
chutten: Should we use `INTERNAL_STORAGE` here?

glean-core/src/core/mod.rs
```diff
@@ -911,13 +932,34 @@ impl Glean {
...
+                "failed to clear interanl storage",
```
chutten:
```suggestion
                "failed to clear internal storage",
```

glean-core/src/database/sqlite/connection.rs
```diff
@@ -0,0 +1,111 @@
...
+                    version += 1;
+                }
+            }
+            Ok(0) => O::create(&mut tx)?,
+            Err(err) => Err(err)?,
+        }
+        // Set the schema version to the highest that we support.
+        // If the current version is higher than ours, downgrade it,
+        // so that upgrading to it again in the future can fix up any
+        // invariants that our version might not uphold.
```
chutten, line 73: So this means we have to be careful to make newer versions transparently downgradable? We could end up with a version 10 db with different columns, constraints, etc. but if the user opens it with code that supports up to version 5 this will treat it like a v5 db and this will in fact tell it that it's a v5 db (when it isn't)?

I mean, what if we decide to split the `labels` column into `label`, `key`, `category`? That would not be permitted here because we couldn't treat the resulting database transparently as any version before the split.

Do we need to document this in the dev book or something?

glean-core/src/database/sqlite/schema.rs
```diff
@@ -0,0 +1,75 @@
...
+    type Error = SchemaError;
+
+    fn setup(conn: &mut rusqlite::Connection) -> Result<(), Self::Error> {
+        conn.execute_batch(
+            "
+             -- we unconditionally want write-ahead-logging mode
+             PRAGMA journal_mode = WAL;
+             -- Sync at the most criticial moments, but not with every write
+             PRAGMA synchronous = NORMAL;
+             -- limit size of the journal. value currently arbitrary. needs refinement.
```
chutten: If it needs refinement, file and ref a bug for that work?

glean-core/src/database/migration.rs
```diff
@@ -0,0 +1,362 @@
...
+    /// # Arguments
+    ///
+    /// * `lifetime` - The metric lifetime to iterate over.
+    /// * `transaction_fn` - Called for each entry being iterated over.
+    ///                      It is passed two arguments: `(key: String, metric: &Metric)`.
+    ///
+    /// # Returns
+    ///
+    /// The number of metrics that failed to deserialize from the stored data.
+    pub fn iter_store<F>(&self, lifetime: Lifetime, mut transaction_fn: F) -> i32
```
chutten, line 115: i32? Isn't this guaranteed to be non-negative?

chutten: Oh. This is because it goes into a CounterMetric::add eventually, isn't it. *cries in bug 1982071

glean-core/src/database/migration.rs
```diff
@@ -0,0 +1,362 @@
...
+    pub failed_metrics: i32,
+    /// The number of metrics stored in SQLite after a migration run
+    pub metrics_in_sql: i32,
+    /// The duration for one full migration run at startup
+    pub duration: Duration,
+}
+
+/// Migrate the `rkv` database to SQL.
+///
+/// Returns the number of migrated metrics.
```
chutten: nit: Returns more than that now

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,671 @@
...
+    fn size_of(&self, _ops: &mut malloc_size_of::MallocSizeOfOps) -> usize {
+        // FIXME: Can we get the allocated size of the connection?
+        0
+    }
+}
+
+impl std::fmt::Debug for Database {
+    fn fmt(&self, fmt: &mut std::fmt::Formatter) -> std::fmt::Result {
+        fmt.debug_struct("Database")
+            .field("conn", &self.conn)
```
chutten: We're hiding the other fields because... I can't think of a reason to hide the other fields. Why?

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,671 @@
...
+
+/// Calculate the database size from all the files in the directory.
+///
+///  # Arguments
+///
+///  *`path` - The path to the directory
+///
+///  # Returns
+///
+/// Returns the non-zero combined size of all files in a directory,
```
chutten:
```suggestion
/// Returns the non-zero combined size in bytes of all files in a directory,
```

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,671 @@
...
+                }
+            }
+        }
+    }
+
+    NonZeroU64::new(total_size)
+}
+
+pub fn sqlite_open(path: &Path) -> std::result::Result<(Connection, LoadState), Error> {
+    // TODO: Make this more robust, use the correct errors and see how we can test all the branches
```
chutten: This TODO either needs to be addressed or a follow-up bug filed and referenced

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,671 @@
...
+impl Database {
+    /// Initializes the data store.
+    ///
+    /// This opens the underlying SQLite store and creates
+    /// the underlying directory structure.
+    pub fn new(
+        data_path: &Path,
+        _delay_ping_lifetime_io: bool,
+        _ping_lifetime_threshold: usize,
+        _ping_lifetime_max_time: Duration,
```
chutten, line 153: (not a bug): The cleanup is gonna be so satisfying

glean-core/src/database/sqlite.rs
```diff
@@ -0,0 +1,671 @@
...
+            match migration::try_migrate(&path, &db) {
+                Ok(Some(state)) => {
+                    log::debug!("Migration done. state={state:?}");
+                    db.migration_state = Some(state);
+                }
+                Ok(None) => {
+                    log::debug!("No migration.");
+                }
+                Err(e) => {
+                    db.migration_error = Some(());
```
chutten: nit: It feels like `migration_error` ought to take ownership of `e` here, even if we then ignore it and just record 1.

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -251,181 +248,91 @@ impl TestGetValue for DualLabeledCounterMetric {
...
+        log::warn!(
+            "Metric {:?}: label length {} exceeds maximum of {}",
+            metric_id,
             label.len(),
             MAX_LABEL_LENGTH
         );
-        record_error(glean, meta, ErrorType::InvalidLabel, msg, None);
+        false
+    } else if label.contains(RECORD_SEPARATOR) {
+        log::warn!("Metric {metric_id:?}: Label cannot contain the ASCII record separator character (0x1E)");
```
chutten, line 333: Whoops, forgot to add this limitation to the label limits docs. Lemme file... https://bugzilla.mozilla.org/show_bug.cgi?id=2048978

glean-core/src/storage/mod.rs
```diff
@@ -120,25 +118,42 @@ impl StorageManager {
...
         };
 
-        storage.iter_store_from(Lifetime::Ping, store_name, None, &mut snapshotter);
-        storage.iter_store_from(Lifetime::Application, store_name, None, &mut snapshotter);
-        storage.iter_store_from(Lifetime::User, store_name, None, &mut snapshotter);
+        if let Err(e) = storage.iter_store(Lifetime::Ping, store_name, &mut snapshotter) {
+            log::debug!("could not snapshot ping lifetime store: {e:?}");
+        }
+        if let Err(e) = storage.iter_store(Lifetime::Application, store_name, &mut snapshotter) {
+            log::debug!("could not snapshot ping lifetime store: {e:?}");
```
chutten:
```suggestion
            log::debug!("could not snapshot application lifetime store: {e:?}");
```

glean-core/src/storage/mod.rs
```diff
@@ -120,25 +118,42 @@ impl StorageManager {
...
+        }
+        if let Err(e) = storage.iter_store(Lifetime::User, store_name, &mut snapshotter) {
+            log::debug!("could not snapshot ping lifetime store: {e:?}");
```
chutten:
```suggestion
            log::debug!("could not snapshot user lifetime store: {e:?}");
```

glean-core/src/common_metric_data.rs
```diff
@@ -68,51 +70,61 @@ pub struct CommonMetricData {
     ///
     /// Disabled metrics are never recorded.
     pub disabled: bool,
-    /// Dynamic label.
+    /// Label for this metric.
```
chutten: merge conflict comment misposition

glean-core/src/common_metric_data.rs
```diff
@@ -68,51 +70,61 @@ pub struct CommonMetricData {
...
+            MetricLabel::Label(label) => write!(f, "{label}"),
+            MetricLabel::KeyOnly(key, category) => {
+                write!(f, "{key}{RECORD_SEPARATOR}{category}")
+            }
+            MetricLabel::CategoryOnly(key, category) => {
+                write!(f, "{key}{RECORD_SEPARATOR}{category}")
+            }
+            MetricLabel::KeyAndCategory(key, category) => {
+                write!(f, "{key}{RECORD_SEPARATOR}{category}")
+            }
```
chutten:
```suggestion
            MetricLabel::Static(label)
            | MetricLabel::Label(label) => write!(f, "{label}"),
            MetricLabel::KeyOnly(key, category)
            | MetricLabel::CategoryOnly(key, category)
            | MetricLabel::KeyAndCategory(key, category) => {
                write!(f, "{key}{RECORD_SEPARATOR}{category}")
            }
```

glean-core/src/internal_metrics.rs
```diff
@@ -361,11 +358,27 @@ impl UploadMetrics {
...
+
+    /// The number of metrics stored in SQLite after a migration run
+    pub metrics_in_sqlite: CounterMetric,
+
+    /// Number of metrics that failed to deserialize from storage
+    /// while iterating the Rkv database for migration.
+    pub failed_metrics: CounterMetric,
+
+    /// The duration for one full migration run at startup
+    pub migration_duration: TimespanMetric,
```
chutten: May want to change this to `timing_distribution` as the other state and error metrics account for multiple migrations in the same ping

glean-core/tests/sqlite.rs
```diff
@@ -0,0 +1,216 @@
...
+    let temp = {
+        let (glean, temp) = new_glean(None);
+        drop(glean);
+        temp
+    };
+
+    {
+        let path = temp.path().join("db").join("glean.sqlite");
+        fs::remove_file(&path).unwrap();
+        fs::write(&path, "not sqlite").unwrap();
```
chutten, line 48: (not a bug): I suppose if Magritte's PR is ever merged to make "not sqlite" a valid sqlite db, this test will fail until we find a new sentinel...

glean-core/tests/sqlite.rs
```diff
@@ -0,0 +1,216 @@
...
+        fs::remove_file(&path).unwrap();
+
+        let conn = rusqlite::Connection::open(path).unwrap();
+        conn.execute("CREATE TABLE telemetry (a TEXT)", ()).unwrap();
+    }
+
+    let (glean, _temp) = new_glean(Some(temp));
+
+    let client_id = clientid_metric().get_value(&glean, None);
+    assert!(client_id.is_some());
```
chutten, line 76: Is there an error/loadstate metric we could also validate here?

glean-core/tests/sqlite.rs
```diff
@@ -0,0 +1,216 @@
...
+        let path = temp.path().join("db").join("glean.sqlite");
+        let conn = rusqlite::Connection::open(path).unwrap();
+        conn.execute("DROP TABLE telemetry", ()).unwrap();
+        conn.execute("CREATE TABLE telemetry (a TEXT)", ()).unwrap();
+    }
+
+    let (glean, _temp) = new_glean(Some(temp));
+
+    let client_id = clientid_metric().get_value(&glean, None);
+    assert!(client_id.is_some());
```
chutten, line 101: ditto, can we check an error metric or something

glean-core/tests/sqlite.rs
```diff
@@ -0,0 +1,216 @@
...
+            session_mode: glean_core::SessionMode::Auto,
+            session_sample_rate: 1.0,
+            session_inactivity_timeout_ms: 1_800_000,
+        };
+        let glean = Glean::new(cfg);
+        assert!(glean.is_err());
+    }
+}
+
+// TODO:
```
chutten: TODO should either be resolved or a follow-up bug filed and referenced

chutten (approved):
r+wc

Possible later augmentation: on filesystem errors creating the db, go in-memory (https://sqlite.org/inmemorydb.html) so we're still able to report the failure.

badboy:
> on filesystem errors creating the db, go in-memory

https://bugzilla.mozilla.org/show_bug.cgi?id=2049284

glean-core/rlb/src/private/object.rs
badboy: make it work, make it right, make it fast.
(but also: it's for the profiler, so already not the default use case)

glean-core/src/database/sqlite/connection.rs
badboy: That's copy-pasta from [the skv bits](https://searchfox.org/firefox-main/rev/c280a3de79038960073359b0b4777c676320b76a/third_party/skv/src/connection.rs#116) and I have in fact not thought about it deeply.

I don't think we should/need to support downgrades.
We _do_ have to be careful with any schema changes in the future though. And yes, that should be documented.

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=2049289

PR #3508, opened 2026-06-19
Keep `rkv_load_error` active while Rkv-powered Glean is still in use

badboy:
This just restores what was previously there.

### DIFF
--- a/glean-core/metrics.yaml
+++ b/glean-core/metrics.yaml
@@ -955,6 +955,23 @@ glean.database:
       - glean-team@mozilla.com
     expires: never
 
+  rkv_load_error:
+    type: string
+    description: |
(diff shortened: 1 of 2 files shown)


chutten (approved):
This isn't _new_ collection, is it? (If so, you know the Data Review drill). This is just undoing a prior removal?

badboy:
> This isn't _new_ collection, is it? (If so, you know the Data Review drill). This is just undoing a prior removal?

Correct. I replaced it with `glean.database.load_error` in 6bc4e3a8177, but I kinda forgot that this will mark it as gone in the dictionary once we land it in `main`.
While we haven't migrated fully over we should keep it in-tree so that it is marked as active.

PR #3512, opened 2026-06-22
bug 2049040 - Accept 0-duration samples in timing_distribution metrics

chutten:
They go into the underflow bucket.

This means a sample of 1 becomes 1x (where x is the time_unit), a duration of 1ns is still promoted to 1x (where x is the time_unit), but a sample of 0 or a duration of 0 remains 0 to count in the 0 (underflow) bucket.

(( draft for now because I haven't 1) Added more tests for 0 values for all APIs of all the metrics to make sure our behaviours aren't surprising and don't change out from under us, or 2) thought deeply about the ramifications of reporting counts of zeroes in their special zero buckets. ))

Alternatives not fully considered:
* Record it as a zero for the purposes of `sum`, but put it in the bucket for 1x (where x is time_unit) like the rest of the < 1x values

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,6 +2,9 @@
 
 [Full changelog](https://github.com/mozilla/glean/compare/v69.0.0...main)
 
+* General
+  * Accept 0-duration samples in timing_distribution metrics ([bug 2049040](https://bugzilla.mozilla.org/show_bug.cgi?id=2049040))
+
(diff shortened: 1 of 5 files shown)


chutten:
Another alternative not considered: treating a sample of `0` as zero, but treating durations of 0 as 1`x`. Would limit the code changes necessary while also fixing the issue I had in Desktop.

PR #3513, opened 2026-06-23
Bug 2049295 - Validate that the database connection on open is usable for reads and writes

badboy:
This handles the case where, for whatever reason, the database is actually locked externally and no transaction can be run against it.

This enables the `database_externally_locked` test.

### DIFF
--- a/glean-core/src/error.rs
+++ b/glean-core/src/error.rs
@@ -9,7 +9,7 @@ use std::result;
 
 use rkv::StoreError;
 
-use crate::database::sqlite::SchemaError;
+use crate::database::sqlite::{OpenError, SchemaError};
 
 /// A specialized [`Result`] type for this crate's operations.
 ///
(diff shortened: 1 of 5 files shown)


glean-core/src/error.rs
```diff
@@ -173,15 +173,18 @@ impl From<rusqlite::Error> for Error {
...
+    fn from(error: OpenError) -> Error {
         match error {
-            SchemaError::Sqlite(err) => Error {
-                kind: ErrorKind::SQLite(err),
+            OpenError::IncompatibleVersion(v) => Error {
+                kind: ErrorKind::Schema(SchemaError::UnsupportedSchemaVersion(v)),
             },
-            err => Error {
-                kind: ErrorKind::Schema(err),
+            // TODO
```
chutten: TODO should have a bug number

chutten (approved):

glean-core/src/error.rs
badboy: copy-pasta. That's already handled in the construction of the error.

PR #3524, opened 2026-06-29
Bug 2048978 - docs: Specify that a record separator cannot be used in labels

badboy:
[doc only]

### DIFF
--- a/docs/user/_includes/label-limits.md
+++ b/docs/user/_includes/label-limits.md
@@ -3,7 +3,7 @@
 * The list of labels is limited to:
   * 16 different dynamic labels if no static labels are defined.
     Additional labels will all record to the special label `__other__`.
-    These labels may contain any UTF-8 characters.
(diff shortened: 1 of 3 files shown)


docs/user/_includes/label-limits.md
```diff
@@ -3,7 +3,7 @@
 * The list of labels is limited to:
   * 16 different dynamic labels if no static labels are defined.
     Additional labels will all record to the special label `__other__`.
-    These labels may contain any UTF-8 characters.
+    These labels may contain any UTF-8 characters except a record separator (`0x1E`).
```
travis79, line 6: Please tell me someone isn't using unprintable ASCII in labels....

travis79 (approved):

docs/user/_includes/label-limits.md
badboy: ~~they can't!~~
well, they can. Some use things like addon IDs, which are certainly more random (but at least not _arbitrary_ characters either)

PR #3533, opened 2026-07-07
Remove documentation about Qt/QML integration

badboy:
It's been officially removed in 2023: https://github.com/mozilla/glean.js/pull/1810

[doc only]


Dexterp37 (approved):
Docs were still around as the VPN team was using this. However, this seems to no longer be the case!

PR #3534, opened 2026-07-09
Bug 2053453 - glean-sym: Implement the EventMetric

badboy:
This is the bare minimum.
This basically duplicates what the RLB does, without using the traits or other niceties.
But it does work in the test.


glean-core/glean-sym/src/types.rs
```diff
@@ -136,3 +138,75 @@ pub enum HistogramType {
...
 
 pub type CowString = std::borrow::Cow<'static, str>;
+
+pub trait ExtraKeys {
+    /// List of allowed extra keys as strings.
+    const ALLOWED_KEYS: &'static [&'static str];
+
+    /// Convert the event extras into 2 lists:
+    ///
+    /// 1. The list of extra key indices.
```
chutten: It's not indices, it's the keys themselves

samples/glean-sym-test/app.py
```diff
@@ -15,37 +15,49 @@ def library_name(name):
...
+        amount = 31
+        services_record(amount)
+
+        xul.submit()
+        xul.shutdown()
+
+        # Check that
+        # * We submitted one ping only
+        # * It's the `prototype` ping
+        # * It contains 2 metrics with the expected values
```
chutten: More than 2 now

chutten (approved):
r+wc

PR #3540, opened 2026-07-15
Bug 2055182 - Run database maintenance after migration

badboy:
Based on the same maintenance tasks appservices runs on the places DB


travis79 (approved):
I guess this would be hard to cover via a test 🤔, oh well, it isn't probably a likely failure point so it should be okay.

travis79 (approved):

glean-core/tests/sqlite_migration.rs
```diff
@@ -218,3 +225,39 @@ fn migration_fails() {
...
+    drop(glean);
+
+    let db_file = db_path.join("glean.sqlite");
+    let db_file_size = fs::metadata(db_file).unwrap().size();
+
+    // Unvacuumed the database is _smaller_, because the migrated data is in the WAL file.
+    // It's around 20k bytes.
+    // Vacuumed & checkpointed the WAL transactions are merged into the database.
+    let vacuumed_database_size = 32768;
+    assert_eq!(db_file_size, vacuumed_database_size);
```
badboy: to be honest I don't like this test. It's way too fragile and will break if we add any other migration metrics.

travis79: 🤔 Yeah, the 32768 here seems magically dangerous. What if you just check to see if it is bigger and didn't use the comparison to a magic/const/whatever? Just enough to show that the transactions have been applied to the database?

badboy: We don't really have anything to check it against.

PR #3541, opened 2026-07-15
Bug 2055209 - Implement glean-noop as a feature of glean-sym

badboy:
Needs _some_ testing. We can probably extend the glean-sym sample test.
All we need to test is that it doesn't crash, doesn't call Glean, doesn't set metrics.


badboy:
Testing added.

travis79 (approved):

PR #3549, opened 2026-07-22
Bug 2049317 - Update to use zig 0.16 and maturin 1.15 for packaging additional Python wheels

badboy:


badboy:
one issue filed as https://github.com/PyO3/maturin/issues/3274

badboy:
~~I have a test run of this on Windows ARM as provided by GitHub actions.~~
~~CI code is here: https://github.com/fnordig/test-repo/blob/main/.github/workflows/build-test-deploy.yml#L94-L131~~
~~Latest run: https://github.com/fnordig/test-repo/actions/runs/30012410646/job/89223629709~~

This is currently indeed broken because of the upstream issue I filed, fixed and which got merged:
The definition file exports _only_ the `PyInit_glean._uniffi` symbol, hiding all others, most notably `UNIFFI_META_NAMESPACE_GLEAN` which uniffi-bindgen uses to generate the bindings. We'll have to wait for a release.

badboy:
I now [have a test](https://github.com/fnordig/test-repo/blob/12774cb39c195b01330ad32f187a740886a73634/.github/workflows/build-test-deploy.yml#L90-L173) run using upstream maturin (from git), with my patch: https://github.com/fnordig/test-repo/actions/runs/30039673820

This now looks good.

badboy:
Turns out that whole `__chkstk` thing isn't needed. The issue was that maturin was hiding all those symbols. With the upstream fix all symbols are visible again and thus linking works.

Note: still requires a new maturin release first.

PR #3562, opened 2026-07-29
iosbuild: Add a parameter to skip xcpretty to get verbose xcodebuild output

badboy:
xcpretty will hide the log output, which occasionally is useful when debugging. In this case skipping the prettifying of the output makes it easier to see what's going on.

### DIFF
--- a/bin/iosbuild
+++ b/bin/iosbuild
@@ -14,6 +14,11 @@ DESTINATION="platform=iOS Simulator,name=iPhone 17"
 
 cmd="${1:-}"
 target="${2:-sdk}"
+xcpretty=xcpretty
+
+if [[ "$3" == "--verbose" ]]; then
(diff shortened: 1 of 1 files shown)


chutten:
(reassigning as I don't know xcode enough or bash enough to rule on this)

bin/iosbuild
```diff
@@ -14,6 +14,11 @@ DESTINATION="platform=iOS Simulator,name=iPhone 17"
 
 cmd="${1:-}"
 target="${2:-sdk}"
+xcpretty=xcpretty
+
+if [[ "$3" == "--verbose" ]]; then
+    xcpretty=cat
```
travis79, line 20: a...   _pretty cat_. 🤩

travis79 (approved):

PR #3564, opened 2026-07-31
Bug 2059503 - Ensure errors are not put into the client_info or any other internal section

badboy:


glean-core/rlb/tests/schema.rs
```diff
@@ -174,6 +175,16 @@ fn validate_against_schema() {
...
 
+    // string JUST outside the limits, forcing this to record an error.
+    let term = "a".repeat(255) + "b";
+    glean::update_attribution(AttributionMetrics {
+        source: Some("rlb-tests".into()),
+        medium: Some("cargo-test".into()),
+        campaign: Some("testing".into()),
+        term: Some(term),
+        content: None,
+    });
```
jeddai, line 186: Does this subsequently test that the error doesn't show up in the client_info or other sections?

badboy: yup! This test specifically fills a ping with data and then checks the resulting payload against the (bundled) ping schema.
The test with this addtions fails without the fix from this PR

jeddai (approved):

PR #3565, opened 2026-07-31
Python: Remove any use of setuptools & ensure we use an updated pip

badboy:


badboy:
I have the hopes that this maybe fixes the intermittent issues we've seen

glean-core/build/src/lib.rs
```diff
@@ -137,9 +137,10 @@ impl Builder {
         } else {
             let venv = VirtualEnv::new(&sh, "py3-glean_parser")?;
 
+            // Ensure pip is the latest version
+            venv.pip_upgrade("pip")?;
+
             let glean_parser = format!("glean_parser~={GLEAN_PARSER_VERSION}");
-            // TODO: Remove after we switched glean_parser away from legacy setup.py
```
chutten, line 141: So, uh, _have_ we switched glean_parser away from legacy setup.py?

Oh, is this referencing mozilla/glean_parser#340 ... _from 2021_?!

Cargo.toml
```diff
@@ -37,3 +37,6 @@ opt-level = 3
 
 [profile.dev.package.flate2]
 opt-level = 3
+
+[patch.crates-io]
+xshell-venv = { git = "https://github.com/badboy/xshell-venv", branch = "venv-in-out-dir" }
```
chutten: Uh, what's this xshell-venv stuff doing in here?

glean-core/build/src/lib.rs
badboy: ... 🙈

Cargo.toml
badboy: testing! (gone in a second)

chutten (approved):

PR #3566, opened 2026-07-31
Bug 2046158 - Update to xshell-venv to always use `OUT_DIR` for the venv now

badboy:


travis79 (approved):

travis79:
Does just updating to the latest version force the `OUT_DIR`?

badboy:
> Does just updating to the latest version force the `OUT_DIR`?

yup, I landed [this commit](https://github.com/badboy/xshell-venv/commit/810fb460af96420597bfc9fe407c1e2c6ed943c0) and released it as 2.0

PR #3572, opened 2026-08-04
Bug 1652528 - Implement the custom distribution for iOS

badboy:


badboy:
This is not a breaking change and thus we can land this now separate from #3573.

samples/ios/app/metrics.yaml
```diff
@@ -265,3 +265,39 @@ party:
           type: array
           items:
             type: number
+
+# These remain unused in the sample app,
+# but it's a good way to ensure the code that is generated for them compiles.
+codegen_test:
```
travis79, line 271: 😢 what, no 99 luftballon metric?

travis79 (approved):

PR #3575, opened 2026-08-04
Bug 2060553: Add submitted_pings database table and add get/store methods

jeddai:
Based on #3545, will be rebased when that merges.

### DIFF
--- a/glean-core/src/database/sqlite.rs
+++ b/glean-core/src/database/sqlite.rs
@@ -238,6 +238,11 @@ impl Database {
             }
         }
 
+        db.conn.write(|tx| {
+            tx.execute("INSERT INTO migration (id, state) VALUES (1, 'done') ON CONFLICT(id) DO UPDATE SET state = excluded.state", [])?;
+            Ok::<(), rusqlite::Error>(())
+        })?;
+
         Ok(db)
     }
 
--- a/glean-core/src/database/sqlite.rs
+++ b/glean-core/src/database/sqlite.rs
@@ -2,6 +2,17 @@
 // License, v. 2.0. If a copy of the MPL was not distributed with this
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
+use chrono::{DateTime, FixedOffset};
+use connection::Connection;
+use malloc_size_of::MallocSizeOf;
+use rusqlite::fallible_iterator::FallibleIterator;
+use rusqlite::types::{FromSql, FromSqlError, FromSqlResult, ToSqlOutput, ValueRef};
+use rusqlite::OptionalExtension;
+use rusqlite::Transaction;
+use rusqlite::{params, ToSql};
+use rusqlite::{Error as SqlError, ErrorCode};
+use schema::Schema;
+pub use schema::SchemaError;
 use std::fmt::{self, Display};
 use std::fs;
 use std::num::NonZeroU64;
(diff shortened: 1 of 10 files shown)


glean-core/src/database/sqlite.rs
```diff
@@ -2,31 +2,31 @@
...
-
-use connection::Connection;
 use schema::Schema;
 pub use schema::SchemaError;
+use std::fmt::{self, Display};
+use std::fs;
+use std::num::NonZeroU64;
+use std::path::Path;
+use std::str;
+use std::time::Duration;
```
badboy: We usually keep the `std` imports on top, separated by a newline from crate uses.

glean-core/src/database/sqlite/schema.rs
```diff
@@ -62,6 +62,14 @@ impl ConnectionOpener for Schema {
                UNIQUE(id, ping, labels)
              );
              CREATE TABLE migration(id INTEGER PRIMARY KEY, state TEXT NOT NULL);
+             CREATE TABLE submitted_pings(
+               document_id TEXT PRIMARY KEY,
+               ping TEXT NOT NULL,
+               date_submitted DATETIME NOT NULL,
+               date_uploaded DATETIME,
+               value BLOB
```
badboy: We usually call this `payload`

glean-core/src/database/sqlite/schema.rs
```diff
@@ -92,6 +100,22 @@ impl ConnectionOpener for Schema {
...
+                // Clients upgrading to schema 3 don't have the table or index
+                tx.execute_batch(
+                    "CREATE TABLE submitted_pings(
+                           document_id TEXT PRIMARY KEY,
+                           ping TEXT NOT NULL,
+                           date_submitted TIMESTAMP NOT NULL,
+                           date_uploaded TIMESTAMP,
+                           value BLOB
+                         );
+                         CREATE INDEX submitted_pings_ping on submitted_pings(ping);",
```
badboy: Indentation here is a bit weird (admittedly I also messed that up in my other patch).
I guess I don't care super much about the exact indentation, as long as: it's either 2 or 4 spaces & the `CREATE INDEX` is aligned with the `CREATE TABLE`)


```suggestion
                    "
                    CREATE TABLE submitted_pings(
                      document_id TEXT PRIMARY KEY,
                      ping TEXT NOT NULL,
                      date_submitted TIMESTAMP NOT NULL,
                      date_uploaded TIMESTAMP,
                      value BLOB
                    );
                    CREATE INDEX submitted_pings_ping on submitted_pings(ping);
                    ",
```

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +500,106 @@ impl Database {
             .unwrap_or(false)
     }
 
+    pub fn get_all_submitted_pings(&self) -> Vec<SubmittedPing> {
```
badboy, line 504: Hmm, do we have plans to ever trim this table? (Doesn't have to happen here, but we should probably file a bug. We don't want an ever-growing database)

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +500,106 @@ impl Database {
...
+                            uploaded_date: r.get(3).unwrap(),
+                            value: r.get(4).unwrap(),
+                        })
+                    })
+                    .collect()
+            })
+            .unwrap_or_default()
+    }
+
+    pub fn store_submitted_ping(
```
badboy, line 576: I assume this will later be called from the actual uploader?

badboy (commented):
Did a first pass. No major blockers I think. I want to compare it to my prototype that I have somewhere (or at least what I put in my document).

glean-core/src/database/sqlite.rs
jeddai: There are plans to trim the table to only pings that were submitted in the last 30 days. I'll get that bug filed

glean-core/src/database/sqlite.rs
jeddai: Precisely!

glean-core/src/database/sqlite/schema.rs
jeddai: I think `fmt` did that, I'll get that fixed.

PR #3576, opened 2026-08-06
Swift: Test to verify accumulateSamples & accumulateSingleSample work for timing distributions

badboy:
The docs said accumulateSamples and accumulateSingleSample didn't exist on Swift, but that's not true.
Now verified by a test and updated docs.

While I'm there I also reordered `cancel` above everything else, which seemed much more logical and is also how it's ordered for labeled timing distributions.

### DIFF
--- a/glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
+++ b/glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
@@ -202,4 +202,27 @@ class TimingDistributionTypeTests: XCTestCase {
 
         XCTAssertNil(metric.testGetValue())
     }
+
+    func testTiminingDistributionAccumulatesSamples() {
+        let metric = TimingDistributionMetricType(CommonMetricData(
+            category: "telemetry",
+            name: "timing_distribution",
+            sendInPings: ["store1"],
+            lifetime: .ping,
+            disabled: false
+        ), .nanosecond)
+
+        XCTAssertNil(metric.testGetValue())
+
+        // Accumulate a few values
+        metric.accumulateSamples([1, 2])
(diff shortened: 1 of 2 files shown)


glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
```diff
@@ -202,4 +202,27 @@ class TimingDistributionTypeTests: XCTestCase {
...
+        ), .nanosecond)
+
+        XCTAssertNil(metric.testGetValue())
+
+        // Accumulate a few values
+        metric.accumulateSamples([1, 2])
+        metric.accumulateSingleSample(3)
+
+        // Check that data was properly recorded.
+        // We can only check the count, as we don't control the time.
```
chutten: Wrong comment, I think?

glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
```diff
@@ -202,4 +202,27 @@ class TimingDistributionTypeTests: XCTestCase {
...
+        let snapshot = metric.testGetValue()!
+        let sum = snapshot.values.values.reduce(0, +)
+        XCTAssertEqual(3, sum)
```
chutten: The snapshot ought to have a `sum` in it. And the sum of 1, 2, and 3 should be 6?

chutten (changes requested):

glean-core/ios/GleanTests/Metrics/TimingDistributionMetricTests.swift
badboy: oh hey, that's on me for copy-pasta.

chutten (approved):

PR #3577, opened 2026-08-06
Bug 2037387 - Wire mozbuild through to link against sqlite

badboy:


Makefile
```diff
@@ -95,7 +95,7 @@ test-python: build-python ## Run all Python tests
 lint: lint-rust lint-kotlin lint-swift lint-yaml lint-python
 
 lint-rust: ## Run cargo-clippy to lint Rust code
-	cargo clippy --all --all-targets --all-features -- -D warnings -A unknown-lints
+	cargo clippy --all --all-targets -- -D warnings -A unknown-lints
```
chutten: Is there a way to run it on the prior subset of features?

chutten (approved):

Makefile
badboy: yeah we can list them out I guess. Will do.

PR #3578, opened 2026-08-07
Typo fix

badboy:
[doc only]


travis79 (approved):
I guess the spellcheck task doesn't check the Makefile?

PR #3579, opened 2026-08-07
sqlite: Simplify read "transactions"

badboy:
See https://sqlite.org/lang_transaction.html for details on transactions in SQLite.

Any single SQL statement is automatically a transaction. It gets upgraded if it's a write, otherwise it's a read transaction.

Now rusqlite, on top of it, automatically rolls back `Transactions` unless `commit()` is called. It's doing that by executing `ROLLBACK`. Even if this transaction only ever did reads it does that. And even if that then doesn't really have to do much it does more than necessary.

We can skip all that.
The connection is already behind a lock, so we already don't run into concurrency issues on that access.

### DIFF
--- a/glean-core/src/common_metric_data.rs
+++ b/glean-core/src/common_metric_data.rs
@@ -243,7 +243,7 @@ impl CommonMetricDataInternal {
     /// No error is recorded in the database if the check fails.
     /// Extract the validated label, if any, using [`LabelCheck::label`].
     /// Record an error, if any, using [`LabelCheck::record_error`].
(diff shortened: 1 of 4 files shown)


travis79 (approved):
I'm all for simplifying things, but let's please document this decision somewhere including "why we did this" and what the impact is.

badboy:
> I'm all for simplifying things, but let's please document this decision somewhere including "why we did this" and what the impact is.

It's documented in the commit message! (and of course GitHub is exceptionally good at hiding it): https://github.com/mozilla/glean/pull/3579/changes/0fd5e68344c055061108a34e1d8053c804a7e0bc

PR #3580, opened 2026-08-07
bug 2060890 - Early Event Acceleration

chutten:
Add a new configuration property and Server Knob to accelerate the first few "events" pings each session by decreasing the number of `event` records needed to trigger a `max_capacity`-reason "events" ping submission.

### DIFF
--- a/glean-core/src/event_database/mod.rs
+++ b/glean-core/src/event_database/mod.rs
@@ -10,7 +10,7 @@ use std::io::BufReader;
 use std::io::Write;
 use std::io::{self, BufRead};
 use std::path::{Path, PathBuf};
-use std::sync::{Arc, Mutex, RwLock};
+use std::sync::{atomic, Arc, Mutex, RwLock};
 use std::{fs, mem};
 
 use chrono::{DateTime, FixedOffset, Utc};
@@ -111,6 +111,9 @@ pub struct EventDatabase {
     event_store_files: RwLock<HashMap<String, Arc<File>>>,
     /// A lock to be held when doing operations on the filesystem
     file_lock: Mutex<()>,
+    /// How many "events" pings have been submitted,
+    /// as estimated from how often the "events" store is snapshotted and cleared.
+    events_pings_submitted: atomic::AtomicUsize,
 }
 
 impl MallocSizeOf for EventDatabase {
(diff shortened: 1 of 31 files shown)


glean-core/src/lib.rs
```diff
@@ -181,6 +181,8 @@ pub struct InternalConfiguration {
     /// Inactivity timeout in milliseconds for AUTO mode before a new session starts.
     /// Default: 1 800 000 ms (30 minutes).
     pub session_inactivity_timeout_ms: u64,
+    /// The number of "events" pings to accelerate each session, plus one.
```
badboy, line 185: Why this `plus one`?

glean-core/src/event_database/mod.rs
```diff
@@ -351,8 +361,24 @@ impl EventDatabase {
                 let event_json = serde_json::to_string(&event).unwrap(); // safe unwrap, event can always be serialized
                 store.push(event);
                 self.write_event_to_disk(store_name, &event_json);
-                if store_name == "events" && store.len() == glean.get_max_events() {
-                    submit_max_capacity_event_ping = true;
+                if store_name == "events" {
+                    if store.len() == glean.get_max_events() {
```
badboy, line 365: This `==` is https://bugzilla.mozilla.org/show_bug.cgi?id=2061479, right?

badboy (changes requested):

glean-core/src/lib.rs
chutten: That's how the math works. `factor = pings + 1` because the derivation of `a` ( `max_events = a * factor^2` ) means that when `factor = ping`, there is no acceleration left, because the limit is again `max_events`.

Or, concretely, for `max_events = 500` and `factor = 5` (our recommendations), the fifth ping has a threshold of `500` (`y = 20 * 5^2 = 500`) ([ref](https://docs.google.com/document/d/1_pcsUPvus72Xumu-bh1Ir4IO_dNLpA83syZKaDdQcHc/edit?tab=t.0#heading=h.c9px9pyb6bta))

glean-core/src/event_database/mod.rs
chutten: Correctamundo. I'd have changed it here but I haven't double-checked that nothing bad will happen.

badboy (approved):
Needs a rebase for the changelog entry, but then I guess good to go (it's not active until the configuration is set anyway, so it's safe to land now)

PR #3584, opened 2026-08-11
Close the database on shutdown

badboy:
This ensure we actually close the database connection, which is a good idea to do when we can.
Renamed the function to reflect better what it is doing.


chutten (approved):
Does it need a CHANGELOG? Not sure it does.

PR #3585, opened 2026-08-11
Add submitted pings table

jeddai:
Re-opening from https://github.com/mozilla/glean/pull/3575


jeddai:
Some changes:
- Updated the PR to not require the `time` module for rusqlite by using Integers to store the dates
- Added the mechanism to actually store the pings when they're submitted (still need to test this)

glean-core/src/database/sqlite/schema.rs
```diff
@@ -90,6 +98,24 @@ impl ConnectionOpener for Schema {
...
+                        document_id TEXT PRIMARY KEY,
+                        ping TEXT NOT NULL,
+                        date_submitted INTEGER NOT NULL,
+                        date_uploaded INTEGER,
+                        payload BLOB
+                    );
+                    CREATE INDEX submitted_pings_ping on submitted_pings(ping);
+                    ",
+                )?;
+                tx.execute("INSERT INTO migration (id, state) VALUES (2, 'done') ON CONFLICT(id) DO UPDATE SET state = excluded.state", [])?;
```
badboy:
```suggestion
```

This shouldn't be needed here.

glean-core/src/database/sqlite.rs
```diff
@@ -72,6 +73,41 @@ impl MallocSizeOf for Database {
     }
 }
 
+#[allow(unused)]
+pub struct SubmittedPing {
+    pub document_id: String,
+    pub ping: String,
+    pub submitted_date: SqliteDatetime,
+    pub uploaded_date: Option<SqliteDatetime>,
+    pub value: Option<Vec<u8>>,
```
badboy:
```suggestion
    pub payload: Option<Vec<u8>>,
```

Probably best to stay consistent between code and the SQL table and call it `payload` everywhere

glean-core/src/database/sqlite.rs
```diff
@@ -72,6 +73,41 @@ impl MallocSizeOf for Database {
...
+}
+
+impl SubmittedPing {
+    pub fn value(&self) -> Option<JsonValue> {
+        self.value
+            .as_ref()
+            .map(|v| rmp_serde::from_slice(v).expect("IMPOSSIBLE: Deserializing value failed"))
```
badboy:
```suggestion
    pub fn payload(&self) -> Option<JsonValue> {
        self.payload
            .as_ref()
            .map(|v| rmp_serde::from_slice(v).expect("IMPOSSIBLE: Deserializing value failed"))
```

I wonder if we should be a bit more defensive here and not `.expect` that deserialization works? Afterall we're handling a rather arbitrary byte array.
We should only store valid things, but who knows what breaks in between.

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +484,195 @@ impl Database {
...
+            VALUES
+                (?1, ?2, ?3, ?4, ?5)
+            ON CONFLICT(document_id) DO UPDATE SET
+                ping = excluded.ping,
+                date_submitted = excluded.date_submitted,
+                date_uploaded = excluded.date_uploaded,
+                payload = excluded.payload
+            "#;
+            let mut stmt = tx.prepare_cached(insert_sql)?;
+            let encoded = rmp_serde::to_vec(&value).expect("IMPOSSIBLE: Serializing metric failed");
```
badboy: Is there a benefit for storing this message-packed here? Could we just store the raw json?

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +484,195 @@ impl Database {
...
+            })
+            .unwrap_or_default()
+    }
+
+    /// Returns all submitted pings in the `submitted_pings` table that match a supplied ping name.
+    ///
+    /// # Arguments
+    ///
+    /// * `ping` - The name of the pings to return.
+    pub fn get_submitted_pings(&self, ping: &str) -> Vec<SubmittedPing> {
```
badboy: Hm, what do you think about naming this differently to make it clear that it's only fetching the pings of a certain name?

`get_submitted_pings_by_name` or similar (not particular happy with that one).

badboy: Is this function actually used anywhere?
I only see a single use in a test, but it's not reachable from the outside, so no integration could even call this.

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +484,195 @@ impl Database {
...
+    ///
+    /// # Returns
+    ///
+    /// A `usize` representing the number of rows updated.
+    pub fn mark_ping_as_uploaded(&self, document_id: &str, date_uploaded: DateTime<Utc>) -> usize {
+        let update_submitted_pings_sql = r#"
+        UPDATE submitted_pings
+        SET date_uploaded = ?1
+        WHERE document_id = ?2
+        "#;
```
badboy: This can probably be inline with a normal string.

```suggestion
        let update_submitted_pings_sql = "UPDATE submitted_pings SET date_uploaded = ?1 WHERE document_id = ?2";
```

(run `make fmt-rust` afterwards)

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +484,195 @@ impl Database {
...
+                ping,
+                SqliteDatetime(date_submitted),
+                date_uploaded.map(SqliteDatetime),
+                encoded
+            ])?;
+            Ok(())
+        })
+    }
+
+    /// Removes rows from the `submitted_pings` table where the `date_submitted` is older than the supplied date.
```
badboy:
```suggestion
    /// Remove stored submitted pings that are older than `before_time` (or 30 days if not specified)
```

glean-core/src/database/sqlite.rs
```diff
@@ -448,6 +484,195 @@ impl Database {
...
+    /// # Arguments
+    ///
+    /// * `before_time` - An optional date – when supplied uses that date as the oldest date_submitted we should keep.
+    ///     Defaults to 30 days if `None` is supplied.
+    ///
+    /// # Returns
+    ///
+    /// An empty `Result`.
+    pub fn cleanup_submitted_pings(&self, before_time: Option<DateTime<Utc>>) -> Result<()> {
+        let time = before_time.unwrap_or(Utc::now() - Duration::from_secs(2592000));
```
badboy: This way we don't have to re-guess what the magic number `2592000` means.

```suggestion
        let days_30 = Duration::from_secs(30 * 24 * 60 * 60); 
        let time = before_time.unwrap_or_else(|| Utc::now() - days_30);
```

(also using `unwrap_or_else` and a clouse so we save us the `Utc::now` call if not needed)

glean-core/src/metrics/ping.rs
```diff
@@ -353,6 +353,18 @@ impl PingType {
...
 
+                if glean.store_submitted_pings_enabled {
+                    if let Err(e) = glean.storage().store_submitted_ping(
+                        ping.doc_id,
+                        &self.0.name,
+                        Utc::now(),
+                        None,
+                        ping.content.clone(),
+                    ) {
+                        log::warn!("{}", e);
```
badboy: The error message here would be coming directly from SQLite. We should extend that to make it understandable (by future-us).


```suggestion
                        log::warn!("Storing a submitted ping failed: {e}");
```

badboy (changes requested):
this is starting to look pretty good (did a full review this time).
A few more things.

Right now getting those pings back from the outside is not implemented, but I assume that's gonna be done in a followup?

glean-core/src/database/sqlite.rs
jeddai: Ah I meant to include a method for it, same for `get_all_submitted_pings`. I'll fix that. 

And yeah, I think `get_submitted_pings_by_name` is at least better than not noting that it's by name.

CHANGELOG.md
```diff
@@ -22,8 +27,8 @@
     * Implement glean-noop as a feature of glean-sym ([#3541](https://github.com/mozilla/glean/pull/3541))
     * Support pings ([#3544](https://github.com/mozilla/glean/pull/3544))
     * Implement the event metric ([#3534](https://github.com/mozilla/glean/pull/3534))
-    * BREAKING CHANGE: Switch from a noop feature to an `active` feature ([#3583](https://github.com/mozilla/glean/pull/3583))
 * iOS
+    * BREAKING CHANGE: Switch from a noop feature to an `active` feature ([#3583](https://github.com/mozilla/glean/pull/3583))
```
badboy, line 31: This moving of the changelog line looks like a mistake.

glean-core/tests/sqlite.rs
```diff
@@ -3,9 +3,9 @@
 // file, You can obtain one at https://mozilla.org/MPL/2.0/.
 
 mod common;
-use std::fs;
-
 use crate::common::*;
+use chrono::Utc;
+use std::fs;
```
badboy, line 8: Let's try to stick to the grouping as before, which means the `use chrono::Utc` should be down on line 10, next to the `use glean_core`

glean-core/src/database/sqlite/schema.rs
```diff
@@ -53,15 +53,24 @@ impl ConnectionOpener for Schema {
...
+                value BLOB,
+                UNIQUE(id, ping, labels)
+            );
+            CREATE TABLE migration(id INTEGER PRIMARY KEY, state TEXT NOT NULL);
+            CREATE TABLE submitted_pings(
+                document_id TEXT PRIMARY KEY,
+                ping TEXT NOT NULL,
+                date_submitted INTEGER NOT NULL,
+                date_uploaded INTEGER,
+                upload_failed BOOLEAN NOT NULL,
```
badboy, line 70: A thing I'm only thinking about now: Should this field be the positive, so `upload_succeeded` instead?
Or should this even be a date? As in the last time the upload failed for this ping (null if upload succeeded, at which point `date_uploaded` is set to a proper time)

This comes down to what we want to do with this field and how much information we want to expose

badboy (commented):
No major blocker anymore. Couple of questions.

You can also rebase this branch on `main` and squash down the commits.

glean-core/src/database/sqlite/schema.rs
jeddai: I've updated it to be a date! The `date_uploaded` field is a positive signal

CHANGELOG.md
```diff
@@ -1,5 +1,10 @@
 # Unreleased changes
 
+* Add mechanisms for storing and retrieving submitted pings ([#3585](https://github.com/mozilla/glean/pull/3585)).
+    * Add new `submitted_pings` table to the SQLite database.
+    * Add methods to store, retrieve, update, and clear stored submitted pings.
+    * Update Ping and uploader implementations to store and update submitted pings as appropriate.
```
badboy: Hm, this seems to be in the wrong place now, but I can fix that up on merge.

badboy (approved):

badboy:
Rebased and fixed up the changelog, will merge when CI passes

PR #3588, opened 2026-08-14
Update Cargo.lock for the benchmarks

badboy:
That's always missed when we do the update. Oops.


chutten (approved):
Any thoughts on how to make this less easily forgettable?

badboy:
> Any thoughts on how to make this less easily forgettable?

We already build it on CI, so I added the `--locked` flag there, which will make CI fail if the Cargo.lock is not up-to-date.

PR #3590, opened 2026-08-17
Bug 2062257 - sqlite: Use an in-memory database when delay_ping_lifetime_io=true

badboy:
This restores `delay_ping_lifetime_io` behavior (not touching disk until `persist_ping_lifetime_data` is called, or in auto-mode when the counter/time is hitting the limit)

The local benchmark shows it's twice as fast compared to current `main` (so compared to always hitting disk).

---

~~Not yet ready for use. Should get some testing (and fixing at least one clippy lint).~~
~~Also some refactoring to not use the same string constant everywhere (and instead use a const for that)~~


glean-core/src/database/sqlite.rs
```diff
@@ -769,6 +987,67 @@ impl Database {
...
+            self.conn.write(|tx| tx.execute_one(persist_sql))?;
+
+            // We can reset the write-counter. Current data has been persisted.
+            self.ping_lifetime_count.store(0, Ordering::Release);
+            self.ping_lifetime_store_ts.replace(Instant::now());
+        }
+        Ok(())
+    }
+
+    fn persist_ping_lifetime_data_if_full(&self, tx: &mut Transaction) -> Result<()> {
```
badboy, line 966: Except for the actual `INSERT INTO` at the end this is exactly the same as the old implementation that was removed.

glean-core/src/database/sqlite.rs
```diff
@@ -63,6 +65,32 @@ pub struct Database {
 
     /// Set when a database migration attempt failed.
     pub(crate) migration_error: MigrationResult,
+
+    /// If the `delay_ping_lifetime_io` Glean config option is `true`,
+    /// we will save metrics with 'ping' lifetime data in a map temporarily
+    /// so as to persist them to disk using rkv in bulk on demand.
+    ping_lifetime_data: Option<()>,
```
badboy: this shouldn't be an `Option<()>` obviously. A bool would do. Need to fix that up.

glean-core/src/database/sqlite.rs
```diff
@@ -63,6 +65,32 @@ pub struct Database {
 
     /// Set when a database migration attempt failed.
     pub(crate) migration_error: MigrationResult,
+
+    /// If the `delay_ping_lifetime_io` Glean config option is `true`,
+    /// we will save metrics with 'ping' lifetime data in a map temporarily
+    /// so as to persist them to disk using rkv in bulk on demand.
```
badboy: Ooops, yeah, rkv is gone.

glean-core/rlb-tests/tests/it.rs
```diff
@@ -79,6 +79,10 @@ fn delayed_ping_data() {
     let mut lines = payload.lines();
     assert!(lines.any(|line| line.contains("\"url\":") && line.contains("/prototype/")));
     assert!(payload.contains("\"test.metrics.sample_counter\":10"));
+    assert!(
+        payload.contains("\"test.metrics.labeled_counter\":{\"label\":10}}}"),
+        "{payload}"
```
badboy, line 84: This prints the full payload in case the assert fails. Helps with debugging.

glean-core/src/database/sqlite/schema.rs
```diff
@@ -108,3 +108,30 @@ pub enum SchemaError {
...
+    table: &str,
+) -> Result<(), rusqlite::Error> {
+    tx.execute(
+        "ATTACH DATABASE ':memory:' AS :database",
+        named_params! {":database": database},
+    )?;
+    tx.execute(
+        &format!(
+            r#"
+             CREATE TABLE {database}.{table}(
```
badboy, line 127: Must be the same schema as the main table. I don't know yet how to best unify that.

glean-core/src/database/sqlite.rs
```diff
@@ -33,6 +36,22 @@ use super::ConnExt;
 mod connection;
 mod schema;
 
+const DEFAULT_TABLE: &str = "telemetry";
+const IN_MEMORY_DATABASE: &str = "lifetime_ping";
+const IN_MEMORY_TABLE: &str = "lifetime_ping.telemetry";
+
+#[test]
+fn consts_are_correct() {
```
badboy, line 44: Best way to ensure the above consts are actual substrings of each other. Can't do that in const (without external crates)

glean-core/src/database/sqlite.rs
```diff
@@ -306,6 +367,23 @@ impl Database {
...
+    fn load_ping_lifetime_data(&self) {
+        if !self.ping_lifetime_data {
+            return;
+        };
+
+        let copy_sql =
+            "INSERT INTO lifetime_ping.telemetry SELECT * FROM telemetry WHERE lifetime = 'ping'";
+        let res = self.conn.write(|tx| tx.execute_one(copy_sql));
+        if let Err(err) = res {
+            log::error!("Could not load ping lifetime data into memory: {err:?}");
```
badboy, line 382: Wonder if maybe we should disable `ping_lifetime_data` if we can't even load data into it?

glean-core/src/database/sqlite/schema.rs
chutten: Need to also sync with the doc in database.md that documents the schema.

Until we figure out a better idea, can do what I've seen in m-c: Put in a comment `// This must remain in sync with X found at Y.rs otherwise EBADTHINGS` and think about how the tests would react.

glean-core/src/database/sqlite/schema.rs
```diff
@@ -108,3 +108,30 @@ pub enum SchemaError {
...
+    database: &str,
+    table: &str,
+) -> Result<(), rusqlite::Error> {
+    tx.execute(
+        "ATTACH DATABASE ':memory:' AS :database",
+        named_params! {":database": database},
+    )?;
+    tx.execute(
+        &format!(
+            r#"
```
chutten: Why raw? `schema.rs` doesn't need it

glean-core/src/database/sqlite.rs
```diff
@@ -63,6 +82,32 @@ pub struct Database {
 
     /// Set when a database migration attempt failed.
     pub(crate) migration_error: MigrationResult,
+
+    /// If the `delay_ping_lifetime_io` Glean config option is `true`,
+    /// we will save metrics with 'ping' lifetime data in memory only,
+    /// and persist them to disk in bulk on demand.
+    ping_lifetime_data: bool,
```
chutten: You can fit all ping lifetime data into a single `bool` ?!?!

(( More seriously, maybe there's a better name here like `ping_lifetime_data_in_memory`. Or, heck, keep the `delay_ping_lifetime_io` naming. ))

glean-core/src/database/sqlite.rs
```diff
@@ -63,6 +82,32 @@ pub struct Database {
...
+
+    /// The last time the `lifetime=ping` data was flushed to disk.
+    ///
+    /// Data is flushed to disk automatically when the last flush was more than
+    /// `ping_lifetime_max_time` ago.
+    ///
+    /// Only relevant if `delay_ping_lifetime_io` is set to `true`,
+    ping_lifetime_store_ts: Cell<Instant>,
+
+    /// After what time to auto-flush. 0 disables it.
```
chutten:
```suggestion
    /// After what delay to auto-flush. 0 disables it.
```

glean-core/src/database/sqlite.rs
chutten: Good idea. We should also probably try to report an error (though how we'd do that without a db that can even use memory is a mystery)

glean-core/src/database/sqlite.rs
```diff
@@ -757,10 +901,36 @@ impl Database {
             if let Err(e) = res {
                 log::warn!("Could not clear store for all lifetimes: {:?}", e);
             }
+
+            if self.ping_lifetime_data {
+                let clear_sql =
+                    "DELETE FROM lifetime_ping.telemetry WHERE lifetime = ?1 OR lifetime = ?2 OR lifetime = ?3";
```
chutten: Uh, but it's literally only going to have one lifetime in it, no? I mean, I suppose we might change our minds about that later, but for now at least?

glean-core/src/database/sqlite.rs
```diff
@@ -769,6 +939,67 @@ impl Database {
...
+        let persist_sql = r#"
+            INSERT INTO telemetry SELECT * FROM lifetime_ping.telemetry WHERE true
+            ON CONFLICT(id, ping, labels) DO UPDATE SET
+              lifetime = excluded.lifetime,
+              value = excluded.value
+        "#;
+        tx.execute_one(persist_sql)?;
+
+        self.ping_lifetime_count.store(0, Ordering::Release);
+        self.ping_lifetime_store_ts.replace(Instant::now());
```
chutten: Ah, and it's because we already have a transaction that this doesn't just call `persist_ping_lifetime_data`. I suppose you already considered and decided against factoring this out to a helper that both would call?

chutten (approved):

glean-core/src/database/sqlite.rs
badboy: > You can fit all ping lifetime data into a single bool ?!?!

what is a lifetime if not a decision between yes and no &lt;/philosophical>

glean-core/src/database/sqlite.rs
badboy: Oh. yes.

PR #3606, opened 2026-08-20
Bug 2053775 - CI: Run Rust tests on GhA

badboy:
This is a start to what will be full CI on GitHub Actions.
We do not yet block on this!

---

We start with a decision taks right from the get go: We will run specific tasks only when the corresponding files have changed.
That will speed up CI significantly.
E.g. typo fixes in docs or Python-only files won't trigger a full Kotlin run, nor does it trigger a Rust rebuild.

With this first change however we only migrate the Rust tests, those only happen when *.rs files or Cargo files change


.github/workflows/test.yml
```diff
@@ -0,0 +1,96 @@
...
+        with:
+          toolchain: ${{ matrix.rust-version }}
+
+      - name: Install nextest
+        run: |
+          .github/workflows/crate-ci.sh \
+            --git nextest-rs/nextest\
+            --crate nextest \
+            --tag "cargo-nextest-0.9.140" \
+            --filename "cargo-nextest-0.9.140-universal-apple-darwin.tar.gz"
```
badboy: We won't need to install that many tools really, most things just come with the runners we use.
So for now I'll keep this inline, but should we need more we might want to handle this differently.

travis79 (approved):

PR #3611, opened 2026-08-27
Fix MetricsPingSchedulerTest failing between midnight and 4am

rvandermeulen:
`MetricsPingScheduler` reads "now" from `Calendar.getInstance()`, which Robolectric cannot shadow, while `DateUtils.isToday()` does see the clock set by `SystemClock.setCurrentTimeMillis()`. Tests that install a fake time therefore got a mix of the simulated and real clocks.

The visible effect is in `schedule()`: startup collection only happens when `isAfterDueTime(now)` is true, and `now` was the real time. Between midnight and `DUE_HOUR_OF_THE_DAY` (4am) local, the ping was scheduled for later in the day instead of collected immediately, so the mock web server never received it, `takeRequest` timed out after 20s and `!!` threw an NPE.

This affects `Data recorded before Glean inits must not get into overdue pings` and `Glean must preserve lifetime application metrics across runs`. It looks like a flake but is deterministic — it depends only on the local hour the suite runs at, which is a ~4h window per day for a UTC CI.

`getCalendarInstance()` already existed as a seam, but the instance Glean builds in `startMetricsPingScheduler()` is not reachable from tests, so this adds a test-only `clockOverride` that it consults. It is null in production, so there is no behavior change outside tests. `GleanTestRule` sets it alongside the `SystemClock` call it already makes and clears it in `finished()`.

Verified by running the full `:glean:testDebugUnitTest` suite plus `detekt` and `ktlint` at 00:24 and 00:28 local — inside the window where these tests failed at 00:02.


rvandermeulen:
For the record, this has been annoying me for AGES.

badboy (approved):

PR #3612, opened 2026-08-27
Update to Gradle 9.5.0

rvandermeulen:
Updates Gradle from 8.14.3 to 9.5.0, and moves off the APIs that AGP 9 removes while staying on AGP 8.

9.5.0 is as far as we can go for now: Gradle 9.6.0 removes `org.gradle.api.problems.internal.InternalProblems`, which AGP 8.x's `com.android.internal.library` depends on, and 8.13.2 is the last AGP 8 release. Going further requires the AGP 9 upgrade, which is deliberately not part of this.

**Breaking change for consumers of the Glean Gradle plugin.** Gradle 9 removed `Project.exec()`, which JetBrains' `gradle-python-envs` calls internally. That plugin is unmaintained, 0.0.31 is its only recent release, and there is no Gradle 9 compatible version, so the Miniconda bootstrap is gone. The plugin now always creates a virtual environment with the standard library `venv` module, which is the path that already existed for offline builds. Applications need a Python 3.9 or later interpreter on the `PATH`, or `GLEAN_PYTHON` set, unless they supply their own environment via `ext.gleanPythonEnvDir`. They no longer need to apply `com.jetbrains.python.envs` themselves.

Note that on Debian and Ubuntu this also means `python3-venv` has to be installed, since a bare `python3` there has no `ensurepip`. That is what the first commit does for our own Android CI jobs, and downstream consumers building on similar images will need the same. mozilla-central already sets `gleanPythonEnvDir` everywhere, so it is unaffected.

The rest of the stack prepares for AGP 9. Registering the uniffi bindings through `androidComponents` also let the lint baseline go away: the generated directory was being added to the Kotlin source set by hand, which is why lint treated the bindings as ordinary source and needed three `NewApi` suppressions.

Along the way this fixes three latent missing task dependencies, all of which had been masked by warm build directories and scheduling that happened to work:

- `copyGleanNative` could copy the native libraries before Cargo had built them
- the metrics generating tasks could run before the Python environment existed
- the Maven artifact check ran before `publish`, found nothing, and passed

The last one is worth a second look: `bin/check-artifact.sh` treated "no artifact found" as success, so the release check in `module-build` would have gone green while verifying nothing. It now fails instead of skipping whenever it has nothing to inspect.

These were all found while trying to enable the configuration cache, which is **not** part of this PR. That turned up a problem outside this repo: with the cache on, a consumer task with a correct `dependsOn` on `rust-android-gradle`'s cargo task still observes an empty output directory on a cold build, which suggests the plugin does not declare `rustJniLibs` in a way Gradle's parallel scheduler respects. The compatibility fixes are kept here since they are improvements in their own right, and enabling the cache can be revisited once that is sorted out.

Verified on a clean tree with a cold Python environment: full build, all three unit test variants, detekt, ktlint, offline builds, publishing, and the `module-build` and `android-build` CI task sequences.

Not covered: Windows, where the interpreter path changed from the conda layout to `Scripts/python`.


glean-core/android/src/test/java/mozilla/telemetry/glean/TestUtil.kt
```diff
@@ -62,7 +62,7 @@ internal fun checkPingSchema(content: JSONObject) {
     val os = System.getProperty("os.name")?.lowercase()
     val pythonExecutable =
         if (os?.indexOf("win")?.compareTo(0) == 0) {
-            "${BuildConfig.GLEAN_MINICONDA_DIR}/python"
+            "${BuildConfig.GLEAN_MINICONDA_DIR}/Scripts/python"
```
badboy: If we're not relying on miniconda anymore, should we rename this variable to reflect that (it just stores the `PYTHON_ENV_DIR` now)

glean-core/android/build.gradle
```diff
@@ -194,29 +170,46 @@ dependencies {
...
 ext.configurePublish()
 
-def generateUniffiBindings = tasks.register("generateUniffiBindings") {
+// Gradle 9 removed Project.exec().
+interface InjectedExecOps {
+    @Inject ExecOperations getExecOps()
+}
+def execOps = objects.newInstance(InjectedExecOps).execOps
+
+// AGP owns the output location, so the task exposes it as a wired property.
```
badboy: What's a "wired property"?

badboy (commented):
On a first review this looks reasonable. As said privately already I'm not a Gradle pro, so this might as well be just a shallow understanding of all that's going on.
But the build is what counts, so if this builds locally and on CI it seems reasonable ok to me.

Got some questions before we land this.

As also mentioned, we're in the process of landing Glean v70 on Desktop, so I'm trying to hold back on merging bigger things into `main` for the moment, in case we need other changes there (and patch releases otherwise are annoying). So this will land once that's done.

glean-core/android/src/test/java/mozilla/telemetry/glean/TestUtil.kt
rvandermeulen: Renamed to `GLEAN_PYTHON_ENV_DIR` (the BuildConfig field, its two uses, and the comment above it that was apologizing for the old name). While sweeping for the rest I also dropped the `gleanCondaDir` project property the plugin still set as an alias for `gleanPythonEnvDir`; nothing in mozilla-central or application-services reads it, and the CHANGELOG entry now mentions it. The only conda mentions left are the generic pipenv/conda pointer in the Python dev-setup docs and the dictionary words the CHANGELOG still needs.

glean-core/android/build.gradle
rvandermeulen: Reworded. It meant that AGP assigns the output location when the task is registered through `addGeneratedSourceDirectory`, so the task exposes a `DirectoryProperty` for AGP to fill in rather than choosing a path itself.

badboy (approved):

PR #3614, opened 2026-09-01
Bug 2053775 - CI: Run lints task on GitHub Actions

badboy:
This is the next step after #3606 

It's now running the lint tasks on GitHub Actions:

* doc lints
* yaml lints
* spelling
* licenses
* schema
* code linting (rust, python, kotlin, swift currently missing)


chutten (approved):
Though not the easiest docs to get into, I've now read enough of the GH Actions docs that I feel like I know what this does and can r+

badboy:
> Though not the easiest docs to get into, I've now read enough of the GH Actions docs that I feel like I know what this does and can r+

That's a good point actually, I should have collected and post all the material I based this on :/

PR #3622, opened 2026-09-07
Bug 2053775 - CI: Run the Rust sample & glean-sym sample

badboy:
Nearly missed that in the earlier Rust tests
~~(Currently based on top of #3620 so that it cleanly merges)~~


.github/workflows/test.yml
```diff
@@ -130,6 +130,9 @@ jobs:
         with:
           toolchain: ${{ matrix.rust-version }}
 
+      - name: Install the latest version of uv
+        uses: astral-sh/setup-uv@c771a70e6277c0a99b617c7a806ffedaca235ff9 # v9.0.0
```
chutten: It's not installing the latest version, though, is it.

chutten (approved):

.github/workflows/test.yml
badboy: eeeeh, why do they keep updating in between my commits?
It doesn't, I'll update it here (in the future the dependabot task will send us PRs for updates)

PR #3624, opened 2026-09-08
Bug 2067862 - Fix resetting Glean for tests not clearing dispatcher

travis79:


glean-core/src/lib.rs
```diff
@@ -1355,6 +1355,10 @@ pub fn glean_test_destroy_glean(clear_stores: bool, data_path: Option<String>) {
         // Allow us to go through initialization again.
         INITIALIZE_CALLED.store(false, Ordering::SeqCst);
     } else if clear_stores {
+        // Discard the pre-init queue, which still holds everything recorded so
+        // far and would be replayed by the next initialization (bug 2067862).
+        dispatcher::reset_dispatcher();
```
badboy, line 1360: Oh, yeah, so this is a tricky one I guess. As seen on CI this breaks some tests which _explicitly_ test this behavior:
Assuming Glean has not been initialized, we want to queue some early tasks before then initializing and thus having those earlier tasks applied.

Honestly, I don't know how to solve this right now.

badboy (changes requested):

glean-core/src/lib.rs
travis79: There _is_ a workaround...  if the test class that triggers this adds `testResetFOG` to the `add_setup` function or ensure it gets called and inits Glean before any metrics are recorded.

travis79:
Closing this, since we need to support the behaviour it removes.

PR #3632, opened 2026-09-10
CI: Run Python tasks on more changes

badboy:


travis79 (approved):
The Python 3.12 failure looks like it was just a transient network issue, so r+

PR #3635, opened 2026-09-10
Bug 2070884 - Disable ping lifetime IO delay when data can't be loaded into memory

badboy:
I can't come up with a reasonable test for this, mostly because I can't think of a case where we can't load data into memory.


glean-core/src/database/sqlite.rs
```diff
@@ -378,8 +378,8 @@ impl Database {
             "INSERT INTO lifetime_ping.telemetry SELECT * FROM telemetry WHERE lifetime = 'ping'";
         let res = self.conn.write(|tx| tx.execute_one(copy_sql));
         if let Err(err) = res {
-            // TODO(bug 2070884): Disable `delay_ping_lifetime_io` on error.
-            log::error!("Could not load ping lifetime data into memory: {err:?}");
+            log::error!("Could not load ping lifetime data into memory: {err:?}. Disabling ping lifetime IO delay.");
+            self.delay_ping_lifetime_io = false;
```
chutten, line 382: Musing: ...though I wonder now whether we should. This makes the init and state of this struct kinda messy. There's nothing stopping us from deciding to call this multiple times from multiple places, not just around init time.

Might be unavoidable due to this being a db handler, but.

Anyway.

chutten (approved):

glean-core/src/database/sqlite.rs
badboy: That's at least not a new shortcoming. `load_ping_lifetime_data` was callable before, we just made sure it's only called in a single place. The fact that we now toggle `delay_ping_lifetime_io` on an error doesn't change much.

PR #3637, opened 2026-09-11
Bug 2053775 - CI: Run Swift tests on GitHub Actions

badboy:


.github/workflows/test.yml
```diff
@@ -321,6 +331,94 @@ jobs:
             make build-python-wheel GLEAN_BUILD_TARGET="$target" GLEAN_BUILD_EXTRA="--zig"
           done
 
+  test-swift:
+    runs-on: macos-latest
+    needs: [decide-runs]
+    if: needs.decide-runs.outputs.rust-changed == 'true' || needs.decide-runs.outputs.swift-changed == 'true'
```
chutten, line 337: Does this mean an end to /run-ios? 

Also, running all swift tests on all rust changes? Is that a change?

.github/workflows/test.yml
```diff
@@ -321,6 +331,94 @@ jobs:
...
+          # not the Xcode-bundled clang.
+          echo '[target.aarch64-apple-darwin]' >> ~/.cargo/config.toml
+          echo 'linker = "/usr/bin/cc"' >> ~/.cargo/config.toml
+
+          # List available devices -- allows us to see what's there
+          DEVICES=$(xcrun xctrace list devices 2>&1)
+          echo "$DEVICES"
+          # Pick a device and start it
+          UUID=$(echo "$DEVICES" | grep --max-count=1 'iPhone 17 Simulator (26' | awk -F'[()]' '{print $4}')
+          xcrun simctl boot "$UUID"
```
chutten, line 376: I can rubber-stamp this, but if you want a real review you may need to tap someone who understands mac/ios/swift.

chutten (approved):

.github/workflows/test.yml
badboy: 1. Yes! We run iOS builds by default now
2. Yes! We did that whole /run-ios part and not running by default because of cost. macOS time was costly on CircleCI and we were asked to reduce it. Until someone tells me that about GitHub Actions we will just default to running it.

.github/workflows/test.yml
badboy: It's copied from what we do on CircleCI.

PR #3644, opened 2026-09-17
Bug 2072503 - Add the ability to make the initialization wait for a signal

badboy:
In Firefox for Android the Gecko parts are configuring SQLite to use Gecko allocations (among other things), which MUST happen before any SQLite connection is opened.

Glean doesn't know about that and might race with this when initialized in Fenix.
This new API now allows the Gecko part to signal to Glean it's ok to proceed.

It's behind a feature flag (`wait-for-init`, also on the `glean` crate) and Android only. The Gecko build can enable it, but it's otherwise a no-op on regular builds.


badboy:
Closing. We're gonna handle that explicitly on the Fenix side by ensuring Glean init happens when it _can_ happen.

PR #3646, opened 2026-09-21
Re-enable Rkv storage and put SQLite behind a flag

badboy:
This brings back the old Rkv storage, as snapshotted in https://github.com/mozilla/glean/pull/3509
However this keeps the new API introduced for labeled metrics (no label checks though)
as well as any public API.
This is thus fully compatible with Glean v70.

At the same time it keeps the SQLite backend in-tree, behind a feature flag.
This way we can enable that for developer builds to test it out again.


badboy:
Here's a diff of `glean-core/src/database/rkv.rs` against `glean-core/src/database/mod.rs` from #3509: https://gist.github.com/badboy/7ddc415b3da2d52004ef576becfa3627

As you can see indeed nearly the same.
The other important changes are the label validaation in `glean-core/src/metrics/labeled.rs`, `glean-core/src/metrics/dual_labeled_counter.rs` and `glean-core/src/common_metric_data.rs`

chutten (approved):

glean-core/src/metrics/dual_labeled_counter.rs
```diff
@@ -330,9 +339,168 @@ fn label_is_valid(label: &str, metric_id: &str) -> bool {
...
+) -> (HashSet<String>, HashSet<String>) {
+    let base_identifier = &meta.base_identifier();
+    let mut seen_keys: HashSet<String> = HashSet::new();
+    let mut seen_categories: HashSet<String> = HashSet::new();
+    let mut snapshotter = |_metric_id: &[u8], labels: &[&str], _: &Metric| {
+        if labels.len() == 2 {
+            seen_keys.insert(labels[0].to_string());
+            seen_categories.insert(labels[1].to_string());
+        } else {
+            record_error(
```
travis79, line 469: Does this result in lock contention recording the error in the closure here?

badboy: https://bugzilla.mozilla.org/show_bug.cgi?id=2074132

PR #3650, opened 2026-09-21
iOS: Enable the sqlite backend for the glean-swift build

badboy:


jeddai (approved):

badboy:
I changed the implementation to set the feature explicitly only from the actual Swift build, not in general (Python also uses the `glean-bundle` crate)

jeddai:
Ahh I see, yeah. Still looks reasonable to me!

